PR #149, opened 2014-11-12
mmnormalize tokenized-field-type support + json array subscripting support + foreach support + reset-variable support

janmejay:
NOTE: please read the next post for an updated summary (the summary here is stale, but not disturbing it in order to prevent unnecessary confusion).

This patch-set 
- adds tests for tokenized field-type (mmnormalize)
- adds impl + tests for json array-subscripting
- some tests added earlier weren't included in EXTRA_DIST, it adds them to it.

It goes after https://github.com/rsyslog/liblognorm/pull/8


rgerhards:
If I haven't totally lost track, everything should now be properly merged. If you think something is missing, please open a new pull request. Thx!

PR #186, opened 2014-12-15
omkafka: add stats and dynamic topic

oldmantaiter:
stats:
  - submitted: producer submitted message to broker.
  - failures: producer failed to submit message to broker.

Topic is able to be a template, the topics are stored in the instanceData and the kafka_rd_t object has been moved into instanceData. This is alright because librdkafka is threadsafe and to acheive the dynamic topic results it is required.
The topics are stored in a hashtable due to librdkafka leaking memory when passing in a conf object, plus us keeping a local topic "list" makes the subsequent topic conf and rd_kafka_topic_new calls unnecessary. The move of the topic down the writeKafka is for dynamic parsing of the topic for each message that comes through.

Additional items:
  - Added safety to rd_kafka_t destruction, where it flushes the queued messages before object destruction.
  - Dynamic topic is enabled with dyntopic="on" in the configuration.
  - On HUP the active topics are closed, since we only give the topic hashtable 50 items, this could cause issues when rsyslog is running over a long period of time with lots of dynamically parsed topics from templates.
    - Not sure if 50 is an acceptable number of items for the hashtable, I'm open to either increasing or decreasing it.
    - After the HUP, writeKafka should instantiate a new topic and add it back to the hashtable at a minimal performance hit.

These changes have been run through valgrind and I can confirm there is no leaking (at least within my setup).

### DIFF
--- a/plugins/omkafka/omkafka.c
+++ b/plugins/omkafka/omkafka.c
@@ -38,6 +38,8 @@
 #include "module-template.h"
 #include "errmsg.h"
 #include "atomic.h"
+#include "statsobj.h"
+#include "unicode-helper.h"
 
 MODULE_TYPE_OUTPUT
 MODULE_TYPE_NOKEEP
@@ -47,6 +49,14 @@ MODULE_CNFNAME("omkafka")
  */
 DEF_OMOD_STATIC_DATA
 DEFobjCurrIf(errmsg)
+DEFobjCurrIf(statsobj)
+
+statsobj_t *kafkaStats;
+STATSCOUNTER_DEF(ctrTopicSubmit, mutCtrTopicSubmit);
+STATSCOUNTER_DEF(ctrKafkaFail, mutCtrKafkaFail);
+STATSCOUNTER_DEF(ctrCacheMiss, mutCtrCacheMiss);
+STATSCOUNTER_DEF(ctrCacheEvict, mutCtrCacheEvict);
+STATSCOUNTER_DEF(ctrCacheSkip, mutCtrCacheSkip);
 
 #define MAX_ERRMSG 1024 /* max size of error messages that we support */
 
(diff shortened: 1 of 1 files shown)


rgerhards:
Did you have a change to fix and repush in the mean time?

oldmantaiter:
Sorry about that, had been on vacation and forgot I did not push those up. I just amended my commit and pushed up the formatting fix for the last couple lines (the statistics additions).

rgerhards:
Thx -- will merge it within the next couple of days. It will be part of 8.8.0 (Feb 2015), as we have just begun final touches on 8.7.0.

rgerhards:
Oh, not sure if that was lost in my preparations: did you make sure running out of space on the hash table is handled gracefully? I'll review in any case, but it would be good to get a positive ack from you :-)

oldmantaiter:
Good call, I copied what was in imuxsock where it aborts with a RS_RET_OUT_OF_MEMORY if the return from hashtable_insert equals 0. Is that the safest way to do it? The hashtable is cleared on HUP so (at least in our case) we would HUP every ten minutes to roll log files/topics.

The hashtable is hardcoded to 50 entries currently, might it be better if I make that a configuration variable?

rgerhards:
I need to check that in a bit more depth. I think imuxsock just uses this for the rate-limiter, and falls back to using a default one if there is no space left. That doesn't sound like a solution for kafka. I think it needs to be modelled like in omfile, where old entries are evicted. Will have a new look after the release.

oldmantaiter:
I can look into this and update my branch, I also wanted to add a maxhash configuration option so users can size the hashtable to a size that would make sense for them. Seems like I have some time to do this while the other release is being finalized perhaps?

rgerhards:
Sounds good! And yes,  there is time. Upcoming release is jan, 13th, the
next one six weeks later. Given the fact I need some headroom, an updated
patch in 3 to 4 weeks sounds fine. Maybe it's a good idea to split of the
stats counters, I could merge that very quickly.

Looking at omfile should be useful.  Maybe you can try to keep parameter
names as consistent as possible.

Thanks!

Rainer

Sent from phone, thus brief.
Am 08.01.2015 18:10 schrieb "Tait Clarridge" <notifications@github.com>:

oldmantaiter:
So I "borrowed" the dynaFile structure from omfile, tested this and it is working correctly. Thought I'd push it up for more eyes on it, but I will continue testing and make sure the evictions work properly etc.

I also added more stats (revolving around the dynamic topic cache) and changed some more options to be more inline with what other plugins have.

Let me know if I am on the right track, I'm still kind of new so helpful pointers are always welcome :)

rgerhards:
Just for the records: I checked omfile's license, and it is ASL 2.0, so we don't have any licensing issue due to the copied code here.

rgerhards:
This is now in my development branch. will further look at it and see that I migrate it to the actual master branch.

rgerhards:
This is now in master-candidate and will move to master today if nothing really bad happens.

oldmantaiter:
@rgerhards great! I'm submitting another pull request for rsyslog-doc to describe dynamic topic and add an additional warning to partition.auto. Thanks!

rgerhards:
Thanks! I was unfortunately a bit late with the merge, but at least I managed to get it into todays release. Also a big thanks for the doc contrib!

oldmantaiter:
Not a problem. Completely forgot about the docs until I saw the update in the other pull request.

PR #191, opened 2014-12-24
Added dummyparser

iavael:
Parser module that does nothing


rgerhards:
Sorry for this, but we recently had a big discussion on the mailing list that reached the consensus that it makes no sense to have code inside the project that does not provide any real function. In result, we even removed a couple of modules. I guess you want to have this as an example, but I assume any of the other simple parsers can also act like this -- plus, they are maintained.

iavael:
> I guess you want to have this as an example
Well, not really, but I've go the point.

PR #200, opened 2015-01-09
fix build regression in imkmsg from ee035ef

PaulSD:
Sorry, I missed the header files in both imklog and imkmsg in my commits in https://github.com/rsyslog/rsyslog/pull/198 ... it looks like you fixed imklog.h in commit https://github.com/rsyslog/rsyslog/commit/7e309e38e81993e8752d1e4ec6d8844fcf1e03fb but did not fix imkmsg.h.  This commit fixes imkmsg.h.

Thanks!


rgerhards:
No problem, thx for the heads-up!

PR #203, opened 2015-01-11
bugfix header in platform/redhat/rsyslog.conf

### LINKED ISSUE #202: platform/redhat/rsyslog.conf is invalid
the header in platform/redhat/rsyslog.conf is commented out c-style. it includes a path which has the string "*/" (/usr/share/doc/rsyslog-*/rsyslog_conf.html). that breaks the comment, which breaks the config format, which prevents rsyslogd from starting.

i'm sure it's low priority, i assume most distros include a default config without this problem or else someone would have mentioned it by now. i ran into it with rsyslog from buildroot, which uses this file.

iamb:
the "For more information" doc path includes the string "*/", breaking c-style comment format. fixes #202.

### DIFF
--- a/platform/redhat/rsyslog.conf
+++ b/platform/redhat/rsyslog.conf
@@ -1,13 +1,12 @@
-/* rsyslog configuration file (for Red Hat-based systems)
- * note that most of this config file uses old-style format,
- * because it is well-known AND quite suitable for simple cases
- * like we have with the default config. For more advanced 
- * things, RainerScript configuration is suggested.
- *
- * For more information see /usr/share/doc/rsyslog-*/rsyslog_conf.html
- * or latest version online at http://www.rsyslog.com/doc/rsyslog_conf.html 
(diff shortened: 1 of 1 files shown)


rgerhards:
I didn't notice quickly enough that this PR was to the wrong branch (master vs. master-candidate). I'll try to fixup the mess that I created. Don't wonder if strange messages may appear...

rgerhards:
ok, looks like I could fix it. It is now merged to master-candidate and will migrate to master soon (after testbench runs).

iamb:
thanks much, sorry for the hassle i just ran into this and wanted to throw a quick patch, that was all me putting it on the wrong branch and setting you up.

rgerhards:
no need to be sorry -- this was probably the last incarnation of a too-complex way of doing things ;)

have a look here: http://blog.gerhards.net/2015/01/rsyslog-branches-and-git-history.html

Thanks again for taking the time to contribute! Much appreciated.

PR #209, opened 2015-01-15
Make configure fail when building with "--enable-omkafka" but librdkafka wasn't found

Whissi:
We were only checking for headers but haven't defined an "if-not-found" action.
So when building with "--enable-omkafka" but without librdkafka installed,
compilation would fail.

Now we are checking for librdkafak first via PKG_CHECK_MODULES but will fall back to
AC_CHECK_HEADER for librdkafka packages not providing a .pc file.
If these checks will fail, we will now show a friendly error message.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1374,8 +1374,12 @@ AC_ARG_ENABLE(omkafka,
         [enable_omkafka=no]
 )
 if test "x$enable_omkafka"; then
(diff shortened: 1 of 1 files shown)


Whissi:
When you decide to merge this pull request, please also check the other AC_CHECK_HEADER checks like for *omhdfs* or *gssapi* which seems to miss an *if-not-found* action, too.

PR #214, opened 2015-01-20
Fix for ubuntu-12.04 json_array_looping.sh segfault

janmejay:
It was caused by unchecked use of json_object as linked-hashtable. In that test, first few instances are strings, which causes segfault.

The segfault was showing an unknown frame because it was dereferencing hash_fn from invalid object.

### DIFF
--- a/runtime/rsyslog.h
+++ b/runtime/rsyslog.h
@@ -609,7 +609,8 @@ void rsrtSetErrLogger(void (*errLogger)(const int, const int, const uchar*));
 		json_object_object_get_ex((obj), (key), (retobj))
 #else
 #	define RS_json_object_object_get_ex(obj, key, retobj) \
-		((*(retobj) = json_object_object_get((obj), (key))) == NULL) ? FALSE : TRUE
+		(json_object_is_type(obj, json_type_object) && \
+				(*(retobj) = json_object_object_get((obj), (key))) == NULL) ? FALSE : TRUE
 #endif
 
 #ifndef HAVE_JSON_BOOL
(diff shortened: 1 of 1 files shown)


rgerhards:
sorry for the close -- this seems to happen when I cancel the Travis build.

rgerhards:
I have merged it into master-candidate, where it will be run on the other platforms during the next day. Then, it'll (semi-)automatically be moved to master. Just so that you know.

janmejay:
Going forward, should I create PRs for master or master-candidate?

I remember we kinda went back and forth on this one on mails, hence
clarifying one last time.

Also, assuming whatever you say applies to all related repos too.

On Tue, Jan 20, 2015 at 9:42 PM, Rainer Gerhards <notifications@github.com>
wrote:

> I have merged it into master-candidate, where it will be run on the other
> platforms during the next day. Then, it'll (semi-)automatically be moved to
> master. Just so that you know.
>
> —
> <https://github.com/rsyslog/rsyslog/pull/214#issuecomment-70679976>.
>



-- 
Regards,
Janmejay
http://codehunk.wordpress.com

rgerhards:
the PR to master is fine. I'll (usually) still merge it to master-candidate first, because this gives us the additional testbench runs on more platforms. But when all goes well, I'll do a (more or less daily) merge of master-candidate to master. At this point, the PR is automatically closed (and only then). This happens because the git hash remains intact as I no longer try to avoid merge commits and as such do not need to cherry-pick. I think things have now settled enough, so that I can probably write a workflow description later this week. But still a couple of hopefully minor things to work on.

PR #231, opened 2015-01-28
Fixing TLS auth check

sapiophile:
I think, I found an error of checking whether librelp supports TLS authentication or not.


rgerhards:
As this is not an extremely urgent patch, it needs to go into the next release. Thus I am merging in to master-candidate.

rgerhards:
I had to cherry-pick the patch. Now done. So it is merged now, even though github does not understand that due to the different branches and git hashes. Thanks for the contribution!

PR #236, opened 2015-01-30
adding ability to use rawmsg for mmjsonparse

cooper6581:
Not sure if this is the correct way to go about this or not.  I have applications sending data to rsyslog which doesn't contain a syslog header.  Thanks!


rgerhards:
I have merged this into master-candidate, should migrate to master rather quickly.

PR #246, opened 2015-02-18
Adding new format support for error file in omelasticsearch bulk insert mode

CodeHarsh:
Currently omelasticsearch dumps entire post data and response in error file in case for a failure in bulk insert mode. Wanted to introduce two format flags in action to format data in error file (Both of the flags can be used independently as well as together).

1. erroronly => This flag checks the response stream and logs only failed requests and response.
2. interleaved => This flag interleaves the request to its response and dumps it as pair.

When both flags are used we get only failed request in an interleaved format.

sample: action(server="localhost"
    XXX
    type="omelasticsearch"
    XXX
   errorfile="bad_message.log"
   erroronly="on"
   interleaved="on")

BugFix:
While checking if my code is causing any leaks or not, I found a valgrind error for "pWrkrData->batch.data" not sure if that was intentional or not. but I fixed it anyway.

### DIFF
--- a/plugins/omelasticsearch/omelasticsearch.c
+++ b/plugins/omelasticsearch/omelasticsearch.c
@@ -27,6 +27,7 @@
 #include <stdio.h>
 #include <stdarg.h>
 #include <stdlib.h>
+#include <memory.h>
 #include <string.h>
 #include <curl/curl.h>
 #include <curl/easy.h>
@@ -81,6 +82,8 @@ typedef struct _instanceData {
 	uchar *timeout;
 	uchar *bulkId;
 	uchar *errorFile;
+	sbool errorOnly;
+	sbool interleaved;
 	sbool dynSrchIdx;
 	sbool dynSrchType;
 	sbool dynParent;
@@ -105,7 +108,6 @@ typedef struct wrkrInstanceData {
 	} batch;
 } wrkrInstanceData_t;
 
-
 /* tables for interfacing with the v6 config system */
 /* action (instance) parameters */
 static struct cnfparamdescr actpdescr[] = {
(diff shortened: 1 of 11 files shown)


rgerhards:
I have begun to look into these patches. I haven't reviewed much, but I see that you use C++ style comments ("//"). Can you please change them to C-style, because C++ style breaks compilation on some platforms.

rgerhards:
It also looks like the commits should be a little bit squashed. It's not absolutely vital, but would look better.

CodeHarsh:
I have made changes for comment style and also corrected file name causing Travis failure.
I will try to squash the commits (i am new to git) and update.
Also not sure why threadingmqaq.sh test failed due to my changes.

rgerhards:
Thx! I have restarted the travis test. I guess the threadingmqaq.sh has nothing to do with your changes. I think there is already an issue tracker for that test, as the test harness itself (not the code) seems to have an occasional race here. If the next run is OK, all is fine, else I'll investigate.

CodeHarsh:
Hi Rainer, I have squashed commits into logical entities.

CodeHarsh:
Hi Rainer,
Any update on this?

rgerhards:
I still hope to get this into the next release (Apr 7th). But I admit I've slipped asking for feedback. So let's do that now: @radu-gheorghe any concerns?

radu-gheorghe:
@rgerhards no concerns, it looks good to me. Thanks for the patch, @CodeHarsh!

CodeHarsh:
Thanks @radu-gheorghe.
Very happy to hear this.

rgerhards:
@CodeHarsh If no unexpected problem occurs during the multi-env testbench runs (none expected), we should be ready for 8.9.0 next week :-) Thanks for your help.

Would you also mind an update the doc at https://github.com/rsyslog/rsyslog-doc ?

PR #248, opened 2015-02-21
omkafka: Add automatic partitioning

oldmantaiter:
librdkafka provides an automatic partitioning function that will evenly split the produced messages into all partitions configured for that topic.
To use, set partitions.auto="on" in your omkafka action config. This is instead of specifying the number of partitions on the producer side, where it would be easier to change the kafka configuration on the cluster for number of partitions/topic vs on every machine talking to Kafka via rsyslog.


rgerhards:

PR #250, opened 2015-02-23
imptcp: add config parameter for framing fix (framingfix.cisco.asa)

jgerhards:


rgerhards:
please re-check the contribution, the testbench fails...

PR #257, opened 2015-03-02
ommongodb comment update

grue:
While reading through the ommongodb.c source, I noticed a comment was still referencing MySQL. I've updated this to reflect that `initMongoDB` is initializing a MongoDB connection.

I've also eliminated some trailing whitespace characters.


rgerhards:
merged into master-candidate, will propagate from there to master

PR #265, opened 2015-03-10
configure: Enable valgrind only when explicit declared

Whissi:
rsyslog should only use valgrind when requested with "--enable-valgrind".
This allows you to disable valgrind usage in test suite (and guessing is always bad).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -13,10 +13,6 @@ AC_CONFIG_HEADERS([config.h])
 
 AC_USE_SYSTEM_EXTENSIONS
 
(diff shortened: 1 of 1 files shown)


rgerhards:
I am not sure if that's the right step to do. Let's discuss that on the mailing list.

rgerhards:
done: http://lists.adiscon.net/pipermail/rsyslog/2015-March/040058.html

rgerhards:
Not merged in this form as discussed on mailing list. New PR is coming up.

Issue #144, opened 2014-10-29
imfile doesn't work with ReadMode 2


radu-gheorghe:
Complete description here: http://rsyslog-users.1305293.n2.nabble.com/not-forwarding-when-imfile-has-readmode-2-td7586637.html

Basically, something like this doesn't forward anything:
```
module(load="imfile")

input(type="imfile"
      File="/var/log/cassandra/system.log"
      Tag="cassandra-system:"
      ReadMode="2")

*.* @@logsene-receiver-syslog.sematext.com;RSYSLOG_SyslogProtocol23Format 
```

But changing ReadMode to 0 or setting imfile to polling mode "unlocks" imfile and messages get forwarded.

otisg:
Wondering if this has maybe been fixed in the recent 8.7 release, or maybe in the upcoming 8.8 release?

radu-gheorghe:
@rgerhards I finally managed to add tests that should make testing things easier. There were some bumps in the road, it looks like everything works fine if the file is there before rsyslog is started (like it is with other imfile tests). But it doesn't see new lines being added to the file while in ReadMode 2.

That, plus I forgot to properly use imdiag in the conf :)

Please let me know if I can help with anything. I wanted to write a blog post about using imfile  - and inotify is defintely the way to go - but this issue is in the way :(

rgerhards:
2015-03-20 12:43 GMT+01:00 Radu Gheorghe <notifications@github.com>:

> @rgerhards <https://github.com/rgerhards> I finally managed to add tests
> that should make testing things easier. There were some bumps in the road,
> it looks like everything works fine if the file is there before rsyslog is
> started (like it is with other imfile tests). But it doesn't see new lines
> being added to the file while in ReadMode 2.
>
> That, plus I forgot to properly use imdiag in the conf :)
>
> Please let me know if I can help with anything. I wanted to write a blog
> post about using imfile - and inotify is defintely the way to go - but this
> issue is in the way :(
>
>
>
Thanks, I'll give the patch a try in a separate branch and hope that it
will help me understand what's going on :-)

Rainer

radu-gheorghe:
Thanks, Rainer. I've just done some more tests on 8.8 and it looks like ReadMode 2 doesn't do anything with polling mode either. It just sees that the file has changed and doesn't do anything. Here are the debug logs:

polling:
```
7740.186383856:imfile.c       : strm 0x24b7b80: file 4 read 6 bytes
[...]

PR #275, opened 2015-03-20
added test for imfile ReadMode=2

radu-gheorghe:
See https://github.com/rsyslog/rsyslog/issues/144

At the moment, the test fails because of that bug. A couple of notes:
- if you set ReadMode to 0, the test still fails (more headers than expected), but at least all data gets through
- when starting rsyslog, there's a warning about the deprecated use of StateFile. I guess the clean way around that is to remove the StateFile option from all imfile tests and change diag.sh to remove all imfile-state* files. I can do that as a separate PR if it's needed

### DIFF
--- a/runtime/glbl.c
+++ b/runtime/glbl.c
@@ -109,6 +109,7 @@ static struct cnfparamdescr cnfparamdescr[] = {
 	{ "debug.logfile", eCmdHdlrString, 0 },
 	{ "defaultnetstreamdrivercafile", eCmdHdlrString, 0 },
 	{ "defaultnetstreamdriverkeyfile", eCmdHdlrString, 0 },
+        { "defaultnetstreamdrivercertfile", eCmdHdlrString, 0 },
 	{ "defaultnetstreamdriver", eCmdHdlrString, 0 },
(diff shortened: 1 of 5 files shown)


rgerhards:
On the state files: We should keep this in a couple of tests so that we can test if the legacy functionality still works. IMHO we need to do this as long as we support it. Most probably not for all tests, maybe a single one is sufficient. I would suggest to duplicate the test(s) in question with and without state file.

I'll merge this into one of my temp development branches first, which I will use to look into the issue. When fixed, we can merge everyting into master.

rgerhards:
A quick question: would ASL 2.0 as lincese be fine with you, too? (Else I need to remove the test if we ever manage to move to ASL 2.0, thus the question).

radu-gheorghe:
Yes, ASL2 is fine, sure. I was just copying it from the imfile-basic test.

rgerhards:
I have merged this into my feature branch, which contains the fix. I think both togehter should hit master branch. This will happen once the feature branch passed all tests.

rgerhards:
This is merged, but does not show up as merged probably due to a rebase.

PR #276, opened 2015-03-20
imjournal: fix default message priority

theinric:
Looks like this got mixed up in 1ac20ff.


rgerhards:
merged into maste-candidate, will migrate to master after testbench runs.

Issue #264, opened 2015-03-10
MsgDeserialize out of sync with MsgSerialize for StrucData


adruch:
In msg.c, the MsgSerialize field order does not match MsgDeserialize. Serialize puts structured data labeled as "pszRcvStrucData" after "pszRcvFromIP". Deserialize expects structured data labeled as "pszStrucData" after "localvars".

rgerhards:
a related PR has been merged a couple of days ago. This is fixed in 8.9.0.

PR #283, opened 2015-03-25
Fix field name for pszStrucData in MsgSerialize (Issue #264)

adruch:
Rename the field for serializing pszStrucData so MsgSerialize and MsgDeserialize align.


rgerhards:
Hi, I would like to re-confirm that you actually publish this patch under ASL 2.0 as discussed on the ML. Thanks.

adruch:
Yes. ASL 2.0 is fine.

What is the proper way to label a patch/pull request with a license?

rgerhards:
Thanks - for my needs it's sufficient it's stated here, that clearly is an orginal thought.

PR #289, opened 2015-04-03
Make llDestroy safe

theinric:
Keep the destructed list in a consistent state as the provided
destructor may access the very same list again before llDestroy
finishes.

The previous approach could actually trigger a segmentation violation
error for specific scenarios and configurations.
In one particular case, destructing the list of rulesets lead to
shutdown of an action DA queue and spooling its messages to disk which
in turn triggered a search for the ruleset name of the message trought
the destructed list.

The change probably slightly degrades performace, but that shouldn't
have an observable effect in the current code base.

### DIFF
--- a/runtime/linkedlist.c
+++ b/runtime/linkedlist.c
@@ -90,22 +90,24 @@ rsRetVal llDestroy(linkedList_t *pThis)
 {
 	DEFiRet;
 	llElt_t *pElt;
-	llElt_t *pEltPrev;
 
 	assert(pThis != NULL);
 
 	pElt = pThis->pRoot;
 	while(pElt != NULL) {
-		pEltPrev = pElt;
-		pElt = pElt->pNext;
+		/* keep the list structure in a consistent state as
+		 * the destructor bellow may reference it again
+		 */
+		pThis->pRoot = pElt->pNext;
+		if(pElt->pNext == NULL)
+			pThis->pLast = NULL;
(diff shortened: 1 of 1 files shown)


theinric:
For the record: this bug has been there since at least v5.
Special configuration and a bit of "luck" is needed to reproduce the crash.

A sample config:
$ModLoad imfile
$RuleSet rs
$ActionQueueType LinkedList
$ActionQueueFileName fwd
$ActionQueueSaveOnShutdown on
$ActionResumeRetryCount -1
$ActionQueueSize 10000
$ActionQueueLowWaterMark 2000
$ActionQueueHighWaterMark 8000
$ActionQueueDiscardMark 9750
*.* @@127.0.0.1:1234
$RuleSet RSYSLOG_DefaultRuleset
$InputFileBindRuleset rs
$InputFileName /tmp/infile
$InputFileTag infile:
$InputFileStateFile infile.ste
$InputFileSeverity notice
$InputFileFacility local3
$InputFilePollInterval 10
$InputRunFileMonitor

Steps to reproduce
- the target of the forwarder does not need to exist
- write 10k lines to the input file
- wait for them to get picked up
- shut down the daemon

This should trigger the crash. In my environment with the above configuration, I get about 95% reproducibility.

rgerhards:
thx! merged into master-candidate, will migrate from there to master.

PR #303, opened 2015-04-15
fixes foreach-behavior with async-actions and call + some other smaller issues

janmejay:
- Fixes: Foreach body calling out to async-actions
- Fixes: Foreach body using 'call' to enqueue to other rulesets
- Foreach now checks that the object being iterated upon is really and array
- Fixes: omelasticsearch uninitialized-pointer access in error-logging
- Introduces: concept of queuing to action by copy (param is called action.copyMsg, and is off by default) (until now it only supported queuing by reference). Because 'call' already copies the message, such a control isn't required for 'call'.

Documentation changes coming up, will post link to PR as comment.


janmejay:
Here is the Doc PR: https://github.com/rsyslog/rsyslog-doc/pull/143

rgerhards:
merged into master-candidate with a small nitfix: https://github.com/rsyslog/rsyslog/commit/31f529fb4544453608ac3cdb02b95eaddefc6882

will migrate to master if the extended testbench runs do not point into issues.

Issue #304, opened 2015-04-16
Allow JSON output customization (plain/spaced/pretty)


philsttr:
Currently, when JSON is output ([here](https://github.com/rsyslog/rsyslog/blob/master/runtime/msg.c#L3269)), extra spaces are added after colons/commas/open brackets and before closing brackets.  This creates a lot of extra unnecessary log volume, especially under high load in an enterprise environment.  This is extra-painful when paying for hosted logging providers that charge by the amount of data logged.

It would be great if these extra unnecessary spaces could be disabled through configuration.

Luckily, the json library used allows just such configuration.  rsyslog just needs to take advantage of it.

Currently [`json_object_get_string`](https://json-c.github.io/json-c/json-c-0.10/doc/html/json__object_8h.html#a1f62a660e2619559d1f3f54cd7483ff8) is used, which is equivalent to `json_object_to_json_string_ext(obj, JSON_C_TO_STRING_SPACED)`

This should be changed to use [`json_object_get_string_ext`](https://json-c.github.io/json-c/json-c-0.10/doc/html/json__object_8h.html#a84421dab94ccad42e901e534c6d7b658), which takes flags as input to control the JSON output.  The value passed to the flags parameter should be configurable.  The possible flags are:
* `JSON_C_TO_STRING_PLAIN` - no extra spaces
* `JSON_C_TO_STRING_SPACED` - (default) extra spaces
* `JSON_C_TO_STRING_PRETTY` - extra spaces / lines

rsyslog should allow any of these to be used.

rgerhards:
I was about to merge and realize that we should NOT support *_PRETTY, because multi-line message always cause trouble and we should not deliberatly support something that gets users into trouble. Any comments here?

cooper6581:
@rgerhards You are totally right.  I should have thought about this.  I removed the pretty option in #306 .

rgerhards:
to be released with 8.10.0.

PR #306, opened 2015-04-21
Adding JSON output format options

cooper6581:
This is a proposal for issue #304.  I'm guessing you don't want to pollute the system properties with those format settings, but thought I would take a shot anyway.  Thanks!


rgerhards:
unfortunately, the patch does not work. Maybe it is not checking json-c capabilities (see Travis log)

rgerhards:
I have merged this into master-candidate now. From there, it will migrate automatically into master after successful run of the extended testbench (on different platforms).

Would you mind adding some doc to the rsyslog-doc project as well, so that others can actually use that functionality?

PR #307, opened 2015-04-22
imuxsock: fix a crash when setting a hostname

theinric:
Setting a hostname via the legacy directive would lead to a crash
during shutdown caused by a double-free.


rgerhards:
merged into master-candidate, will migrate to master automatically. To be released with 8.10.0.

PR #308, opened 2015-04-22
Revert "imuxsock: fix a crash when setting a hostname"

rgerhards:
Reverts rsyslog/rsyslog#307 ; merged into wrong branch (not an emergency fix, thus needs to go into master)

### DIFF
--- a/plugins/imuxsock/imuxsock.c
+++ b/plugins/imuxsock/imuxsock.c
@@ -364,7 +364,8 @@ static rsRetVal addInstance(void __attribute__((unused)) *pVal, uchar *pNewVal)
 	inst->bParseHost = UNSET;
 	inst->next = NULL;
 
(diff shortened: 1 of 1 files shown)


theinric:
> Reverts rsyslog/rsyslog#307 ; merged into wrong branch (not an emergency fix, thus needs to go into master)

Sorry about that. I thought this was eligible for the stable branch.

rgerhards:
@theinric my fault, I didn't precisely enough communicate. When we switched to 6-week release cycle, we also said that stable only gets updates in real emergency situations (like security issues). Everything else can and should wait. In essence, it's mostly like before, where we merged into v8-stable, and released when we felt it was time to do so. Now, everything goes into master, and once we release, I update v8-stable based on master.

Usually, I check branches before merging (and adjust), but this time I didn't realize it. Not a big deal.

PR #309, opened 2015-04-22
Make configure checks dependend on build source

Whissi:
We are now checking if we are running from git source. If we do that, we will require `rst2man` and `yacc`.

This PR is based on @davidelang report from http://lists.adiscon.net/pipermail/rsyslog/2015-April/thread.html#40318


Whissi:
Oh, the check for `flex` is missing.

Whissi:
`configure` is now checking for `flex`, too.

PR #314, opened 2015-04-28
Clear two small memleaks

theinric:
Details are in the commits.


rgerhards:
I've check my notes (and my memory ;)): it looks like I did never implement the full unload of many modules because they will be unloaded at rsyslogd termination in any case. Just so that you know.

PR #316, opened 2015-04-29
Testbench: add some tests for the directive $AbortOnUncleanConfig (bad case & good case)

teifler:
This is another test for the directive "$AbortOnUncleanConfig" with a invalid configuration. 

The test works with the return code of the configuration check. If the return code of the configuration check is not equals null then the test passed otherwise the test should returns a fail.


rgerhards:
merged into master-candidate, will migrate to master from there

PR #319, opened 2015-04-30
Typo fix in omfile.c

friedl:


rgerhards:
merged into master-candidate, will migrate to master from there.

PR #340, opened 2015-05-18
bugfix: fix default action naming

theinric:
The default name generated for an action references a global counter
instead of the action-specific number. In practice this means that the
number in the name will be greater by at least one than the assined
number most of the time.

NB: The code generating various action-related messages is using the
action name inconsistently; sometimes the generated name is used,
sometimes the action number is used, thus producing "action N" or
"action N+1" identifiers at different locations in the code.

### DIFF
--- a/action.c
+++ b/action.c
@@ -385,7 +385,7 @@ actionConstructFinalize(action_t *__restrict__ const pThis, struct nvlst *lst)
 	}
 	/* generate a friendly name for us action stats */
 	if(pThis->pszName == NULL) {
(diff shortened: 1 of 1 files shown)


rgerhards:
That whole action naming has evolved and there is a lot of history in it. Probably we should try to find a better default name, e.g. with at least the module in it. But that's for another PR ;)

rgerhards:
will be released with 8.10.0.

theinric:
Thanks.

rgerhards:
I also added a new feature request: https://github.com/rsyslog/rsyslog/issues/342

PR #330, opened 2015-05-13
Omhttpfs - write log to hdfs over http using libcurl

sskaje:
omhttpfs - write log to hdfs over http using libcurl

Developing and being used on centos 6 in my previous workplace (based on Rsyslog 8.6.0).

PR #341, opened 2015-05-19
omhttpfs: some cleanup

rgerhards:
fixing some issue from https://github.com/rsyslog/rsyslog/pull/330


rgerhards:
There is also an issue with older json-c, which is not supported (see https://github.com/rgerhards/rsyslog/commit/b88c88a690378d7f1994dd6209649ae1318afd93)

PR #343, opened 2015-05-19
Fix tests/Makefile to include missing tests in make dist target

Whissi:


rgerhards:
merged into master-candidate, will migrate from there into master.

PR #355, opened 2015-05-21
omkafka concurrency fixes

janmejay:
Plugin omkafka had several data-races, this patchset fixes them.

The symptom was regular crashes under load. Most of them pointed to omkafka(and librdkafka) area (statistically).

This should fix all of them (or all that I could find by reading code).

@rgerhards I may not be capable of seeing more races in this. Im too tainted at this point. So please review carefully, it may have some more.

### DIFF
--- a/plugins/omkafka/omkafka.c
+++ b/plugins/omkafka/omkafka.c
@@ -117,7 +117,9 @@ typedef struct _instanceData {
 	int fdErrFile;		/* error file fd or -1 if not open */
 	pthread_mutex_t mutErrFile;
 	int bIsOpen;
+	pthread_rwlock_t rkLock;
 	rd_kafka_t *rk;
+	int closeTimeout;
 } instanceData;
 
 typedef struct wrkrInstanceData {
@@ -138,7 +140,8 @@ static struct cnfparamdescr actpdescr[] = {
 	{ "confparam", eCmdHdlrArray, 0 },
 	{ "topicconfparam", eCmdHdlrArray, 0 },
 	{ "errorfile", eCmdHdlrGetWord, 0 },
-	{ "template", eCmdHdlrGetWord, 0 }
+	{ "template", eCmdHdlrGetWord, 0 },
+	{ "closeTimeout", eCmdHdlrPositiveInt, 0 }
 };
 static struct cnfparamblk actpblk =
 	{ CNFPARAMBLK_VERSION,
@@ -164,12 +167,14 @@ getPartition(instanceData *const __restrict__ pData)
 }
 
 /* destroy topic item */
+/* must be called with write(rkLock) */
 static rsRetVal
 closeTopic(instanceData *__restrict__ const pData)
 {
 	DEFiRet;
 	DBGPRINTF("omkafka: closing topic %s\n", rd_kafka_topic_name(pData->pTopic));
 	rd_kafka_topic_destroy(pData->pTopic);
+	pData->pTopic = NULL;
 	RETiRet;
 
 }
@@ -235,6 +240,7 @@ static void dynaTopicFreeCache(instanceData *__restrict__ const pData)
 }
 
 /* create the topic object */
+/* must be called with write(rkLock) */
 static rsRetVal
 prepareTopic(instanceData *__restrict__ const pData, const uchar *__restrict__ const newTopicName)
 {
@@ -481,26 +487,43 @@ do_rd_kafka_destroy(instanceData *const __restrict pData)
 	if (pData->rk == NULL) {
 		DBGPRINTF("omkafka: can't close, handle wasn't open\n");
 	} else {
-		DBGPRINTF("omkafka: closing - items left in outqueue: %d\n",
-				  rd_kafka_outq_len(pData->rk));
-		while (rd_kafka_outq_len(pData->rk) > 0)
-			rd_kafka_poll(pData->rk, 10);
+		int queuedCount = rd_kafka_outq_len(pData->rk);
+		DBGPRINTF("omkafka: closing - items left in outqueue: %d\n", queuedCount);
+
+		struct timespec tOut;
+		timeoutComp(&tOut, pData->closeTimeout);
+		
+		while (timeoutVal(&tOut) > 0) {
+			queuedCount = rd_kafka_outq_len(pData->rk);
+			if (queuedCount > 0) {
+				rd_kafka_poll(pData->rk, 10);
+			} else {
+				break;
+			}
+		}
+		if (queuedCount > 0) {
+			DBGPRINTF("omkafka: queue-drain for close timed-out took too long, "
+					 "items left in outqueue: %d\n",
+					 rd_kafka_outq_len(pData->rk));
+		}
+		if (pData->dynaTopic) {
+			/*TBD*/
+		} else {
+			closeTopic(pData);
+		}
 		rd_kafka_destroy(pData->rk);
 		pData->rk = NULL;
 	}
 }
 
-
-static rsRetVal
+/* should be called with write(rkLock) */
+static void
 closeKafka(instanceData *const __restrict__ pData)
 {
-	DEFiRet;
-	if(!pData->bIsOpen)
-		FINALIZE;
-	do_rd_kafka_destroy(pData);
-	pData->bIsOpen = 0;
-finalize_it:
-	RETiRet;
+	if(pData->bIsOpen) {
+		do_rd_kafka_destroy(pData);
+		pData->bIsOpen = 0;
+	}
 }
 
 static void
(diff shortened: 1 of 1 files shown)


janmejay:
Created PR to cover doc changes: https://github.com/rsyslog/rsyslog-doc/pull/157

janmejay:
FWIW have been running this in different environments without crashes for a few days now. So we may have fixed em all.

rgerhards:
@janmejay could you please have a look at this PR: https://github.com/rsyslog/rsyslog/pull/186

Previous to it, each worker instance had it's own kafka handle, which was removed by the PR because "this is not required". I wonder if we could save some of the looking operations be re-instantiating the previous code. I haven't looked in very much detail, but got the strong impression this might be the case.

janmejay:
Looked at it briefly. Resurrecting that version will reduce locking in
omkafka code, but for following reasons, I still think we should not do
that.

It will also make omkafka really bulky, because each worker will have its
own rd_kafka_t, which means it'll have its own rd-kafka-main-thread and a
thread for each broker (they call it broker-main-thread) and its own
dyn-topic-cache (assuming we choose to keep the dyn-topic feature).

According to the design of librdkafka, the current style seems to be the
right way to use it. Infact it goes one step further to maintain
process-wide things (eg. total number of threads created in any librdkafka
instance etc) which is aligned to this kind of usage (sharing rd_kafka_t, I
mean).

But having said that, I'd prefer librdkafka to use an single-threaded
event-loop style architecture (instead of dedicated main and broker-threads
etc), which would be much easier to fit into a variety of usage patterns
(stand-alone apps with multilple threads, stand-alone apps with single
thread with specific requirements around forking, plugin model similar to
rsyslog etc). Its easier to scale by simply adding more threads running the
top-level event loop and by partitioning data between top-level event-loop
threads and has simplicity benefits. But this will require a much larger
change in librdkafka, and I haven't talked to anyone there to see what they
think of this.


On Tue, May 26, 2015 at 6:27 PM, Rainer Gerhards <notifications@github.com>
wrote:

> @janmejay <https://github.com/janmejay> could you please have a look at
> this PR: #186 <https://github.com/rsyslog/rsyslog/pull/186>
>
> Previous to it, each worker instance had it's own kafka handle, which was
> removed by the PR because "this is not required". I wonder if we could save
> some of the looking operations be re-instantiating the previous code. I
> haven't looked in very much detail, but got the strong impression this
> might be the case.
>
> —
> <https://github.com/rsyslog/rsyslog/pull/355#issuecomment-105517579>.
>



-- 
Regards,
Janmejay
http://codehunk.wordpress.com

janmejay:
The downside of current design is that its not very good for dyn-topics.
Steady-state pumping of messages without dyn-topic has almost no
performance-overhead (because its read-lock and not a mutex). With
dyn-topic, it is actually a mutex. I can possibly optimize it if you feel
strongly about it. Dyn-topic equivalent effect for all practical purpose
(and for performance critical scenarios) can be achieved using if-else
rainerscript construct to pump messages to respective non-dyn-topic
actions, so if someone really wants performance, they can get it this way.
If one prefers convenience, dyn-topic is pretty good.

On Wed, May 27, 2015 at 8:06 AM, singh.janmejay <singh.janmejay@gmail.com>
wrote:

> Looked at it briefly. Resurrecting that version will reduce locking in
> omkafka code, but for following reasons, I still think we should not do
> that.
>
> It will also make omkafka really bulky, because each worker will have its
> own rd_kafka_t, which means it'll have its own rd-kafka-main-thread and a
> thread for each broker (they call it broker-main-thread) and its own
> dyn-topic-cache (assuming we choose to keep the dyn-topic feature).
>
> According to the design of librdkafka, the current style seems to be the
> right way to use it. Infact it goes one step further to maintain
> process-wide things (eg. total number of threads created in any librdkafka
> instance etc) which is aligned to this kind of usage (sharing rd_kafka_t, I
> mean).
>
> But having said that, I'd prefer librdkafka to use an single-threaded
> event-loop style architecture (instead of dedicated main and broker-threads
> etc), which would be much easier to fit into a variety of usage patterns
> (stand-alone apps with multilple threads, stand-alone apps with single
> thread with specific requirements around forking, plugin model similar to
> rsyslog etc). Its easier to scale by simply adding more threads running the
> top-level event loop and by partitioning data between top-level event-loop
> threads and has simplicity benefits. But this will require a much larger
> change in librdkafka, and I haven't talked to anyone there to see what they
> think of this.
>
>
> On Tue, May 26, 2015 at 6:27 PM, Rainer Gerhards <notifications@github.com
> > wrote:
>
>> @janmejay <https://github.com/janmejay> could you please have a look at
>> this PR: #186 <https://github.com/rsyslog/rsyslog/pull/186>
>>
>> Previous to it, each worker instance had it's own kafka handle, which was
>> removed by the PR because "this is not required". I wonder if we could save
>> some of the looking operations be re-instantiating the previous code. I
>> haven't looked in very much detail, but got the strong impression this
>> might be the case.
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/rsyslog/rsyslog/pull/355#issuecomment-105517579>.
>>
>
>
>
> --
> Regards,
> Janmejay
> http://codehunk.wordpress.com
>



-- 
Regards,
Janmejay
http://codehunk.wordpress.com

janmejay:
Also, the PR conversation at one point mentions copying omfile's
dyn-entity-cache. I haven't verified this yet, but it may share some of the
bugs we fixed here, so something for us to watch out for.

On Wed, May 27, 2015 at 8:13 AM, singh.janmejay <singh.janmejay@gmail.com>
wrote:

> The downside of current design is that its not very good for dyn-topics.
> Steady-state pumping of messages without dyn-topic has almost no
> performance-overhead (because its read-lock and not a mutex). With
> dyn-topic, it is actually a mutex. I can possibly optimize it if you feel
> strongly about it. Dyn-topic equivalent effect for all practical purpose
> (and for performance critical scenarios) can be achieved using if-else
> rainerscript construct to pump messages to respective non-dyn-topic
> actions, so if someone really wants performance, they can get it this way.
> If one prefers convenience, dyn-topic is pretty good.
>
> On Wed, May 27, 2015 at 8:06 AM, singh.janmejay <singh.janmejay@gmail.com>
> wrote:
>
>> Looked at it briefly. Resurrecting that version will reduce locking in
>> omkafka code, but for following reasons, I still think we should not do
>> that.
>>
>> It will also make omkafka really bulky, because each worker will have its
>> own rd_kafka_t, which means it'll have its own rd-kafka-main-thread and a
>> thread for each broker (they call it broker-main-thread) and its own
>> dyn-topic-cache (assuming we choose to keep the dyn-topic feature).
>>
>> According to the design of librdkafka, the current style seems to be the
>> right way to use it. Infact it goes one step further to maintain
>> process-wide things (eg. total number of threads created in any librdkafka
>> instance etc) which is aligned to this kind of usage (sharing rd_kafka_t, I
>> mean).
>>
>> But having said that, I'd prefer librdkafka to use an single-threaded
>> event-loop style architecture (instead of dedicated main and broker-threads
>> etc), which would be much easier to fit into a variety of usage patterns
>> (stand-alone apps with multilple threads, stand-alone apps with single
>> thread with specific requirements around forking, plugin model similar to
>> rsyslog etc). Its easier to scale by simply adding more threads running the
>> top-level event loop and by partitioning data between top-level event-loop
>> threads and has simplicity benefits. But this will require a much larger
>> change in librdkafka, and I haven't talked to anyone there to see what they
>> think of this.
>>
>>
>> On Tue, May 26, 2015 at 6:27 PM, Rainer Gerhards <
>> notifications@github.com> wrote:
>>
>>> @janmejay <https://github.com/janmejay> could you please have a look at
>>> this PR: #186 <https://github.com/rsyslog/rsyslog/pull/186>
>>>
>>> Previous to it, each worker instance had it's own kafka handle, which
>>> was removed by the PR because "this is not required". I wonder if we could
>>> save some of the looking operations be re-instantiating the previous code.
>>> I haven't looked in very much detail, but got the strong impression this
>>> might be the case.
>>>
>>> —
>>> Reply to this email directly or view it on GitHub
>>> <https://github.com/rsyslog/rsyslog/pull/355#issuecomment-105517579>.
>>>
>>
>>
>>
>> --
>> Regards,
>> Janmejay
>> http://codehunk.wordpress.com
>>
>
>
>
> --
> Regards,
> Janmejay
> http://codehunk.wordpress.com
>



-- 
Regards,
Janmejay
http://codehunk.wordpress.com

rgerhards:
omfile does not support concurrent writes in the same action (to protect the file), so there can be no problem (given the fact that nothing in this area was reported in the past years makes me tend to say this is a solid assertion).

rgerhards:
ok, thx for the explanation of librdkafka. So I think we are ready to merge :-)

rgerhards:
merged to master-candidate, will migrate to master.

PR #356, opened 2015-05-22
test to validate stop in foreach works as expected

janmejay:
Test to validate of stop in foreach.


rgerhards:
merged to master-candidate, will migrate to master from there.

PR #358, opened 2015-05-26
Global variable atomicity

janmejay:
This change ensures global-variables are written and read atomically.

This makes things such as:
    set $/count = $/count + 1;
    set $.i = $/count % 16;
work across multiple threads.

It also kills an unused function.

### DIFF
--- a/contrib/mmcount/mmcount.c
+++ b/contrib/mmcount/mmcount.c
@@ -321,7 +321,7 @@ finalize_it:
 	}
 
 	if(json) {
-		msgAddJSON(pMsg, (uchar *)JSON_COUNT_NAME, json, 0);
+		msgAddJSON(pMsg, (uchar *)JSON_COUNT_NAME, json, 0, 0);
 	}
 ENDdoAction
 
--- a/contrib/mmsequence/mmsequence.c
+++ b/contrib/mmsequence/mmsequence.c
@@ -373,7 +373,7 @@ CODESTARTdoAction
 	if (json == NULL) {
 		errmsg.LogError(0, RS_RET_OBJ_CREATION_FAILED,
 				"mmsequence: unable to create JSON");
-	} else if (RS_RET_OK != msgAddJSON(pMsg, (uchar *)pData->pszVar + 1, json, 0)) {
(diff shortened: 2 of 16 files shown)


rgerhards:
Could you elaborate a little bit about what is the core change in this patch set and why it is needed. I see that the json object is now deep-copied. Did the previous way of doing things have problems in this case:

thread 1: obtain var ptr
thread 2: obtain var ptr
thread 1: read var
thread 1: set new value var
---> did var ptr from thread 1 become corrupted here?

janmejay:
Slightly different, because not thread 1 but thread 2 sees garbage (from
thread 1's pov, its all good):

time 1: thd 1 get x
time 2: thd 2 get x
time 3: thd 1 modify x
time 4: thd 2 read x /* sees garbage, may either have arbitrary value or
may crash */

Basically the patch-set solves 2 problems:
1. makes it explicit when data being put in a variable is expected to
continue to be used by the area of code that produced that data (so we can
make a copy)
2. forces every read to operate within global-lock (by not letting
references to global objects from getting out of lock), this is why we make
a copy (if we didn't, there would be no way to guard that data from
concurrent-reads and modifications).

In order do achieve the 2nd, we need readers to free objects (because now
it can potentially be a copy rather than the original). So now all returned
objects are reference-count-incremented, so that caller can free it in the
same way regardless of it being a copy or not.

On Tue, May 26, 2015 at 6:08 PM, Rainer Gerhards <notifications@github.com>
wrote:

> Could you elaborate a little bit about what is the core change in this
> patch set and why it is needed. I see that the json object is now
> deep-copied. Did the previous way of doing things have problems in this
> case:
>
> thread 1: obtain var ptr
> thread 2: obtain var ptr
> thread 1: read var
> thread 1: set new value var
> ---> did var ptr from thread 1 become corrupted here?
>
> —
> <https://github.com/rsyslog/rsyslog/pull/358#issuecomment-105510352>.
>



-- 
Regards,
Janmejay
http://codehunk.wordpress.com

janmejay:
Have been running this in an environment for more than 12 hours now without any crashes (earlier it was crashing once in <= 6 hours).

rgerhards:
@janmejay I think we describe the same case, you just expected my thread numbers to be different ;)

rgerhards:
merged into master-candidate, will propagate to master

Issue #357, opened 2015-05-22
Crash in imfile when using startmsg.regex


hakman:
I am trying the new startmsg.regex feature to parse some java logs.
I managed to get into some issues, but the latest seems more serious.
This works as expected until the crash. Tried it several times on the same log file and looks like it stops in the same place.
Looking at the backtrace seems that, for some reason, the regexp is applied on a NULL pointer instead of a string.

I can try to prepare a simpler way to reproduce this, but for now, the log file is quite large.
Could you take a look, maybe the backtrace helps?

This is the backtrace from gdb:
```
(gdb) bt
#0  strlen () at ../sysdeps/x86_64/strlen.S:106
#1  0x00007fa23a231d12 in __regexec (preg=preg@entry=0xa31ae8, string=0x0, nmatch=nmatch@entry=0, pmatch=pmatch@entry=0x0, eflags=eflags@entry=0) at regexec.c:245
#2  0x000000000042c840 in strmReadMultiLine (pThis=0x7fa2240021d0, ppCStr=ppCStr@entry=0x7fa2334676a0, preg=preg@entry=0xa31ae8, bEscapeLF=<optimized out>) at stream.c:845
#3  0x00007fa2390f9da5 in pollFile (pLstn=0xa31a70, pbHadFileData=0x0) at imfile.c:576
#4  0x00007fa2390fad98 in in_setupFileWatchStatic (pLstn=0xa31a70) at imfile.c:1510
#5  in_setupInitialWatches () at imfile.c:1526
#6  do_inotify () at imfile.c:1721
#7  0x00007fa2390faee7 in runInput (pThrd=<optimized out>) at imfile.c:1769
#8  0x000000000043d460 in thrdStarter (arg=0xa70e00) at ../threads.c:212
#9  0x00007fa23af3b182 in start_thread (arg=0x7fa23346a700) at pthread_create.c:312
#10 0x00007fa23a24847d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111
```

The input section of my config has nothing special:
```
input(
  type="imfile"
  File="/var/log/jetty.log"
  startmsg.regex="^[[:digit:]]{4}-[[:digit:]]{2}-[[:digit:]]{2}"
  Tag="logsene-receiver:"
  Ruleset="logsene"
)
```

rgerhards:
Would it be possible for you to run it under valgrind?

hakman:
Not sure how to do it, but with some tips, I can try.

[...]

PR #359, opened 2015-05-26
Fix crash when log line is empty

hakman:
Fix for [issue 357](https://github.com/rsyslog/rsyslog/issues/357).
Rsyslog crashes when using startmsg.regex in imfile because of empty log line.


rgerhards:
merged into master-candidate, will migrate from there to master. To be released with 8.11.0.

hakman:
Perfect. Thanks!

Issue #367, opened 2015-05-26
Crash when using property parameter controlcharacters="escape"


hakman:
Using controlcharacters="escape" in a template triggers a crash. This should happen for any log line containing a control character. I will provide a possible fix in a PR soon.

Stack trace:

```
(gdb) run
Starting program: /home/ubuntu/rsyslogd -n
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff0edc700 (LWP 9547)]
[New Thread 0x7ffff06db700 (LWP 9548)]
[New Thread 0x7fffefeda700 (LWP 9549)]
[New Thread 0x7fffef6d9700 (LWP 9550)]
*** Error in `/home/ubuntu/rsyslogd': munmap_chunk(): invalid pointer: 0x00007fffd8092c22 ***

Program received signal SIGABRT, Aborted.
[Switching to Thread 0x7fffef6d9700 (LWP 9550)]
0x00007ffff6bf4cc9 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
56  ../nptl/sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  0x00007ffff6bf4cc9 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007ffff6bf80d8 in __GI_abort () at abort.c:89
#2  0x00007ffff6c31394 in __libc_message (do_abort=do_abort@entry=1, fmt=fmt@entry=0x7ffff6d3fb28 "*** Error in `%s': %s: 0x%s ***\n") at ../sysdeps/posix/libc_fatal.c:175
#3  0x00007ffff6c3c0f7 in malloc_printerr (action=<optimized out>, str=0x7ffff6d3fea8 "munmap_chunk(): invalid pointer", ptr=<optimized out>) at malloc.c:4996
#4  0x000000000041cecf in MsgGetProp (pMsg=pMsg@entry=0x7fffe40cff40, pTpe=pTpe@entry=0x6df9e0, pProp=pProp@entry=0x6dfa00, pPropLen=pPropLen@entry=0x7fffef6d89c4, 
[...]

PR #369, opened 2015-05-26
Fix crash when using property parameter controlcharacters="escape"

hakman:
This is a fix for [issue 367](https://github.com/rsyslog/rsyslog/issues/367).


rgerhards:
quick question: we are moving rsyslog towards ASL 2.0, but msg.c is still under GPL. Could you license your **patch** under ASL 2.0?

hakman:
I am ok with ASL 2.0 for the patch. Should I do anything special to the patch for this to happen?

rgerhards:
no, thx. It's recorded here in the issue tracker, that's good enough

rgerhards:
merged into master-candidate, will migrate from there into master

PR #383, opened 2015-05-28
fix memleak in mmjsonparse plugin

antmat:
During json processing it was not add into message if it was not parsed fully even when json is not null. It resulted in memory leak and this patch fixes this behaviour.


antmat:
Hi! what about this? :-)

rgerhards:
We usually do these kind of things in the finalizer, so I need to change the patch a bit. Currently I am swamped and will do as soon as time is left.

rgerhards:
Just to let you know: this was the perfect place to free() in the code, so I could merge as is. Note, thought that I changed to C++ comment to C-style because this would fail compilation on some platforms otherwise.

antmat:
Fixed  :-)

rgerhards:
I already merged and did the fix myself. So I am closing now. It's merged, albeit it no longer looks so :-)

PR #386, opened 2015-05-31
add sha1 property to message object

bdoin:
This property returns the SHA1 (160-bit) checksums of the raw message.
For users looking after a unique identifier of the message the sha1 can be
prefered over the uuid approach that is random based.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -719,6 +719,23 @@ fi
 AM_CONDITIONAL(ENABLE_UUID, test x$enable_uuid = xyes)
 
 
+# sha1 support
+AC_ARG_ENABLE(sha1,
+        [AS_HELP_STRING([--enable-sha1],[Enable support for sha1 @<:@default=yes@:>@])],
+        [case "${enableval}" in
+         yes) enable_sha1="yes" ;;
+          no) enable_sha1="no" ;;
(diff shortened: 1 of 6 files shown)


andreasstieger:
Can you outline why for a *new* implementation you would use SHA-1 over more recent iterations, e.g. SHA-2/3 family?

rgerhards:
Can you please describe a bit how this should be used. I wonder if a
property is the best solution versus a function. This would also address
different hashes.

Thx Rainer

Sent from phone, thus brief.
Am 31.05.2015 20:17 schrieb "Bruno Coudoin" <notifications@github.com>:

bdoin:
Yes true, supporting several hashes with a function would be better. Do you want me to propose a patch using a function?

rgerhards:
This sounds good, but I realize we must be careful not to introduce any hard dependencies. I've seen you unconditionally include openssl with the first patch. We can't do that, as this would drag in the dependency into any system rsyslog is installed on.

So we either must create a new interface for loadable functions (probably off the board) or hide the dependency inside a loadable module, like we did with the regexp functionality. The latter seems the way to go here.

rgerhards:
I am closing this PR as I cannot merge it as-is for reasons given in my previous posting. Sorry for that.

PR #388, opened 2015-06-02
imfile: Clean up moved files similar to deleted files.

merginator:


hakman:
Just out of curiosity, was there something wrong with this patch?

PR #391, opened 2015-06-05
imfile: remove an unused variable

theinric:


rgerhards:
merged into master-candidate

rgerhards:
I don't know yet why, but this patch causes a regression on some platforms. Issues came up on the extended testbench runs, at least for OpenSUSE and CENTOS 7. git bisect points to this patch, and if I revert it, everything works again. So I revert it for now. Maybe a side-effect of something else otherwise not seen...

theinric:
I think I've spotted the other bug:
The variable preceding the removed one in configSettings_s is int maxLinesAtOnce.
In the legacy cfg, it is parsed with eCmdHdlrSize, which is int64.
I've rebased the branch to fix the type.
DISCLAIMER: I haven't reproduced the crash and can't verify the effects of this change.

PS: The type in the v6 cfg is also an int, but there it is correct. The issue is that the types differ.

rgerhards:
I tried to merge this one again. It looks like the variable is being used in the mean time:

http://build.rsyslog.com:8010/builders/rsyslogcentos7rsyslog/builds/0/steps/shell_2/logs/stdio

Could you please check.

rgerhards:
Well, it looks like the error line is a reference that's not really needed. I will try to remove it and report back.

theinric:
Agreed; that line was added in the revert commit and is just resetting the value.
Thanks for merging.

PR #398, opened 2015-06-08
typo fix: paramter → parameter

mbiebl:
Bug-Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=772490


rgerhards:

PR #410, opened 2015-06-23
rainerscript: function random(max)

janmejay:
- it generates a warning when given a range larger than it can support (we can support it to some extent by scaling, but beyond a measure it'll become sparse with scaling-by-multiplication model, so I decided to not scale it, instead print a warning).
- some code pretty-fication


rgerhards:
merged to master-candidate

PR #411, opened 2015-06-23
bugfix: plug a memleak in imjournal

theinric:
The leak was tiny and shutdown-time only.


rgerhards:

PR #412, opened 2015-06-23
imjournal: don't sanitize empty messages

theinric:
It is an error to pass an empty message to parser.SanitizeMsg() and doing so could crash the daemon. Besides, there's no point in doing so in the first place.


rgerhards:
merged into master-candidate, will migrate from there

PR #413, opened 2015-06-25
Harmonize resetConfigVariables values and defaults

theinric:
Default values for several global and main message queue parameters
were defined in two places which, over time, became inconsistent.
This merges the code used in resetConfigVariables and the rsconf_t
constructor. In cases where values for the same parameter differed,
the value with a more recent commit timestamp was used.

### DIFF
--- a/runtime/rsconf.c
+++ b/runtime/rsconf.c
@@ -122,9 +122,10 @@ static struct cnfparamblk parserpblk =
 /* forward-definitions */
 void cnfDoCfsysline(char *ln);
 
-/* Standard-Constructor
- */
-BEGINobjConstruct(rsconf) /* be sure to specify the object type also in END macro! */
+void cnfSetDefaults(rsconf_t *pThis)
+{
+	pThis->globals.bAbortOnUncleanConfig = 0;
+	pThis->globals.bReduceRepeatMsgs = 0;
 	pThis->globals.bDebugPrintTemplateList = 1;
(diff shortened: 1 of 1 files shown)


rgerhards:
I need to check this in more detail, but I think this was intentional. The reason behind it was that legacy format is, well, legacy and we do not want to change behaviour. If we change defaults, old legacy stuff begins to behave differently. Remember that we do not want people to use that kind of legacy stuff (queue sizes and such) to use when writing new configs. But as I said, I need to check more closely.

theinric:
See this commit for example:
b7dc779ba change main/ruleset queue defaults to be more enterprise-like

It changes the default queue parameters, but only in the constructor. This smells very much like a bug.

> The reason behind it was that legacy format is, well, legacy and we do not want to change behaviour. If we change defaults, old legacy stuff begins to behave differently.

The values in the constructor affect everybody no matter the config version - so a change took place. They were changed, but the ones (at least some) in resetConfigVariables were not.

Consider this thought experiment:
When I put resetConfigVariables at the very top of the configuration, nothing should happen. All the settings should stay the same as if it weren't there, right?
But nobody does this. It is most probably used somewhere in the middle of the configuration. And after that commit above, those configurations suddenly ended up with new parameters for the part before resetConfigVariables and the old ones after it. This can't be correct. :]

rgerhards:
That was convincing ;) merged into master-candidate for extended testbench runs

Issue #419, opened 2015-07-02
Segfault when using startmsg.regex if first log line doesn't match


hakman:
Imfile in multiline mode expects the first line of the log file to match the pattern from startmsg.regex. When this doesn't happen, Rsyslog crashes:
```
Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffefcd6700 (LWP 42700)]
rsCStrAppendStrWithLen (pThis=0x0, psz=0x7fffe40034f0 "Monitor Agent Path: /opt/monitor/lib/monitor-kafka.jar.\r", iStrLen=72) at stringbuf.c:277
277		if(pThis->iStrLen + iStrLen > pThis->iBufSize) {
(gdb) bt
#0  rsCStrAppendStrWithLen (pThis=0x0, psz=0x7fffe40034f0 "Monitor Agent Path: /opt/monitor/lib/monitor-kafka.jar.\r", iStrLen=72) at stringbuf.c:277
#1  0x000000000041e9ac in cstrAppendCStr (pThis=<optimized out>, pstrAppend=<optimized out>) at stringbuf.c:306
#2  0x000000000042c4a9 in strmReadMultiLine (pThis=0x7fffe40021d0, ppCStr=ppCStr@entry=0x7fffefcd36a0, preg=preg@entry=0x6aab08, bEscapeLF=1 '\001') at stream.c:861
#3  0x00007ffff5748da5 in pollFile (pLstn=0x6aaa90, pbHadFileData=0x0) at imfile.c:576
#4  0x00007ffff5749d98 in in_setupFileWatchStatic (pLstn=0x6aaa90) at imfile.c:1510
#5  in_setupInitialWatches () at imfile.c:1526
#6  do_inotify () at imfile.c:1721
#7  0x00007ffff5749ee7 in runInput (pThrd=<optimized out>) at imfile.c:1769
#8  0x000000000043d0d0 in thrdStarter (arg=0x6df450) at ../threads.c:212
#9  0x00007ffff79ab182 in start_thread (arg=0x7fffefcd6700) at pthread_create.c:312
#10 0x00007ffff6cb847d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111
(gdb)
```

rgerhards:
Have you tried this with 8.11.0? I think it contains a fix for it (but I may be wrong).

hakman:
This was tested with 8.11.0. I was also hoping that this version will solve most of my problems.

[...]

PR #421, opened 2015-07-02
Fix segfault when using startmsg.regex if first log line doesn't match

hakman:
Fix for issue #419 
Rsyslog crashes because the first log line is not matched by startmsg.regex and tries to add it to an uninitialised string.


hakman:
Will this be merged for the next release or is some other solution preferred?

rgerhards:
Sorry for the delay, have been busy. I now have it ready for final merge, but noticed that the file affected is still under LGPLv3+. We are moving towards ASL 2.0. Are you willing to release *your contribution* and ASL 2.0? Please let me know.

hakman:
ASL 2.0 is perfect. Thanks!

otisg:
+1 for this.  It just bit us (rsyslog crashed).

PR #423, opened 2015-07-04
fix memory leak in strmCloseFile

schiele:
Here we need to free pThis->pszCurrFName in any case, even if
pThis->bDeleteOnClose is not true.  Otherwise we leak memory if
this stream gets reopened later, for instance after closing by
doSizeLimitProcessing.


rgerhards:
Hi, thanks for the patch. The file in question is still under LGPLv3. We are trying to move towards ASL 2.0. Are you willing to contribute your patch under ASL 2.0?

schiele:
ASL 2.0 is definitely fine for me.

rgerhards:
merged into master-candidate, will migrate to master

PR #434, opened 2015-07-09
Fix race-condition detection in path-creation code

theinric:
The affected code is used to detect a race condition in between
testing for the existence of a directory and creating it if it didn't
exist.  The variable tracking the number of attempts wasn't reset for
subsequent elements in the path, thus limiting the number of
reattempts to one per the whole path, instead of one per each path
element.
This solution was provided by Martin Poole.


rgerhards:
merged into master-candidate

Issue #390, opened 2015-06-05
DoDie written to queue data file


adruch:
The message "DoDie called" is being printed inside the data file for the queue when rsyslog is stopped when running DebugLevel=2 (DEBUG_FULL).

In the function rsyslogdDoDie(), this die message is being written to file descriptor 1. In the debug log, I also noticed that this queue data file has a file descriptor of 1. I believe this extra text could be the reason why I see objDeserialize errors when I start rsyslog again.

rgerhards:
this seems to have been fixed long ago.

PR #457, opened 2015-07-24
Debugoutput: Replaced write calls(Only in debugmode 2)  with printf

alorbach:
In rsyslogdDoDie, the Write function was used on static FD 1 (should
have been STDOUT in normal cases). But appearently it could also
lead to FD of queue files. Replaced the write calls with printf
which ends in STDOUT by default.

closes https://github.com/rsyslog/rsyslog/issues/390


rgerhards:
Does not solve the issue, most probably fd 1 (maybe 2) should not be closed if Debug != DEBUG_OFF.

PR #461, opened 2015-07-28
mmnormalize: liblognorm error messages are now reported

rgerhards:
This is possible due to a new API in liblognorm 1.1.2;
Note that the amount of error messages depends on the version of
liblognorm used. Consequently, liblognorm 1.1.2 is now required
to run mmnormalize.

**Note: merge only after buildbot tests indicate "OK"**


rgerhards:
merged to master-candidate

PR #472, opened 2015-08-10
fixes 2 bugs in omkafka

janmejay:
This fixes a bug which allowed omkafka to get into a state where it wouldn't accept any new messages for publishing.

It also fixes a bug which whould manifest only when using dyntopic and when producer is in i-won't-accept-new-messages state, and caller tries to produce a new message. The dyn-topic-lock wouldn't be unlocked (because of premature fail-jump).

If you know librdkafka at a high level and are interested in details, here is a more detailed explaination:
So 'rk' tries to bound number of objects in flight by maintaining a counter, the max-value of which defaults to 100k. When a new message comes in, it rejects it if it already is at max-messages-in-flight. Now this counter is decremented (and in-flight-msg memory is freed) in callback-poll (which picks things from reply-queue, rd_kafka_op_t for delivery-ack op). If we don't poll when we are already at max-in-flight-messages and there is no poll in flight, it'll basically drain all 100k in-flight message and then start idling (because rk->rk_u.producer->msg_cnt still thinks it has 100k messages in flight, and delivery-acks are yet to be handled, but of-course no one handles them).

### DIFF
--- a/plugins/omkafka/omkafka.c
+++ b/plugins/omkafka/omkafka.c
@@ -788,6 +788,7 @@ writeKafka(instanceData *pData, uchar *msg, uchar *topic)
 	const int partition = getPartition(pData);
 	rd_kafka_topic_t *rkt = NULL;
 	pthread_rwlock_t *dynTopicLock = NULL;
+	int msg_enqueue_status = 0;
 
 	DBGPRINTF("omkafka: trying to send: key:'%s', msg:'%s'\n", pData->key, msg);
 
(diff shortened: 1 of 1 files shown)


oldmantaiter:
@janmejay 
Just dropping a quick comment here in regards to the reason I added dyntopic. It definitely was purely out of convenience, there are scenarios where there might be 50+ different topics coming into a ruleset (as we have it configured in some places) so using an if-else would have become cumbersome (plus not having to reconfigure & reload was a plus).

I've been mainly using a backported version of omkafka for 7.4.7, pre workerInstance, and haven't run into the issues described here and in #355 (although I was worrying about them), I'm glad more people are using this plugin, and finding some critical bugs - which is great because we will be really ramping up the number of messages through omkafka shortly in our environment (currently ~5K msgs/sec on 10 or so topics moving to ~50-60K msgs/sec).

Out of curiosity, how many messages/sec were you pushing through to kafka to see both this and the other issues that you've corrected?

janmejay:
Well, as of now its embarrassingly low, about ~5k msgs/sec, but that is
because we haven't yet taken this generation of the service GA yet.

Once it is GA it'll be ramped-up to well over 20M/min(333K/s) in a few
days. Obviously it is not a single machine, its a cluster. We aren't using
dyn-topics though. But when I fix bugs, I like to reason through all
scenarios, it keeps the logic/line-of-thought clean (especially so with
concurrency bugs). Eventually we plan to pump well over 60M/min.

On Tue, Aug 18, 2015 at 9:26 AM, Tait Clarridge <notifications@github.com>
wrote:

> @janmejay <https://github.com/janmejay>
> Just dropping a quick comment here in regards to the reason I added
> dyntopic. It definitely was purely out of convenience, there are scenarios
> where there might be 50+ different topics coming into a ruleset (as we have
> it configured in some places) so using an if-else would have become
> cumbersome (plus not having to reconfigure & reload was a plus).
>
> I've been mainly using a backported version of omkafka for 7.4.7, pre
> workerInstance, and haven't run into the issues described here and in #355
> <https://github.com/rsyslog/rsyslog/pull/355> (although I was worrying
> about them), I'm glad more people are using this plugin, and finding some
> critical bugs - which is great because we will be really ramping up the
> number of messages through omkafka shortly in our environment (currently
> ~5K msgs/sec on 10 or so topics moving to ~50-60K msgs/sec).
>
> Out of curiosity, how many messages/sec were you pushing through to kafka
> to see both this and the other issues that you've corrected?
>
> —
> <https://github.com/rsyslog/rsyslog/pull/472#issuecomment-132064276>.
>



-- 
Regards,
Janmejay
http://codehunk.wordpress.com

oldmantaiter:
Wow, that's quite a bit of traffic planned. Thanks for fixing up dyn-topic to make it cleaner/safer, I definitely have a lot to learn from these two pull requests.

janmejay:
Pleasure.

On Tue, Aug 18, 2015 at 5:57 PM, Tait Clarridge <notifications@github.com>
wrote:

> Wow, that's quite a bit of traffic planned. Thanks for fixing up dyn-topic
> to make it cleaner/safer, I definitely have a lot to learn from these two
> pull requests.
>
> —
> <https://github.com/rsyslog/rsyslog/pull/472#issuecomment-132192484>.
>



-- 
Regards,
Janmejay
http://codehunk.wordpress.com

PR #474, opened 2015-08-10
Added stream ability to omuxsock

perlei:
Needed stream support over unix-domain-sockets so i though i could share my solution. 

Added a new configuration option 

$OMUxSockType stream   

if this is stream, rsyslog will setup a stream based unixdomain socket. if it is something else then datagrams will  be used (backwards compatible). 

so a sample config could be like: 

$DebugFile /var/log/rsyslog
$DebugLevel 2
module(load="imtcp")
module(load="omuxsock")
input(port="514" type="imtcp" )
$OMUxSockSocket /tmp/socksample
$OMUxSockType stream
*.* :omuxsock:

### DIFF
--- a/plugins/omuxsock/omuxsock.c
+++ b/plugins/omuxsock/omuxsock.c
@@ -59,6 +59,8 @@ DEFobjCurrIf(glbl)
 typedef struct _instanceData {
 	permittedPeers_t *pPermPeers;
 	uchar *sockName;
+	uchar *sockType;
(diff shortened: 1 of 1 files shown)


rgerhards:
would you be willing to change this patch to support the new config system. We do no longer add legacy config statements (due to their complexity) for new features. It would also be great if you could add doc to https://github.com/rsyslog/rsyslog-doc

PR #476, opened 2015-08-11
Fix detection of the GnuTLS package

radosroka:
This commit fixes a number of issues that cropped up from these changes:
23c5b3e replace deprecated GnuTLS functions with newer ones if available
88f0651 bugfix: ommysql did not work when gnutls was enabled
ce560b6 build system: Failure action in AC_CHECK_LIB check for

The solution is based on the assumption that the call to AC_CHECK_LIB()
was added purely to satisfy the subsequent calls to AC_CHECK_FUNCS().
23c5b3e had the unintentional result of libgnutls being added to LIBS
and linked to undesired components.
The next commit removed libgnutls from LIBS which made the following
calls to AC_CHECK_FUNCS() nonfunctional.
Another commit made the processing not stop upon detecting a missing
component which, again, makes the calls to AC_CHECK_FUNCS() malfunction.

The solution is to get rid of AC_CHECK_LIB() and temporarily override
LIBS for use by AC_CHECK_FUNCS().

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -763,18 +763,13 @@ AC_ARG_ENABLE(gnutls,
 if test "x$enable_gnutls" = "xyes"; then
 	PKG_CHECK_MODULES(GNUTLS, gnutls >= 1.4.0)
 	AC_DEFINE([ENABLE_GNUTLS], [1], [Indicator that GnuTLS is present])
-	AC_CHECK_LIB(
-	    [gnutls],
-	    [gnutls_global_init],
-	    [
-	     AC_DEFINE(HAVE_LIB_GNUTLS, 1, [gnutls is available])
-	    ],
-	    [AC_MSG_WARN([gnutls_global_init function missing or not detected])],
-	    []
-	)
+	save_libs=$LIBS
+	LIBS="$LIBS $GNUTLS_LIBS"
 	AC_CHECK_FUNCS(gnutls_certificate_set_retrieve_function,,)
 	AC_CHECK_FUNCS(gnutls_certificate_type_set_priority,,)
+	LIBS=$save_libs
 fi
+
 AM_CONDITIONAL(ENABLE_GNUTLS, test x$enable_gnutls = xyes)
 
 # libgcrypt support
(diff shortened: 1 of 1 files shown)


alorbach:
I am not sure if this is a good Idea, this check was recently added because of a bugfix, see this commit: 
https://github.com/radosroka/rsyslog/commit/88f06516b53a5248839d9520cebe96f48737a16a#diff-67e997bcfdac55191033d57a16d1408a

radosroka:
I started doing this research because rsyslog-8.12 was not able to compile with the newest gnutls which uses gnutls_certificate_set_retrieve_function instead of client-server modification.

I know about that commit, I mentioned it above. It's absolutely unnecessary to check gnutls so many times. I simplify that section because PKG_CHECK_MODULES fails if gnutls is not present or in correct version. So another check is not necessary and it was there because AC_CHECK_LIB() without set if-found-action adds gnutls to Libs... http://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/Libraries.html
I reduced redundant checks and added gnutls manually and I removed gnutls from Libs in the end.

alorbach:
I though the check should only generates a warning (AC_MSG_WARN), not a failure? 
Anyway I remember that this bugfix resulted from a problem on Centos. I will leave this open for Rainer to decide.

theinric:
> I though the check should only generates a warning (AC_MSG_WARN), not a failure?

PKG_CHECK_MODULES() aborts,  by default (so AC_CHECK_LIB() is never reached).
The call to AC_CHECK_LIB() contains AC_MSG_WARN(), which is, as Rado writes above, bogus:
if a check for a library fails, you can't go searching for functions from that library in AC_CHECK_FUNCS().

> Anyway I remember that this bugfix resulted from a problem on Centos. I will leave this open for Rainer to decide.

You're right, lets not rush this and wait for Rainer.

rgerhards:
Thanks for the help with this issue. I admit I constantly struggle with the finer details of autotools. I am sure that the patch will work, but please bear with me for a short while. I had such a bad time with the original commit, I would like to make sure we don't get into that MariaDB hell on RHEL7 again, so I'll have the patch run through a full test suite, even though it sounds absolutely logical it is right --- and you guys know much better what you do in this regard ;) 

As a side-note, it would be great if you at RH could add a test of ommysql on RHEL7 as this would catch issues like the original ones rather quickly.

PR #482, opened 2015-08-14
Add fixed-width option

jite:
Pads a string with spaces up to toChar field if string is shorter, if
string is equal then nothing is done and if string is longer then it
gets truncated.

From mailing list patch proposal.

### DIFF
--- a/runtime/msg.c
+++ b/runtime/msg.c
@@ -3573,7 +3573,7 @@ uchar *MsgGetProp(msg_t *__restrict__ const pMsg, struct templateEntry *__restri
 			if(iTo > 0)
 				--iTo;
 		}
-		if(iFrom == 0 && iTo >=  bufLen) { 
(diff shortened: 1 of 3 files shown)


rgerhards:
I think this is missing code for current-style template definitions. Have not thouroughly checked, though. Will do when I have a bit more time. If it is not present, we need this to be able to merge.

jite:
Yes you're right, it does miss that. I'll update with a new patch.

jite:
@rgerhards New patch proposal uploaded

rgerhards:
merged into master-candidate, will migrate from there to master (after buildbot tests)

rgerhards:
It would be great if you could also update the doc at https://github.com/rsyslog/rsyslog-doc so that others can actually use the feature.

jite:
@rgerhards Absolutely. Should I write that it's available since rsyslog 8.13.0? Does it make it to that?

rgerhards:
yup, it's already in the tree that becomes 8.13.0 :-)

PR #485, opened 2015-08-17
Update rscryutil.c

wizard1024:
Added opportunity to decrypt not closed log file (without appropriate record END: xxx in <name>.encinfo file)

### DIFF
--- a/tools/rscryutil.c
+++ b/tools/rscryutil.c
@@ -277,12 +277,22 @@ doDecrypt(FILE *logfp, FILE *eifp, FILE *outfp)
 	off64_t blkEnd;
 	off64_t currOffs = 0;
 	int r;
+	int fd;
+        struct stat buf;
 
 	while(1) {
 		/* process block */
 		if(initCrypt(eifp) != 0)
 			goto done;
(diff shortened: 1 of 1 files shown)


alorbach:
I get this error when trying to make rscryutil: 
--
rscryutil.c: In function ‘doDecrypt’:
rscryutil.c:290:25: error: label ‘end’ used but not defined
                         goto end;
--
There is no end lable, perhaps it should be goto done; ?

wizard1024:
Thanks. Updated.

alorbach:
Have you tested the patch? 

I have tried it using my sample config, and rscryutil isn't able to decrypt until rsyslog is closed. 
At least in my configuration.

wizard1024:
Yes. Tested on Centos. rsyslogd installed from repo. Typical usage:
http://pastebin.com/tL2UiuYB
Usage of ./rscryutil - usage of modified tool.

alorbach:
Nevermind, was my bad. I was accidentally testing with the installed rscryutil, not the compiled one in my git directory. Works fine, I will merge the patch soon.

Issue #488, opened 2015-08-18
Rotated file is never closed (fd leakage)


ymattw:
When source log file is rotated (renamed and recreated), rsyslog keep the original file open (with new name) and opens the newly created file, it keep doing this whenever the file is rotated but **never** closes old file.  This will cause fd number to reach the max open files limit soon.

Desired behavior is to close the original file once content are sent completely (behave like `heka`), or keep it open for a while (configurable, behave like `fluentd`).

My test environment was CentOS 6.5 with `rsyslog-8.12.0-1.el6.x86_64` installed, I used `imfile` module and forward to remote tcp server.  

```bash
# cat /etc/rsyslog.d/00-file-input.conf
module(load="imfile" mode="inotify")

input(type="imfile"
      File="/dev/shm/rsyslog.tmp"
      Tag="tag1")

# cat /etc/rsyslog.d/30-tcp-output.conf
*.* action(type="omfwd" Target="192.168.172.29" Port="6000" Protocol="tcp" )

# ls -l /proc/`pgrep rsyslogd`/fd
total 0
lrwx------ 1 root root 64 Aug 18 16:33 0 -> socket:[958524]
lrwx------ 1 root root 64 Aug 18 16:33 1 -> /var/log/secure
lr-x------ 1 root root 64 Aug 18 16:48 10 -> /dev/shm/rsyslog.tmp.6
lr-x------ 1 root root 64 Aug 18 16:50 11 -> /dev/shm/rsyslog.tmp.7
lr-x------ 1 root root 64 Aug 18 16:54 12 -> /dev/shm/rsyslog.tmp.8
lr-x------ 1 root root 64 Aug 18 16:54 13 -> /dev/shm/rsyslog.tmp.9
lr-x------ 1 root root 64 Aug 18 16:54 14 -> /dev/shm/rsyslog.tmp
l-wx------ 1 root root 64 Aug 18 16:33 2 -> /var/log/cron
lr-x------ 1 root root 64 Aug 18 16:33 3 -> /proc/kmsg
lr-x------ 1 root root 64 Aug 18 16:34 4 -> /dev/shm/rsyslog.tmp.2
lr-x------ 1 root root 64 Aug 18 16:33 5 -> inotify
lr-x------ 1 root root 64 Aug 18 16:42 6 -> /dev/shm/rsyslog.tmp.3
lr-x------ 1 root root 64 Aug 18 16:46 7 -> /dev/shm/rsyslog.tmp.4
[...]

PR #490, opened 2015-08-19
imfile: fixed an issue not closing rotated files (single files monizt…

alorbach:
…oring).

The previous fix from commit https://github.com/alorbach/rsyslog-1/commit/ad0805703913081b022a5d101de470f2c4c45b00
did solve an issue deleting the listener object from the configuration
table. However this resulted in not closing file handles when files
were rotated. The new fix uses another approach, the listener object
is duplicated before started (like done for dynamic files). It is also
deleted when the file is rotated, which also automatically closes the
file handle.

closes https://github.com/rsyslog/rsyslog/issues/488

### DIFF
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -1539,8 +1539,14 @@ in_setupFileWatchStatic(lstn_t *const __restrict__ pLstn)
 			}
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


alorbach:
No handling will be done as before. It reverts a change made in 8.12.0 and fixes the problem we had there in another way.

ymattw:
Thanks for your quick fix.  I will try to verify asap.

Issue #489, opened 2015-08-18
mmnormalize_{variable,tokenized,regex}.sh tests are failing


Whissi:
The
- mmnormalize_variable.sh
- mmnormalize_tokenized.sh
- mmnormalize_regex.sh

tests are failing:

```
FAIL: mmnormalize_variable.sh
=============================

===============================================================================
[mmnormalize_variable.sh]: basic test for mmnormalize module variable-support
rsyslogd started with pid  25366
starting run 1
Sending file './testsuites/date_time_msg' 1 times.
^M00000001 kb sent
runtime: 0.000
End of tcpflood Run
doing shutdown
imdiag[13500]: mainqueue empty
wait on shutdown
content-check failed
FAIL mmnormalize_variable.sh (exit status: 1)

FAIL: mmnormalize_tokenized.sh
==============================

===============================================================================
[mmnormalize_tokenized.sh]: test for mmnormalize tokenized field_type
rsyslogd started with pid  25409
starting run 1
Sending file './testsuites/tokenized_input' 1 times.
^M00000001 kb sent
runtime: 0.000
End of tcpflood Run
doing shutdown
imdiag[13500]: mainqueue empty
wait on shutdown
content-check failed
FAIL mmnormalize_tokenized.sh (exit status: 1)

FAIL: mmnormalize_regex.sh
==========================

===============================================================================
[mmnormalize_regex.sh]: test for mmnormalize regex field_type
rsyslogd started with pid  25566
starting run 1
Sending file './testsuites/regex_input' 1 times.
^M00000001 kb sent
runtime: 0.000
End of tcpflood Run
doing shutdown
imdiag[13500]: mainqueue empty
wait on shutdown
content-check failed
FAIL mmnormalize_regex.sh (exit status: 1)
```

Confirmed against rsyslog-8.12.0 release tarball and git at 1e26405031390003853f6c11e7221d534dc3a0dd.

Whissi:
OK, this is a problem which only happens with *liblognorm-1.1.2*. It is working fine with *liblognorm-1.1.1*.

Do we need to mask *liblognorm-1.1.2* for rsyslog or does the test requires an update to work with *liblognorm-1.1.2*?

alorbach:
I will look into this.

alorbach:
Are you certain these tests worked with liblognorm-1.1.1? 
I also posted on the mailinglist that at least one of the configs does not work.

Whissi:
100% sure. I just verified again.

[...]

PR #492, opened 2015-08-19
Enable mmnormalize for test coverage within Travis

Whissi:
Travis currently doesn't enable *mmnormalize* so we don't know from our CI that there's a problem with *mmnormalize* (see #489). This PR will enable *mmnormalize*.

Because there **is** currently a problem, Travis will fail. But nevertheless please merge, so we can keep track when this is fixed.

Link-ML: http://lists.adiscon.net/pipermail/rsyslog/2015-August/040997.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,12 +7,13 @@ before_install:
 
 install:
   - travis_retry sudo apt-get install -qq build-essential automake pkg-config libtool autoconf autotools-dev gdb valgrind libdbi-dev libsnmp-dev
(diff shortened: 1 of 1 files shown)


alorbach:
I will leave this open for rainer to verify. I am not sure if the tests were disabled on purpose.

Whissi:
Yep, I also verified before I created the PR. Looks like **I** didn't enable *mmnormalize* when I added initial Travis support (https://github.com/rsyslog/rsyslog/commit/e0d1f430bfe896757752e4d28835cb4b14708fa1) and nobody else did, too.

So it was never deactivated... just never activated :)

rgerhards:
I won't merge into master[-candidate] until it is fixed, because it makes no sense to break other pull requests when we know the reason. However, as you see, im am working on the topic.

PR #493, opened 2015-08-20
Tcp client side keep alives

tinselcity:
Add config options for client side keep-alives for tcp connections.

### DIFF
--- a/tools/omfwd.c
+++ b/tools/omfwd.c
@@ -86,6 +86,11 @@ typedef struct _instanceData {
 	char *port;
 	int protocol;
 	int iRebindInterval;	/* rebind interval */
+        sbool bKeepAlive;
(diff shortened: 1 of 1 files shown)


rgerhards:
hi, thanks for the patch. We do no longer add legacy statements for new features. I can remove them after the merge, or you can do that if you like. Also, it would be good if you could create a PR for the doc update. Doc is kept here: https://github.com/rsyslog/rsyslog-doc

tinselcity:
I'm fine if you want to remove the legacy statements after the merge.
I'll make a PR for doc's with the client side keep-alive configuration parameters.
Thanks

tinselcity:
Added PR for docs:
https://github.com/rsyslog/rsyslog-doc/pull/171

PR #502, opened 2015-08-29
Make it possible to customize librdkafka location

ymattw:
For example, build librdkafka from source and static link.

I've verified the change on CentOS 6.5.  Is there any special reason this was turned off?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1443,7 +1443,7 @@ AC_ARG_ENABLE(omkafka,
          esac],
         [enable_omkafka=no]
 )
-if test "x$enable_omkafka"; then
+if test "x$enable_omkafka" = "xyes"; then
 	PKG_CHECK_MODULES(LIBRDKAFKA, librdkafka)
 	AC_CHECK_HEADERS([librdkafka/rdkafka.h])
 fi
--- a/configure.ac
+++ b/configure.ac
@@ -1444,7 +1444,7 @@ AC_ARG_ENABLE(omkafka,
         [enable_omkafka=no]
 )
 if test "x$enable_omkafka"; then
-	#PKG_CHECK_MODULES(LIBRDKAFKA, librdkafka)
+	PKG_CHECK_MODULES(LIBRDKAFKA, librdkafka)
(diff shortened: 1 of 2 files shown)


rgerhards:
IIRC it did not work with the version of librdkafka that were at hand. Maybe only on some platforms. Need to be careful with the merge, especially as we do not have kafka tests inside the testbench. Can someone confirm this works?

ymattw:
FYI I used latest librdkafka (0.8.6): https://github.com/edenhill/librdkafka/archive/0.8.6.tar.gz for the static linking.  The extra rsyslog configure options used were:

```sh
LIBRDKAFKA_CFLAGS=-I/tmp/rsyslog.deps/include \
LIBRDKAFKA_LIBS=/tmp/rsyslog.deps/lib/librdkafka.a
```

The line `PKG_CHECK_MODULES(LIBRDKAFKA, librdkafka)` in configure.ac was commented out at the very beginning when it was added in commit c26d5e1 made yourself :), not sure if there was any special reason.

Actually I static linked everything to meet our special needs for binary distribution, the output executable `rsyslogd` works fine (so far) on CentOS 6.5.

Let me know if anything else I can help on this.  Thanks!

rgerhards:
I need to check if this works with the default install of mayor distros. As I am thight on time, I probably won't make it for 8.13, but I think it's not an extra-urgent feature. OK with you?

ymattw:
No it's not an urgent feature, I maintain this patch in my own fork right now, so no hurry.  Thank you!

rgerhards:
Sorry, I have to undo aea09800643343ab8b6aa205b0f10a4be676643b because omkafka now no longer builds on a variety of platforms (Ubuntu 14.04 LTS and Fedora for example). You may want to re-enable it via a configure option, looking forward to a PR.

rgerhards:
undo commit: 8a78d0389e733ec17fe9b33895ef6d333a591953

ymattw:
That's ok, my bad wasn't able to test on more platforms, will try to look
into it later.

Issue #500, opened 2015-08-26
Infinite amplified loop in action code


hakman:
An action can go into an infinite loop when it receives an unexpected return code from the output module, during a transaction. What is worse, log lines that fail to be outputted, are duplicated on each new failure.

I found this issue in the elasticsearch module and the result can be easily seen in this [log file](https://gist.github.com/hakman/50d9afa8975ca611dd72#file-rsylog-log).
Reproducing this is easy. Just start Rsyslog using this [config file](https://gist.github.com/hakman/50d9afa8975ca611dd72#file-rsyslog-conf) and send a plain text log line over UDP:
```echo 'my-host my-process:hello world!' | nc -4u -w0 localhost 514```
On a fast connection, this gets out of control very fast.

I see 2 issues here:
1. omelasticsearch doesn't handle HTTP errors (400+ status codes) - I provided #505 for this
2. action code doesn't handle well unknown error codes from transactions.

otisg:
Ouch.  Nice catch!
Which version of rsyslog?

hakman:
I did my debugging by building the master branch, so latest version at least should be affected.

hakman:
This comment is a summary of my view of the issue in action.c code.

In some failure cases, the return value for a transaction in omelasticsearch can be the value of endTransaction->curlPost->checkResult, which can be anything, in this case RS_RET_ERR (but can be any other error code):
https://github.com/rsyslog/rsyslog/blob/v8.12.0/plugins/omelasticsearch/omelasticsearch.c#L981

actionTryCommit will return the value from endTransaction in the output module in case of "permanent failure", in this case RS_RET_ERR:
https://github.com/rsyslog/rsyslog/blob/v8.12.0/action.c#L1198

At this point actionCommit is already in a loop and, unless it receives an expected iRet or action state, it will loop here forever:
https://github.com/rsyslog/rsyslog/blob/v8.12.0/action.c#L1243

IMHO, here we should call actionCommitted() before FINALIZE, to force cleanup (or some new function that also handles the failed counters):
https://github.com/rsyslog/rsyslog/blob/master/action.c#L1201

Maybe something similar would be needed here:
https://github.com/rsyslog/rsyslog/blob/v8.12.0/action.c#L1057

otisg:
We just got our Elasticsearch hammered by this rsyslog/omelasticsearch bug last night.

rgerhards:
[...]

PR #505, opened 2015-09-02
omelasticsearch: Fix CURL ignoring HTTP errors

hakman:
By default, CURL doesn't check if the HTTP request to elasticsearch fails. This leads to extra unnecessary processing and can trigger #500.


rgerhards:

hakman:
This was my initial investigation:
https://github.com/rsyslog/rsyslog/issues/500

If it is not clear, I can try to look at it again.

Issue #506, opened 2015-09-02
tcpflood may fail with ***buffer overflow detected *** on some architectures


karibou:
Running the testbench on ubuntu i386 & powerpc leads to FTBS for 8.12.0. This is caused by some test using tcpflood that returns a *** buffer overflow error *** from glibc and core dumps with the following trace : 

TEST: [wr_large_sync.sh]: test for file writing for large message sets
$OMFileAsyncWriting off
rsyslogd started with pid  18781
starting run 1
Sending 4000 messages.
*** buffer overflow detected ***: ./tcpflood terminated
======= Backtrace: =========
/lib/i386-linux-gnu/libc.so.6(+0x6a199)[0xf74e0199]
/lib/i386-linux-gnu/libc.so.6(__fortify_fail+0x68)[0xf7570cb8]
/lib/i386-linux-gnu/libc.so.6(+0xf8e3a)[0xf756ee3a]
./tcpflood[0x804ae0b]
./tcpflood[0x804af15]
/lib/i386-linux-gnu/libpthread.so.0(+0x61aa)[0xf777b1aa]
/lib/i386-linux-gnu/libc.so.6(clone+0x5e)[0xf755ffde]
======= Memory map: ========
08048000-0804d000 r-xp 00000000 fd:01 267581                             /«PKGBUILDDIR»/tests/tcpflood
0804d000-0804e000 r--p 00004000 fd:01 267581                             /«PKGBUILDDIR»/tests/tcpflood
0804e000-0804f000 rw-p 00005000 fd:01 267581                             /«PKGBUILDDIR»/tests/tcpflood
08fa2000-08fc3000 rw-p 00000000 00:00 0                                  [heap]
f6900000-f6921000 rw-p 00000000 00:00 0 
f6921000-f6a00000 ---p 00000000 00:00 0 
f6ab1000-f6acd000 r-xp 00000000 fd:01 135162                             /lib/i386-linux-gnu/libgcc_s.so.1
f6acd000-f6ace000 r--p 0001b000 fd:01 135162                             /lib/i386-linux-gnu/libgcc_s.so.1
f6ace000-f6acf000 rw-p 0001c000 fd:01 135162                             /lib/i386-linux-gnu/libgcc_s.so.1
f6ad4000-f6ad5000 ---p 00000000 00:00 0 
f6ad5000-f72d7000 rw-p 00000000 00:00 0                                  [stack:18793]
[...]

PR #507, opened 2015-09-02
testbench: Fix potential buffer overflow detection

karibou:
On some architectures, calculation of edLen may lead to a
negative value when the value returned by rand() is less
than extraDataLen away from RAND_MAX. The negative value
passed to memset() triggers a glibc error. Forcing the
cast of rand() to be unsigned avoid the return of a negative
value.

Closes: #506

Signed-off-by: Louis Bouchard <louis.bouchard@canonical.com>

### DIFF
--- a/tests/tcpflood.c
+++ b/tests/tcpflood.c
@@ -398,7 +398,7 @@ genMsg(char *buf, size_t maxBuf, int *pLenBuf, struct instdata *inst)
 			}
 		} else {
 			if(bRandomizeExtraData)
-				edLen = ((long) rand() + extraDataLen) % extraDataLen + 1;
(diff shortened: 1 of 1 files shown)


mbiebl:
I did occasionally get buffer overflows in tcpflood, like in [1]. Haven't been able to reproduce it locally. So thanks for tracking it down!

[1] https://buildd.debian.org/status/fetch.php?pkg=rsyslog&arch=mipsel&ver=8.9.0-3&stamp=1431386336

karibou:
mbiebl : I forwarded my Ubuntu fix to one of your debian bug.  It was rather sporadic as it only happens when rand() + extraDataLen > RAND_MAX

PR #510, opened 2015-09-03
omtesting: fix unit labels

theinric:
An argument given in microseconds was incorrectly labeled as given in milliseconds.

### DIFF
--- a/plugins/omtesting/omtesting.c
+++ b/plugins/omtesting/omtesting.c
@@ -14,9 +14,9 @@
  *
  * CURRENT SUPPORTED COMMANDS:
  *
- * :omtesting:sleep <seconds> <milliseconds>
+ * :omtesting:sleep <seconds> <microseconds>
  *
- * Must be specified exactly as above. Keep in mind milliseconds are a millionth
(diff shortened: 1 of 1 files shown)


theinric:
Rainer, thanks for merging.
There's a [post](http://lists.adiscon.net/pipermail/rsyslog/2015-September/041110.html) on the rsyslog ML related to this PR; there are two tests using this sleep function and I don't have enough insight to determine whether they need to be fixed.

rgerhards:
Thanks for bringing this up again in https://github.com/rsyslog/rsyslog/pull/604

I guess I have overlooked the reply or not correctly understood it. I think the tests need fixing, and maybe that's the cause of a long-standing bug. Question is if we really should fix the tests or omtesting (actually make it use ms).

rgerhards:
and had another look again. These sleeps are actually used to delay processing, to make sure action processing is so slow a queue builds up. From that, it looks like indeed only the comment was wrong, but a 0.0001 second wait time is correct.

theinric:
OK, thanks for checking.

PR #516, opened 2015-09-08
Czmq bind connect

bluca:
Some context here: this is a bug, where the pub/sub sockets are not binded/connected as they should be. So a user has to prebend the endpoint string with a ">" for a sub socket or a "@" for a pub socket, otherwise it will not work.


taotetek:
@buca - thanks so much for the contribution, I'll make space to look at this and see if we can get it in the next release!

taotetek:
:thumbsup:

bluca:
No problem, thanks for merging!

PR #517, opened 2015-09-08
omczmq: implement topics for pub socket

bluca:
As the titles says, this PR enables omczmq to publish to specific topics.

There is actually a bug, where the "" topic is not set, so whatever the first string is, it will be used by the pub socket as the topic.

### DIFF
--- a/contrib/omczmq/omczmq.c
+++ b/contrib/omczmq/omczmq.c
@@ -60,6 +60,7 @@ typedef struct _instanceData {
 	uchar *tplName;
 	char *beacon;
 	int beaconport;
+	char *topicList;
 } instanceData;
 
 typedef struct wrkrInstanceData {
@@ -74,7 +75,8 @@ static struct cnfparamdescr actpdescr[] = {
 	{ "authtype", eCmdHdlrGetWord, 0 },
 	{ "clientcertpath", eCmdHdlrGetWord, 0 },
 	{ "servercertpath", eCmdHdlrGetWord, 0 },
-	{ "template", eCmdHdlrGetWord, 0 }
+	{ "template", eCmdHdlrGetWord, 0 },
+	{ "topics", eCmdHdlrGetWord, 0 }
 };
 
 static struct cnfparamblk actpblk = {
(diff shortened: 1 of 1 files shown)


taotetek:
@bluca So - the way I've been using the plugin is to construct topic strings on the output using rsyslog templates.  So, for instance, to create a topic based off of "hostname.programname" - 

```
template(name="pubsub_host_tag" type="list") {
  property(name="hostname")
  constant(value=".")
  property(name="syslogtag" position.from="1" position.to="32")
  constant(value="@cee:")
  constant(value="{")
  property(name="hostname")
  constant(value="\",\"severity\":\"")
  property(name="syslogseverity-text")
  constant(value="\",\"facility\":\"")
  property(name="syslogfacility-text")
  constant(value="\",\"syslogtag\":\"")
  constant(value="\",")
  property(name="$!all-json" position.from="2")
} 
```

This allows for the construction of dynamic topics per message based on the actual message content.  Then I can attach to the output and subscribe to specific logs from a certain host, etc.

If setting a static topic on the output solves a problem for you that this technique doesn't solve, I'm happy to look over the PR.

bluca:
@taotetek - interesting configuration, hadn't thought of that. Maybe it's worth adding it to the README :-)

The reason I implemented the topic on the pub side is mainly because, since they can be configured in the config in the sub side, as a user I expected them to be configurable on the pub side as well, and I was a bit confused when I noticed that it wasn't the case.

I don't think there's a use case that can't be covered by the configuration you provided. It would only be for simplicity, for simple use cases. Eg: if you just want to tag the message with a particular key (for example the hostname) that is fixed at config time.

taotetek:
@bluca better documentation is definitely on my list of things to do :)  I'm not opposed to this PR if having a simple way to add a static topic in the config is helpful to some people.  My only caveat would be that it doesn't impact the other use case of using a template.  

Empty string on a sub simply means "get all messages" so there's no need to send any special bytes from the publisher.  If you want to remove the extra sendm since it does have some performance impact and then test to your satisfaction, I'm happy to merge the PR.

Cheers!

bluca:
@taotetek - thanks for being open to this! I think it's nice to have both possibilities.
I've refactored as discussed. If no topics parameter is passed, behaviour is unchanged. I tested this on Debian Jessie, and using your template, no topics and having the client subscribe to the hostname works.

taotetek:
@bluca works for me!  :thumbsup:

bluca:
Perfect, thanks!

PR #518, opened 2015-09-08
imtcp: Added logerror output when MaxListeners limit is reached.

alorbach:
Added errmsg.LogError when configured listener limit is reached in input modules using tcpsrv.c.


rgerhards:
Somehow I get errors on buildbot/debian when applying this patch. Doesn't look like something is wrong, but still it fails if and only if I merge this one. Will now revert master-candidate.

rgerhards:
Note: this is **not** merged even though github tells it is (I found no way to change that back).

PR #536, opened 2015-09-28
libksi 3.4.0.1

vii5ard:
The latest libksi release has some compatibility breaking changes.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1107,8 +1107,8 @@ AC_ARG_ENABLE(gt-ksi,
         [enable_gt_ksi=no]
 )
 if test "x$enable_gt_ksi" = "xyes"; then
-	PKG_CHECK_MODULES(GT_KSI, libksi >= 3.2.2.0)
-	PKG_CHECK_MODULES(GT_KSI, libksi < 3.3.0.0)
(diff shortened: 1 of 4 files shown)


rgerhards:
we need to wait for the libksi versioning to be in place before we can merge

rgerhards:
We should also close https://github.com/rsyslog/rsyslog/issues/581 before merging this (even though it may not strictly be required for 3.4.0.1).

alorbach:
Going to take care on issue https://github.com/rsyslog/rsyslog/issues/581 after merging because I can't compile with libksi 3.4.0.x without this patch.

rgerhards:
@alorbach  I think this merge needs to be undone. As the versioning is not yet correct, we cannot build this version without getting into real big trouble. Anything that depends on this patch needs to wait until it can finally be merged.

alorbach:
Pullrequest has been reverted with this commit until the versioning issue is being solved: 
https://github.com/rsyslog/rsyslog/commit/0fe7f1aa2837ebb8515083af91032d8fbdc6c381

alorbach:
Pullrequest finally merged now.

PR #538, opened 2015-09-30
Fixed lookup-table reload bug

janmejay:
It wasn't picking up new table-length.


rgerhards:
merged into master-candidate

PR #544, opened 2015-10-06
remove unofficial lookup table code

rgerhards:
This code officially never existed and was, in its incomplete
state, added as a convenience to some users.

Background (from ML posting):
Sorry, folks, good intent always seems to find someone who turns it
into negative. I was yesterday contacted by a distro maintainer who
wants to turn this bug in the officially non-existant lookup table
feature into a CVE and insists that it is a vuln even after the
argument that the feature never oficially existed.

It looks like it was a bad idea to merge potentially useful yet
incomplete code into the main branch (and documenting it to be not
present). It looks like I need to re-think my stance on experimental
features.

Anyhow, I really don't want to support the argument that something
non-existing can be a CVE. As such, I will create a new
master-insecure branch, which will be a clone of the current master
branch. Then I'll remove the lookup table code, so that the code base
matches the documentation. I really don't want to create a general
principle here that we need to create CVEs (and patched) for something
that was just added as a convenience for a handful of folks who were
ready to take a risk.

If there is sufficient interest, we can consider officially adding
this feature to the January 8.15 release iff it is ready by then.
@janmejay: please let me know if you would like to continue with your
work on lookup tables under this new situation.

As soon as I have time, I'll check what else needs to be removed. Not
sure about the ./contributed branch, because the project cannot
guarantee at all this is bug-free. It's documented to be so, but if
that is not sufficient, it should probably live only in the
master-insecure branch.


rgerhards:
We should **not** merge this PR until shortly before release. Then, we can see if it is actually needed.

janmejay:
Just to validate, yes I'll finish it as planned.

--
Regards,
Janmejay

PS: Please blame the typos in this mail on my phone's uncivilized soft
keyboard sporting it's not-so-smart-assist technology.

rgerhards:
excellent. Please create a PR when done. I'll also leave that one open up until close to the next release, and then we'll decide what gets merged into which branches. Does that sound OK with you?

janmejay:
Rainer, can you please create a branch that I can work on? Let us apply the
original patch on that, but we won't merge it to master until ready.

--
Regards,
Janmejay

PS: Please blame the typos in this mail on my phone's uncivilized soft
keyboard sporting it's not-so-smart-assist technology.

rgerhards:
you can use master-insecure. It'll fast-forward to master during that timeframe.

janmejay:
Awesome.

On Thu, Oct 8, 2015 at 8:58 PM, Rainer Gerhards <notifications@github.com>
wrote:

> you can use master-insecure. It'll fast-forward to master during that
> timeframe.
>
> —
> <https://github.com/rsyslog/rsyslog/pull/544#issuecomment-146579268>.
>



-- 
Regards,
Janmejay
http://codehunk.wordpress.com

janmejay:
Rainer, the proposal in lookup(2) function section says:

indexvalue (expression)
The value to be looked up. While this is an arbitrary RainerScript expression, it's final value is always converted to a string in order to conduct the lookup. For example, "**lookup(table, 3+4)" would be exactly the same as "lookup(table, "7")**". In most cases, indexvalue will probably be a single variable, but it could also be the result of all RainerScript-supported expression types (like string concatenation or substring extraction). Valid samples are "lookup(name, $fromhost-ip & $hostname)" or "lookup(name, substr($fromhost-ip, 0, 5))" as well as of course the usual "lookup(table, $fromhost-ip)".

How will this work for array and sparseArray? Shouldn't we take the given expr and internally interpret it appropriately? Eg. if user passes an expression that evaluates to "75", we interpret it as int32 75, then lookup can proceed for say sparseArray with key 75.

Is the proposal intentionally written to coerce the expr to string?

janmejay:
also, should values always be strings?

janmejay:
Another question: why represent key as unsigned-int? Won't representing it as signed-int allow ppl to use negative numbers more conveniently?

rgerhards:
It's a long time (3?4? years) since I have written this and I would need some more time to think about the details, time that I unfortunately do not currently have. Maybe @davidelang can jump in here, many of these were ideas of him.

To move forward, however, I would recommend that we first stabilize at what we currently have, without adding any additional features (well, probably the defaults). This would enable us to support the functionality, which already is useful for many use cases. 

After that, we could go for the bigger things.

The small things I could probably merge quickly, whereas my time is currently very limited to review complex PRs.

janmejay:
Well, Im have already started making changes for for the whole thing, so
let me finish it off, I'll be done by tomorrow, I guess. It'll be simple
set of changes, not too involved to merge + well covered by tests.

I'll keep values string for now, and will maintain uint32, we can change it
to int or long in the next cut. This means for now I'll stick to what the
proposal says pretty-much.


On Thu, Oct 29, 2015 at 2:38 PM, Rainer Gerhards <notifications@github.com>
wrote:

> It's a long time (3?4? years) since I have written this and I would need
> some more time to think about the details, time that I unfortunately do not
> currently have. Maybe @davidelang <https://github.com/davidelang> can
> jump in here, many of these were ideas of him.
>
> To move forward, however, I would recommend that we first stabilize at
> what we currently have, without adding any additional features (well,
> probably the defaults). This would enable us to support the functionality,
> which already is useful for many use cases.
>
> After that, we could go for the bigger things.
>
> The small things I could probably merge quickly, whereas my time is
> currently very limited to review complex PRs.
>
> —
> <https://github.com/rsyslog/rsyslog/pull/544#issuecomment-152119740>.
>



-- 
Regards,
Janmejay
http://codehunk.wordpress.com

davidelang:
in rsyslog we don't really have types, and even numbers need to be passed as strings. The idea was that for the basic mode of operation we are doing an array/sparse array with a numeric key, so whatever is put in that position in the lookup() call needs to be reduced to a number before it can be used.

sparse arrays (and the ability to use something like fromhost-ip) was to allow us to do a Geo-IP lookup against something like the maxmind free database. This is also why the table load/reload was speced  to be async, that's a lot of data to load and parse, we didn't want to stop all log processing while this was happening.

In addition to the pure numeric lookup, we want a "lookup by string" mode so that you could do something like lookup the hostname and get back a value ('prod' vs 'dev') for example. Thinking about this recently, I wonder if this should really be done as part of the table lookup, of if it should be structured more the way that liblognorm does things. List all the strings and turn them into a parse tree and just walk down that parse tree with the key until you get a value (or run out of tree, in which case we need to have an option to give an error/default value or the next lower value as per sparse arrays). It may even be that in the long term this works better for the lookup for large arrays than simple numeric arrays, but simple numeric arrays are so well understood and fast that it's hard to believe that something more complex will beat them.

janmejay:
The type coercion makes sense, we internally coerce it to int when we need
to.

For IP we'll likely need a little helper-fn in rainerscript, so we know
user wants a binary-representation of an IP eg. User wants 0x0E0E0E0E
instead of "14.14.14.14".

set uint_ip = bin_ip($fromhost-ip)

Then pass uint_ip to lookup. We can build this fn later, it doesn't involve
changes to lookup table itself.

If fromhost-ip is internally represented as number, this won't be needed
(haven't seen implementation yet).

Parsing the value to known annotated values also makes sense, but that
probably doesn't belong in lookup table.

--
Regards,
Janmejay

PS: Please blame the typos in this mail on my phone's uncivilized soft
keyboard sporting it's not-so-smart-assist technology.

rgerhards:
I have thought over this PR, and I think we can leave it out for one more release cycle, especially as I have not heared again on the need to provide a vuln note. That we we can later merge in @janmejay's new PR.

janmejay:
proposal-impl PR: https://github.com/rsyslog/rsyslog/pull/578

rgerhards:
I am closing this unmerged as we have PR #578 and I am confident that will be the final solution.

PR #558, opened 2015-10-21
imudp: Fixed a dbgprintf in processPacket function.

alorbach:
rcvBuf is not terminated with a NULL Byte, so the dbgprintf was
printing the whole buffer.


rgerhards:
I'll merge, but the printf may no longer be portable. So I potentially will revert to the old one if we get trouble. The "no-so-correct" print so far was considered OK given the lack of better options. I admit, I didn't know about the %.* format, but again, it sounds like it is non-standard. so let's see...

rgerhards:
merged to master-candidate

Issue #576, opened 2015-11-02
Fixed width option in 8.13.0 causes whitespace in property expansion


fastcat:
Commit f3bd7a2 (Add fixed-width option) appears to have introduced a regression in the property expansion code.

I have a configuration that allows syslog from some trusted hosts to be directed to separate files based on the source hostname, stripping off a common prefix, using a configuration line like:

`$template FooFileName,"/var/log/foo-%fromhost:4:$:lowercase%.log"`

(the common prefix is three chars log in this case, so it starts at 4 and runs to the end)

With 0.8.12, this worked properly, and created `/var/log/foo-sysa.log` files, but with 0.8.13, it is instead creating `/var/log/foo-sysa .log` files (note the space).

Looking at the diffs between 0.8.12 and 0.8.13, commit f3bd7a2 looks like the only plausible cause of this error, specifically the changes around https://github.com/rsyslog/rsyslog/blame/master/runtime/msg.c#L3652

Compare with the comment and code at https://github.com/rsyslog/rsyslog/blob/master/template.c#L1072

PR #577, opened 2015-11-02
Address off-by-one issues introduced in f3bd7a2 resulting in extra whitespace in property expansions

fastcat:
Fix issue #576: at the time `iTo` is being set based on `bufLen`, `iFrom` and `iTo` have been readjusted to be zero based instead of one based.  This issue predates f3bd7a2, but that commit then made it insert spaces if `iLen` referenced positions beyond the end of the string, which exposed this older mistake.

### DIFF
--- a/runtime/msg.c
+++ b/runtime/msg.c
@@ -3621,9 +3621,9 @@ uchar *MsgGetProp(msg_t *__restrict__ const pMsg, struct templateEntry *__restri
 			 */
 			; /*DO NOTHING*/
 		} else {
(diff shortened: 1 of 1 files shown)


rgerhards:
Quick question: we are moving towards ASL 2.0. Are you OK with releasing your patch under ASL 2.0?

fastcat:
Yes, you may use this patch under ASL 2.0, and/or the current LGPL v3 listed in that file's header.

rgerhards:
merged to master-candidate, will migrate to master from there

Issue #575, opened 2015-11-02
Segmentation Fault at gnutls_handshake Callback Function - gtlsClientCertCallback


futuernorn:
When an inbound connection is received by an instance of rsyslogd using the TCP Syslog Input Module, a segmentation fault occurs. The binary run was built from [source](https://github.com/rsyslog/rsyslog) on Ubuntu 14.04 (commit: `c1238936ff6b5567fa8e671e0eddc7c67da6129c`) using these flags:

```
./configure --enable-gnutls --enable-rsyslogd --enable-impstats --enable-uuid --enable-imfile --enable-imptcp --enable-debug
```

Though the issue occurs with the `8.13.0-0adiscon1trusty1` packaged version as well. The specific assignment that is triggering the segmentation fault is `st->key.x509 = pThis->ourKey;` in [runtime/nsd_gtls.c](https://github.com/rsyslog/rsyslog/blob/master/runtime/nsd_gtls.c#L254). From a strace of the process, it looks like the associated key file is being read in without issue:

```
 140760 29084 19:08:23 open("[/some/location/with/certs/cert].key", O_RDONLY <unfinished ...>
 140761 29083 19:08:23 clock_gettime(CLOCK_REALTIME,  <unfinished ...>
 140762 29084 19:08:23 <... open resumed> )     = 11
 140763 29083 19:08:23 <... clock_gettime resumed> {1446404903, 76399423}) = 0
 140764 29084 19:08:23 fstat(11,  <unfinished ...>
 140765 29083 19:08:23 write(1, "4903.076399423:main Q:Reg/w0  : Action 4 transitioned to state: rdy\n", 68 <unfinished ...>
 140766 29084 19:08:23 <... fstat resumed> {st_dev=makedev(202, 0), st_ino=230614, st_mode=S_IFREG|0600, st_nlink=1, st_uid=0, st_gid=0, st_blksize=4096, st_blocks=8, st_size=1679, st_atime=2015/11/01-16:30:58, st_mtime=2015/07/0
 140766 9-14:38:50, st_ctime=2015/11/01-16:30:58}) = 0
 140767 29083 19:08:23 <... write resumed> )    = 68
 140768 29084 19:08:23 read(11,  <unfinished ...>
 140769 29083 19:08:23 clock_gettime(CLOCK_REALTIME,  <unfinished ...>
 140770 29084 19:08:23 <... read resumed> "-----BEGIN [<KEY>...]"..., 1679) = 1679
[...]

PR #585, opened 2015-11-06
Possible fix for #575

csoutherland:
The callback given to [gnutls_certificate_set_retrieve_function](https://github.com/csoutherland/rsyslog/blob/6272aa12bcc025cb4ac285dfccae9ecc7418425d/runtime/nsd_gtls.c#651) is expected to set the appropriate server certificate during its execution. When the callback is not specified, it appears that [gnutls loops through the available credentials in order to find an appropriate certificate](https://gitlab.com/gnutls/gnutls/blob/gnutls_2_12_23/lib/auth_cert.c#L2084). Since we appear to be [using the real certificate](https://github.com/csoutherland/rsyslog/blob/6272aa12bcc025cb4ac285dfccae9ecc7418425d/runtime/nsd_gtls.c#1696) for the client side communication as well, it stands to reason that similar code should safely solve this problem. The only caveat is that this depends on my assumption that this is, in fact, the correct key to use.

### DIFF
--- a/runtime/nsd_gtls.c
+++ b/runtime/nsd_gtls.c
@@ -643,6 +643,17 @@ gtlsInitSession(nsd_gtls_t *pThis)
 
 	pThis->sess = session;
 
+#	if HAVE_GNUTLS_CERTIFICATE_SET_RETRIEVE_FUNCTION 
+	/* store a pointer to ourselfs (needed by callback) */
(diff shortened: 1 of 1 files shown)


alorbach:
Your patch looks good, I merged it into the master-candidate branch.

PR #592, opened 2015-11-10
contrib/omczmq and contrib/imczmq improvements

taotetek:
* omczmq - suspend / Retry handling - the output plugin can now recover from some error states due to issues with plugin startup or message sending
* omczmq - refactored topic handling code for ZMQ_PUB output to be a little more efficient
* omczmq - added ability to set a timeout for sends
* omczmq - set topics can be in separate frame (default) or part of message frame (configurable)
* omcmzq - code cleanup
* imczmq - code cleanup
* imczmq - fixed a couple of cases where vars could be used uninitialized
* imczmq - ZMQ_ROUTER support
* imczmq - Fix small memory leak from not freeing sockets  when done with them

### DIFF
--- a/contrib/imczmq/imczmq.c
+++ b/contrib/imczmq/imczmq.c
@@ -48,26 +48,28 @@ DEFobjCurrIf(glbl)
 DEFobjCurrIf(prop)
 DEFobjCurrIf(ruleset)
 
- 
+
+/* Arguments to pass to poll handler */
+/* --------------------------------- */ 
 typedef struct _pollerData_t {
 	ruleset_t *ruleset;
-	thrdInfo_t *thread;
+	thrdInfo_t *thread; 
 } pollerData_t;
 
 
 struct instanceConf_s {
-	bool is_server;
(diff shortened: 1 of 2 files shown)


rgerhards:
@taotetek any concerns in merging this (I saw you self-assigned it, thus the question)?

taotetek:
@rgerhards No concerns merging.  There will be more PRs for these plugins for the 8.15 release, but what's here so far is complete and has been tested. I couldn't remember the etiquette on merging here and wasn't sure if I should merge or not :)

rgerhards:
I was just not sure because you assigned yourself, what I took as indication you are working on it. As far as purely zmq is concerned, I am fine with you merging directly, because I think there is noone else who can actually verify a patch. On the other hand, I also merge happily, which means I'll also keep track of maintaining the ChangeLog.

Will merge later today, will wait until the testbench machines become free.

rgerhards:
merged into master-candidate, will migrate to master

PR #604, opened 2015-11-18
imjournal: fix off-by-one error in name cmprs

portante:
Looks like a long standing off-by-one comparison issue with the
invocation of strncmp.  The constant string is only 3 or 6 characters
long, but the comparison length is 4 or 7.

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -234,7 +234,6 @@ readjournal() {
 
 	const void *get;
 	const void *pidget;
-	char *parse;
 	size_t length;
 	size_t pidlength;
 
@@ -335,64 +334,7 @@ readjournal() {
 		/* get length of journal data prefix */
 		prefixlen = ((char *)equal_sign - (char *)get);
 
-		/* translate name fields to lumberjack names */
-		parse = (char *)get;
-
-		switch (*parse)
-		{
-		case '_':
-			++parse;
-			if (*parse == 'P') {
-				if (!strncmp(parse+1, "ID=", 4)) {
-					name = strdup("pid");
-				} else {
-					name = strndup(get, prefixlen);
-				}
-			} else if (*parse == 'G') {
-				if (!strncmp(parse+1, "ID=", 4)) {
-					name = strdup("gid");
-				} else {
-					name = strndup(get, prefixlen);
-				}
-			} else if (*parse == 'U') {
-				if (!strncmp(parse+1, "ID=", 4)) {
-					name = strdup("uid");
-				} else {
-					name = strndup(get, prefixlen);
-				}
-			} else if (*parse == 'E') {
-				if (!strncmp(parse+1, "XE=", 4)) {
-					name = strdup("exe");
-				} else {
-					name = strndup(get, prefixlen);
-				}
-			} else if (*parse == 'C') {
-				parse++;
-				if (*parse == 'O') {
-					if (!strncmp(parse+1, "MM=", 4)) {
-						name = strdup("appname");
-					} else {
-						name = strndup(get, prefixlen);
-					}
-				} else if (*parse == 'M') {
-					if (!strncmp(parse+1, "DLINE=", 7)) {
-						name = strdup("cmd");
-					} else {
-						name = strndup(get, prefixlen);
-					}
-				} else {
-					name = strndup(get, prefixlen);
-				}
-			} else {
-				name = strndup(get, prefixlen);
-			}
-			break;
-
-		default:
-			name = strndup(get, prefixlen);
-			break;
-		}
-
+		name = strndup(get, prefixlen);
 		CHKmalloc(name);
 
 		prefixlen++; /* remove '=' */


portante:
So while this is a simple fix, it potentially breaks compatibility, as those field names are now changed.  Perhaps we should either remove this field mapping code entirely, or put the field mapping under a new switch to turn it on explicitly.

rgerhards:
I would still think that it is a fix, but am also somewhat unsure. @davidelang @theinric what do you think?

rgerhards:
postponing to 8.16 as we do not yet have sufficient feedback

rgerhards:
Moving this again one more release goal forward. If we do not get feedback by then, I think we will implement it.

portante:
Summary of mailing list discussion:
 * Fix off-by-one behavior under an opt-in switch in 8.x
 * Off-by-one behavior fix default under 9.x with switch to fallback to 8.x behavior (turned off)
 * Apply conservative JSON property under 9.x with switch to fallback to 8.x behavior

Why not have opt-in for both under one switch for 8.x, and just leave it that way for 9.x?

davidelang:
the off-by-one is fixing a bug and doesn't change behaviour as I understand it. This fix should go in as standard in 8.x

changing behaviour (not showing properties that we use to populate standard properties) should not become the default until we have a major version change (if then, there are arguments on both sides of this one)

portante:
Perhaps I don't understand the way the code works, but the off-by-one fix appears to read that `_PID` becomes "`pid`", `_CMDLINE` becomes `cmd`, etc.  At least in the deployments that I have, this would break my environments because I would no longer have the uppercase names for the variables.

davidelang:
right now, rsyslog doesn't care about case in variable names, no the change could be to add the leading _ if that's missing, but it wouldn't change the case.

However, if it does change what's visible to the user, I didn't understand that from the patch description.

portante:
@davidelang, I have updated the patch description, and rebased it.  Please double check that I have understood the code's behavior properly.  Thanks!

portante:
@davidelang, actually, rsyslog does care about the case in variable names.  It preserves them first and foremost, but it performances searches for lowercase variable names only when referenced from templates.  I found out about that behavior because of `imjournal`: I was able to get all `imjournal` generated uppercase variables emitted in a template via `$!all-json` and the like, but could not reference an uppercase variable by itself.

theinric:
Sorry for the delay...

I generally agree with what was said on the mailing list; keep the current behavior and if possible add a switch to enable the translation. Without it, the relevant code can be just deleted.

Regarding the second commit, I'm not sure I like it that much.
It kind of makes sense not to duplicate the message in $!MESSAGE to save memory but the other fields are more borderline. E.g. the PID field is only used as part of the identifier, there's no standalone (JSON) PID field, so it's not really a duplicate.
Some other fields are parsed in some way and the change prevents the verbatim values from being preserved.
Additionally, this makes existence of some fields conditional - can we test for existence of a certain variable before referencing it?
... And not actually conditional. The fetched_* variables seem redundant as all the parsing is unconditional - either the field is present and parsed twice or it isn't and not parsed at all. Or am I missing something?

Back to the original issue: to me, the original parsing code seems needlessly complex and it's not surprising that such bugs happen. I probably won't get to test whether the switch()es actually improve performance but even if they do, I don't think it justifies the loss of readability.

rgerhards:
@portante So Peter, what do you think? Make it optionally or just plainly drop the non-working code?

portante:
I'll update this request with the non-working code removed, and that way we have no behavioral changes, and no features added.  This will level-set the code base for any future proposal for enhancements to `imjournal`.

Sounds reasonable?

rgerhards:
Very well!

Thanks

Sent from phone, thus brief.
Am 02.02.2016 13:20 schrieb "Peter Portante" <notifications@github.com>:

davidelang:
new features are acceptable, it's existing behavior changes that we try to avoid 
at (almost) all costs.

David Lang

portante:
@rgerhards, huh, that queue-persist.sh test failed again.  Bugger all.

rgerhards:
From contributing file (relatively new content):

Also, patches are requested to not break the testbench. Unfortunately, the
current testbench has some racy tests, which are still useful enough so that
we do not want to disable them until the root cause has been found. If your
PR runs into something that you think is not related to your code, just sit
back and relax. The rsyslog core developer team reviews PRs regularly and
restarts tests which we know to look racy. If the problem persists, we will
contact you.

theinric:
@portante Yeah, queue-persist.sh sometimes fails. I was looking into it some time ago but lagged insight into its design. I've brought up a question regarding this on the ml:
http://lists.adiscon.net/pipermail/rsyslog/2015-September/041110.html

@rgerhards could you perhaps shed some light here? The occasional fails are not a big problem in your local setup but are a showstopper for automated builds downstream.

rgerhards:
I would love to shed some light on it, but I really can't do much. It takes quite a while to investigate each of these cases. We have tracked down many of those cases. They all seem to be related to the fact that we are not able to properly detect when the engine has totally finished processing. In all cases seen so far, we had a race where sometime processing is stopped too early and thus we had some message loss. For the remaining issues (there is a small *set*), we assume that's also the reason, among other because this usually happens only in slower environments (like the testbench machines). Queue-persists.sh is quite complex, so the risk of race is higher than for others. I'd really love to have these racy test fixed, but it's again a question of where to put the effort. After all, the testbench has become much more reliable and complete over the past month. Help with the problematic tests would definitely be appreciated.

theinric:
Sure, I don't mean analysis of all the failures. For me, as the email states, the only failing tests were valgrind related and queue-persist.sh (perhaps I was just lucky).
queue-persist.sh uses ":omtesting:sleep 0 1000" and the documentation in omtesting turned out to be wrong so I just wanted to confirm that the intended delay is 0.001 and not 1 sec.
Potentially correcting the sleep interval wouldn't really fix the synchronization issue but, in my mind, the test would be in a more consistent state to build upon.

rgerhards:
re-checked 0.001 is correct (see #510 for more info)  . But let me check the termination wait times...

PR #609, opened 2015-11-19
pmciscoios: Support for IOSXR type syslog messages and dots (.) before timestamps

c0by:
Enable IOSXR message format support, this can be enabled with the present.xr parser option.

Unsynchronised clocks can append dots (.) or asterisk (*) before the timestamps.


c0by:
Also includes FreeBSD support for omelasticsearch

rgerhards:
would you be willing to update the doc at the rsyslog-doc project so that others can actually make use of this feature?

c0by:
yep sure, I'll send in an update to the docs shortly

PR #618, opened 2015-11-24
bugfix: RFC5424 did not handle NILVALUE in datestamp

rgerhards:
Note: this is untested and needs some more testing and verification before it actually can be merged. It's an extreme border case, so there is no hard need to merge quickly.

see also http://lists.adiscon.net/pipermail/rsyslog/2015-November/041555.html

### DIFF
--- a/tools/pmrfc5424.c
+++ b/tools/pmrfc5424.c
@@ -6,7 +6,7 @@
  *
  * File begun on 2009-11-03 by RGerhards
  *
- * Copyright 2007-2014 Rainer Gerhards and Adiscon GmbH.
+ * Copyright 2007-2015 Rainer Gerhards and Adiscon GmbH.
  *
  * This file is part of rsyslog.
  *
@@ -245,7 +245,11 @@ CODESTARTparse
 	 */
 
 	/* TIMESTAMP */
-	if(datetime.ParseTIMESTAMP3339(&(pMsg->tTIMESTAMP),  &p2parse, &lenMsg) == RS_RET_OK) {
+	if(lenMsg >= 2 && p2parse[0] == '-' && p2parse[1] == ' ') {
+		memcpy(&pMsg->tTIMESTAMP, &pMsg->tRcvdAt, sizeof(struct syslogTime));
+		p2parse += 2;
+		lenMsg -= 2;
+	} else if(datetime.ParseTIMESTAMP3339(&(pMsg->tTIMESTAMP),  &p2parse, &lenMsg) == RS_RET_OK) {
 		if(pMsg->msgFlags & IGNDATE) {
 			/* we need to ignore the msg data, so simply copy over reception date */
 			memcpy(&pMsg->tTIMESTAMP, &pMsg->tRcvdAt, sizeof(struct syslogTime));


tools/pmrfc5424.c
```diff
@@ -245,7 +245,11 @@ CODESTARTparse
 	 */
 
 	/* TIMESTAMP */
-	if(datetime.ParseTIMESTAMP3339(&(pMsg->tTIMESTAMP),  &p2parse, &lenMsg) == RS_RET_OK) {
+	if(lenMsg >= 2 && *p2parse == '-' && *(p2parse+1) == ' ') {
```
dubek: I think it'll be clearer as `p2parse[0] == '-' && p2parse[1] == ' '` (similar to line 225 above).

dubek:
It works good!

I tried to see how to add a unit test for this new case, but when I run `make check` I get `All 0 tests passed`, so I'm not sure how to proceed.

Here are the details of my manual verification (on RHEL 6.7 x86_64):

Config:

```
module(load="imuxsock")
module(load="imklog")
module(load="imudp")
module(load="pmrfc5424")
module(load="omfile")

input(type="imudp" port="9514" ruleset="myparser")

ruleset(name="myparser" parser="rsyslog.rfc5424") {
    action(type="omfile" file="/tmp/rsyslog/mylog")
    action(type="omfile" file="/tmp/rsyslog/rsyslogdebug" template="RSYSLOG_DebugFormat")
}
```

Test script:

```
echo '<11>1 2000-01-01T15:00:00.000Z 1.2.3.4 myhost - L - first message text' | nc -u -w1 127.0.0.1 9514
echo '<11>1 - 1.2.3.4 myhost - L - second message text' | nc -u -w1 127.0.0.1 9514
```

Content of `mylog` before the fix (master):

```
2000-01-01T15:00:00.000Z 1.2.3.4 myhost first message text
2015-11-24T10:42:39.938548-05:00 localhost  - 1.2.3.4 myhost - L - second message text
```

Content of `mylog` after the fix:

```
2000-01-01T15:00:00.000Z 1.2.3.4 myhost first message text
2015-11-24T10:46:15.076873-05:00 1.2.3.4 myhost second message text
```

rgerhards:
You need to give the --enable-testbench configure option. There was a reason I don't really remeber why its not on by default. I think it was a distro thingy.

you are right on the code, I'll modify to patch to use the same syntax all over the place. Will then merge soon. Thanks for testing.

rgerhards:
@dubek I have made the code change. It wouldn't hurt if you give the code one more test before I merge. Note that I have rewritten the branch.

dubek:
I can confirm that the new code is working OK. I sent three UDP packets:

1. with full timestamp
2. with `-` in the timestamp field (NILVALUE)
3. with invalid timestamp (`XYZ` in the timestamp field)

This is the content of `mylog` which is what I expected:

```
2000-01-01T15:00:00.000Z 1.2.3.4 myhost first message text
2015-11-24T11:16:59.965612-05:00 1.2.3.4 myhost second message text
2015-11-24T11:17:00.970546-05:00 localhost  XYZ 1.2.3.4 myhost - L - third message text
```

rgerhards:
thx, will merge soon.

dubek:
I tried adding a test case (`tests/testsuites/rfc5424-5.parse1`) with NILVALUE for timestamp and check it with `make check TESTS=parsertest.sh`, but the matching of the expected resulting log line is literal, and the resulting line will be different in each invocation (because it takes the received timestamp which is the current time).

Possible solutions:

1. Support a "fake wall time" mode for testing, thus producing an expected constant log line. Maybe [libfaketime](https://github.com/wolfcw/libfaketime) can help (see their "Absolute time" mode).
2. In `nettester`, similar to expanding `~H` in the expected response: support more variables or regexp matching.

Anyway, nothing really simple right now as far as I can see.

Issue #481, opened 2015-08-14
variable names case problem


davidelang:
variable names in templates get forced to be lower case but the variables created by parsers are not always lower case

for example, $!MemTotal won't find the value in { "MemTotal": 8976 } but it will find { "memtotal": 8976 }

this is in v8.12

portante:
So `grammar/lexer.l` uses the case-insensitive option. :(  This means the lexer never sees an uppercase character.  Turn that off, and then stuff starts breaking.  This might be too invasive.

rgerhards:
The route to take here is make sure that variable names are stored in lower case by whoever generates them. We may want to enforce that in those functions that add them.

Back in the dark ages of rsyslog everything was case-sensitive. This caused a lot of trouble. I guess the introduction of (JSON) variables did not properly ensure that everything is lowercase, like it is at other places inside the code.

davidelang:
On Wed, 25 Nov 2015, Rainer Gerhards wrote:

> The route to take here is make sure that variable names are stored in lower case by whoever generates them. We may want to enforce that in those functions that add them.
>
> Back in the dark ages of rsyslog everything was case-sensitive. This caused a lot of trouble. I guess the introduction of (JSON) variables did not properly ensure that everything is lowercase, like it is at other places inside the code.

Is JSON supposed to be case sensitive in it's names? if so and we collapse the 
case, we can end up loosing data from JSON messages if they have two fields that 
differ by case. It would be nice to learn that JSON is not supposed to be case 
sensitive.

Remember that we have only limited control over the JSON we receive. logs from 
logstash will have field names that start wtih @ for example (@timestamp is the 
equivalent of our $timereceived)

David Lang

rgerhards:
ok, you are right. But than it's a much harder question. This means we need to re-introduce case sensitivity in rsyslog :-( something for next year march+

portante:
@davidelang, @rgerhards, perhaps we add an option to mmjsonparse, imjournal, and others, such that they fold variables to lowercase when creating them?  This would be fairly easy to do, and bridge the gap until we can be confident of a solution for case-sensitivity in variable names.

I am trying the case-sensitivity out in a [branch](https://github.com/portante/rsyslog/tree/case-sensitive-vars).  I'll be building a set of packages for EPEL 7 where I can run this to see how it fairs in my environment.

[...]

PR #622, opened 2015-11-25
Accept leading underscores for variable names

### LINKED ISSUE #603: Unable to reference leading underscore variables created off root message property
If I have:
```
# cat /etc/rsyslog.conf.test
set $!this = $!foo;
set $!that = $!_bar;
```

Then when I try to parse this configuration I receive the following error with 8.14.0:
```
# /usr/sbin/rsyslogd -N1 -f /etc/rsyslog.conf.test
rsyslogd: version 8.14.0, config validation run (level 1), master config /etc/rsyslog.conf.test
rsyslogd: error during parsing file /etc/rsyslog.conf.test, on or before line 2: invalid character '$' in expression - is there an invalid escape sequence somewhere? [v8.14.0 try http://www.rsyslog.com/e/2207 ]
rsyslogd: error during parsing file /etc/rsyslog.conf.test, on or before line 2: invalid character '!' in expression - is there an invalid escape sequence somewhere? [v8.14.0 try http://www.rsyslog.com/e/2207 ]
rsyslogd: error during parsing file /etc/rsyslog.conf.test, on or before line 2: invalid character '_' in expression - is there an invalid escape sequence somewhere? [v8.14.0 try http://www.rsyslog.com/e/2207 ]
rsyslogd: error during parsing file /etc/rsyslog.conf.test, on or before line 2: syntax error on token ';' [v8.14.0 try http://www.rsyslog.com/e/2207 ]
rsyslogd: CONFIG ERROR: could not interpret master config file '/etc/rsyslog.conf.test'. [v8.14.0 try http://www.rsyslog.com/e/2207 ]
```

If I change `$!_bar` on line 2 to be `$!bar` it works, and will work if I change it to `$!then!_bar`.

Unfortunately, the `imjournal` module uses the exact names of trusted variables defined at [systemd.journal-fields](http://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html), which all have leading underscores.

Is this expected behavior?

portante:
The grammar now allows for leading underscores for variable names. This
addresses a problem stemming from the fact that the imjournal module
will create variables using the given name from systemd. All "trusted"
variables in systemd begin with a leading underscore.

Fixes #603.

### DIFF
--- a/grammar/lexer.l
+++ b/grammar/lexer.l
@@ -141,7 +141,7 @@ int fileno(FILE *stream);
 <EXPR>0[0-7]+ |			/* octal number */
 <EXPR>0x[0-7a-f] |		/* hex number, following rule is dec; strtoll handles all! */
 <EXPR>([1-9][0-9]*|0)		{ yylval.n = strtoll(yytext, NULL, 0); return NUMBER; }
-<EXPR>\$[$!./]{0,1}[@a-z][!@a-z0-9\-_\.\[\]]*	{ yylval.s = strdup(yytext+1); return VAR; }
+<EXPR>\$[$!./]{0,1}[@a-z_][!@a-z0-9\-_\.\[\]]*	{ yylval.s = strdup(yytext+1); return VAR; }
 <EXPR>\'([^'\\]|\\['"\\$bntr]|\\x[0-9a-f][0-9a-f]|\\[0-7][0-7][0-7])*\'	 {
 				   yytext[yyleng-1] = '\0';
 				   unescapeStr((uchar*)yytext+1, yyleng-2);
--- a/outchannel.c
+++ b/outchannel.c
@@ -235,7 +235,7 @@ struct outchannel *ochAddLine(char* pName, uchar** ppRestOfConfLine)
 
 
 /* Find a outchannel object based on name. Search
- * currently is case-senstive (should we change?).
+ * currently is case-sensitive (should we change?).
  * returns pointer to outchannel object if found and
  * NULL otherwise.
  * rgerhards 2004-11-17
--- a/runtime/msg.c
+++ b/runtime/msg.c
@@ -539,87 +539,87 @@ propNameToID(uchar *pName, propid_t *pPropID)
 
 	/* sometimes there are aliases to the original MonitoWare
 	 * property names. These come after || in the ifs below. */
-	if(!strcmp((char*) pName, "msg")) {
+	if(!strcasecmp((char*) pName, "msg")) {
 		*pPropID = PROP_MSG;
-	} else if(!strcmp((char*) pName, "timestamp")
-		  || !strcmp((char*) pName, "timereported")) {
+	} else if(!strcasecmp((char*) pName, "timestamp")
+		  || !strcasecmp((char*) pName, "timereported")) {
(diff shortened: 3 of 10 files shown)


portante:
That looks like a sporadic failure of sorts.  Let me know if I have misunderstood the failure mode, and I'll try to address it further.

portante:
This also seems to fix #481 as well, @davidelang.

portante:
@theinric, this is one I'd like us to track closely.

rgerhards:
This addresses character case handling, merge with care after very careful review.Maybe we should cherry-pick the patch for leading underscores.

portante:
@rgerhards, definitely feel free to cherry-pick, which is why I did not collapse this into one patch.

The leading underscores is a problem unto itself, but I added the second patch for the mixed case references because my use case for the leading underscores is `imjournal`, which uses uppercase names, and I still can't reference the names without them.

FWIW, I have successfully deployed these two patches on the 8.14 code branch, and they work great.

portante:
@rgerhards, I further broke this fix out so that you can see that the only change that appears to be necessary is where we stop lowercasing variable name references from templates.

This is a bug today, as one cannot reference uppercase variable names, even though the parser is allowing them.

I am going to add a switch to turn this on as an opt-in.

portante:
Okay, so this PR has three patches in it:

1. "Accept leading underscores for variable names"
  * simple change to allow those characters
  * should **not** pose any backward compatibility issues
2. "Ensure non-variable property names are case-insensitive"
  * simple change to ensure property name matching is performed in a case-insensitive manner
  * should **not** pose any backward compatibility issues 
3. "Fix upper and mixed case variable references"
  * The only problem found with upper and mixed case variable name references occurs in templates
  * All template name references are initially left as provided by the parser (NOT lowercased)
  * If the template option of "`option.casesensitivity`" is "`off`", the default, then the field and property names are forced lowercase (as was the case previously), **maintaining existing behavior**
  * If the template option of "`option.casesensitivity`" is "`on`", then the field and property names are used as-is (this is "opt-in" new behavior)

Feedback welcome.

portante:
@rgerhards, any further thought on this patch?  @theinric?

rgerhards:
I think we are ready to merge, but I didn't manage last week due to unexpected problems with the other PRs. Now I need to look into some other stuff, first. This may require me to back off merging for another two weeks :-( But we should make it for 8.17 in any case.

urbaniak:
:+1:

Looking forward for the next release. ;)

rgerhards:
@portante I have lost a bit track of this ... did you also add accompanying doc to rsyslog-doc? If not, it would be great if you could do so. The merge should hopefully happen today (running throught the final multi-platform testbench right now).

portante:
@rgerhards, I have not waded into the documentation at all.  I can certainly take a stab at that.

rgerhards:
I guess at least this page needs some update: https://github.com/rsyslog/rsyslog-doc/blob/master/source/rainerscript/global.rst

portante:
@rgerhards, I thought it was a template directive modification, I updated the documentation there.  Let me know if that was not what you were looking for, or if I am missing something.

rgerhards:
**I** got lost in the many PRs I have merged recently. Somehow remembered it was a global option...

Issue #605, opened 2015-11-19
Test failures on a 2 socket box with v8.14 branch


portante:
Just thought I would post the test failures that I found compiling and running on a Fedora 21 2 socket box:

```
==========================================
   rsyslog 8.14.0: tests/test-suite.log
==========================================

# TOTAL: 163
# PASS:  154
# SKIP:  5
# XFAIL: 0
# FAIL:  4
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

SKIP: sndrcv_udp.sh
===================

===============================================================================
[sndrcv_udp.sh]: testing sending and receiving via udp

FAIL: queue-persist.sh
======================

[queue-persist.sh]:
testing memory queue persisting to disk, mode LinkedList
rsyslogd started with pid  9895
injecting 5000 messages
imdiag[13500]: 5000 messages injected
There must exist some files now:
total 2172
-rw------- 1 pportant pportant 1048728 Nov 19 01:26 mainq.00000001
-rw------- 1 pportant pportant 1048728 Nov 19 01:26 mainq.00000002
-rw------- 1 pportant pportant  114622 Nov 19 01:26 mainq.00000003
-rw------- 1 pportant pportant     523 Nov 19 01:26 mainq.qi
rsyslogd started with pid  9917
imdiag[13500]: mainqueue empty
-------------------------------------------------------------------------------
testing memory queue persisting to disk, mode FixedArray
rsyslogd started with pid  9963
injecting 5000 messages
imdiag[13500]: 5000 messages injected
There must exist some files now:
total 2172
-rw------- 1 pportant pportant 1048728 Nov 19 01:26 mainq.00000001
-rw------- 1 pportant pportant 1048728 Nov 19 01:26 mainq.00000002
-rw------- 1 pportant pportant  114178 Nov 19 01:26 mainq.00000003
-rw------- 1 pportant pportant     523 Nov 19 01:26 mainq.qi
rsyslogd started with pid  9984
imdiag[13500]: mainqueue empty
read value 20, but expected value 19
sequence error detected

SKIP: imuxsock_logger_root.sh
=============================

[imuxsock_logger_root.sh]: test trailing LF handling in imuxsock
This test must be run as root with no other active syslogd

SKIP: imuxsock_traillf_root.sh
==============================

[imuxsock_traillf_root.sh]: test trailing LF handling in imuxsock
This test must be run as root with no other active syslogd

SKIP: imuxsock_ccmiddle_root.sh
===============================

[...]

PR #623, opened 2015-11-25
Fix mmnormalize_regex failure for rule w/ noeol

portante:
Looks like at least on Fedora 21 with liblognorm 1.1.2-2, this test
fails because of the missing newline character before EOF.

Partial fix for the failure reported in #605.


portante:
@theinric, this we should get into Fedora, too.

rgerhards:
sorry, I think this crossed: https://github.com/rsyslog/rsyslog/pull/627
I merged from top to bottom of the list ;)

rgerhards:
Thanks for the fix. @theinric: I suggest you do NOT enable regex by default. IMHO its a bad idea. See also: https://github.com/rsyslog/liblognorm/issues/143

rgerhards:
merged to master-candidate

PR #624, opened 2015-11-25
Add tests for variable name comparison regressions

### LINKED ISSUE #617: Local variable comparisons do not work
This does not appear to work on v8.14.0:
```
set $.foo = "foo";
set $.bar = "foo";
if ($.foo == $.bar) then {
    *.* /var/log/rsyslog-debug.log;RSYSLOG_DebugFormat
}
```

portante:
Don't change the first character of the variable name, since we need
that depending on its name space. Instead, determine the proper root
variable string that should be used.

Fixes #617.

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -125,6 +125,12 @@ TESTS +=  \
 	rscript_wrap3.sh \
 	rscript_re_extract.sh \
 	rscript_re_match.sh \
+	rscript_eq.sh \
+	rscript_ge.sh \
+	rscript_gt.sh \
+	rscript_le.sh \
+	rscript_lt.sh \
+	rscript_ne.sh \
 	rs_optimizer_pri.sh \
 	cee_simple.sh \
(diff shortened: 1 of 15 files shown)


portante:
I'll update with a unit test shortly.

portante:
Okay, I have posted a number of tests ahead of an actual fix.

The first patch, "Enable rscript comparison tests", `48321f9`, just adds a few existing comparison tests as a base line.  The second patch, "Tests to verify string variable comparisons", `92f4f08`, adds explicit and exhaustive testing of json, local, and global variables.

The third test fails, which is trying to mirror what is reported in #617.  But I still have not figured out why it fails.

rgerhards:
moving to 8.16, as the releated changes also don't make it into 8.15.

portante:
I updated the title as this PR has morphed from originally trying to fix a problem to just enabling a set of tests to help catch regressions in variable name comparisons.

PR #629, opened 2015-11-26
free the tokener 

c6226:


rgerhards:
We need to carefully review the patch set. Out of my head, I think the third commit (msgbuf) is wrong, because msgbuf was not duplicated.

c6226:
Sorry，my log description is not accurate, I mean  pmsgbuf need to release。when the lenRcv bigger than  the sizeof(msgbuf)，the psmsgbuf is malloc from system ，so i think  it need to free。

PR #630, opened 2015-11-26
omhttpfs: If realloc function fails, the original block is not freed. The orignal block pointed to by mem->reply will be leaked

haixingood:
The "mem->reply" is assigned for realloc function`s return value directly. If realloc function fails then mem->reply is assigned as NULL, but the original block is not freed. The original block pointed to by mem->reply will be leaked.


rgerhards:
merged this into my private travis-omhttpfs branch, will migrate from there into the official master branch.

side-note: the code probably can be improved, I am not sure that discarding the old buffer is a good idea. However, I am short of time right now and as this is just a contributed module, I cannot go fully through analysis and potentially changing that.

PR #635, opened 2015-11-28
imzmq: fix Memory leak

xushengping:
here shoud free subscribes , becase   CHKiRet(parseSubscriptions(es_str2cstr(pvals[i].val.d.estr, NULL),  &inst->subscriptions)) have no way to free  
the memory which malloc by es_str2cstr


taotetek:
Looks good!

rgerhards:
merged, but github doesn't detect that due to the way the PR was crafted.

PR #636, opened 2015-11-28
omkafka: fix memory leaks

hiteric:
Here param should be freed, because the following func:
 CHKiRet(processKafkaParam(es_str2cstr(pvals[i].val.d.ar->arr[j], NULL),&pData->confParams[j].name,&pData->confParams[j].val)); 
malloc memory with es_str2cstr, but without free.


rgerhards:
I have checked, the patch is wrong and leads to a double-free. "param" is memory inside the parameter block and is freed on parameter block destruction by the framework.

rgerhards:
@portante +1

I'll refactor the patch, but thanks for submitting.

rgerhards:
replaced by https://github.com/rgerhards/rsyslog/commit/18bb6a5423ac7af0178d068607b69488a6900317

PR #637, opened 2015-11-30
omzmq: free memory before return

szjack:


szjack:
hi all,is this match better ?

taotetek:
Looks good!  Sorry for the late reply and thanks so much!

PR #640, opened 2015-11-30
fix some issues shown in clang static analyzer and improve testbench

rgerhards:


rgerhards:
merged into master-candidate

Issue #439, opened 2015-07-13
FAIL: imtcp-tls-basic-vg.sh on Debian 8


alorbach:
Test Fails on Debian

--
[imtcp-tls-basic-vg.sh]: testing imtcp in TLS mode - basic test
==26108== Memcheck, a memory error detector
==26108== Copyright (C) 2002-2013, and GNU GPL'd, by Julian Seward et al.
==26108== Using Valgrind-3.10.0 and LibVEX; rerun with -h for copyright info
==26108== Command: ../tools/rsyslogd -C -n -irsyslog.pid -M../runtime/.libs:../.libs -f./testsuites/imtcp-tls-basic.conf
==26108== 
rsyslogd started with pid  26108
startup-vg still running
starting run 1
Sending 50000 messages.

00050000 messages sent
runtime: 0.050
End of tcpflood Run
imdiag[13500]: mainqueue empty
==26108== 
==26108== Process terminating with default action of signal 11 (SIGSEGV): dumping core
==26108==  General Protection Fault
==26108==    at 0x506128B: __lll_unlock_elision (elision-unlock.c:33)
==26108==    by 0x42B533: asyncWriterThread (stream.c:1284)
==26108==    by 0x50580A3: start_thread (pthread_create.c:309)
==26108==    by 0x609204C: clone (clone.S:111)
==26108== 
==26108== HEAP SUMMARY:
==26108==     in use at exit: 1,073,067 bytes in 2,274 blocks
==26108==   total heap usage: 108,993 allocs, 106,719 frees, 44,263,696 bytes allocated
==26108== 
==26108== 288 bytes in 1 blocks are possibly lost in loss record 1,138 of 1,208
==26108==    at 0x4C2AD10: calloc (vg_replace_malloc.c:623)
==26108==    by 0x4010FD1: allocate_dtv (dl-tls.c:296)
==26108==    by 0x40116DD: _dl_allocate_tls (dl-tls.c:460)
==26108==    by 0x5058C27: allocate_stack (allocatestack.c:589)
==26108==    by 0x5058C27: pthread_create@@GLIBC_2.2.5 (pthread_create.c:495)
==26108==    by 0x42C90E: wtpStartWrkr (wtp.c:433)
==26108==    by 0x42C90E: wtpAdviseMaxWorkers (wtp.c:476)
==26108==    by 0x4317AD: qqueueAdviseMaxWorkers (queue.c:369)
[...]

PR #643, opened 2015-11-30
stream driver: Fixed double mutex unlock when timeout in asyncWri…

alorbach:
…terThread

Also corrected mutex locking when timeout occurs, and removed never reached
mutex unlock. This also fixes issue:
https://github.com/rsyslog/rsyslog/issues/439


rgerhards:
duplicate

PR #647, opened 2015-12-01
omczmq: can now set timeout properly

taotetek:
* sendtimeout now works properly.
* action will suspend on timeout error if sendtimeout is set.


taotetek:
@rgerhards This is ready for merge at your convenience.

PR #650, opened 2015-12-01
omczmq: make sendtimeout not required with reasonable default

taotetek:
* sendtimeout now works properly on conjunction with suspend / resume


rgerhards:
merged into master-candidate

PR #652, opened 2015-12-02
Bury es_str2cstr() calls under explicit Debug test

portante:
In rainerscript.c there was a wayward debug statement emitting a es str,
where the es string was converted to a c-string before printing, and
then freeing the memory. It would seem doing that on the fast path is
unnecessary, so we moved it under an explicit Debug test.

Searching for other uses of es_str2cstr() calls in dbgprintf() calls
(lowercase), which would mean that the memory allocation always occurs
even if the debugging is not enabled, we found two in omprog and
mmexternal.

In the modules that we touched, we switched to use DBGPRINTF() were
possible and logical, to avoid processing any parameters on the fast
path.

### DIFF
--- a/grammar/rainerscript.c
+++ b/grammar/rainerscript.c
@@ -112,7 +112,7 @@ tokenToString(const int token)
 	case CMP_STARTSWITH: tokstr ="CMP_STARTSWITH"; break;
 	case CMP_STARTSWITHI: tokstr ="CMP_STARTSWITHI"; break;
 	case UMINUS: tokstr ="UMINUS"; break;
-	default: snprintf(tokbuf, sizeof(tokbuf), "%c[%d]", token, token); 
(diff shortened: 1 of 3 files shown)


portante:
@rgerhards, could we consider this for 8.15?

rgerhards:
yes, definitely, just forgot to add the label. I am right now focussed on getting static analyzer full integrated into the testbench and that's occupying most of the time that I can currently spend.

rgerhards:
@portante I see you did well by fixing some ugly outputs. I also see they were not indented, what means they were ment to be temporal and to be removed, what I obviously forgot to do. I'll need a bit more review to delete those lines after the merge, just FYI.

rgerhards:
Note: I have removed the messages in question. This leads to merge conflicts, which I will resolve. That was the easiest way for me to address this.

PR #653, opened 2015-12-02
Add common tokens in tokenToString

portante:
A static buffer is used in tokenToString() when creating a string to
represent an unrecognized token. There are a number of operators and
datatypes that were unrecognized, so this was used often. This change
avoids multiple threads writing over the same string, confusing the
debug output.

Note that this does not solve the problem, as this can still happen if
new datatypes or operators are added in the future, but it does take
care of the common cases known today.


rgerhards:
merged to master-candidate

PR #664, opened 2015-12-04
imfile: Trim line when length of line over specified bytes

JindongChen:
Add an opt-in flag 'trimLineOverBytes' to determine if trim line when length of line over specified bytes number. 
Default value of 'trimLineOverBytes' is 0, means never trim line. 
If 'trimLineOverBytes' is positive number, we will truncate the line which length is greater than 'trimLineOverBytes'.

### DIFF
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -102,6 +102,7 @@ typedef struct lstn_s {
 	int iFacility;
 	int iSeverity;
 	int maxLinesAtOnce;
+	uint32_t trimLineOverBytes;
 	int nRecords; /**< How many records did we process before persisting the stream? */
 	int iPersistStateInterval; /**< how often should state be persisted? (0=on close only) */
 	strm_t *pStrm;	/* its stream (NULL if not assigned) */
@@ -129,6 +130,7 @@ static struct configSettings_s {
 	int iSeverity;  /* notice, as of rfc 3164 */
 	int readMode;  /* mode to use for ReadMultiLine call */
 	int64 maxLinesAtOnce;	/* how many lines to process in a row? */
+	uint32_t trimLineOverBytes;  /* 0: never trim line, positive number: trim line if over bytes */
 } cs;
 
 struct instanceConf_s {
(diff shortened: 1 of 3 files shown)


rgerhards:
Would you consider adding related doc to the rsyslog-doc project?

davidelang:
how is this different than what happens at maxmessagesize?

When you hit maxmessagesize, what's been read so far becomes the message, and anything left becomes the start of the next message.

It looks like this function does the same thing (although it adds a newline at the end of the truncated message)

Is the intent of this to truncate the message at a point smaller than maxmessagesize so that there is room to add header info and send it out without additional truncation?

JindongChen:
When you hit trimLineOverBytes, the message what's been read so far is send out, and anything left is discarded. 

It does not add a newline, but a char '\n' at end of the truncated message. String *ppCStr need it.

As you said, maxmessagesize does not truncate message. And i'm not sure maxmessagesize can be used in imfile plugin. I don't set maxmessagesize(default value should be 8k) explicitly, but can see message length over 10k, so I implement trimLineOverBytes by myself.

davidelang:
reading the code, I don't see anything that continues reading the message to through it away. I could be missing it.

I see that it will stop reading and process the message, but the pointer into the input remains at the same place so when the next message starts being read it picks up the remainder of the first message.

This is exactly what happens when an incoming message exceeds maxmessagesize (including the remainder of the message starting the next message)

So I think it would make more sense to make imfile respect maxmessagesize the way other things do than to implmenent a new function to do the same thing.

That said, it would make sense to add a flag that says something like "if you hit maxmessagesize, keep reading until you hit the normal 'end of message' and throw it all away", and possibly have two variations of this, one that creates a log from rsyslog saying that it did so, and one that does so silently.

It may also make sense to have a max inbound message size as opposed to just max message size, so that you can try to be sure that your output is going to be < maxmessagesize after you do things like add header info that may be missing.

JindongChen:
The whole line message has been read into *ppCStr, rsCStrTruncate throw unwanted message away.
```
rsCStrTruncate(*ppCStr, cstrLen(*ppCStr) - trimLineOverBytes);
```
maxmessagesize has widespread influence and many limited.  For example:
```
Do not set this below 1k, as it would cause interoperability problems with other syslog implementations.
```
Ref: http://www.rsyslog.com/doc/v8-stable/configuration/global/index.html

trimLineOverBytes aims to specific target - truncation without more restrictions.

PR #670, opened 2015-12-05
Update build steps for RH/CentOS/Fedora

portante:
These are the complete steps needed to build on a RHEL / CentOS / Fedora box (but only verified on a Fedora 21 box).


portante:
BTW, at least on Fedora, one needs to set `CFLAGS` to include the output of the following: `$(pkg-config --cflags glib)`.

rgerhards:
merged to master-candidate, note on CFLAGS added

PR #681, opened 2015-12-07
Derive glib-2.0 build flags for mmgrok at config

### LINKED ISSUE #680: mmgrok: confusion between glib and glib2 to be used
Looks like `glib-devel` is require, but the library link line references `glib-2.0`.

portante:
Instead of requiring the user to set the proper CFLAGS before invoking
./configure, we add CFLAGS and LIBS values as determined by pkg-config
during configuration setup.

We also update the requirements to be glib-2.0 instead of just plain
glib.

Fixes #680.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+dist: trusty
+
 language: c
 
 compiler:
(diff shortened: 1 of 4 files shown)


rgerhards:
Would it be possible to build this under Travis as well?

portante:
@rgerhards, sure let me take a stab at it.

portante:
@rgerhards, I am not sure how to get `grok` and `libgrok-dev` packages installed on the Ubuntu instances created via Travis.  Any ideas?  I believe those are the right package names, but not entirely sure.

portante:
@rgerhards, okay, looks like using Ubuntu "Trusty Tahr" distribution fixes the problem of not finding `grok` and `libgrok-dev`.

Not sure if we want to make that holistic switch.

rgerhards:
@portante ah, I didn't know there were packages for libgrok at all. I'll have a look. The clang static analyzer run already builds under thrusty thar. But I had some problems with the other tests. Ideally, we would run all tests under both platforms (they differ in fine details, what can help detect races). As a side note, I run many more test at  http://build.rsyslog.com:8010/waterfall?tag=rsyslog-rsyslog It's unfortunately not yet as nicely integrated into github as travis. The buildbot tests run based on master-candidate, so they are the final check before merge into master happens.

rgerhards:
I'll merge this into my experimental branch for now and see if I can get the build going.

rgerhards:
merged via https://github.com/rsyslog/rsyslog/pull/685

PR #685, opened 2015-12-08
make mmgrok build under travis

rgerhards:
includes some fixes to the build system as well as fixes to mmgrok itself.

Includes https://github.com/rsyslog/rsyslog/pull/681

DO NOT MERGE IF TRAVIS FAILS. We are trying to solve an issue with the build process.


portante:
So now we are down to just an issue with the clang build...

PR #687, opened 2015-12-09
Remove leading whitespace in front of CPP directives

portante:
Seems like the leading whitespace is not a real problem, but noticed it reviewing another patch and thought I'd submit the cleanup for consideration.

### DIFF
--- a/contrib/pmpanngfw/pmpanngfw.c
+++ b/contrib/pmpanngfw/pmpanngfw.c
@@ -92,11 +92,11 @@ BEGINparse
     uint64 log_type;
     int j;
 CODESTARTparse
-    #define CSV_DELIMITER '\t'
-    #define STATE_FIELD_START 0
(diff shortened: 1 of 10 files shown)


rgerhards:
More out of curiosity: Is this a requirement? I tend to indent the hash together with the rest as I think this is more readable (and I know it's not done consistently).

portante:
I don't think this is a requirement, see https://gcc.gnu.org/onlinedocs/gcc-2.95.3/cpp_1.html#SEC3.  So if you don't feel this is work taking in, we can just drop this.

rgerhards:
After thinking over it, I think we should drop it. C90 and C99 support either style. To me, the code looks more readable with the hash NOT at the start of the line. Does this sound reasonable to you?

portante:
No problem by me, thanks.

PR #695, opened 2015-12-11
omkafka: fix potentially negative partition number

rgerhards:
bases on problem identification done by Tait Clarridge


oldmantaiter:
@rgerhards Thanks, this is definitely a better solution. Once merged I'll respond to the mailing list user to have them try again and see if there is another place that negative partitions could exist.

rgerhards:
@oldmantaiter thanks, I am now in the process of merging, should be in master in roughly an hour or so. And once again thanks for providing the real patch - I just did the much smaller work of refactoring it a bit:-)

PR #703, opened 2015-12-16
Remove use of sizeof(char) and sizeof(uchar)

portante:
Remove the use `sizeof(char)` or `sizeof(uchar)` in calculations for memory
allocation or string length.  There are no known platforms for which
`sizeof(char)` or `sizeof(uchar)` is not 1, and c99 has defined `sizeof(char)`
to be `1` (section 6.5.3.4 of c99).

### DIFF
--- a/action.c
+++ b/action.c
@@ -387,7 +387,7 @@ actionConstructFinalize(action_t *__restrict__ const pThis, struct nvlst *lst)
 	}
 	/* generate a friendly name for us action stats */
 	if(pThis->pszName == NULL) {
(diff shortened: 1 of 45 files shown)


rgerhards:
@portante just to make sure because some files still under GPLv3 may be touched: are you OK with releasing your patch under ASL? (we plan to migrate to ASL in the medium term).

portante:
@rgerhards, yes, thanks.

rgerhards:
thx!

PR #707, opened 2015-12-16
Add missing json_null{,_array} tests to dist target

Whissi:
These tests were not included in the rsyslog-8.15.0 release tarball.


Whissi:
It was just the *queue-persist.sh* test which failed.

Issue #718, opened 2015-12-29
Rsyslog crashes in imfile module (memory corruption)


zacharyzhao:
### Summary

Seen in v8.14 and reproduced in master code (commit 8e974a3) as well.

OS information:

- CentOS 6.5 (x86_64)
- gcc 4.4.7

Backtrace below, looks like the stack is overflowed.  An example coredump is
accessible at [dropbox](https://www.dropbox.com/sh/0r2kjdxbmp7jvhl/AADfhgoQ7iKxbQLtqzWGky7Wa/rsyslog?dl=0).

Originally found with imfile + omkafka and more easier to reproduce when

- Use imfile + omfile
- Output is blocked (for example make the partition stores the output file pretty small)
- Rename/remove the input file

### Configuration files used to reproduce the problem

```
# more *conf
::::::::::::::
00-template.conf
::::::::::::::

template(name="format_json" type="list") {
    constant(value="{\"host\":\"")
    property(name="hostname")
    constant(value="\"")

    constant(value=",\"ip\":\"")
    constant(value="10.255.255.2")
    constant(value="\"")

    constant(value=",\"app-name\":\"")
    property(name="app-name" caseConversion="lower")
    constant(value="\"")

    constant(value=",\"file\":\"")
    property(name="$!metadata!filename")
    constant(value="\"")

    constant(value=",\"msg\":\"")
    property(name="msg" format="json")
    constant(value="\"")

    constant(value=",\"time\":\"")
    property(name="timestamp" dateFormat="rfc3339")
    constant(value="\"")

    constant(value="}")
}

::::::::::::::
01-ruleset.conf
::::::::::::::

module(load="builtin:omfile")

ruleset(name="rule_omfile") {
    action(type="omfile"
           template="format_json"
           file="/dev/shm/rsyslog-debug-output.log"

           action.resumeRetryCount="-1"
           queue.type="LinkedList"
           queue.size="100000"
           queue.spoolDirectory="/export/servers/rsyslog/var/spool"
           queue.filename="queue"
           queue.maxfilesize="64M"
           queue.maxdiskspace="4G"
           queue.highwatermark="99000"
           queue.lowwatermark="80000"
           queue.saveonshutdown="on"
           queue.dequeuebatchsize="16"
          )
}

::::::::::::::
11-input-file.conf
::::::::::::::

module(load="imfile" mode="inotify")

input(type="imfile"
      ruleset="rule_omfile"
      addmetadata="on"
      tag="test"
[...]

PR #720, opened 2015-12-31
Remove bad use of inotify_event structure

portante:
Fixes # 718.

We also clean up whitespace in a separate commit so that `git diff --check` is helpful in the future.

### DIFF
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -12,11 +12,11 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *       http://www.apache.org/licenses/LICENSE-2.0
  *       -or-
  *       see COPYING.ASL20 in the source distribution
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -174,7 +174,7 @@ struct modConfData_s {
 static modConfData_t *loadModConf = NULL;/* modConf ptr to use for the current load process */
 static modConfData_t *runModConf = NULL;/* modConf ptr to use for the current load process */
 
-#if HAVE_INOTIFY_INIT 
+#if HAVE_INOTIFY_INIT
 /* support for inotify mode */
 
 /* we need to track directories */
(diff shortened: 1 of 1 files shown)


portante:
@Dieken, because the entire call is now under the Debug flag, so there is no need to use the macro which does the same.

rgerhards:
@portante sorry for possible post-vacation blureness: I fail to see how this fixes #718 As far as I see, the "mocked-up" ev only is used for display purposes. Could you point me bluntly to the point? ;)

Note that I am not saying the new method is better - it *is*. But I want to make sure #718 is actually fixed.

On the file display function: that was originally meant to be a temporary debug aid. I will remove it with https://github.com/rsyslog/rsyslog/commit/f4e9da0e6231e618104851c17f68b9f0ccb4e9d4. I know this will cause merge conflicts, but I'll solve them manually, so nothing you need to bother about.

ymattw:
Well.. See the first comment under #718, the original contributor wrote
very detailed information in his commit log.

zacharyzhao:
`evFileHelper.name` is not allocated and not used at all, in typical
case its address is the same as the local variable `etry` in the same
stack (compiler optimization), assigning to `evFileHelper.name[0]`
actually corrupts the etry data structure and crashes the program.

Here's an example output from gdb session.

```
(gdb) p &evFileHelper.name
$6 = (char (*)[]) 0x7ffff6dfc930
(gdb) p &etry
$7 = (wd_map_t **) 0x7ffff6dfc930
(gdb)
```

See clearly in my commit: 9213531

rgerhards:
@zacharyzhao ah, now I see it - I hadn't looked at #719 at all, because that was already closed (abandoned IMO). But now I am a bit confused on given proper credits: looking at #718, #719 and #720 I think it would at least be useful to give credit to @zacharyzhao and @portante . Also, could @portante include the helpful commit text from #719 into his commit? I think that would be useful in the future.

portante:
I hope @zacharyzhao gets the attribution.  I just quickly saw a small step of clean up.  I'll update the commit comment with the initial text from #719 referencing @zacharyzhao.

ymattw:
I would suggest to merge #719 first because that's a minimal fix.  IMHO if you want to clean up useless variable, fix indention problem, or debug logging, better to do that in a separate pull request.

portante:
@ymattw the simple fix does not appear to be right, as the `len` field is still set to a value that does not represent that allocated length of that structure.  I believe removing the variable entirely is the best path forward to eliminate confusion.

rgerhards:
@portante if you like, I can merge things together manually. I'd indeed start with #719 and then merge this one on top. I need to do conflict resolution in any case, so that's probably not much extra work for me. Let me know.

portante:
This is fine by me.  I am not worried about having my name on any patch, but would like to see the confusing piece of this code removed.  Thanks!

rgerhards:
2016-01-14 18:07 GMT+01:00 Peter Portante <notifications@github.com>:

> This is fine by me. I am not worried about having my name on any patch,
>

I never wanted to imply this, sorry if it sounded so. I know that's not the
case!

My main intent was to safe you from doing work that's not strictly
necessary. I'll see that I can do the merge tomorrow (unfortunately quite
busy :-( ).

Thanks for all your help,
Rainer

Issue #729, opened 2016-01-12
Add a UTC equivalent of $NOW family of system properties


rgerhards:
This is useful if someone wants to add a normalized timestamp to templates. [request via Adiscon support customer]

portante:
+1

Could we also add a modify to convert a given timestamp field in UTC?

rgerhards:
I would suggest to open a new issue tracker for this. It's unfortunately a pretty complex beast to do.

portante:
Would you do something like, `$NOWUTC`?

rgerhards:
upcoming: https://github.com/rgerhards/rsyslog/commit/4d73026d0f804ae106436334ba84275967a7f656

rgerhards:
We actually need to treat this and #731 separately, unfortunately. The reason is that internally the $NOW family of system properties does not generate a timestamp but rather a string. So we cannot use the usual date modifiers with it, the format is fixed. And, yes, we should probably look at some point into refactoring the timestamp-handling code inside rsyslog...

Issue #730, opened 2016-01-12
Add a way to emit timestamps in UTC


portante:
From #729, could we add a way to specify timestamps be emitted in UTC?  Like we have `date-rfc3339`, we could have `dateutc-rfc3339`.  For example, perhaps we have today:
```
template(name="tgtmpl" type="string" string="%timegenerated:::date-rfc3339%")
```
and the possible enhancement might look like:
```
template(name="tgtmpl" type="string" string="%timegenerated:::dateutc-rfc3339%")
```

rgerhards:
Note: as UTC is requested (and not any unspecified timezone), this is simpler than usual TZ conversion. However, it's still more complicated if we don't know the zone (RFC3164). However, the code currently assumes local TZ in that case, and we could extend that to the new format specification.

PR #731, opened 2016-01-12
Add system time properties in UTC

rgerhards:
closes https://github.com/rsyslog/rsyslog/issues/729

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ before_install:
   - travis_retry sudo apt-get update -qq
 
 install:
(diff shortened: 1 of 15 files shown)


rgerhards:
NOTE: I hold merging this PR because we may find a more generic solution while thinking over https://github.com/rsyslog/rsyslog/issues/730. However, we should merge if can't get this changed in time for the next release.

rgerhards:
See comment added to #730: $NOW needs to be treated separately, so I go ahead and merge this PR.

PR #739, opened 2016-01-13
cleanup: remove dbgprintf's not intended to stay

rgerhards:
those starting at column 0 are temporary debug aids during development.
Some of them I intend to keep in for a version or two. Some of them
I tend to forget later. This is a cleanup of those. Most are removed,
some are promoted to be permanently there - now indicated by proper
indention.

### DIFF
--- a/action.c
+++ b/action.c
@@ -63,7 +63,7 @@
  * beast.
  * rgerhards, 2011-06-15
  *
- * Copyright 2007-2015 Rainer Gerhards and Adiscon GmbH.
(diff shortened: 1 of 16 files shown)


ymattw:
Hi, if this is not urgent, I suggest to hold off merging because it will
likely cause lots of conflict when you try merging other pending pull
requests, for example, several PRs to imfile. Just a reminder.  Thanks.

rgerhards:
@ymattw Actually, those other commits triggered this one. Once I merge them, I will no longer find what was intended to be temporary ;) So I will merge this one, even though I know that means I have to deal with manual conflict resolution.

PR #762, opened 2016-01-29
Fix statan findings

rgerhards:
new version of clang static analyzer seems to find some new issues...


portante:
FWIW, looks good to me.

Issue #769, opened 2016-02-02
omjournal does not support templating


bobthemighty:
Hi,

I'm trying to set up rsyslog to handle json-logstash messages from the syslog docker log driver.

I want to ship raw json to redis or kafka for later processing, but also put a human-readable version of the log into the local journal.

I've got a first, hackish approach to a templated omjournal and I'm happy to fix it up and open a PR, but would you prefer:

1)
``` 
template(name="foo" type="list") {
   property(name="some-expr" outname="message")
   property(name="some-expr" outname="severity")
}

action(type="omjournal" template="foo")
```

where structured data are automatically mapped to the appropriate systemd-journal field, or

2)

```
template(name="journal-message" type="string" string="some-expr") 
template(name="journal-severity" type="string" string="some-expr")
action(type="omjournal" 
    message-template="journal-message" 
    severity-template="journal-severity"
)
```

where the user configures a separate template for each item or 

3) Something else entirely?

I think as an end-user, my preference is 1, but it seems less flexible.

Thoughts?

bobthemighty:
Actually, option 1 makes it really simple to add in arbitrary fields, so I'm going to plough on with that.

rgerhards:
implemented via #770

PR #770, opened 2016-02-02
Adds simple templating to omjournal

bobthemighty:
As per #769 I wanted to transform log lines before they were sent to the journal. 

I've hacked about a bit, and have something that works, based on the ommongodb module. Each entry in the OMSR_TPL_AS_JSON is dumped into the journal. The caveat is that if you don't include a MESSAGE field, then you don't see anything useful.

```
template(name="journaljson" type="list"){
 property(name="$!@message" outname="MESSAGE") 
 property(name="$!@fields!levelname" outname="LEVEL") 
 property(name="$!@fields!funcName" outname="CODE_FUNC")
 property(name="$!@fields!lineno" outname="CODE_LINE")
 property(name="$!@fields!filename" outname="CODE_FILE")
 property(name="syslogtag" outname="SYSLOG_IDENTIFIER")
}

action(type="omjournal" template="journaljson")
```

If you don't provide a template, then the module continues to behave as before.

Seems to build under both gcc and clang, though the static analyzer complains about a possible uninitialised value when I'm freeing the pointers to the iovec strings. I'm 99.5% convinced that it's a false positive, but if you've any suggestions on how I clean that up, I'm all ears.

I'd also really like to add a new property replacer for converting log level names (INFO, WARN, FATAL etc.) to syslog severity levels. Any pointers would be warmly welcomed.

### DIFF
--- a/plugins/omjournal/omjournal.c
+++ b/plugins/omjournal/omjournal.c
@@ -35,12 +35,16 @@
 #include <errno.h>
 #include <unistd.h>
 #include "conf.h"
+#include "cfsysline.h"
+#include <json.h>
 #include "syslogd-types.h"
 #include "srUtils.h"
 #include "template.h"
 #include "module-template.h"
 #include "errmsg.h"
 #include <systemd/sd-journal.h>
+#include "unicode-helper.h"
+#include <sys/uio.h>
 
 MODULE_TYPE_OUTPUT
 MODULE_TYPE_NOKEEP
@@ -54,22 +58,30 @@ DEF_OMOD_STATIC_DATA
 
 
 typedef struct _instanceData {
+    uchar *tplName;
 } instanceData;
 
 typedef struct wrkrInstanceData {
 	instanceData *pData;
 } wrkrInstanceData_t;
 
+static struct cnfparamdescr actpdescr[] = {
+	{ "template", eCmdHdlrGetWord, 0 }
+};
+static struct cnfparamblk actpblk =
+	{ CNFPARAMBLK_VERSION,
+	  sizeof(actpdescr)/sizeof(struct cnfparamdescr),
+	  actpdescr
+	};
+
+
 struct modConfData_s {
 	rsconf_t *pConf;	/* our overall config object */
 };
-static modConfData_t *loadModConf = NULL;/* modConf ptr to use for the current load process */
 static modConfData_t *runModConf = NULL;/* modConf ptr to use for the current exec process */
 
 BEGINbeginCnfLoad
 CODESTARTbeginCnfLoad
-	loadModConf = pModConf;
-	pModConf->pConf = pConf;
 ENDbeginCnfLoad
 
 BEGINendCnfLoad
@@ -102,11 +114,14 @@ ENDcreateWrkrInstance
 
 BEGINisCompatibleWithFeature
 CODESTARTisCompatibleWithFeature
+	if(eFeat == sFEATURERepeatedMsgReduction)
+		iRet = RS_RET_OK;
 ENDisCompatibleWithFeature
 
 
 BEGINfreeInstance
 CODESTARTfreeInstance
+    free(pData->tplName);
 ENDfreeInstance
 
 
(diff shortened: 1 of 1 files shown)


rgerhards:
Thanks for the patch.

I would suggest that you check the reasoning static analyzer provides: often, it then becomes clear that it is indeed a true problem. If it is a false positive, you should normally be able to easily work around it. But I have seen that travis didn't err out -- this may be a good sign, it may also be an indication of a problem in the travis environment.

I would also appreciate if you could squash commits. Those that you only used to correct yourself should better go away. If you haven't done this yet: ```git rebase -i``` is your friend.

Finally, it would be great if you could also update the documentation in our sister project https://github.com/rsyslog/rsyslog-doc

rgerhards:
I had forgotten: have you already run static analyzer locally and looked at the elaborative report (the one shown via scan-view)? If not, that's definitely the next step to take.

bobthemighty:
Hi Rainer,

Happy to oblige. Can you comment briefly on the property replacer?

I have log4j style level names in my incoming logs, and I want to map them
to syslog severity levels so that they end up correctly tagged in the
journal. I can see that there's some functionality in msg.c in this area,
but nothing that's obviously exposed to end-users. Am I missing something?
If not, would you be happy to accept a patch that added a new pair of
property replacers, one to map severity ints to names, and another to map
names to ints?

I *have* run the static analyser with scan-view. The warning I get tells me
that I may be trying to free an uninitialised value when I'm traversing an
array of iovec and freeing the string each one points to. I'll do a bit
more digging.

Thanks,

 -- Bob.

rgerhards:
There are the syslogfacility-text and syslogseverity-text properties. They should do the trick?

regarding static analyer: iovec may actually be a false positive. We still should try very hard to find a work-around, as otherwise we get into trouble with automatted testing. There is some limited advise in the static analyzer doc itself (commenting out code during analyzer run as a last resort).

bobthemighty:
My problem is that the incoming message will have been written with a fixed
severity: my data are coming from the docker syslog driver, which isn't
inspecting application logs.

If a docker container logs to stdout, then the syslog driver will send the
resulting message with severity LOG_INFO, if the container logs to stderr,
then the driver logs with LOG_ERR.

An application in the container will usually have chosen an appropriate log
level of its own accord, and that's the data I want to pull from the
incoming message.

On Wed, Feb 3, 2016 at 10:25 AM, Rainer Gerhards <notifications@github.com>
wrote:

> There are the syslogfacility-text and syslogseverity-text properties. They
> should do the trick?
>
> regarding static analyer: iovec may actually be a false positive. We still
> should try very hard to find a work-around, as otherwise we get into
> trouble with automatted testing. There is some limited advise in the static
> analyzer doc itself (commenting out code during analyzer run as a last
> resort).
>
> —
> <https://github.com/rsyslog/rsyslog/pull/770#issuecomment-179149717>.
>



-- 
Q. How many members of a demographic group does it take to perform a
specified task?

A. A finite number; one to perform the task, and the remainder to act in a
manner stereotypical of the group in question.

rgerhards:
2016-02-03 11:34 GMT+01:00 bobthemighty <notifications@github.com>:

> My problem is that the incoming message will have been written with a fixed
> severity: my data are coming from the docker syslog driver, which isn't
> inspecting application logs.
>
> If a docker container logs to stdout, then the syslog driver will send the
> resulting message with severity LOG_INFO, if the container logs to stderr,
> then the driver logs with LOG_ERR.
>
> An application in the container will usually have chosen an appropriate log
> level of its own accord, and that's the data I want to pull from the
> incoming message.
>
>
would you mind taking this question to the rsyslog mailing list? Reason is
that we don't pollute this bug tracker, especially as I would need to go
pretty off-topic. Also, other members may have advise and for even others
it is educating.

Rainer

bobthemighty:
@rgerhards I've rewritten the code to an uglier form that's more amenable to static analysis.

rgerhards:
@bobthemighty I feel your pain, and have had it often enough myself. But the long-term benefits, I think, justify it. It would be much nicer if one could tell the static analyzer that this is a false positive, tough. I myself have currently two things that I am working  very hard on a workaround, which I have not yet found :-(

PR #771, opened 2016-02-03
Fix lexer bug affecting hex number literals.

Elemecca:
This fixes a mistake in the lexer rule for hexadecimal number literals that prevented it from recognizing hex numbers containing more than one digit or the digits `8` or `9`.


portante:
@Elemecca, nice catch!

PR #774, opened 2016-02-04
travis: enable elasticsearch tests in testbench

rgerhards:
enable tests plus apply fixes for "in the meantime undetected problems". The whole things turns out to be much more work than originally anticipated.


rgerhards:
Please **do not merge** at the moment. I need to add some more fixes, but would not like to withdraw this PR for now.

PR #776, opened 2016-02-05
PROTOTYPE fix for tpls references

portante:
@rgerhards this is just a hack fix from reviewing the code.  I tried to compile it on Fedora but have not built libfastjson for Fedora yet (hopefully that will land there soon).

This is just meant to share what I saw, feel free to delete.


rgerhards:
fixed a little differently: https://github.com/rgerhards/rsyslog/commit/879dd4872ba4feb38476a5533258fa107dfc05b0

I thought this makes clear that ```tpls==NULL``` is actually a different use case (based on analysis of the rest of the code).

PR #779, opened 2016-02-08
adds addceetag option to imfile

taotetek:
We have use cases where people are using imfile to read in pure JSON files.  This pr adds an "addceetag" option to imfile that will prepend each line from a pure json log file with "@cee:" so that the contents can be used with mmjsonparse, etc.


lobsterdore:
This is a great addition :+1:

taotetek:
@lobsterdore I've back ported it as a patch to our in house 8.16 build and it's working very nicely.

PR #790, opened 2016-02-15
Output Module for AMQP 1.0-compliant brokers

kgiusti:
Hello,

I'd like to contribute an output module I have written for use with AMQP 1.0 brokers.  I've tested this against the QPID broker, and the QPID Dispatch Router.

Thanks,

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,6 @@ matrix:
 before_install:
   #- travis_retry sudo apt-get install -qq python-software-properties
   - travis_retry sudo add-apt-repository ppa:adiscon/v8-stable -y
-  - travis_retry sudo add-apt-repository ppa:qpid/released -y
   - travis_retry sudo apt-get update -qq
   - mysql -u root -e "CREATE USER 'rsyslog'@'localhost' IDENTIFIED BY 'testbench';"
   - mysql -u root -e "GRANT ALL PRIVILEGES ON * . * TO 'rsyslog'@'localhost'; FLUSH PRIVILEGES;"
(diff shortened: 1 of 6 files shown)


rgerhards:
I would like to build this under travis. Can you tell me which pacakges I need for ubuntu 12.04 (and 14.04 if it is different)?

kgiusti:
Ah, yes I should update the README with additional detail.

The Proton libraries are not currently in any of the Ubuntu base repos.   The Apache QPID project does have its own PPA up on launchpad - you can get the most recent stuff there:

sudo add-apt-repository ppa:qpid/released
sudo apt-get install libqpid-proton3-dev

That should work for both Trusty and Precise.
For Centos7 and Fedora the latest proton packages are available via epel.
Let me know if you have any problems.

rgerhards:
It looks like there is a problem with libqpdi-proton3-dev. See here:

https://travis-ci.org/rgerhards/rsyslog/jobs/110741422

For a broader overview, have a look at my merge-PR #790

kgiusti:
Ah, sorry - I really should've tried a travis build before I issued the pull request...

I think the problem you've hit is that there needs to be an apt-get update run after the repository is added.  There's also a weird dependency bug with the qpid-proton3-dev package.  I've updated the README with more details regarding installing pre-built packages.

I've also hacked up the rsyslog .travis.yml to build with the module and it now passes.  Feel free to use it as a reference:

https://github.com/kgiusti/rsyslog/commit/488fdccf6bd6b1e82b5d19bc23b339f74b2d4b3c

Let me know if you hit any issues, thanks.

rgerhards:
I have merged based on the previous state and the new information. I'll now see that I can merge in the remaining changes...

rgerhards:
looks like this was straightforward, it's running the integration tests once again. I think it should work witout any problems.

Side-note: would you consider contributing to https://github.com/rsyslog/rsyslog-doc as well? While it is nice to have a readme, users usually don't look at the source tree and so almost noone will know about this feature until it is part of the official doc.

kgiusti:
Awesome - thanks!

I'll create that documentation and send you a pull request shortly

PR #793, opened 2016-02-16
integration PR for #706

rgerhards:
This is a branch to let #706 run through the enhanced travis tests.


rgerhards:
We should wait if https://github.com/rsyslog/rsyslog/issues/794 brings some positive results and then merge this into this branch. Else we really don't do any verification.

Issue #794, opened 2016-02-16
travis: at least compile ommongodb under travis


rgerhards:
best would be to have some tests as well, but that's another issue...

PR #795, opened 2016-02-16
travis: enable building ommongodb

rgerhards:
see also https://github.com/rsyslog/rsyslog/pull/793
closes https://github.com/rsyslog/rsyslog/issues/794


rgerhards:
**Note well: this will be merged via #793**

PR #798, opened 2016-02-17
A CEE cookie shall only be added when using the CEE format

volter:
I think that the flow is wrong here and that the CEE cookie is currently added, regardless whether you pick json, json-elasticsearch or cee for impstats.

### DIFF
--- a/runtime/statsobj.c
+++ b/runtime/statsobj.c
@@ -274,7 +274,7 @@ getStatsLineCEE(statsobj_t *pThis, cstr_t **ppcstr, const statsFmtType_t fmt, co
 
 	CHKiRet(cstrConstruct(&pcstr));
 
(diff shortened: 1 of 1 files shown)


rgerhards:
While I think you are actually right, I wonder how many script we break if we apply the patch. Let me get some feedback from the ML first. My gut feeling is that we need to add an option to remove the "@cee".

volter:
Following up Thomas' comment: What if you declared JSON and CEE to be synonyms (as they effectively are now) and add other formats, like PURE-JSON and PURE-JSON-Elastisearch that exclude the cookie? That should be perfectly compatible.

rgerhards:
ML list consensus is: be bold enough to apply the breaking change

see http://lists.adiscon.net/pipermail/rsyslog/2016-February/042081.html

Issue #732, opened 2016-01-12
Errors with TLS connections in 8.15


Kryai:
After upgrading to 8.15 I have started to get errors in my logs on my RHEL 6.5 rsyslog server that uses TLS.

```
Jan 11 13:01:02 syslog.err rsyslog_server rsyslogd-2078: unexpected GnuTLS error -50 in nsd_gtls.c:530: The request is invalid.  [v8.15.0 try http://www.rsyslog.com/e/2078 ]
Jan 11 13:01:02 syslog.err rsyslog_server rsyslogd-2089: netstream session 0x7f77a8190520 from 172.30.108.105 will be closed due to error  [v8.15.0 try http://www.rsyslog.com/e/2089 ]

Jan 11 13:02:02 syslog.err rsyslog_client rsyslogd-2078: unexpected GnuTLS error -53 in nsd_gtls.c:1628: Error in the push function.  [v8.11.0 try http://www.rsyslog.com/e/2078 ]
Jan 11 13:02:02 syslog.info rsyslog_client rsyslogd-2359: action 'action 8' resumed (module 'builtin:omfwd') [v8.11.0 try http://www.rsyslog.com/e/2359 ]
Jan 11 13:02:02 syslog.info rsyslog_client rsyslogd-2359: action 'action 8' resumed (module 'builtin:omfwd') [v8.11.0 try http://www.rsyslog.com/e/2359 ]
```

I've narrowed down the issue to the following commit in 8.15
https://github.com/rsyslog/rsyslog/commit/1394e0bec2373e85e003b7320c6d12870424d483

Reverting that commit, I've been error-free for a day now, previously errors would be generated hourly. I'm not knowledgeable enough to understand the full extent of the commit, and so further investigation will be required. I can help test any changes as needed.

alorbach:
Strange, I don't see how the commit hould cause these unexpected TLS error's. 
A full debuglog would be helpful.

dbray925:
I am seeing this exact same error, repeatedly throughout the day. I was lucky enough to capture a debug entry as well:

rsyslogd-2078: unexpected GnuTLS error -50 in nsd_gtls.c:530: The request is invalid.  [v8.16.0 try http://www.rsyslog.com/e/2078 ]
rsyslogd-2089: netstream session 0x7f7afc5e8e30 from IP.REMOVED. will be closed due to error  [v8.16.0 try http://www.rsyslog.com/e/2089 ]


```
[...]

PR #801, opened 2016-02-18
gnutls: Fix for issue #732

alorbach:
Commit https://github.com/rsyslog/rsyslog/commit/1394e0bec2373e85e003b7320c6d12870424d483
changed how the receive buffer was freed in rcv() call. However Retry handling
was not considered properly. 

Closes https://github.com/rsyslog/rsyslog/issues/732

### DIFF
--- a/runtime/nsd_gtls.c
+++ b/runtime/nsd_gtls.c
@@ -1592,8 +1592,9 @@ Rcv(nsd_t *pNsd, uchar *pBuf, ssize_t *pLenBuf)
 	*pLenBuf = iBytesCopy;
 
 finalize_it:
-	if (iRet != RS_RET_OK) {
-		/* in this case, we also need to free the receive buffer, if we
+	if (iRet != RS_RET_OK && 
+		iRet != RS_RET_RETRY) {
+		/* We need to free the receive buffer in error error case unless a retry is wanted. , if we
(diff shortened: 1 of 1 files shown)


runtime/nsd_gtls.c
```diff
@@ -1592,13 +1592,16 @@ Rcv(nsd_t *pNsd, uchar *pBuf, ssize_t *pLenBuf)
...
-		/* in this case, we also need to free the receive buffer, if we
+	if (iRet != RS_RET_OK && 
+		iRet != RS_RET_RETRY) {
+		/* We need to free the receive buffer in error error case unless a retry is wanted. , if we
 		 * allocated one. -- rgerhards, 2008-12-03 -- moved here by alorbach, 2015-12-01
 		 */
 		*pLenBuf = 0;
-		free(pThis->pszRcvBuf);
-		pThis->pszRcvBuf = NULL;
+		if(pThis->pszRcvBuf != NULL) {
```
rgerhards: This is not necessary, free(NULL) is defined to be a valid (no-)operation. I just point this out in case you consider this an important part of the fix. Else I would suggest to remove it again (for clarity).

alorbach: In this case thee if(pThis->pszRcvBuf != NULL) can be removed.

rgerhards: will you amend the PR or should I do so after merging (amending would be much cleaner)?

PR #821, opened 2016-02-24
Ksi extract update

alorbach:
Fixes multiple KSI issues in rsgtutil


rgerhards:
There are unfortunately some output/temporary files committed to git:
* tests/rsyslog.out.compare
* tools/rscryutil
* tools/rscryutil.1
Please fix so that I can merge.

rgerhards:
addition: there may be other temporary files, these were just the once that I found when trying to merge. I suggest to check the commits to make sure no such files are included.

alorbach:
Fixed the mess, sry ;)

Issue #786, opened 2016-02-15
test dynstats_overflow-vg.sh sometimes fails


rgerhards:
This test sometimes fails, a typical failure is given below. This might be related to the actual thing that is being tested, but it is also likely caused by some race inside the testbench itself or inside rsyslog. Especially in the later case, this can point into a more severe, general problem. As such, we should try to nail down what is going on [note that at this time we know we have some other racy tests as well, I begin to add issue trackers from them whenever they come up].

@janmejay although there is only a limited chance this might be related to the actual functionality, I would still appreciate if you could find some time to have a look as well.

This can most easily be reproduce by running the test under the loop script like this:
```
./loop.sh ./dynstats_overflow-vg.sh
```
Note: you need to be inside the ./tests directory and $srcdir must be set appropriately.

**Failures seen:**

```
content-missing assertion failed, some line in 'rsyslog.out.stats.log' matched pattern 'corge'
FAIL!
```

**another one:**
```
content-check failed to find 'foo 017 -6', content is
 [origin software="rsyslogd" swVersion="8.17.0.master" x-pid="9736" x-info="http://www.rsyslog.com"] start -1
dynstats: bucket 'msg_stats' is being reset [v8.17.0.master try http://www.rsyslog.com/e/2041 ] -1
foo 001 0
bar 002 0
baz 003 0
foo 004 0
baz 005 0
foo 006 0
quux 007 -6
corge 008 -6
quux 009 -6
foo 010 0
corge 011 -6
grault 012 -6
foo 013 0
dynstats: bucket 'msg_stats' is being reset [v8.17.0.master try http://www.rsyslog.com/e/2041 ] -1
dynstats: bucket 'msg_stats' is being reset [v8.17.0.master try http://www.rsyslog.com/e/2041 ] -1
 [origin software="rsyslogd" swVersion="8.17.0.master" x-pid="9736" x-info="http://www.rsyslog.com"] rsyslogd was HUPed -1
corge 014 0
grault 015 0
quux 016 0
dynstats: bucket 'msg_stats' is being reset [v8.17.0.master try http://www.rsyslog.com/e/2041 ] -1
foo 017 0
corge 018 0
dynstats: bucket 'msg_stats' is being reset [v8.17.0.master try http://www.rsyslog.com/e/2041 ] -1
 [origin software="rsyslogd" swVersion="8.17.0.master" x-pid="9736" x-info="http://www.rsyslog.com"] exiting on signal 15. -1
FAIL!
```

[...]

PR #825, opened 2016-02-28
testbench: temporarily disable dynstats_overflow-vg.sh

rgerhards:
... because this test fails most of the time on slower machines.
As we know there is an issue, it doesn't make sense to let many
CI builds fail while we work on the timing issue.

see also https://github.com/rsyslog/rsyslog/issues/786


rgerhards:
no longer needed, as https://github.com/rsyslog/rsyslog/pull/827 fixes the core problem.

PR #827, opened 2016-02-29
EXPERIMENT: join always, except already exited, proper predicate loop

portante:
This is just a quick attempt to code-up some changes that I saw reviewing the thread termination handling.  I have to spend some more time thinking about this, but the gist of the changes is to work the non-cancel termination under a predicate loop over `bIsActive` to avoid spurious wakeups, only allowing the thread itself to change the state of `bIsActive`.

This passed all Travis checks in my tree, for what its worth (https://travis-ci.org/portante/rsyslog).

### DIFF
--- a/threads.c
+++ b/threads.c
@@ -2,7 +2,7 @@
  *
  * This file implements threading support helpers (and maybe the thread object)
  * for rsyslog.
- * 
+ *
  * File begun on 2007-12-14 by RGerhards
  *
  * Copyright 2007-2012 Adiscon GmbH.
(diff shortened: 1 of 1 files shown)


rgerhards:
Thanks for working on this. Visual inspection lets me think it could nail the problem. I'll give it a try on the slow machines as soon as I can get hold of one (I need to essentially stop it's use for buildbot to do that).

rgerhards:
100% success on that machine (ok 100 out of 100 runs ;)). Now let's do the regular merge process...

Issue #741, opened 2016-01-16
Segfaults using type=imfile if File has no leading /


karl-wiebe-op:
Using both 8.9.0 and 8.15.0, if an input of type imfile has a File= without a leading / (as in the case of a typo), rsyslogd will abort on several Linux distros. The culprit would seem to be https://github.com/rsyslog/rsyslog/blob/master/plugins/imfile/imfile.c#L665 which returns the wrong index, as the presence of a / is assumed. Perhaps a more graceful syntax error would be appropriate?

rgerhards:
fixed via https://github.com/rsyslog/rsyslog/pull/840

PR #840, opened 2016-03-03
fix possible buffer overflow in imfile config parser

taotetek:
If an imfile config accidently used a relative path name with no directory separators, the imfile config parsing routine would cause a buffer overflow.  This PR changes getBasename to return a -1 if a directory separator is not found in the File argument, and subsequently return a config error handling the condition a little more gracefully.


rgerhards:
closes https://github.com/rsyslog/rsyslog/issues/741

rgerhards:
@portante good idea, I have applied your rewording right after the merge

PR #850, opened 2016-03-05
AF_UNIX support in imptcp module

nbrownus:
I need to use SOCK_STREAM AF_UNIX sockets and making imuxsock support it was shaping up to be way more complicated than using something that already deals with sessions. Wanted to get your eyes on this before I wrote up any documentation.

To use it:
```
module(load="imptcp")
input(type="imptcp" path="/path/to/stream.sock")
```

To test it:
```
logger -Tu /path/to/stream.sock "Hello there"
```

### DIFF
--- a/plugins/imptcp/imptcp.c
+++ b/plugins/imptcp/imptcp.c
@@ -166,7 +166,7 @@ static struct cnfparamblk modpblk =
 
 /* input instance parameters */
 static struct cnfparamdescr inppdescr[] = {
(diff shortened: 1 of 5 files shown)


rgerhards:
I still need a bit until I can look up the details, but this sounds like a very useful addition.

rgerhards:
It would also be great if you could add a bit to the testbench. I think it should be fairly easy to clone one of the logger tests.

PR #851, opened 2016-03-05
Fix processing of partial messages over TCP

theinric:
Commit b0e8ce6c added a new variable to track processing of a single
message split into multiple packets. Several bits of code weren't
ported to the new variable and could cause mishandling.
Messages not terminated by a NL were discarded upon connection
termination.

### DIFF
--- a/runtime/tcps_sess.c
+++ b/runtime/tcps_sess.c
@@ -67,7 +67,7 @@ static rsRetVal Close(tcps_sess_t *pThis);
 /* Standard-Constructor */
 BEGINobjConstruct(tcps_sess) /* be sure to specify the object type also in END macro! */
 		pThis->iMsg = 0; /* just make sure... */
-		pThis->bAtStrtOfFram = 1; /* indicate frame header expected */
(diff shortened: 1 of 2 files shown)


theinric:
> would this also apply to the code in imptcp.c?

@portante, there seems to be some benign redundant code but not an obvious bug. The behavior is probably the same as it was in imtcp: partial messages are discarded when the session closes. But the code in imptcp doesn't indicate that this was ever supposed to be handled as opposed to imtcp.
I haven't actually run any tests with imptcp yet.

rgerhards:
@theinric would it be possible to also propose a testbench test that guards against further regressions? Or as an alternative a description of what such a test should do to trigger that problem (I would then try to add one myself). Or would it be sufficient to send a single message without LF and then terminate the connection?

theinric:
> would it be possible to also propose a testbench test that guards against further regressions? 

I had this in mind but I don't have much spare time right now so I'd rather not promise anything.

> Or as an alternative a description of what such a test should do to trigger that problem (I would then try to add one myself). Or would it be sufficient to send a single message without LF and then terminate the connection?

Yes, that should be enough.

PR #863, opened 2016-03-09
fixes incorrect formatting of stats when CEE/Json format is used

janmejay:
This happens because we generate stats making assumption that every statsobj will have atleast one counter. This removes that assumption, so we generate valid json regardless of statsobj having counters (and origin field).

Reported here: http://lists.adiscon.net/pipermail/rsyslog/2016-March/042209.html


rgerhards:
thx for the quick action, much appreciated!

Issue #873, opened 2016-03-11
timegenerated-{uxtimestamp,dateordinal}.sh are failing on x86


Whissi:
On x86 *timegenerated-uxtimestamp.sh* and *timegenerated-dateordinal.sh* are failing:

```
==========================================
   rsyslog 8.17.0: tests/test-suite.log
==========================================

# TOTAL: 218
# PASS:  209
# SKIP:  7
# XFAIL: 0
# FAIL:  2
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: timegenerated-uxtimestamp.sh
==================================

[timegenerated-uxtimestamp]:
Tue Mar  1 12:00:00 UTC 2016
***SUBTEST: check 1970-01-01
rsyslogd started with pid  9734
starting run 1
Sending 1 messages.

00000001 messages sent
runtime: 0.000
End of tcpflood Run
imdiag[13500]: mainqueue empty
***SUBTEST: check 2000-03-01
rsyslogd started with pid  9758
starting run 1
Sending 1 messages.

00000001 messages sent
runtime: 0.000
End of tcpflood Run
imdiag[13500]: mainqueue empty
***SUBTEST: check 2016-01-01
rsyslogd started with pid  9782
starting run 1
Sending 1 messages.

00000001 messages sent
runtime: 0.000
End of tcpflood Run
imdiag[13500]: mainqueue empty
***SUBTEST: check 2016-02-29
rsyslogd started with pid  9806
starting run 1
Sending 1 messages.

00000001 messages sent
runtime: 0.000
End of tcpflood Run
imdiag[13500]: mainqueue empty
***SUBTEST: check 2016-03-01
rsyslogd started with pid  9830
starting run 1
Sending 1 messages.

00000001 messages sent
runtime: 0.000
End of tcpflood Run
imdiag[13500]: mainqueue empty
***SUBTEST: check 2016-03-03
rsyslogd started with pid  9854
starting run 1
Sending 1 messages.

00000001 messages sent
runtime: 0.000
End of tcpflood Run
imdiag[13500]: mainqueue empty
***SUBTEST: check 2016-12-31
rsyslogd started with pid  9878
starting run 1
Sending 1 messages.

00000001 messages sent
runtime: 0.000
End of tcpflood Run
imdiag[13500]: mainqueue empty
***SUBTEST: check 2017-01-01
rsyslogd started with pid  9902
starting run 1
Sending 1 messages.

00000001 messages sent
[...]

PR #875, opened 2016-03-11
Run year-2038 tests only on systems which supports those dates

Whissi:
Proposal for https://github.com/rsyslog/rsyslog/issues/873


Whissi:
Thanks for letting me now. I updated the PR:

- Re-enabled `. $srcdir/diag.sh exit` in *now_family_utc.sh* (was commented out)

- Changed `rsyslog_testbench_require_y2k38_support` in *faketime_common.sh* to exit with `. $srcdir/diag.sh exit` instead of just `exit 0`

PR #878, opened 2016-03-14
Enable socket bindings before network interface is up

Werkov:
Allow rsyslog to bind UDP ports even w/out specific interface being up at the moment.
Rather a proof of concept patch (originally by Marius) to open/track discussion of solution.

Alternatively, rsyslog could be ordered after networking, however, that might have some negative side effects. Also IP_FREEBIND is [recommended][1] by systemd documentation.

Cc: @nirmoy

[1]: https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/#whatdoesthismeanformeadeveloper

### DIFF
--- a/plugins/imudp/imudp.c
+++ b/plugins/imudp/imudp.c
@@ -116,6 +116,10 @@ struct instanceConf_s {
 	int ratelimitInterval;
 	int ratelimitBurst;
 	int rcvbuf;			/* 0 means: do not set, keep OS default */
+	/*  0 means:  IP_FREEBIND is disabled
+	1 means:  IP_FREEBIND enabled + warning disabled
+	1+ means: IP+FREEBIND enabled + warning enabled */
+	int ipfreebind;
 	struct instanceConf_s *next;
 	sbool bAppendPortToInpname;
 };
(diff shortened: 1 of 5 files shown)


nirmoy:
As discussed with @rgerhards earlier, we need some more modification to  this patch(warning message, a config option to disable it).  Currently I am working on it, once I am done I will supersede this PR with new one.

Werkov:
> once I am done I will push those changes to your( @Werkov ) branch. 

Good, I've added you as collaborator to the repository, so you should have permissions to that branch.

nirmoy:
Hi @rgerhards  Please review the patches.

rgerhards:
Would you care to change the numerical values into ```#define ``` constants? I know it's inconsistent in the existing code base, but it would be a step towards getting an overall better state...

nirmoy:
Please review this patch for replacing numerical values to defines. I have added defines in net.h, do you think its okay? nirmoy/rsyslog@ bfbaf0f331177960f5ad302bd4caf89a7b601535

rgerhards:
excellent, thanks. Looks very well.

rgerhards:
For the ChangeLog: whom is to receive credits? No problems with multiple folks, but it's not really clear at the moment to me ;)

Werkov:
Nirmoy did the most, Marius came up with the idea. Put them in the changelog, I was just coordinating :-)

rgerhards:
I guess we need to work a bit more on portability, see http://build.rsyslog.com:8010/builders/rsyslog%20solaris11sparc%20rsyslog/builds/311/steps/shell_4/logs/stdio

I think it's sufficient to define ```IP_FREEBIND``` to 0 if it is undefined. Will probably follow that route (but first wait what the rest of the buildbot environment says).

rgerhards:
wow, thx, you were fast :-)

PR #890, opened 2016-03-19
add FreshStartTail input parameter to imfile

curu:
when deploy rsyslog to a large number of servers. We may only care about new logs after the deployment. In this case, logs generated before should be discarded, otherwise, there may be vast useless message burst on the central rsyslog receiver.

set **FreshStartTail**  to *on* (**default: off**)cause imfile to seek to the end/taill of input file at fresh start, which means only new log events will be processed.

### DIFF
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -116,6 +116,7 @@ typedef struct lstn_s {
 	sbool reopenOnTruncate;
 	sbool addMetadata;
 	sbool addCeeTag;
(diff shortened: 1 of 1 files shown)


rgerhards:
It would be great if you could also update the doc at https://github.com/rsyslog/rsyslog-doc/blob/master/source/configuration/modules/imfile.rst

curu:
@rgerhards , doc updated: https://github.com/rsyslog/rsyslog-doc/pull/235

PR #898, opened 2016-03-22
omkafka: makes kafka-producer on-HUP restart optional

janmejay:
As of now, omkafka kills and re-creates kafka-producer on HUP. This is not always desirable. This change introduces an action param (reopenOnHup="on|off") which allows user to control re-cycling of kafka-producer. 

It defaults to on (for backward compatibility). Off allows user to ignore HUP as far as kafka-producer is concerned.

omkafka error-log-file is still re-opened on HUP (regardless of the value of this flag).

### DIFF
--- a/plugins/omkafka/omkafka.c
+++ b/plugins/omkafka/omkafka.c
@@ -127,6 +127,7 @@ typedef struct _instanceData {
 	pthread_rwlock_t rkLock;
 	rd_kafka_t *rk;
 	int closeTimeout;
+	int bReopenOnHup;
 } instanceData;
 
 typedef struct wrkrInstanceData {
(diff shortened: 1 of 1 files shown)


janmejay:
Will update the documentation once this is merged (in case param name changes, it'll prevent some re-work).

rgerhards:
thanks. I think the param name is fine.

side-note: would you be able to add at least one basic test for omkafka... just asking ;)

janmejay:
We'll have to run kafka for that. It'll be very useful to have though. I'll do it.

Let us think through some more details so I can wire things up:
- Any thoughts on pulling the binary? Do we have a public-cache where we manually push binaries (Rsyslog devs download Kafka binary and push it to the public-mirror)? or do we pull binaries down from releases (from corresponding projects)? 
- How do we deal with common dependencies like JVM and Scala for environments like FreeBSD and Solaris?

Binaries that are not provided by build-env (distros) will have to be made available by us. Things like Elasticsearch, Kafka, HDFS etc all fall in this category. Let us figure out the way forward, then I'll help put it together for Kafka.

In case we already have thought-through this stuff in context of some other dependencies I can read up the setup and figure it out.

rgerhards:
My current stance at this is that I first try to find a distro-provided package. That works for many components rather well. Next I try to find project-created ones. Works somewhat. As a fallback (not yet really done) we can build dependencies from source, but if that is very time-consuming or error-prone, that's IMHO a reason to not run dynamic tests for the component in question.

I do not try to have everything on any build machine. So FreeBSD and Solaris will get a considerable less share of tests (not ideal, but better than nothing). I try to get as much things done on travis, because the travis result is what contributors see when they submit a PR -- all the buildbot tests are much harder to convey (I whish we could setup buildbot to act on PRs, but that's another thing nobody finds time for).

rgerhards:
oh, and we should probably move this to a new issues tracker, as I want to merge this here soon (and we are really going off-topic, shame on me...).

rgerhards:
opened #900 for the tests

Issue #877, opened 2016-03-14
Feature request: add option to sync periodically from streams


portante:
Right now it appears that `runtime/stream.c`'s `strmPhysWrite()` calls `fsync()` on every right when the `sync` option is set.

Have folks considered adding the ability to periodically call `fsync()`, like perhaps once a minute by default?

rgerhards:
Correct me if I am wrong, but I would assume the OS has long written out the data if we use a one minute interval. IMHO the sole purpose of this option is to ensure that data is **actually written to disk after the write call** (well, close to it...).

@davidelang what do you say?

davidelang:
how long a write can linger in memory varies by the OS, the filesystem used, OS settings, and how busy it is doing various things.

I would agree that a 1 min fsync cycle is long enough to probably not be useful, and I'll point out that there are OS/filesystem combinations that have pathological fsync behavior when there are ongoing writes, especially to the same files. So I would be opposed to doing this by default.

we already have the checkpoint interval that lets you tell rsyslog not to fsync after every message but instead every X messages. Why does this not address the need?

portante:
On Mon, Mar 14, 2016 at 4:12 AM, davidelang <notifications@github.com>
wrote:

> how long a write can linger in memory varies by the OS, the filesystem
> used, OS settings, and how busy it is doing various things.
>
> I would agree that a 1 min fsync cycle is long enough to probably not be
> useful,
>
The N second fsync would be an upper bound on the time limit to do the
fsync.  What we have today, as far as I can tell, at least by looking at
straces, is that a write() system call to a file is immediately followed by
an fdatasync(), and then an fsync() of the parent directory.

The write()/fdatasync() is effectively O_DIRECT, so we might want to
consider just opening the log files for writing that way and skip the
buffer cache entirely.

I believe an fsync() on the parent directory is only needed when the
directory entries are changed, files created, or deleted, so I am not sure
that fsync() should be performed with every write().

> and I'll point out that there are OS/filesystem combinations that have
> pathological fsync behavior when there are ongoing writes, especially to
> the same files. So I would be opposed to doing this by default.
>
I agree, I don't think we need to change rsyslog defaults.

> we already have the checkpoint interval that lets you tell rsyslog not to
> fsync after every message but instead every X messages. Why does this not
> address the need?
>
It does address the need to some extent, but we are filling the buffer
cache and then using fdatasync()/fsync() to persist the writes, which seems
[...]

PR #904, opened 2016-03-24
Fsync cursor value to disk

portante:
Without this change, the imjournal state file contents are not
guaranteed to be on disk.  So the notion of persistence is broken
without this change.

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -385,6 +385,7 @@ persistJournalState () {
 			if (fprintf(sf, "%s", cursor) < 0) {
 				iRet = RS_RET_IO_ERROR;
 			}
+			fsync(fileno(sf));
 			fclose(sf);
 			free(cursor);
 		} else {


rgerhards:
I think we should make this optional and the default be the current mode.

portante:
@rgerhards, but I don't see how we can say data was persisted to disk if we have not performed an `fsync()` on the data.

One of guys here lost a bunch of logs leading up to a crash because the data was no persisted.  And then found that rsyslog could not read the systemd journal anymore because the `imjournal.state` file was zero.  Which is a hall mark of no `fsync()`.

rgerhards:
Well, doesn't that mean we distrust the OS? If data is considered persisted only if we do an fsync, wouldn't we then need to do an fsync for each and every write, no matter where it is done? I assume your problem occured after a fatal power outage? Otherwise, I don't see why the OS wouldn't have written the data. 

On the "cannot read due to corrupt file" - I think this is something we should try to solve in imjournal, I guess that can happen due to various "bad things", and if so imjournal should IMHO try to recover by reading from start (at least as a config option). Does that sound fair?

rgerhards:
side-note: I think this boils down to the same (still unanswered) question on the use case I had in https://github.com/rsyslog/rsyslog/issues/877 I don't understand why in your environment fsync() is so important. Nowadays, things should run on a UPS with proper shutdown procedures in place and so I don't see where you could loose the data.

portante:
This is why I posted the article: http://lwn.net/Articles/457667/

portante:
While working in OpenStack Swift for while, it became pretty clear that persisting data to disk is NOT the same as writing it the buffer cache.  In order to ensure data is in a stable or durable state, you have to force the OS to push it out through the page cache, and onto the disk.

davidelang:
portante is correct that the data is not guaranteed safe on disk until the fsyncs happen .

The problem is that doing the fsyncs is so insanely expensive that it's not the right thing to do it by default.

By default, rsyslog has it's queues in ram, so even if it were to persist things like this to disk, unless you also used disk-only (not disk assisted) queues, you will still loose a lot of the most recent logs.

In my opinion, the right answer is to have things sent off-box ASAP (including from inside containers) and have relay systems on each subnet receive the logs and handle them. These relay boxes are not as likely to go down at the same time as the other systems, so it's far less likely that you loose the critical logs (you may loose a few that haven't been sent, but no matter what you do you have that problem)

note that the journal has the same problem, it doesn't fsync data to disk either, so the data may get lost before it even gets to rsyslog if it has to go through the journal first.

I've posted elsewhere about my experiences making rsyslog run in the ultra-reliable mode. They are a bit old now, but I doubt if things have changed much. A run-of-the-mill PC with sata  could handle gig-e wire speeds normally (~400K messages/sec), but in ultra-reliable mode a high-end machine with a PCI-E SSD was 2000 messages/sec on ext3, 4000 on xfs, and 8000 on ext2 with the simplest possible config.

I firmly believe that this could be improved with a re-write if the disk queue code, but that's a fairly major project.

Adding fsync's to spot places like this is NOT the right idea.


what filesystem are you running? some filesystems have more protection than others against empty files being created like this. It may be that we can create an option that overwrites the file in a more reliable manner (mmap of the file and just overwrite the data instead of truncate/rewrite???)

I'll do another post on data reliability. I really should do a full article on the topic, I've been dancing around doing so for a few years.

portante:
@davidelang, this `fsync()` is done once at the point where `imjournal` is trying to persist the data that the admin requested be done in the configuration.  This is not about the log processing pipeline.

It would seem your above note is directed at #877, which we can debate over there.

davidelang:
what did the admin do to say that they wanted the data persisted in this config?

portante:
See the `PersistStateInterval` option of `imjournal`, http://www.rsyslog.com/doc/v8-stable/configuration/modules/imjournal.html

davidelang:
Ok, I read persiststateinterval to be a weaker thing than you do. What it's doing is recording where you are in the file, but it's still not a guarantee that the data will be on disk. Without this, rsyslog only records where it is at shutdown. This tells rsyslog to record it every X messages (the overhead of updating this every message would be prohibitive, even without doing a fsync, although mmap could potentially reduce this to a sane level)

This is protecting against rsyslog crashes, not system crashes.

adding fsyncs to protect against system crashes if significantly more overhead, and a different guarantee. I see how you are seeing this from the point of view you are coming from, but I think there is a lot of value in this option without the fsyncs, so we would need a different option for fsyncs.

rgerhards:
I know that data is not guaranteed safe until fsync(). However, I am doing this for 30+ years now and except in the early days with shaky power sources, you could always assume that the OS successfully persists written data except for fatal problems. These fatal problems always were related to power failures, which really should not be a problem for over 20 years now, I think, The remaining issues were disk subsystem failure (where fsync() doesn't really help) or a fatal glitch in the OS that caused a system panic. With system panic, fsync() is questionable as a solution, because it's basically a race when the OS glitch happens.

Has the status of Linux really so much deterioated that we can no longer rely on data to actually be persisted to disk without fsync(). I don't mean for power fail. I mean for regular operations. If that's the case, then it makes of course sense to put a fsync() after each and every IO request.  If that's actually the case for mainstream Linux, then this patch here makes sense. If so, that also means I begin to hope that Microsoft increases its market share again, and I would also hope that my personal records are not being processed on Linux by service providers... 

If the majority of Linux is still reliable, then this should be an optional call, even though it may not have a performance impact. But again: if we need it **here**, we need it at a myriad of other places as well...

davidelang:
when reading data from a file via imfile or imjournal, and rwirting it out, here is an approximation of the process and where datalos is possible. (this ignores the fact that the systemd journal has potential for loosing data as well due to similar issues)

This is a simplified explination of the process

1. data is read from the source and placed into the main queue (RAM, so vulnerable from this point)
2. the state file is updated to record how much was read (takes time to get out to disk, so where you are could be lost)
3. the log may get moved to an action queue (another RAM queue, so still vulnerable)
4. the log is read from the queue, processed by the om module. The module creates the output string and passes it to the OS buffers and considers it complete (omrelp  and a few others are exceptions to this, everything else assumes the OS will do something with it, and failures outside rsyslog can cause the data to be lost)

Adding a fsync to step #2 without addressing all the other steps means that you are still going to loose data in crashes

portante:
@davidelang, @rgerhards, it seems that folks are pretty satisfied with how rsyslog works today.  I'll close it.

rgerhards:
@portante If there is a platform that has reliablilty problems (openstack you mention), then I think it **makes sense** to add support for those cases. I just don't think it makes sense to force this on all platforms. So I don't think it is good to just close the PR. I still suggest to add support for this, but make it optional. But of course, I leave this to your decision.

davidelang:
If the system crashes, data in the system buffers may be lost (this applies to windows and Linux).

All Operating systems are caching/buffering much more aggressively than they did 30 years ago, it's a large part of where the performance increases have come from. Some filesystems provide more risk in crashes than others

fsync only matters in the case of system crashes, fi the OS isn't crashing/freezing, then fsync is not needed, the OS guarantees a coherent view of the filesystem from it's memory buffers, no matter what's on disk. If you go back a few years to see the fuss over ext4 'loosing' KDE config files, you will find that it was happening to people who were overlocking their systems and crashing them. With VMs, people frequently 'crash' the VM rather than waiting for a proper shutdown.

now, in the case of this file, if rsyslog is truncating the file and then writing data to it, then there is a window where rsyslog could crash and the file would be empty, but if rsyslog is just overwriting the contents, then a rsyslog crash would not result in an empty file, only a OS crash.

rgerhards:
2016-03-24 13:34 GMT+01:00 davidelang <notifications@github.com>:

> If the system crashes, data in the system buffers may be lost (this
> applies to windows and Linux).
>
> All Operating systems are caching/buffering much more aggressively than
> they did 30 years ago, it's a large part of where the performance increases
> have come from. Some filesystems provide more risk in crashes than others
>

full ack, but that doesn't mean that they lose data on a regular basis. I
understood @portante so that on openstack "you never know if anything will
survive"  (to exagregate it a bit). I am fine with data loss und
exceptional conditions, but become very nervous if the same thing happens
under **normal** conditions. Maybe that's the new trend. I've seen it, for
example, with ElasticSearch. You can hope that data you submit and get
acked makes it to the data store... but often it doesn't. As long as it is
only for analytics and search, this may be ok. But if your really depend on
it...

Rainer

davidelang:
we do a fopen with the mode wb, a quick google search isn't telling me what the b means, but w says to truncate the file. This does leave a window where rsyslog could crash and leave an empty file.

I think that for things like this were we have a small amount of data that we want to update frequently, instead of opening, writing/closing the file, we shoudl instead mmap the file, update the memory in an order that guarantees that corruption can be detected, and then have three levels of persisting the data to disk

1. let the OS do it as it chooses (except when we exit)
2. every X messages tell the OS that we want it to write the data out, but don't fsync
3. every X messages tell the OS to write the data out NOW and fsync it (fdatasync in thsi case)

to make the updating safe, we can take advantage of a few properties of modern systems.

1. properly aligned writes to memory of word size are atomic (this allows 64 bit values in cases we really care about)
2. writes to disk happen in block sized chunks, min 512 bytes, 4k is common.

what's the range of this cursor value? if it's <64 bits, we can do this very reliably.

davidelang:
databases have traditionally been very paranoid about getting data to disk, doing a lot of fsyncs, writing things to queues so that if the updates don't happen to the raw files they can recover, etc. This also makes them 'very slow' This is the ACID guarantees, they guarantee that if the database says that the write completed, it will be there even after a crash.

all the new NOSQL databases gain their performance by relaxing the ACID guarantees. They all do this is slightly different ways with different tradeoffs.

I think this is more visibile today for a few reasons

1. things are being managed by people who aren't DBAs, and in many cases don't have the sysadmin experience to know about this sort of issue, so they get caught by surprise when things happen (and are vocal about blaming X for loosing their data). This isn't bashing them, it's usually something that just hasn't come up. Even people working for decades in the field may not have run into the issue enough to understand it (as this discussion is showing)

2. with more stuff running in VMs, especially in 'the cloud', the people running the system are more isolated from the realities of running hardware, and the hosting party may stop/crash/restart the system without notice (for any or no reason as far as the people running the system know). In the past, the people running the system handled things down to the power level, so only power outages were a surprise to them, everything else was planned and the system was shutdown much more cleanly.

davidelang:
by the way, I think the mmap approach will be faster/less overhead as well as being more reliable.

It's possible that I am mistaken on this and mmap has more overhead than I think it does, but avoiding the open/close syscalls will be a win.

davidelang:
this lwn.net article touches on mmap http://lwn.net/Articles/357767/

If I understand things correctly, if you just mmap a file with MAP_SHARED, any changes to the data will get written back at some point in the future at the convenience of the OS (a write makes the page as 'dirty' and the OS will eventually write it out), you can do a msync() in either async mode (which means you don't really know when/if the write completes) or in sync mode (effectively fsync and you know when the data is safe on disk)

This is part of the Single Unix Specification and has been for many years, so it should be safe and reliable across all *nix flavors. But it's worth checking with *BSD, Solaris and AIX guys just to be on the safe side.

portante:
@rgerhards, I want to make it really clear: OpenStack Swift does not have reliability problems.  I only said that I had worked on OpenStack Swift and learned about the concerns of data durability during that time.  I did not say that OpenStack Swift has reliability problems, or that ANY of the platforms on which it is supported has reliability problems.

rgerhards:
OK, I then understood it incorrectly, but "While working in OpenStack Swift for while, it became pretty clear that persisting data to disk is NOT the same as writing it to the buffer cache." sounded to me like a reliability problem. But I'll remove the reference from the other post.

Giving the rest of the discussion, the use case so seems to be that a power off without proper shutdown looks like what needs to be supported. That indeed requires much more work all over the code.

rgerhards:
I opened an issue tracker (#910) for the hardening, as this is far outside the scope of this PR.

PR #905, opened 2016-03-24
Silence unknown statement errors in debug logs

portante:
When running with debugging enabled, we'll often see the following
error:

    error: unknown stmt type %u during iterateAll

Where "%u" is 4006 or 4007. This patch silences those errors by
adding them to the "ignore" case.


rgerhards:
It should be noted that these can simply be ignored without problems because the function in question just iterates over all action objects, and SET and UNSET (the statements in question) do not affect this iteration in any way. Execution of them happens via a different path. The iteration is just used to activate actions at startup and for HUP processing.

PR #918, opened 2016-03-29
External plugin for sending metrics to SPM Monitoring SaaS

radu-gheorghe:
We use this internally and we thought we should share it. Really easy to implement - looks similar to the Solr plugin in the sense that it pushes batches of messages via HTTP POST. It's useful when you want to plot and alert on numerical data that you extract from logs (like the metrics generated by impstats). Example screenshot:
screen shot 2016-03-29 at 4 28 57 pm


otisg:
Nice, @radu-gheorghe!

PR #919, opened 2016-03-30
testbench: fix only partially executed test

rgerhards:
this is a regression from some month ago, when we changed
the shell calls. Only the first of three tests was executed.


rgerhards:
do **not** merge at the moment. I notice the patch is incomplete

PR #946, opened 2016-04-09
fix json API issues in ommongodb and omjournal

rgerhards:


plugins/omjournal/omjournal.c
```diff
@@ -181,111 +181,105 @@ ENDtryResume
...
+{
+	struct iovec *iov;
+	const char *key;
+	const char *val;
+	size_t key_len;
+	size_t val_len;
+	size_t vec_len;
+	size_t i;
+
+	*argc = json_object_object_length(json);
```
portante, line 194: Would it be worth making sure we are going to be successful before we write to external arguments?  For example, if we have a number of elements, but memory allocation fails, we have written a positive number to `*argc`, but then are returning NULL.  Perhaps instead we should write `*argc` when just before we `return`?

rgerhards, line 194: sounds useful, but could you please open another issue for this. I am working on a different topic (json changes) and wouldn't like to finish this first.

rgerhards, line 194: oh, wait, I misunderstood you. Let me check, you are right, there is something fishy...

rgerhards, line 194: it's not a problem here, actually the return code is what is checked. But it's cleaner to change the sequence.

PR #948, opened 2016-04-09
omjournal cleanup

rgerhards:
@portante can you please have a look at the last commit (a21ff7e)


portante:
That looks clean in omjournal.  Nice work wrestling all this, well worth the effort!

rgerhards:
note: the failed "default" CI was a test with buildbot (we are trying to tie github and buildbot better together).

PR #955, opened 2016-04-14
Fixed the dynstats unusedMetricTtl bug

janmejay:
This was responsible for non-resettable counters being reset on TTL expiry. This portion was missed out in the original implementation.

This doesn't need any documentation update (it just fixes existing deviation from expected behavior and docs already describes the correct behavior).

### DIFF
--- a/plugins/imdiag/imdiag.c
+++ b/plugins/imdiag/imdiag.c
@@ -36,6 +36,8 @@
 #include <netdb.h>
 #include <sys/types.h>
 #include <sys/socket.h>
+#include <pthread.h>
+#include <semaphore.h>
 #if HAVE_FCNTL_H
 #include <fcntl.h>
 #endif
@@ -55,6 +57,7 @@
 #include "queue.h"
 #include "lookup.h"
 #include "net.h" /* for permittedPeers, may be removed when this is removed */
+#include "statsobj.h"
 
 MODULE_TYPE_INPUT
 MODULE_TYPE_NOKEEP
@@ -68,6 +71,7 @@ DEFobjCurrIf(netstrm)
 DEFobjCurrIf(errmsg)
 DEFobjCurrIf(datetime)
 DEFobjCurrIf(prop)
+DEFobjCurrIf(statsobj)
 
 /* Module static data */
 static tcpsrv_t *pOurTcpsrv = NULL;  /* our TCP server(listener) TODO: change for multiple instances */
@@ -76,6 +80,18 @@ static prop_t *pInputName = NULL;	/* there is only one global inputName for all 
 static prop_t *pRcvDummy = NULL;
 static prop_t *pRcvIPDummy = NULL;
 
+statsobj_t *diagStats;
+STATSCOUNTER_DEF(potentialArtificialDelayMs, mutPotentialArtificialDelayMs)
+STATSCOUNTER_DEF(actualArtificialDelayMs, mutActualArtificialDelayMs)
+STATSCOUNTER_DEF(delayInvocationCount, mutDelayInvocationCount)
+
+static sem_t statsReportingBlocker;
+static long statsReportingBlockStartTimeMs = 0;
+static int allowOnlyOnce = 0;
+DEF_ATOMIC_HELPER_MUT(mutAllowOnlyOnce);
+pthread_mutex_t mutStatsReporterWatch;
+pthread_cond_t statsReporterWatch;
+int statsReported = 0;
 
 /* config settings */
 struct modConfData_s {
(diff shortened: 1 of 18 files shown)


janmejay:
Something seems to have gone wrong with the PR build on travis (package installation failed).

rgerhards:
@janmejay FYI: to retrigger, I often simply do a "git commit --amend" without any change. It creates a new SHA, and that triggers the build.

janmejay:
Ah, nice trick! I'll do that next time.

rgerhards:
something seems to be discovered by the dynstats_prevent_premature_eviction-vg.sh test: it either fails with data errors (like the current one), or a failure that points into premature termination and a related memory leak. It's most probably timing related, the fedora23x32 machines is a *very* slow box. But this test consistently fails, so I think it would be good to have a look at it.

rgerhards:
note: you should be able to re-trigger a buildbot built by changing the SHA hash.

rgerhards:
Yes, there is definitely some problem, I tried on my workstation (not slow ;)). This is on Ubuntu 16.04 LTS Beta 2:

```
rger@ubuntu16:~/proj/rsyslog/tests$ ./loop.sh ./dynstats_prevent_premature_eviction-vg.sh

###### RUN:  1 success 0 fail 0
===============================================================================
[dynstats_prevent_premature_eviction-vg.sh]: test for ensuring metrics are not evicted before unused-ttl with valgrind
==90628== Memcheck, a memory error detector
==90628== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==90628== Using Valgrind-3.11.0 and LibVEX; rerun with -h for copyright info
==90628== Command: ../tools/rsyslogd -C -n -irsyslog.pid -M../runtime/.libs:../.libs -f/home/rger/proj/rsyslog/tests/testsuites/dynstats_reset.conf
==90628== 
rsyslogd started with pid  90628
startup-vg still running
will wait for stats push
waiting for stats file 'rsyslog.out.stats.log' to be created
waiting for stats file 'rsyslog.out.stats.log' to be created
waiting for stats file 'rsyslog.out.stats.log' to be created
waiting for stats file 'rsyslog.out.stats.log' to be created
waiting for stats file 'rsyslog.out.stats.log' to be created
waiting for stats file 'rsyslog.out.stats.log' to be created
waiting for stats file 'rsyslog.out.stats.log' to be created
waiting for stats file 'rsyslog.out.stats.log' to be created
waiting for stats file 'rsyslog.out.stats.log' to be created
waiting for stats file 'rsyslog.out.stats.log' to be created
stats push registered
injecting msg payload from: /home/rger/proj/rsyslog/tests/testsuites/dynstats_input_1
imdiag[13500]: 1 messages injected
imdiag[13500]: 1 messages injected
injecting msg payload from: /home/rger/proj/rsyslog/tests/testsuites/dynstats_input_2
imdiag[13500]: 1 messages injected
imdiag[13500]: 1 messages injected
injecting msg payload from: /home/rger/proj/rsyslog/tests/testsuites/dynstats_input_3
imdiag[13500]: 1 messages injected
imdiag[13500]: 1 messages injected
imdiag[13500]: mainqueue empty
will wait for stats push
stats push registered
doing shutdown
imdiag[13500]: mainqueue empty
wait on shutdown
==90628== 
==90628== HEAP SUMMARY:
==90628==     in use at exit: 22,189 bytes in 126 blocks
==90628==   total heap usage: 3,081 allocs, 2,955 frees, 1,401,918 bytes allocated
==90628== 
==90628== LEAK SUMMARY:
==90628==    definitely lost: 0 bytes in 0 blocks
==90628==    indirectly lost: 0 bytes in 0 blocks
==90628==      possibly lost: 0 bytes in 0 blocks
==90628==    still reachable: 22,189 bytes in 126 blocks
==90628==         suppressed: 0 bytes in 0 blocks
==90628== Reachable blocks (those to which a pointer was found) are not shown.
==90628== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==90628== 
==90628== For counts of detected and suppressed errors, rerun with: -v
==90628== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
rsyslogd run exited with 0
content-check failed to find 'baz=2' inside 'rsyslog.out.stats.log'
FAIL!
```

rgerhards:
sorry, there was something wrong with my system, probably a partial build. I got sceptic when the test  consistently failed, cleaned up things, and now it works (but that does not yet mean I can't reproduce).

rgerhards:
500 of 500 runs on my workstation VM without any problems. I now need to see if I can get interactive on a slow machine (maybe monday).

janmejay:
It can fail on a very slow machine.

But it has to be slow enough to have scheduling delays of the order of
100ms. I think there is something I can do to make it resilient to that
kind of delay too. Let me give it a shot on Monday.

janmejay:
A memory leak is certainly not expected (it should have been a
data-assertion failure due to slowness).

rgerhards:
The memory leak is related to thread shutdown. I tried to find out if it relates to some other occasional failures I see, but could not yet confirm. I think the data failure is just not reported in those cases because valgrind is checked first and aborts on failure.

I sometimes used the slow machines intentionally, as they showcased some real bugs which were otherwise extremely hard to find ;)

janmejay:
In this case the timing issue is not a real bug. It is due to stats flush
being delayed more than expected. The data issues happen when impstat was
supposed to start at x+4s but actually started at x+4.2s and test's
tolerance falls short.

rgerhards:
2016-04-16 17:46 GMT+02:00 Janmejay Singh <notifications@github.com>:

> In this case the timing issue is not a real bug. It is due to stats flush
> being delayed more than expected. The data issues happen when impstat was
> supposed to start at x+4s but actually started at x+4.2s and test's
> tolerance falls short.
>

Maybe we can just increase some wait time inside the test? I would prefer
if this runs on the slow slave as well, because it always fails and so I
would need to disable the slave (maybe just this test on the slave).

TBH I can't see, though, if the memory error goes away or if there is an
easy workaround for it.

janmejay:
I'll try to tune it a little on Monday after looking at the failure trace.

In worst case we can always build coordination mechanism that will make
test timing independent.

janmejay:
I think we should enhance imdiag to coordinate with impstats. This will allow everyone to write rock-solid tests around stats regardless of machines being slow or fast.

Working on it.

rgerhards:
sounds like a good plan!

janmejay:
Switched that particular test to use explicit coordination instead of time-based coordination. If this goes well, we can switch more tests that appear unstable to this new coordination mechanism.

janmejay:
@rgerhards This time build didn't trigger on fedora23x32 box. That is the only build that I care about right now, and it didn't build it this time :-(

janmejay:
The builder page says the buildslave is offline.

rgerhards:
Should be back soon.

Sent from phone, thus brief.
Am 18.04.2016 21:18 schrieb "Janmejay Singh" <notifications@github.com>:

rgerhards:
Looks like an exceptionally bad day... Need to check tomorrow morning, sorry...

janmejay:
Np, I'm going to sleep too.

janmejay:
Ok, I think we are good here.

We have this elaborate coordination mechanism now, it is capable of "blocking stats publishing", 
"allowing only one batch of stats to be published and blocking it again after that" and "removing the block completely(back to free form stats publishing)".

The most natural form for these tests is to be time-coordinated, because that is how impstats works in the real world. But when tests get flaky, we can always switch to this (which leaves nothing at the mercy of elapsed-time based coordination).

As of now, 2 tests are migrated over, premature-eviction-protection test and overflow test. Others are all working the old way.

rgerhards:
Sounds great. I'd also say leave the others as they are. I will also evaluate the usefulness of the slow environments, but will first address some dangling issues we have (less frequently, unfortunately) on some other test. In the past, most of them pointed to real issues.

janmejay:
I think we should definitely keep slow environments. It allow one to look at things in a different light, cause interleavings that we are unlikely to see otherwise (very useful for a maintaining a solid test suite).

Issue #927, opened 2016-03-30
Failing test git master: privdropuser.sh


Whissi:
```
rsyslogd: ID for user '' could not be found or error [v8.18.0.master try http://www.rsyslog.com/e/3003 ]
rsyslogd started with pid  80255
imdiag[13500]: mainqueue empty
message indicating drop to gid 250 is missing:
 [origin software="rsyslogd" swVersion="8.18.0.master" x-pid="80255" x-info="http://www.rsyslog.com"] start
ID for user '' could not be found or error [v8.18.0.master try http://www.rsyslog.com/e/3003 ]
 [origin software="rsyslogd" swVersion="8.18.0.master" x-pid="80255" x-info="http://www.rsyslog.com"] exiting on signal 15.
FAIL privdropuser.sh (exit status: 1)
```
```
# id 250
uid=250(portage) gid=250(portage) groups=250(portage)
```

rgerhards:
Could you have a look at the shell script if there is something non-portable for your platform?

Whissi:
Interesting, thought it could be a permission problem, because the user running the test suite was already limited. But when running as root, all 4 privdrop tests are failing. Yes, I'll look into it.

rgerhards:
Its"dropping " to the same user to avoid permission issues.

Sent from phone, thus brief.
Am 30.03.2016 19:03 schrieb "Thomas D." <notifications@github.com>:

Whissi:
OK. The problem is that the checks are relying on `$USER` variable which isn't guaranteed to exist.
We should use `$UID` instead or maybe even better `$EUID`.

But because of the problems with the other tests (when running tests as root these tests cannot work) I want to prepare the following changes:

- Create *privs.common.sh* like I did for the timestamp tests
This script will check/validate ID

- The user should have the possibility to specify a user to use in this test.

- To keep it running without user interaction I am thinking about the following logic:
  1. If the user has set an ID we should use the user's value (after validation)
  2. If the user didn't specify an ID, try to use "rsyslog" first, "syslog" second, "daemon" third and finally current `$EUID`

- If we don't end with a valid UID we can work with, skip the tests


What do you think? Should I prepare a PR?

Whissi:
Mh, we can only try to use different users when running as *root*. Normal users don't have the privileges to run as different user. So if running as non root, we can only use `$EUID`... however does it make sense to test in that case because we are not really dropping privs...

rgerhards:
[...]

PR #956, opened 2016-04-15
Make privdrop tests more portable and robust

Whissi:
This is my proposal.

But this fix for issue #927 will uncover a new problem: When we are really testing impersonating another user (which root can only do per default) privdrop tests will fail because the user we drop to probably cannot write to *$SRCDIR/*. So the tests are failing because files like *rsyslog.out.log* or *rsyslogd.started* cannot be created and the test will run into a timeout.

This will be the case when running the testbench as root and for example the user didn't specify a user we should use so that we will fallback to *daemon* user.

Making *tests* world writeable would address this problem but only if the source path is also accessible by the user we will drop to (not guaranteed).

From my current view the right solution would be to create these files in */tmp* (sub dir). This dir could be world writeable and due to the fact that we control the whole path we wouldn't risk that the path isn't accessible.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,9 @@
 ------------------------------------------------------------------------------
 Version 8.18.0 [v8-stable] 2016-04-19
+- testbench: When running privdrop tests testbench tries to drop
+  user to "rsyslog", "syslog" or "daemon" when running as root and
+  you don't explict set RSYSLOG_TESTUSER environment variable.
+  Make sure the unprivileged testuser can write into tests/ dir!
 - templates: add option to convert timestamps to UTC
   closes https://github.com/rsyslog/rsyslog/issues/730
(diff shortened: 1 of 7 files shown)


rgerhards:
It looks like some files are missing in EXTRA_DIST.

Whissi:
Not forgotten, I just wanted to test your Travis `make dist` work ;-)
Thanks!

- *tests/privdrop_common.sh* added to *tests/Makefile.am*
- *ChangeLog* entry added to tell users about *RSYSLOG_TESTUSER* environment variable and that *tests/* must currently be writable for the unprivileged user when running testbench as root

rgerhards:
hehe...

rgerhards:
Note the Solaris test fails due to a glitch on that buildbot slave (not under our control) - we have disabled it for further builds.

PR #959, opened 2016-04-16
Test to see if we can trivially fix the timing issue in #955

rgerhards:
includes #955


rgerhards:
ok, this was an obvious fail ;)

PR #961, opened 2016-04-18
WIP - Enhancements to imfile

portante:
Please consider this a Work-In-Progress and do not merge as is.  It needs some better test coverage of the changes at the least.

Each of the four changes should be fairly contained so that they could be cherry picked if desired.

The last two I am interested in seeing them land in some form: explicit ratelimiting for imfile and offset metadata for messages read from files.

### DIFF
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -468,11 +468,10 @@ static rsRetVal enqLine(lstn_t *const __restrict__ pLstn,
 	MsgSetInputName(pMsg, pInputName);
 	if (pLstn->addCeeTag) {
 		size_t msgLen = cstrLen(cstrLine);
-		const char *const ceeToken = "@cee:";
-		size_t ceeMsgSize = msgLen + strlen(ceeToken) +1;
+		size_t ceeMsgSize = msgLen + CONST_LEN_CEE_COOKIE + 1;
 		char *ceeMsg;
 		CHKmalloc(ceeMsg = MALLOC(ceeMsgSize));
-		strcpy(ceeMsg, ceeToken);
+		strcpy(ceeMsg, CONST_CEE_COOKIE);
 		strcat(ceeMsg, (char*)rsCStrGetSzStrNoNULL(cstrLine));
 		MsgSetRawMsg(pMsg, ceeMsg, ceeMsgSize);
 		free(ceeMsg);
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -105,6 +105,8 @@ typedef struct lstn_s {
 	uint32_t trimLineOverBytes;
 	int nRecords; /**< How many records did we process before persisting the stream? */
 	int iPersistStateInterval; /**< how often should state be persisted? (0=on close only) */
+	int ratelimitInterval;
+	int ratelimitBurst;
 	strm_t *pStrm;	/* its stream (NULL if not assigned) */
 	sbool bRMStateOnDel;
 	sbool hasWildcard;
@@ -129,6 +131,8 @@ static struct configSettings_s {
 	uchar *pszBindRuleset;
 	int iPollInterval;
 	int iPersistStateInterval;	/* how often if state file to be persisted? (default 0->never) */
+	int ratelimitInterval;
+	int ratelimitBurst;
 	int iFacility; /* local0 */
 	int iSeverity;  /* notice, as of rfc 3164 */
 	int readMode;  /* mode to use for ReadMultiLine call */
@@ -145,6 +149,8 @@ struct instanceConf_s {
 	uchar *pszBindRuleset;
 	int nMultiSub;
 	int iPersistStateInterval;
+	int ratelimitInterval;
+	int ratelimitBurst;
 	int iFacility;
 	int iSeverity;
 	sbool bRMStateOnDel;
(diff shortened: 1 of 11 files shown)


plugins/imfile/imfile.c
```diff
@@ -490,8 +490,12 @@ static rsRetVal enqLine(lstn_t *const __restrict__ pLstn,
 	MsgSetTAG(pMsg, pLstn->pszTag, pLstn->lenTag);
 	msgSetPRI(pMsg, pLstn->iFacility | pLstn->iSeverity);
 	MsgSetRuleset(pMsg, pLstn->pRuleset);
-	if(pLstn->addMetadata)
+	if(pLstn->addMetadata) {
+		uchar sOffset[80];
```
richm, line 494: 80 seems like overkill - 23 should suffice, or just use 32 for a nice power of 2.

runtime/obj.c
```diff
@@ -667,8 +667,8 @@ rsRetVal objDeserializeProperty(var_t *pProp, strm_t *pStrm)
 finalize_it:
 	if(Debug && iRet != RS_RET_OK && iRet != RS_RET_NO_PROPLINE) {
 		strm.GetCurrOffset(pStrm, &offs);
-		dbgprintf("error %d deserializing property name, offset %lld, step %d\n",
-			  iRet, offs, step);
+		dbgprintf("error %d deserializing property name, offset %llu, step %d\n",
+			  iRet, (long long unsigned)offs, step);
```
richm, line 670: Why does this have to change from %lld to %llu?

runtime/stream.h
```diff
@@ -216,6 +217,6 @@ strmGetCurrFileNum(strm_t *pStrm) {
 /* prototypes */
 PROTOTYPEObjClassInit(strm);
 rsRetVal strmMultiFileSeek(strm_t *pThis, int fileNum, off64_t offs, off64_t *bytesDel);
-rsRetVal strmReadMultiLine(strm_t *pThis, cstr_t **ppCStr, regex_t *preg, sbool bEscapeLF);
+rsRetVal strmReadMultiLine(strm_t *pThis, cstr_t **ppCStr, regex_t *preg, sbool bEscapeLF, off64_t *offset);
```
richm, line 221: This is a binary incompatible change to the plugin API, but it looks like imfile is the only one that uses this API, at least in the distributed rsyslog source code.  What is the policy here?  Usually this entails adding a new function e.g. strmReadLineWithOffset or something like that, and change the implementation of strmReadLine to call strmReadLineWithOffset with NULL or a dummy parameter for the offset argument.

runtime/obj.c
portante, line 670: This consistent with other locations in the code emitting 64 bit unsigned integers.

runtime/stream.h
portante, line 221: I don't know what the policy in rsyslog for support of plugins not maintained in the source tree.  Is there an rsyslog development release type packaging out there?  Or are folks just compiling against the checked-out git tree?

Certainly it would be easy to write such a compatibility layer to do that if we have to support external plugins.

richm, line 221: I don't know either, just trying to anticipate what the rsyslog maintainers will object to.

rgerhards, line 221: Some time (years) ago, we tried to modularize everything, so that thrid parties could make use of it. However, this hasn't really worked out. First of all, nobody really used it. Secondly, it gets very complicated in regard to version mixes. So there is no need to preserve these interfaces. In fact, as a very low priorty TODO item I have a refactoring of this to something more simple and efficient (regular dynamic linking).

richm:
This PR is marked as "waitingOnResponse" - what response is being waited for?

rgerhards:
@richm See initial posting: @portante said he wants to add tests before this is to be considered for merge.

rgerhards:
@portante Do you still consider this WiP and plan to add some tests? If not, I would see if we could finally merge it...

portante:
@rgerhards, yes I should finish this, and would like to finish this, but would love a little help to get the tests right.

rgerhards:
@portante I can gladly help. Maybe I could create a model commit for a new tests that you suggest? In any case, it might make sense to wait one or two weeks before rebasing and continuing, as @alorbach is currently working on a big overhaul of imfile that will permit wildcards in directory names.

portante:
I should probably rebase none-the-less, in preparation to shake out any bit rot in the mean time.

Waiting for @alorbach's commit sounds like a good idea.

rgerhards:
closing this as the code bases have greatly diverged. If you'd like to continue on it,please open a new PR.

rgerhards:
well... I re-open as I think I may cherry-pick some of the commits. Let me see before final close...

portante:
@rgerhards, sorry for letting this languish.  I hope we can merge these four changes, as they would be really useful.

rgerhards:
it would definitely be helpful if you could do a rebase. The Code base has
changed quite a bit...

rgerhards:
I'll probably cherry-pick some commits to resolve this from being dormant.

On the ratelimit commit, I am not sure this is good to do it that way: if we read a large file on startup, won't we discard messages?

portante:
@rgerhards, in the end, I think we want to be able to read the file without discarding messages just at a slower rate.

davidelang:
while you are working on imfile, we recently added metadata to indicate where in 
the file we are, can we add metadata to give the file size at the time (so that 
monitoring software can tell if you are falling behind on the file if it's rate 
limtied)

rgerhards:
Yeah, thought so. The rate limiter rsyslog currently provides does not
provide that. It just discards excessive messages.

Rainer

Sent from phone, thus brief.

Am 21.10.2017 20:32 schrieb "Peter Portante" <notifications@github.com>:

rgerhards:
The functionality of commit 21712df was added in this PR https://github.com/rsyslog/rsyslog/pull/1758

rgerhards:
added separate issue https://github.com/rsyslog/rsyslog/issues/1883 for the rate-limiting functionality

rgerhards:
Will merge remaining two commits via #1884, then close this PR.

PR #966, opened 2016-04-19
Resolves that rsyslog waits after reboot

radosroka:
Rsyslog waits after reboot due to endless loop,
because wtiGetState(wti_t *pThis) checks only wti state bIsRunning,
but when the thread is canceling by pthread_cancel,
this flag is not unset so we do need check if the thread
terminated some other way.

The bug appears when we reboot system and Rsyslog has lot of works to do(close tcp connections, save all messages to disk etc.), so some timeouts ran out and Rsyslog needs to cancel threads.

I think it could be harder to reproduce with newer Rsyslog but problem still exists. It's simply reproducible on v7.4.7.

Here is the reproducer:
- Client:
  /etc/rsyslog.d/tcpsender.conf

$ActionQueueFileName tcprule # unique name prefix for spool files
$ActionQueueMaxDiskSpace 1g   # 1gb space limit (use as much as possible)
$ActionQueueSaveOnShutdown on # save messages to disk on shutdown
$ActionQueueType Direct   # run asynchronously
$ActionResumeRetryCount -1    # infinite retries if host is dow 
_._ @@192.168.122.1:20000
- Server:
  /etc/rsyslog.conf

$ModLoad imjournal # provides access to the systemd journal
$ModLoad imtcp
$InputTCPServerRun 20000 
$ActionFileDefaultTemplate RSYSLOG_TraditionalFileFormat
$IMJournalIgnorePreviousMessages on

_.info;mail.none;authpriv.none;cron.none                /var/log/messages
authpriv._                                              /var/log/secure
mail.\*                                                  -/var/log/maillog
cron.\*                                                  /var/log/cron
_.emerg                                                 :omusrmsg:_
uucp,news.crit                                          /var/log/spooler
local7.\*                                                /var/log/boot.log

Steps to reproduce:
- Server:
  setenforce 0
  systemctl stop firewalld
  systemctl start rsyslog
- Client:

disable selinux permanently
systemctl disable rsyslog

systemctl start rsyslog
while true; do logger "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"; done
# after few seconds ctrl+C

systemctl reboot

If you cannot reproduce, let the while running longer.
Sometimes is good to remove /var/lib/rsyslog/tcprule\* before test.

### DIFF
--- a/runtime/wti.c
+++ b/runtime/wti.c
@@ -160,9 +160,10 @@ wtiCancelThrd(wti_t *pThis)
 		DBGOPRINT((obj_t*) pThis, "canceling worker thread\n");
 		pthread_cancel(pThis->thrdID);
 		/* now wait until the thread terminates... */
-		while(wtiGetState(pThis)) {
+		while(!pthread_kill(pThis->thrdID, 0)) {
 			srSleep(0, 10000);
 		}
+	wtiSetState(pThis, 0);
 	}
 
 	RETiRet;


runtime/wti.c
```diff
@@ -160,9 +160,10 @@ wtiCancelThrd(wti_t *pThis)
 		DBGOPRINT((obj_t*) pThis, "canceling worker thread\n");
 		pthread_cancel(pThis->thrdID);
 		/* now wait until the thread terminates... */
-		while(wtiGetState(pThis)) {
+		while(!pthread_kill(pThis->thrdID, 0)) {
```
portante, line 163: I get the intention of this fix here: wait for thread termination by looking for the continued existence of the thread instead of spinning on our state variable which the thread may never have a chance to change.

Why not call `pthread_join()` here without a loop after issuing a `pthread_cancel()`?  I don't understand the comments on lines 134 - 141, since a `pthread_cancel()` with a `pthread_join()` seems sufficient.  If a `pthread_kill()` will eventually fail because it returns `ESRCH`, won't a `pthread_join()` eventually return as well?  This is the reason the `pthread_join()` method exists, I believe.

So the man page for pthread_kill() (see [linux's]() and the [opengroup's](http://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_kill.html)) states that implementations might return `ESRCH` when a thread no longer exists, but does not guarantee it, and the use of a thread ID that no longer exists in some implementations could cause a segmentation fault.  So this method of canceling a thread that is "better than" `pthread_cancel()` is a bit dubious to begin with here.

If for whatever reason, we must call `pthread_kill()` in a loop, why not add the loop where we first called `pthread_kill()` (after checking for an error status from that `pthread_kill()` first) in a bounded way and then move on to `pthread_cancel()`?

So I would think that if we issue `pthread_kill()` at line 153 above, there is no guarantee that even in a 10 ms window the "killed" thread gets a chance to run and exit before this thread checks the state on line 158.  Having a bounded loop before line 158 would address that?  If the `pthread_kill()` on line 153 returns an error that the thread does not exist, `ESRCH`, then we don't need to do any of this, right?

radosroka, line 163: Yes, you are right It is option as well, but pthread_kill with signal 0 only checks if thread is alive so we can do a loop. If pthread_kill fails than we brake the loop and continue to shutdown.

portante, line 163: @radosroka but that is what `pthread_join()` does.

theinric:
> Rsyslog waits after reboot due to endless loop, because wtiGetState(wti_t *pThis) checks only wti state bIsRunning,
> but when the thread is canceling by pthread_cancel, this flag is not unset so we do need check if the thread terminated some other way.

I haven't done any extensive testing yet but I've read a large part of the relevant code and it seems to take this eventuality into account.
The wtpWorker() function pushes a cleanup handler to take care of setting the flag in case of the thread gets canceled: [pthread_cleanup_push(wtpWrkrExecCancelCleanup, pWti);](https://github.com/rsyslog/rsyslog/blob/v8.18.0/runtime/wtp.c#L388)

I don't see any obvious race except maybe line 390. But due to threads being spawned with the PTHREAD_CANCEL_DEFERRED type by default, this shouldn't be an issue either.
I don't have a good explanation for this issue so far.

rgerhards:
I have looked at the code and the patch, and unfortunately do not see where the code fails and why the patch should solve this. Any more insight would be appreciated.

tomassykora:
There is a new pull request for this issue here: #1274

rgerhards:
replaced by https://github.com/rsyslog/rsyslog/pull/1274

Issue #900, opened 2016-03-22
testbench: add tests for omkafka


rgerhards:
if possible. see also #898.

janmejay:
Cool, let us aim for linux-only setup.

Because distros will build this on different flavours of Linux distro provided package will make it complicated. I am thinking of distro provided binaries. 

Even if we harden it and get it working perfectly, building from source is not really going to be very good user experience.

Distro provided binaries sounds good?

janmejay:
Also, this will require JVM. We'll expect user to install JVM using distro provided mechanism. That itself has a bunch of complexity depending on which distro of java the user chooses. On travis we can't escape this one, so we'll pick openjdk (distro provided binary) unless its too old to run kafka 0.8.x.

rgerhards:
I think the key aim is to get things running on travis.

Sent from phone, thus brief.
Am 22.03.2016 13:28 schrieb "Janmejay Singh" <notifications@github.com>:

rgerhards:
Just some extra thoughts now that I have a better keyboard ;)

I do not fully understand distro-provided packages vs. binaries. Do you really want not to use the packaging system, but rather extract the binaries and use them? That sounds much more complicated to me.

If you have not already done so, have a look at the .travis.yml and the script under ./tests/travis. I setup all that I can so that at least the compile can happen (and thus the static analyzer run) and, if possible, dynamic testing is also possible. I think we don't need to be 100% aligned with what the actual user does, but at least be able to run the test. That's a really big plus.

janmejay:
Sorry, let me clarify the terms:
- distro provided packages: dependencies we install using
package-manager(apt) in travis (or stuff users install using their own
package manager while building from source)
- project provided binaries: archives containing binaries that projects
release (make available for download alongside versioned source-archives).

In java-world, its a fairly common thing for projects to make releases that
provide both (usually separate) source and binary archives (tar.gz, zip
etc), because almost no java-based-projects have architecture-specific
binaries (the binaries in this case are JVM opcode, not machine opcode),
this works for everyone(across architectures, as long as they have JVM).

For instance, someone looking to use Kafka can fetch kafka-binary-archive
(say from a link such as this:
https://www.apache.org/dyn/closer.cgi?path=/kafka/0.8.2.2/kafka_2.11-0.8.2.2.tgz),
unarchive it, setup appropriate config files and fire-up the startup script.

So when I say distro-provided-binary vs. project-provided-binary, I don't
[...]

PR #967, opened 2016-04-19
Kafka integration test

janmejay:
New flag to enable kafka tests (--enable-kafka-tests)

configure fails when:
- omkafka is disabled and kafka-tests is enabled.
- kafka-tests is enabled but JVM is not found

The test is self-sufficient. One can enable kafka-tests and just run the test. It downloads zookeeper and kafka binaries the first time (it then keeps them cached). It starts and stops kafka through controls available via diag.sh and makes assertions by draining the kafka-topic to a file. No elaborate kafka or zookeeper setup is required.

Impl in diag.sh has some sleeps, because process startup of ZK and Kafka takes time.


janmejay:
@rgerhards This is an unrelated failure. Do you want me to trigger the build again?

rgerhards:
Sent from phone, thus brief.
Am 19.04.2016 14:35 schrieb "Janmejay Singh" <notifications@github.com>:
>
> @rgerhards This is an unrelated failure. Do you want me to trigger the
build again?

I have an eye on those things as long as the testbench still has quirks. So
if you are sure it's unrelated, just relax. I investigate if problems
persist and will post in such rate cases.

Rainer

janmejay:
I wasn't sure if I should do a force-push in unrelated-failure scenarios (because it'll re-trigger all environments, eat build capacity etc). 

In general should go ahead and trigger the entire thing again or do you prefer to intervene in such situations and selectively trigger the failing build (or investigate etc)?

rgerhards:
I have no problem with resource usage, but I prefer to intervene in such cases, because sometimes another build that worked fails. That said, I am still working on getting the remaining quircks out of the testbench, and have made progress (only the very hard ones are left ;)).

janmejay:
Cool, so in the future when I come across unrelated failures I'll drop you a note.

BTW, this PR has a base-testcase for omkafka (which uses static topic name).

A bunch more tests are possible like dynamic-topic, key-based partitioning (as opposed to random partitioning), using compression etc.

This infra allows us to write those tests, but I haven't written them yet. It will require some changes for multi-broker tests (involving multiple kafka instances). All single-broker scenarios can be covered with what we have.

Will create a separate PR for more complex tests (still happy-path tests, but exercising more features) around omkafka.

rgerhards:
2016-04-19 15:47 GMT+02:00 Janmejay Singh <notifications@github.com>:

> Cool, so in the future when I come across unrelated failures I'll drop you
> a note.
>

Yeah, I usually monitor PRs pretty closely and look into such things
whenever they occur. But depending on what I do, it might take a day until
I notice. So if it is critical, just let me know,. If it is not critical,
just do nothing. I'll post a note to the Pr when I think it is not a
testbench but a real test probelm (like in this case here).


> BTW, this PR has a base-testcase for omkafka (which uses static topic
> name).
>
> A bunch more tests are possible like dynamic-topic, key-based partitioning
> (as opposed to random partitioning), using compression etc.
>
> This infra allows us to write those tests, but I haven't written them yet.
> It will require some changes for multi-broker tests (involving multiple
> kafka instances). All single-broker scenarios can be covered with what we
> have.
>
> Will create a separate PR for more complex tests (still happy-path tests,
> but exercising more features) around omkafka.
>

Looking forward to further work. Note that I will begin to merge again
soon, but I am working on the daily build packages and CI and won't merge
until I have a decent milestone in that regard.

Rainer

janmejay:
Sounds good.

configure.ac
```diff
@@ -1475,7 +1475,22 @@ AC_ARG_ENABLE(omkafka,
...
+         esac],
+        [enable_kafka_tests=no]
+)
+AM_CONDITIONAL(ENABLE_KAFKA_TESTS, test x$enable_kafka_tests = xyes)
+
 if test "x$enable_omkafka" = "xyes"; then
+	if test "x$enable_kafka_tests" = "xyes"; then
+		AX_PROG_JAVAC #we don't need javac, but macro documentation says JAVAC *must* be checked before JAVA
+		AX_PROG_JAVA
+	fi
```
rgerhards, line 1501: do we need this for the build (I don't think so, I think it built so far) or for the testbench test? If it is the latter, I think we should check it in the test case only, so that we keep folks able to build it on a machine that does not have the full runtime environment.

janmejay, line 1501: This will run only if the user enables kafka-tests. It is disabled by default, so it won't affect anyone unless they enable kafka-tests.

If they enable it, this will fail in an expected way which will pin-point the cause of failure.

I felt its better to fail upfront (people expect to catch incompatibilities in configure, right?).

The test silently ignoring it will make it more hidden, isn't it?

rgerhards, line 1501: 2016-04-19 18:02 GMT+02:00 Janmejay Singh notifications@github.com:

> In configure.ac
> https://github.com/rsyslog/rsyslog/pull/967#discussion_r60259790:
> 
> >  if test "x$enable_omkafka" = "xyes"; then
> > -   if test "x$enable_kafka_tests" = "xyes"; then
> > -       AX_PROG_JAVAC #we don't need javac, but macro documentation says JAVAC _must_ be checked before JAVA
> > -       AX_PROG_JAVA
> > -   fi
> 
> This will run only if the user enables kafka-tests. It is disabled by
> default, so it won't affect anyone unless they enable kafka-tests.

sorry, I think I should call it a day. I have overlooked the second if
condition...

Rainer

janmejay, line 1501: The other option is to not check it in configure, but fail the test when it can't find JVM.

I can switch it to that if you prefer.

Or we can change the help-text "enable Kafka specific tests in testbench" to "enable Kafka specific tests in testbench (requires JVM)".

rgerhards, line 1501: no,no,no all is well. My fault.

2016-04-19 18:07 GMT+02:00 Janmejay Singh notifications@github.com:

janmejay, line 1501: I think I should add "(requires JVM)" anyway. If an unsuspecting user who does not understand autotools at all enables kafka-tests and isn't comfortable reading configure.ac, it can help identify where the dependency is coming from.

What do you think?

janmejay:
I felt its a good idea to make the dependency more visible through enable-flag doc. Java is a rather unusual dependency in a C based project. So this may help some endusers.

janmejay:
@rgerhards I'll be traveling the next few days and will not have access to my computer. In case you need it reverted (or something else changed), I'll do it once I am back on monday.

janmejay:
Added dependencies on readlink and wget.

rgerhards:
closes #900

PR #969, opened 2016-04-19
fix memleak in worker shutdown

rgerhards:


rgerhards:
fix may also included in #1155

rgerhards:
fix is provided by #1155, so there is no need any longer to check why this PR failed.

PR #975, opened 2016-04-21
ommysql and ompgsql should include netdb.h

mamash:
MAXHOSTNAMELEN is defined in netdb.h on (at least) SunOS. Fixes build.

### DIFF
--- a/plugins/ommysql/ommysql.c
+++ b/plugins/ommysql/ommysql.c
@@ -33,6 +33,7 @@
 #include <signal.h>
 #include <errno.h>
 #include <time.h>
+#include <netdb.h>
 #include <mysql.h>
 #include "conf.h"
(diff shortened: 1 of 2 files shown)


rgerhards:
note: the CI errors are due to work on the CI system itself and/or a partial different merge. I am investigating.

mamash:
Thanks!

rgerhards:
Thanks for contributing! ;-) Merge took a bit as I had trouble with the CI environment...

mamash:
I guess I should prepare the rest of the patches that we've accumulated over in pkgsrc then... :)

rgerhards:
That would be definitely appreciated.

Sent from phone, thus brief.
Am 22.04.2016 22:07 schrieb "Filip Hajny" <notifications@github.com>:

PR #980, opened 2016-04-22
travis: do a special run which merges with master

rgerhards:
this guards against cross-PR regressions and also ensures that our
newest travis settings are used


rgerhards:
Before merging, we need to check if this is really required with travis. To proof this, re-run the merge script when master has changed.

Issue #688, opened 2015-12-09
imtcp-tls-basic-vg.sh is currently failing in git master


Whissi:
Hi,

I am currently testing git master (fa5a6c9f02460548693179d22d07d331f94ef4e5) with valgrind testbench enabled:
```
FAIL: imtcp-tls-basic-vg.sh
===========================

===============================================================================
[imtcp-tls-basic-vg.sh]: testing imtcp in TLS mode - basic test
==43068== Memcheck, a memory error detector
==43068== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==43068== Using Valgrind-3.11.0 and LibVEX; rerun with -h for copyright info
==43068== Command: ../tools/rsyslogd -C -n -irsyslog.pid -M../runtime/.libs:../.libs -f./testsuites/imtcp-tls-basic.conf
==43068==
rsyslogd started with pid  43068
startup-vg still running
starting run 1
Sending 50000 messages.
^M00050000 messages sent
runtime: 0.013
End of tcpflood Run
imdiag[13500]: mainqueue empty
./diag.sh: line 166: wait: pid 43068 is not a child of this shell
rsyslogd run exited with 127
scanf error in index i=49152
sequence error detected
FAIL imtcp-tls-basic-vg.sh (exit status: 1)
```

I repeated 3 times and the test was always failing.

```
============================================================================
Testsuite summary for rsyslog 8.15.0.master
============================================================================
# TOTAL: 179
# PASS:  173
# SKIP:  5
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
```

alorbach:
Which Linux version are you using? 
II am on Ubuntu 14.04, test is not failing. 
Also no fails on our buildbot as far as I know. 

Perhaps it is a timing issue? I have seen such fails when rsyslog is not finished properly.

Whissi:
This is happening on my Gentoo build cluster.

Looks like you are correct regarding a timing issue. Inspired by your recent commit d782c9108b4b37d2 I added `sleep 10` after the the *tcpflood* command which seems to be a working workaround.

But that's a bad hack, isn't it? :)
[...]

PR #984, opened 2016-04-23
imdiag: avoid races in detecting queue empty status

rgerhards:
Previously, we just did three checks with some longer waits in between.
This was insufficient to avoid races, as we saw in practice. Now we
have slightly changed that approach and require 10 times in a row
an empty queue to deduce it atcually is empty.

closes https://github.com/rsyslog/rsyslog/issues/688


rgerhards:
ubuntu failed due to KSI service not available, so no real issue

Issue #996, opened 2016-04-29
Thread stack overflow in tcpsrv when using the imtcp module, when receiving a message (when running in a musl environment)


ashleysommer:
I was debugging a segfault I was getting when running rsyslog in an Alpine docker container. Alpine uses musl rather than glibc for a smaller container size.

I kept hitting a segfault when I had the rsyslog daemon configured to recieve incoming TCP communication using the imtcp module. The problem does not occur using the UDP net stream driver.
I reported the problem to the Alpine bug tracker here; http://bugs.alpinelinux.org/issues/5504
I experienced the same bug versions 8.9 and 8.16 of rsyslog.
I ran the exact same build and configuration on an ubuntu machine (with glibc) and there was no problem.

I finally came across someone with a related problem in this thread: https://github.com/lovell/sharp/issues/242 in which someone points out that there is a 100k array pushed to the stack (to use as a string buffer), and that overflows musl's 80k default stack size.
More info about the 80k stack size limitation here: http://wiki.musl-libc.org/wiki/Functional_differences_from_glibc#Thread_stack_size

I installed valgrind in my container to see if I could find the problem. I got this in the `doReceive` function of the `tcpsrv` module:
```
35 Thread 4 in:imtcp:
35 Invalid write of size 8
35 at 0x66E29E1: doReceive (tcpsrv.c:561)
35 Address 0x40a7be8 is on thread 4's stack
35
35 Invalid write of size 8
35 at 0x12D541: do_dbgprint (debug.c:883)
35 Address 0x4097a08 is on thread 4's stack
.... and lots more.
```
I looked up the code in `tcpsrv.c`, and I found:
```char buf[128*1024]; /* reception buffer - may hold a partial or multiple messages */```
A 128k char buffer! It is overflowing the stack and then reading and writing at invalid thread memory addresses.

I then went to check if the stack size for the thread could possibly be increased using `pthread_attr_setstacksize` but I found that technique is actually already used! Searching the whole codebase there are three usages of `pthread_attr_setstacksize(&thread, 4096*1024)` which creates threads with a 4M stack size, they are:
1) In `imptcp` module in `modInit` when creating worker threads (which I'm not using, rather I'm using the imtcp module)
2) in `imudp` in `runInput` when creating worker threads (again, I'm not using this module)
3) in `tcpsrv` in `startWorkerPool` when creating the worker threads. (this applies to my case)

[...]

PR #997, opened 2016-04-29
core: set default stack size to 4MiB

rgerhards:
Note that this is below the usual 10MiB default, but should cause
no issues (output threads already use 4MiB stack size, for example).
This also addresses issues introduced by micro-libc's which only
provide very limited stack space by default.

closes https://github.com/rsyslog/rsyslog/issues/996


rgerhards:
Please wait for response on #996 before merging.

Issue #1005, opened 2016-05-03
testbench: sometimes an instance hangs at end of test


rgerhards:
It seems to be primarily the case in performance-intense tests, I've seen this sometimes with valgrind. Here is a sample: http://build.rsyslog.com:8010/builders/rsyslog%20ubuntu%20rsyslog/builds/486/steps/shell_4/logs/stdio

We need to check if our shutdown predicate is sufficient -- doesn't look so.

PR #1006, opened 2016-05-03
testbench: sometimes an instance hangs at end of test

rgerhards:
shutdown detection was based on pid file, we should check the
actual pid. We can't use wait, because we want to be able to
timeout. This patch should fairly portable handle the situation.

closes https://github.com/rsyslog/rsyslog/issues/1005

### DIFF
--- a/tests/diag.sh
+++ b/tests/diag.sh
@@ -215,7 +215,20 @@ case $1 in
    'wait-shutdown')  # actually, we wait for rsyslog.pid to be deleted. $2 is the
    		# instance
 		i=0
-		while test -f rsyslog$2.pid; do
+		ls -l rsyslog*
+		out_pid=`cat rsyslog$2.pid.save`
+		if [[ "x$out_pid" == "x" ]]
+		then
+			terminated=1
+		else
+			terminated=0
+		fi
+		while [[ $terminated -eq 0 ]]; do
+			ps -p $out_pid
+			if [[ $? != 0 ]]
+			then
+				terminated=1
+			fi
 			./msleep 100 # wait 100 milliseconds
 			let "i++"
 			if test $i -gt $TB_TIMEOUT_STARTSTOP
@@ -226,6 +239,8 @@ case $1 in
 			   exit 1
 			fi
 		done
+		unset terminated
+		unset out_pid
 		if [ -e core.* ]
 		then
 		   echo "ABORT! core file exists"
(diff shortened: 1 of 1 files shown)


portante:
Looks good to me, but one of the failures was because the ps command line was not valid for that platform, and in another failure I also see that we get error messages on stderr from `cat` when the pid file does not exist, so we might want to use `cat rsyslog$1.pid 2> /dev/null`.

rgerhards:
I am still working on this, you really need the testbench to fully get to all details. The missing file in cat is indication of a race, so it is a real problem. I am glad you mention the portability issue. Looks like I need to write a small program to get around this (easier than trying how to resolve that via command line tools IMHO).

tests/diag.sh
```diff
@@ -284,11 +298,13 @@ case $1 in
 		   echo Shutting down instance 2
 		fi
 		. $srcdir/diag.sh wait-queueempty $2
+		cp rsyslog$2.pid rsyslog$2.pid.save
```
portante: So if something else can remove the pid file, won't this potentially fail?

I think you have to check success/failure of `cp` here and then perform the kill conditionally off the saved pid file, no?

The read of the pid file needs to be atomic, something like:

```
rsyslog_pid="$(cat rsyslog$2.pid 2> /dev/null)"
if [ "x$rsyslog_pid" != "x" ]; then
    ./msleep 1000
    kill $rsyslog_pid
fi
```

rgerhards: That "something else" is rsylog, when it is killed (2 lines below). So this copy should be safe. But still, there is something missing overall ;)

portante:
But if rsyslog exits abnormally early, won't it clean up its pid file still in some cases?

rgerhards:
No, it happens late in shutdown.

Sent from phone, thus brief.
Am 04.05.2016 13:12 schrieb "Peter Portante" <notifications@github.com>:

rgerhards:
Let me clarify a bit more: what we try to do is check the actual rsyslog
shutdown happened. Previously, we just checked that the pid file was
deleted, as this is really very late in the process. However, I begun to
notice some "hanging" instances over the weekend. This seems to be related
to timing, which changed for some reason (maybe package updates,
maybesomething else). In any case, slow test (mostly with valgrind) seem to
need a bit time after the pid file was deleted. I can see that. I just hope
that it is the cause of the problems we see, and it is something that
definitely should be fixed.

So we need to ensure that the pid has gone away. Before shutdown, we can
record the pid, so that we have it when when need to wait for termination
(sometimes the pidfile is already gone when we reach that point). If
rsyslog abort, the pid file is never removed.

Hope that gives a better overview of the whole thing.

PR #1013, opened 2016-05-06
more diagnostics around coordinating with stats-flush (in response to it still failing on very slow env)

janmejay:
Let us not merge this just yet. Have created this PR so I can see it build on fedora 23 (slowenv)


janmejay:
@rgerhards fedora 23x32 seems offline

janmejay:
travis failiure seems like a false one

janmejay:
Ready for merge.

janmejay:
BTW, the test-code code also had a bug, which this fixes. We weren't using tv_nsec at all, we were dividing tv_sec by 10^6 to get mills of the nano portion, which was wrong. It could not have failed any test, but it would have published slightly incorrect stats (imdiag produces wait stats). 

It wouldn't have been broken anything, but we have fixed it anyway, thanks for fedora23x32 :-)

PR #1020, opened 2016-05-10
CStr: reset a previously cached sz string when appending

### LINKED ISSUE #874: imfile: multiline buffer corruption
I'm testing a setup where I collect logs using rsyslog imfile and pass them to kafka. Some of the logs (~30%) are multiline. I'm seeing a problem where around 1% of the multiline logs are corrupted. The symptoms are that the first line of the multiline log message is intact, but the rest of the message is corrupt. Initially I discovered the problem because logstash was unable to parse the corrupt logs after retrieving them from kafka. I've tried to isolate the problem by saving the log messages to a local file using omfile, on the source host. The problem appears there as well, so I believe we can safely leave kafka and logstash out of the picture.

**Example of good log message**
Original log lines:
```
1457562691510 [E(6)] INFO  LookupClient  - UPDATE api-test.jobrstreaming.com/192.168.185.222:443 sLen:256 cLen:52
ConnectivityInfoUpdate(guid:2B 24 97 74 F5 EB D2 AE B4 1B 1C 01 A3 36 24 9F, gatewayType:2, publicAddress:InetAddress(address:C0 4B B2 81, port:8001), privateAddress:InetAddress(address:C0 4B B2 81, port:8001), isDialup:false, uploadBandwidth:0, justRelay:true, externalAS:8075, internalAS:0)
```
Log message after collection by imfile and output with omfile using a json template (I've indented the json structure to make it easier to read):
```
{"@timestamp":"2016-03-09T22:34:41.367305+00:00",
  "host":"jobr-test-euw-relay-ry1-0",
  "type":"relay-service",
  "file":"\/opt\/relay-service\/current\/logs\/main.log",
  "message":"1457562691510 [E(6)] INFO  LookupClient  - UPDATE api-test.jobrstreaming.com\/192.168.185.222:443 sLen:256 cLen:52ConnectivityInfoUpdate(guid:2B 24 97 74 F5 EB D2 AE B4 1B 1C 01 A3 36 24 9F, gatewayType:2, publicAddress:InetAddress(address:C0 4B B2 81, port:8001), privateAddress:InetAddress(address:C0 4B B2 81, port:8001), isDialup:false, uploadBandwidth:0, justRelay:true, externalAS:8075, internalAS:0)\\n"
}
```
_(The newline marker `\\n` is misplaced, but I've reported that in a separate issue #843. It should have been placed before the second line instead of after, i.e. between `cLen:52` and `ConnectivityInfoUpdate`.)_

**Example of corrupt log message**
Original log lines:
```
1457569891545 [E(7)] INFO  LookupClient  - UPDATE api-test.jobrstreaming.com/192.168.185.222:443 sLen:256 cLen:52
ConnectivityInfoUpdate(guid:2B 24 97 74 F5 EB D2 AE B4 1B 1C 01 A3 36 24 9F, gatewayType:2, publicAddress:InetAddress(address:C0 4B B2 81, port:8001), privateAddress:InetAddress(address:C0 4B B2 81, port:8001), isDialup:false, uploadBandwidth:0, justRelay:true, externalAS:8075, internalAS:0)
```
Output log message:
```
{"@timestamp":"2016-03-10T00:31:31.548801+00:00",
  "host":"jobr-test-euw-relay-ry1-0",
  "type":"relay-service",
  "file":"\/opt\/relay-service\/current\/logs\/main.log",
  "message":"1457569891545 [E(7)] INFO  LookupClient  - UPDATE api-test.jobrstreaming.com\/192.168.185.222:443 sLen:256 cLen:52\u0000\u0000\u0000\u0000\u0000\u0000\u0000q\u0001\u0000\u0000\u0000\u0000\u0000\u0000 \u0001\u0000D\u0013\u0000\u0000 \u0001\u0000D\u0013\u0000\u0000(7)] DEBUG WebClient  - LOOKUP REQ: https:\/\/api-test.jobrstreaming.com:443\/v1\/lookup\/2b249774f5ebd2aeb41b1c01a33X\u0001\u0000D\u0013\u0000\u0000 \u0000\u0000\u0000\u0000\u0000\u0000\u0000H\u0001\u0000D\u0013\u0000\u0000H\u0001\u0000D\u0013\u0000\u0000pClient  - UPDATE api-test.jobrstreaming.com\/192.168.185.222:443 sLen:256 cLen:52\u00002aeb41b1c01a33 \u0000\u0000\u0000\u0000\u0000\u0000\u0000d\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0010 \u0001D\u0013\u0000"
}
```
As you can see, everything after the first line of the message (after `cLen:52`) is corrupt. This is true for all corrupted log messages. You can also see bits and pieces of other log lines in the corrupt data. (In one case I could see the beginning of the next log line from the file in the middle of the corrupt data.)

As I mentioned, this happens in about 1% of the multiline log messages. Single-line log messages are unaffected. I ran the system for a couple of days and got 28 corrupted log messages out of a total of 2980 multiline log messages.

I'm running this on Ubuntu 14.04.1, using rsyslog (and rsyslog-kafka) version 8.16.0-0adiscon1trusty1 from the http://ppa.launchpad.net/adiscon/v8-stable/ubuntu repository. Below is the configuration I used:

```
module(load="imfile")

main_queue(
  queue.workerthreads="1"      # threads to work on the queue
  queue.dequeueBatchSize="100" # max number of messages to process at once
  queue.size="10000"           # max queue size
)

template(name="debug_json" type="list") {
  constant(value="{")
  property(name="hostname" format="json")
  constant(value="\",\"type\":\"")
  constant(value="\",\"file\":\"")
  property(name="$!metadata!filename" format="json")
  constant(value="\",\"message\":\"")
  property(name="msg" format="json")
  constant(value="\"}\n")
}

ruleset(name="kafka_logstash_debug") {
  action(
    type="omfile"
    file="/mnt/relay.log"
    template="debug_json"
  )
  stop
}

$RepeatedMsgReduction off
input(type="imfile"
  file="/opt/relay-service/current/logs/main.log"
  tag="relay-service"
  PersistStateInterval="100"
  startmsg.regex="^[0-9]{13} "
  addMetaData="on"
  ruleset="kafka_logstash_debug"
)
$RepeatedMsgReduction on
```

mhyllander:
When appending to CStr, any previously saved sz string (pszBuf) needs to be reset so it will be re-created the next time a sz string is requested.

Fixes #874.

### DIFF
--- a/runtime/stringbuf.c
+++ b/runtime/stringbuf.c
@@ -534,6 +534,15 @@ rsRetVal cstrTrimTrailingWhiteSpace(cstr_t *pThis)
 	if(i != (int) pThis->iStrLen) {
 		pThis->iStrLen = i;
 		pThis->pBuf[pThis->iStrLen] = '\0'; /* we always have this space */
+
+		if(pThis->pszBuf != NULL) {
+			/* in this case, we adjust the psz representation
+			 * by writing a new \0 terminator - this is by far
+			 * the fastest way and outweights the additional memory
+			 * required.
+			 */
+			 pThis->pszBuf[pThis->iStrLen] = '\0';
+		}
 	}
 
 done:	return RS_RET_OK;
--- a/runtime/stringbuf.c
+++ b/runtime/stringbuf.c
@@ -281,6 +281,10 @@ rsRetVal rsCStrAppendStrWithLen(cstr_t *pThis, const uchar* psz, size_t iStrLen)
 	memcpy(pThis->pBuf + pThis->iStrLen, psz, iStrLen);
 	pThis->iStrLen += iStrLen;
 
+	/* reset any cached sz string */
+	free(pThis->pszBuf);
+	pThis->pszBuf = NULL;
+
 finalize_it:
 	RETiRet;
 }
--- a/runtime/stringbuf.h
+++ b/runtime/stringbuf.h
@@ -32,6 +32,7 @@
 #ifndef _STRINGBUF_H_INCLUDED__
 #define _STRINGBUF_H_INCLUDED__ 1
 
+#include <stdlib.h>
 #include <assert.h>
 #include <libestr.h>
 
@@ -82,6 +83,10 @@ static inline rsRetVal cstrAppendChar(cstr_t *pThis, uchar c)
 
 	/* ok, when we reach this, we have sufficient memory */
 	*(pThis->pBuf + pThis->iStrLen++) = c;
+
+	/* reset any cached sz string */
+	free(pThis->pszBuf);
+	pThis->pszBuf = NULL;
 
 finalize_it:
 	return iRet;


mhyllander:
The problem actually occurs in this line: https://github.com/rsyslog/rsyslog/blob/master/plugins/imfile/imfile.c#L480.
The cached sz string `pszBuf` is old and shorter than the current buffer `pBuf`. cstrLen returns the length of pBuf, therefore MsgSetRawMsg copies data past the end of pszBuf.

rgerhards:
can you please rebase to current master branch? It looks like there is a patch missing that makes the testbench fail.

portante:
@mhyllander, where does `MsgSetRawMsg` reference `pszBuf`?  I see it reference `pszRawMsg`.  Do you mean `MsgSetTAG`?

Not sure I see how this fix helps yet.

mhyllander:
It's `rsCStrGetSzStrNoNULL(cstrLine)` that initializes and returns `cstrLine->pszBuf`. Apparently the problem is that `rsCStrGetSzStrNoNULL` sometimes gets called after the first line of a multiline message has been collected. This initializes `cstrLine->pszBuf` from `cstrLine->pBuf` so that it contains the first line. Then the next line is appended to `cstrLine->pBuf`, but `cstrLine->pszBuf` is not reset, so it still contains only the first line of the message.

When we get to the call to `MsgSetRawMsg`, `rsCStrGetSzStrNoNULL(cstrLine)` will return the (cached) `cstrLine->pszBuf` with the first line only, while `cstrLen(cstrLine)`returns the length of the multiple lines in `cstrLine->pBuf`. `MsgSetRawMsg` thus receives a string with the first line only, but a length that can be much longer. And so `MsgSetRawMsg` reads more data than is available from the string.

Besides all that, if you just consider `rsCStrGetSzStrNoNULL`, you see that it will only initialize `pThis->pszBuf` the first time it's called. If you later append another string to `pThis` using `rsCStrAppendStrWithLen`, it's only logical that `pThis->pszBuf` must be reset so that `rsCStrGetSzStrNoNULL` will return the complete string the next time it's called.

runtime/stringbuf.c
```diff
@@ -281,6 +281,10 @@ rsRetVal rsCStrAppendStrWithLen(cstr_t *pThis, const uchar* psz, size_t iStrLen)
 	memcpy(pThis->pBuf + pThis->iStrLen, psz, iStrLen);
 	pThis->iStrLen += iStrLen;
 
+	/* reset any cached sz string */
+	free(pThis->pszBuf);
```
portante, line 285: If we need this here, do we also need this after line 256 above?

mhyllander, line 285: You mean in `rsCStrExtendBuf`? I don't think so, in that case nothing is being appended to `pBuf`, it's just being re-allocated, so it doesn't get out of sync with `pszBuf`.

mhyllander:
I have also found out why only about 1% of the multiline messages were affected. I've been tracing all calls to `rsCStrGetSzStrNoNULL`, and found that only one occurs "randomly" outside the normal flow of control. 

It's this line: https://github.com/mhyllander/rsyslog/blob/master/plugins/imfile/imfile.c#L635: 

```
		if(pLstn->iPersistStateInterval > 0 && pLstn->nRecords++ >= pLstn->iPersistStateInterval) {
			persistStrmState(pLstn);
			pLstn->nRecords = 0;
		}
```
`persistStrmState` eventually calls `rsCStrGetSzStrNoNULL` deeper down the stack, and this would occur after the first line of a multiline message has been read and held while the previous message is processed. And since I have `PersistStateInterval="100"` this affects on average 1 out of 100 of the multiline messages.

runtime/stringbuf.c
portante, line 285: `pszBuf` is allocated to be the same size as `pBuf`, right?  So if we reallocate `pBuf`, why would we want to keep `pszBuf` around that is of a different size?

mhyllander, line 285: `pBuf` is a buffer of size `iBufSiz`, that contains a string of length `iStrLen`. Initially `iBufSiz==iStrLen`. `pszBuf` is allocated as size `iStrLen+1`. When you extend `pBuf` the content is unchanged (`iBufSiz` increases but `iStrLen` is the same). There is therefore no need to reset `pszBuf`. 
`pszBuf` contains a cached version of the string in `pBuf`, and it will become stale if the string in `pBuf` is modified. It's only when you actually change the _content_ of `pBuf` that you need to reset `pszBuf`.

rgerhards, line 285: I cannot look at this in-depth at the moment, but some piece of information: after cstrFinalize (or similar name) has been called, the buffer should be read-only. It's not being enforced, but this may be the root cause of the issue. I'll look in more depth when I am finished with the higher priority things.

mhyllander, line 285: Please see my comment about `persistStrmState`. The root cause here is that `persistStrmState` calls `rsCStrGetSzStrNoNULL` when only the first line of a multiline message has been read from the stream. This initializes `pszBuf` to the current contents of `pBuf`, which is the first line. When the rest of the lines are appended to `pBuf`, this is not reflected in `pszBuf`. `pszBuf` becomes inconsistent with `pBuf` (shorter), and it results in `MsgSetRawMsg` reading past the end of the allocated memory.

It's pretty straight-forward to me. I realize this is code from 2005, but it's being used in a new way by imfile that uncovered this bug. 

Regardless how the bug is actually triggered, it's logically obvious that if you change the contents of `pBuf`, then that change must also be reflected in `pszBuf`. One way to do this is to reset `pszBuf` so that `rsCStrGetSzStrNoNULL` will re-initialize it the next time it is called.

davidelang, line 285: On Wed, 11 May 2016, Magnus Hyllander wrote:

> > @@ -281,6 +281,10 @@ rsRetVal rsCStrAppendStrWithLen(cstr_t _pThis, const uchar_ psz, size_t iStrLen)
> >      memcpy(pThis->pBuf + pThis->iStrLen, psz, iStrLen);
> >      pThis->iStrLen += iStrLen;
> > -    /\* reset any cached sz string */
> > -    free(pThis->pszBuf);
> 
> Please see my comment about `persistStrmState`. The root cause here is that `persistStrmState` calls `rsCStrGetSzStrNoNULL` when only the first line of a multiline message has been read from the stream. This initializes `pszBuf` to the current contents of `pBuf`, which is the first line. When the rest of the lines are appended to `pBuf`, this is not reflected in `pszBuf`. `pszBuf` becomes inconsistent with `pBuf` (shorter), and it results in `MsgSetRawMsg` reading past the end of the allocated memory.
> It's pretty straight-forward to me. I realize this is code from 2005, but it's being used in a new way by imfile that uncovered this bug.

I think that the right answer here isn't to mess with the variables, but rather 
make it so that persistStrmState doesn't get called at each newline, but rather 
it only gets called at the end of a message.

This assumes that we can do this without loosing the data that was just read 
that tells us that we have a new message, not a continuation of the old one.

David Lang

mhyllander, line 285: In other words, you want to leave the bug in the code and work around it elsewhere. But it won't work: if all the messages in the file are multiline there will be no time you can call persistStrmState without triggering the bug.
(And this is because of how multiline parsing with regex works, you can't process a message until it's complete, which happens when you've read the first line of the next message. And you don't even know if the last line you've read is the start of a multiline message or not.)

rgerhards, line 285: 2016-05-11 9:26 GMT+02:00 Magnus Hyllander notifications@github.com:

> In runtime/stringbuf.c
> https://github.com/rsyslog/rsyslog/pull/1020#discussion_r62802165:
> 
> > @@ -281,6 +281,10 @@ rsRetVal rsCStrAppendStrWithLen(cstr_t _pThis, const uchar_ psz, size_t iStrLen)
> >     memcpy(pThis->pBuf + pThis->iStrLen, psz, iStrLen);
> >     pThis->iStrLen += iStrLen;
> > -   /\* reset any cached sz string */
> > -   free(pThis->pszBuf);
> 
> In other words, you want to leave the bug

as I said, I need to review this more in-depth. BUT IIRC there is NO BUG in
stringbuf, it is just that the internal API is incorrectly called. Note
that there are many, many more callers of this internal API, and the change
affects many things (e.g. does it create a memleak? How do the many more
NULL writes affect performance especially given the fact that in 99.99% of
the cases it is not needed).

> in the code and work around it elsewhere.

IMHO wrong assumption. There is one place that does it wrong. Wouldn't that
be the best place for a fix?

mhyllander, line 285: OK, take you're time, I understand if you're concerned with performance. I did handle the memleak by freeing `pszBuf` before setting it to NULL. But leaving a bug like that in the code and working around it elsewhere will probably bite you some other time and place.

I can think of other ways to handle the problem instead of freeing `pszBuf`, e.g. setting a flag that instructs `rsCStrGetSzStrNoNULL` to re-create the sz string by reallocating `pszBuf`. If you want I can implement that variant instead.

mhyllander:
I found two more places where pszBuf was not being reset or handled consistently.

If you're interested, I have made an alternative implementation in a separate branch, using a flag to re-create pszBuf when needed. https://github.com/mhyllander/rsyslog/tree/feature/stringbuf-append-v2

runtime/stringbuf.c
```diff
@@ -534,6 +534,15 @@ rsRetVal cstrTrimTrailingWhiteSpace(cstr_t *pThis)
...
 		pThis->iStrLen = i;
 		pThis->pBuf[pThis->iStrLen] = '\0'; /* we always have this space */
+
+		if(pThis->pszBuf != NULL) {
+			/* in this case, we adjust the psz representation
+			 * by writing a new \0 terminator - this is by far
+			 * the fastest way and outweights the additional memory
+			 * required.
+			 */
+			 pThis->pszBuf[pThis->iStrLen] = '\0';
```
mhyllander, line 544: Regarding this code and the twin place in `rsCStrTruncate`: If you don't reset `pszBuf` when you append data to `pBuf` (as I have suggested), you can actually write the `'\0'` byte into unknown memory after `pszBuf`:

``` C
cstr_t *thisLine = NULL;
rsCStrConstructFromszStr(&thisLine, "foo"); /* pBuf = "foo" */
rsCStrGetSzStrNoNULL(thisLine);                /* pszBuf = "foo\0" */
rsCStrAppendStr(thisLine, "bar ");                 /* pBuf  = "foobar ", pszBuf = "foo\0" */
cstrTrimTrailingWhiteSpace(thisLine);           /* pBuf = "foobar", plus writing \0 after the end of pszBuf */
```

The last call will actually do `thisLine->pszBuf[6] = '\0'`, but `pszBuf` is only 4 bytes long ("foo\0").

(In my alternative implementation I check that `pszBuf`is not stale before writing the null byte.)

davidelang:
@mhyllander 

You are not hearing what Rainer is saying.

The bug isn't in the handling of the pszBuf variable, but rather in the timing of when the function can be called.

The function is only supposed to be called when the message in the buffer is complete and there will never be any changes after this point. However currently it can be called anytime there is a newline.

Historically, this didn't make any difference, because a newline indicated the end of the message.

As we have gained the ability to have different things that indicate the end of the message, the fact that this function is getting called on newlines instead of at the end of the message has been missed.

So the fix is to make sure that this function is only called when the message is complete, not when a newline is seen.

This does change the definition of the checkpoint setting a bit. Instead of being 'every X lines of the file', it's 'every X messages in the file', but I don't think that really hurts anything (if anything, I think it's a good change, you don't want to restart in the middle of a message after all)

As for the other places that you think this is being done wrong, we will need to check and see if they are a similar case of the function being called incorrectly.

P.S. this isn't just a performance question to change this, it's a multi-threaded correctness issue. This structure is supposed to be read-only after this function is called, if it can be modified after this point, that exposes a lot of interesting race conditions.

mhyllander:
OK, then I have at least identified the problem for you, and I will continue using my patch until you come up with an acceptable solution. Perhaps it is as you say that `rsCStrGetSzStrNoNULL` should not be called until the string has been completed, because if it is, then there are methods that can potentially both read and write beyond the end of `pszBuf`.

We can debate whether it should be called a bug or not, but stringbuf is now being used in ways that were not expected when it was originally written. And `pszBuf`is not handled consistently throughout the stringbuf code, compare `cstrTrimTrailingWhiteSpace` with `rsCStrTruncate`.

There are also other inconsistencies from what I can see. A couple of methods (`cstrGetSzStr` and `cstrGetSzStrNoNULL`) assume that `cstrFinalize` first be called to add a terminating null byte to `pBuf`. Those methods must be legacy since most of the stringbuf methods handle strings that can contain null bytes, and rely on iStrLen to know the length of the string (case in point `rsCStrGetSzStrNoNULL` which also replaces null bytes with blanks when constructing the sz string).

So maybe it's time to do an overhaul of stringbuf. As I see it, the stringbuf code is not very robust and can give unpredictable results, as demonstrated by imfile.

mhyllander:
I'm closing this PR in favor of #1026.

PR #1021, opened 2016-05-10
imfile: append newline marker before the next line in a multiline message

### LINKED ISSUE #843: imfile: missing newline after first line of multiline message
I have begun testing rsyslog to read log files with imfile and forward them to kafka/logstash with omkafka. I noticed that when I have multiline messages, there is no newline character after the first line.

I had a look at the code, function [strmReadMultiLine](https://github.com/rsyslog/rsyslog/blob/master/runtime/stream.c#L829) I believe, and it looks to me like newline characters are appended after lines that don't match `startRegex` ([line 862-885](https://github.com/rsyslog/rsyslog/blob/master/runtime/stream.c#L862)).

I believe what you probably wanted was to append the newline *before* appending the next line, so that you end up with newlines between each line, but not after the last line, which matches matches how single-line messages are handled.

mhyllander:
Fixes #843.

### DIFF
--- a/runtime/stream.c
+++ b/runtime/stream.c
@@ -873,12 +873,12 @@ strmReadMultiLine(strm_t *pThis, cstr_t **ppCStr, regex_t *preg, sbool bEscapeLF
 		} else {
 			if(pThis->prevMsgSegment != NULL) {
 				/* may be NULL in initial poll! */
-				CHKiRet(cstrAppendCStr(pThis->prevMsgSegment, thisLine));
 				if(bEscapeLF) {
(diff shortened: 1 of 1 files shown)


portante:
Looks good to me, thanks!

rgerhards:
@mhyllander I guess the current testbench did not catch the problem. Will try to add a test which does, so that we can be sure we won't miss it again.

rgerhards:
side-note: the testbench failure is most probably due to some missing patch in the version this is based on. We had a problem for a couple of days with master branch recently (this was only discovered after parts of the CI environment got upgraded, so we didn't notice it when it actually was introduced).

PR #1031, opened 2016-05-12
support to connect with mysql unix socket on ommysql plugin

Joungkyun:
ommysql plugin can't connect with mysql unix socket. This patch is enable to connect with unix socket.

***Expected:***
```syslog
*.* :ommysql:/tmp/mysqld.sock,Syslog,syslogwriter,topsecret
```

### DIFF
--- a/plugins/ommysql/ommysql.c
+++ b/plugins/ommysql/ommysql.c
@@ -224,9 +224,9 @@ static rsRetVal initMySQL(wrkrInstanceData_t *pWrkrData, int bSilent)
 
 		/* check unix socket */
 		if (pData->dbsrv[0] == '/') {
-			strcpy(usock, pData->f_dbsrv);
+			strcpy(usock, pData->dbsrv);
 			usock_p = usock;
-			strcpy(pData->f_dbsrv, "localhost");
+			strcpy(pData->dbsrv, "localhost");
 		}
 
 		if(mysql_real_connect(pWrkrData->hmysql, pData->dbsrv, pData->dbuid,
--- a/plugins/ommysql/ommysql.c
+++ b/plugins/ommysql/ommysql.c
@@ -224,7 +224,7 @@ static rsRetVal initMySQL(wrkrInstanceData_t *pWrkrData, int bSilent)
 		/* check unix socket */
 		if (pData->dbsrv[0] == '/') {
 			size_t pathlen = strlen(pData->dbsrv);
-			memset(pData->dbusock, 0, _DB_MAXFILEAPTH+1)
+			memset(pData->dbusock, 0, _DB_MAXFILEAPTH+1);
 
 				if (pathlen > _DB_MAXFILEAPTH)
 					strncpy(pData->dbusock, pData->dbsrv, _DB_MAXFILEAPTH);
--- a/plugins/ommysql/ommysql.c
+++ b/plugins/ommysql/ommysql.c
@@ -229,7 +229,7 @@ static rsRetVal initMySQL(wrkrInstanceData_t *pWrkrData, int bSilent)
 				if (pathlen > _DB_MAXFILEAPTH)
 					strncpy(pData->dbusock, pData->dbsrv, _DB_MAXFILEAPTH);
 				else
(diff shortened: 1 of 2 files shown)


plugins/ommysql/ommysql.c
```diff
@@ -219,6 +220,20 @@ static rsRetVal initMySQL(wrkrInstanceData_t *pWrkrData, int bSilent)
...
+
+		/* check unix socket */
+		if (pData->dbsrv[0] == '/') {
+			size_t pathlen = strlen(pData->dbsrv);
+			memset(pData->dbusock, 0, _DB_MAXFILEAPTH+1)
+
+				if (pathlen > _DB_MAXFILEAPTH)
+					strncpy(pData->dbusock, pData->dbsrv, _DB_MAXFILEAPTH);
+				else
+					strncpy(pData->dbusock, pData->dbsrv);
```
rgerhards: Sorry if I overlook the obvious, but where do you actually use dbusock?

runtime/syslogd-types.h
```diff
@@ -43,6 +43,7 @@
 #define	_DB_MAXDBLEN	128	/* maximum number of db */
 #define _DB_MAXUNAMELEN	128	/* maximum number of user name */
 #define	_DB_MAXPWDLEN	128 	/* maximum number of user's pass */
+#define	_DB_MAXFILEAPTH	2048	/* maximun size of file path */
```
rgerhards: if we need that large, wouldn't it be better to dynamically allocate it?

rgerhards:
Can you please also provide a sample on how this is used with current configuration? [action(type="ommyslq" ...)]

Joungkyun:
Thank you for your interest in this patch.

This patch is being used from AnNyung LInux distribution 2.0 from 2011.

And, Korea social commerce company that has 1.2 billion dollars of annual sales is using a combination of rsyslog and LogAnalyzer. The clients of rsyslog is about 300 servers. (In 2011, I was setting)

rgerhards:
Sorry, I cannot merge PRs which do not work with the current config system. So I need new syntax to check. Ideally, you would also contribute a testbench test.

Joungkyun:
dbsrv variable is limited to 64.
Therefore, if the length of the unix domain socket is greater than 64, it may be a problem.

Need to be remake in consideration of this problem.

P.S.
And, sorry, I don't understand your mention: "***do not work with the current config system***".
In this patch, unix domain socket is used instead of db hostname.

use follows:
*.* :ommysql:***/tmp/mysqld.sock***,Syslog,syslogwriter,topsecret
instead of
*.* :ommysql:***mysqlserver.example.com***,Syslog,syslogwriter,topsecret

Joungkyun:
Rewirte this patch..

And I think, the length of the unix domain socket is sufficient degree 2048.
If the length of the path is more than 2048, it seems that malicious use.

davidelang:
On Sat, 14 May 2016, JoungKyun Kim wrote:

> And, sorry, I don't understand your mention: "***do not work with the current config system***".
> In this patch, unix domain socket is used instead of db hostname.
>
> use follows:
> *.* :ommysql:***/tmp/mysqld.sock***,Syslog,syslogwriter,topsecret
> instead of
> *.* :ommysql:***mysqlserver.example.com***,Syslog,syslogwriter,topsecret

the current config system would look like

action(type="ommysql", ....)

David Lang

rgerhards:
Sent from phone, thus brief.
Am 14.05.2016 11:16 schrieb "JoungKyun Kim" <notifications@github.com>:
>
> dbsrv variable is limited to 64.
> Therefore, if the length of the unix domain socket is greater than 64, it
may be a problem.

Yes, this is why I said you need to malloc it.

In general, I doubt that the patch really works, as the variable you say is
never used. A new testbench test who prooves that would be most helpful. I
can do or when I have time, but you need to conceive me why and how it
works ;-)
>
> Need to be remake in consideration of this problem.
>
> P.S.
> And, sorry, I don't understand your mention: "do not work with the
current config system".

Joungkyun:
After test, request again about this function. Plez close this request.
Thanks

PR #1041, opened 2016-05-18
Output Module TCL

fcr--:
An output module that calls a TCL procedure for each message.  This module can be used as an alternative to omshell/ompipe when fast output with custom processing or ad-hoc integration is needed.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -253,6 +253,11 @@ if ENABLE_OMAMQP1
 SUBDIRS += contrib/omamqp1
 endif
 
+# omtcl
+if ENABLE_OMTCL
+SUBDIRS += contrib/omtcl
+endif
+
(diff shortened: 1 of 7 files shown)


rgerhards:
Thanks for the contribution. It would be great if you would also add some doc at https://github.com/rsyslog/rsyslog-doc so that users will become aware the module exists.

rgerhards:
@rgerhards TODO: add module to CI, if possible

rgerhards:
@fcr-- I have begun to work on merging, unforutnately the Travis fails, as it looks due to some missing files:

https://travis-ci.org/rgerhards/rsyslog/jobs/134210718

rgerhards:
now trying to merge via #1168, will give up if that does not work out.

PR #1042, opened 2016-05-18
refactoring of stringbuf.c

rgerhards:
includes PR https://github.com/rsyslog/rsyslog/pull/1029 to show the original cause is fixed. Replaces #1026.

### DIFF
--- a/runtime/stringbuf.c
+++ b/runtime/stringbuf.c
@@ -8,7 +8,7 @@
  * begun 2005-09-07 rgerhards
  * did some optimization (read: bugs!) rgerhards, 2009-06-16
  *
- * Copyright (C) 2007-2012 Adiscon GmbH
+ * Copyright (C) 2007-2016 Adiscon GmbH
  *
  * This file is part of the rsyslog runtime library.
  *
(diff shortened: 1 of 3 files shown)


mhyllander:
I tried this patch in my test environment and it works fine! Thanks for the quick work!

runtime/stringbuf.c
```diff
@@ -383,40 +382,15 @@ rsRetVal rsCStrSetSzStr(cstr_t *pThis, uchar *pszNew)
...
 uchar*  rsCStrGetSzStrNoNULL(cstr_t *pThis)
 {
-	size_t i;
-
 	rsCHECKVALIDOBJECT(pThis, OIDrsCStr);
 
 	if(pThis->pBuf == NULL)
 		return (uchar*) "";
 
 	if(pThis->pBuf != NULL)
```
mhyllander: I guess this check `!= NULL` is not needed...

rgerhards: thx, indeed this is not necessary :-) Will ammend the commit.

side-note: if I have time the next two days, I'll see that I do some more refactoring, there is sufficient potential ;)

Issue #89, opened 2014-06-13
omelasticsearch : usehttps with self-signed certificate


Xat59:
Hi team,

related to my comments in this closed issue : https://github.com/rsyslog/rsyslog/pull/44 , I have to use the omelasticsearch with usehttps="on" with a self-signed certificate.

I'm able to push datas in command line using curl with --insecure option.

Would it be possible to add this feature in omelasticsearch please ?

Thanks a lot !
Regards.

rgerhards:
Better (very) late than never... I finally added the ```allowUnsignedCerts="true"``` option. If there is still interest, please test and let me know the outcome. Otherwise, I'll merge as experimental and wait for others to report. Thanks!

Xat59:
Thank a lot team . I will use it really soon :)

rgerhards:
I merge the PR because of the new release (and as I have not heard anything bad so far).

@Xat59 if you get results (either good or bad), please still post them! :-)

PR #1050, opened 2016-05-24
omelasticsearch: add option to permit unsigned certs (experimentally)

rgerhards:
This adds plumbing as suggested by Joerg Heinemann and Radu Gheorghe,
but is otherwise untested. Chances are good it works. If you use it,
please let us know your experience and most importantly any bug
reports you may have.

closes https://github.com/rsyslog/rsyslog/issues/89


rgerhards:
Note: we need to **add doc** when we merge. But let's wait a day or two for feedback on #89

Issue #1054, opened 2016-05-27
Async output file writing does not respect flushInterval


rgerhards:
When async writing is enabled, parameter ```flushInterval``` is not properly honored.

see also: http://kb.monitorware.com/post26545.html

PR #1055, opened 2016-05-27
fix omfile async writing flush functionality

rgerhards:
see also https://github.com/rsyslog/rsyslog/issues/1054

### DIFF
--- a/runtime/stream.c
+++ b/runtime/stream.c
@@ -1309,8 +1309,8 @@ asyncWriterThread(void *pPtr)
 				continue;
 			}
 			bTimedOut = 0;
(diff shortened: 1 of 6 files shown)


runtime/stream.c
```diff
@@ -1293,6 +1312,9 @@ asyncWriterThread(void *pPtr)
 			timeoutComp(&t, pThis->iFlushInterval * 1000); /* *1000 millisconds */
```
portante: This timeout computation could be moved under the `if` below, right?

rgerhards: good catch! Will change with next commit.

Issue #727, opened 2016-01-11
If the redis server is unavailble module omhiredis failed with Segmentation fault


missnebun:
#my test config

module(load="imtcp") 
module(load="imudp")
module(load="omhiredis")

input(type="imudp" port="514" ruleset="remote")
input(type="imtcp" port="514" ruleset="remote")

ruleset(name="remote"
    queue.type="FixedArray"        # we'll have a fixed memory queue for this ruleset
    queue.spoolDirectory="/var/spool/rsyslog" #Spool directory
    queue.filename="remote-input"
    queue.maxdiskspace="10m"
    queue.maxfilesize="5m"
    queue.highwatermark="800" #Start spilling to disk at this size
    queue.lowwatermark="400"   #stop spilling when it gets back to this size
    queue.size="120" # capacity of the queue
    queue.saveonshutdown="on"
    queue.workerthreads="1"
    queue.dequeuebatchsize="100"
    )
  {
    action(
      type="omhiredis"
      name="redis2050-fwd"
      server="10.250.24.205"
      mode="queue"         
      key="rsyslog_logstash"
      )
    action(
      type="omhiredis"
      name="redis2093-fwd"
      server="10.250.24.71"
      mode="queue"
      key="rsyslog_logstash"
	  action.execOnlyWhenPreviousIsSuspended="on"
      )
  }

  
  
  
  2732.137948661:imtcp.c        : epoll returned 1 entries
2732.137981398:imtcp.c        : epoll push ppusr[0]: 0x7f25a06fee90
2732.137996289:imtcp.c        : tcpsrv: ready to process 1 event entries
2732.138007097:imtcp.c        : tcpsrv: processing item 0, pUsr 0x7f25a06fee90, bAbortConn
2732.138097541:imtcp.c        : New connect on NSD 0x7f25a06ff120.
2732.138143215:imtcp.c        : dnscache: entry (nil) found
2732.141204500:imtcp.c        : adding nsdpoll entry 0/0x7f258c003550, sock 12
2732.141232063:imtcp.c        : New session created with NSD 0x7f258c003550.
2732.141243401:imtcp.c        : doing epoll_wait for max 128 events
2732.141255427:imtcp.c        : epoll returned 1 entries
2732.141264062:imtcp.c        : epoll push ppusr[0]: 0x7f258c003550
2732.141273041:imtcp.c        : tcpsrv: ready to process 1 event entries
[...]

PR #1062, opened 2016-06-10
Simplest fix for the omhiredis segfault

bobthemighty:
This is a brutally simple fix for the omhiredis segfault - if we receive an error from the redis context, we don't try to free the reply object. See https://github.com/redis/hiredis/issues/122 for more info.

I'd like to do a deeper reworking of the module, though, and try to reduce the number of network roundtrips. Comments suggest that the original author had this in mind.

I'm slightly confused by the use of the transaction handlers. Is there a good example of a transactional, batching output module that I can use for inspiration? In particular, what ought I to do if endTransaction fails? How many times should I expect a retry?

Is there a way to introduce a randomised backoff rather than immediately retrying the redis connection?

### DIFF
--- a/contrib/omhiredis/omhiredis.c
+++ b/contrib/omhiredis/omhiredis.c
@@ -232,12 +232,18 @@ ENDdoAction
 BEGINendTransaction
 CODESTARTendTransaction
 	dbgprintf("omhiredis: endTransaction called\n");
+    redisReply* reply;
 	int i;
 	pWrkrData->replies = malloc ( sizeof ( redisReply* ) * pWrkrData->count );
 	for ( i = 0; i < pWrkrData->count; i++ ) {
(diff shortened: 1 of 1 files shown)


bobthemighty:
Specifically, *this* segfault: https://github.com/rsyslog/rsyslog/issues/727

pemontto:
I tested this by pushing some logs to Redis, stopping Redis then pushing more logs.

This will prevent a segfault for sure, but it loses logs pushed while Redis is down and doesn't seem to reconnect or send logs after Redis has been started again.

bobthemighty:
Thanks for the feedback, @pemontto - the initial pull was more or less a proof-of-concept. I'd like to do a deeper reworking of the omhiredis module, but I've got some conceptual things I need to clear up first.

I'll close this PR and see if I can build something more robust.

PR #1064, opened 2016-06-11
omamqp1: bugfix - supply port number to message bus address

kgiusti:
Fixes a bug that will prevent connecting to the message bus if a port number other than 5672 is in use.


portante:
Seems like a good fix.

Issue #974, opened 2016-04-21
data can be lost when problems in commitTransaction happen


rgerhards:
Thanks to Kane Kim for identifying this.

The full begin of this story is on the rsyslog mailing list (e.g. http://lists.adiscon.net/pipermail/rsyslog/2016-March/042379.html), but we should have it on github especially as this may involve coordinated development. Short summary:

During end transaction processing, errors are not handled as good as possible. This is already indicated by a TODO comment: https://github.com/rsyslog/rsyslog/blob/master/action.c#L1236

The comment is partly outdated (see ML), but the core idea is correct. I think it would also be the proper spot to begin to extend things.

The best plan to address this is:

The plan acutally is:
1. get a testbench test to expose that problem -- this greatly eases debugging  line of thought is to update omfile, which is needed anyhow, so that we can induce errors via it
2. add missing functionality to write error file this does not directly address the issue, but is useful and ensures we are on the right path
3. check the end transaction endpoint. After review of old notes and some thinking, I think that      part of the "not yet implemented" comment is actually right. It's not as easy as suggested by it,  but we probably can do better.

During implementation, specifics of engine shutdown need to be considered. Let me stress that I think having a testbench test case is vital in order to apply any changes, as we must make sure this works well on many platforms, which almost rules out manual testing.

I would definitly and strongly recommend to tackle this beast in the order given above. I would also go for low-hanging fruits at first. David Lang has suggested advanced transaction error handling on the ML. This makes sense, but there are many subtle issues to consider. Unfortunately, I cannot elaborate much here, but the bottom line is that we need to have actions explicitely opt-in to use that functionality. So I strongly suggest to do anything advanced only after the base functionality described here is solid.

It should be save if the implementation blocks for the configured number of retries (and watching shutdown state) at the line mentioned above. One might look at the equivalent code path in the non-transactional case.

rgerhards:
moved the error file enhancement to its own issue tracker #1836 so that we can finally close the core issue here.

PR #1070, opened 2016-06-21
fix issues in action commitTransaction processing

rgerhards:
This also includes testbench enhancements.

see #500, #974 for more details

### DIFF
--- a/action.c
+++ b/action.c
@@ -1283,14 +1283,13 @@ actionCommit(action_t *__restrict__ const pThis, wti_t *__restrict__ const pWti)
 			ABORT_FINALIZE(RS_RET_FORCE_TERM);
 		} else if(iRet == RS_RET_SUSPENDED) {
 			iRet = actionDoRetry(pThis, pWti);
-			if(iRet != RS_RET_OK) {
+			if(iRet == RS_RET_FORCE_TERM) {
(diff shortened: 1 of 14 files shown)


action.c
```diff
@@ -1257,11 +1277,22 @@ actionCommit(action_t *__restrict__ const pThis, wti_t *__restrict__ const pWti)
 	bDone = 0;
 	do {
 		iRet = actionTryCommit(pThis, pWti);
-		DBGPRINTF("actionCommit, in retry loop, iRet %d\n", iRet);
+		DBGPRINTF("actionCommit, action %d, in retry loop, iRet %d\n",
+			pThis->iActionNbr, iRet);
+		if(iRet == RS_RET_FORCE_TERM) {
```
portante, line 1282: Should we then remove the same test on line 1292?

rgerhards, line 1282: no, because RS_RET_FORCE_TERM really tells us to finish as quickly as possible. So it is better to check before each potentially time-consuming task. But I appreciate your note!

side-note: while this code is better than what was there previously, some more refactoring is strongly desired and on my todo list. I just want to push this here as quickly as possible, as it definitely solves a number of hard-to-track issues (that we have seen on the issue trackers and mailing list).

portante:
@rgerhards, nice, that makes sense.  Thanks.

PR #1074, opened 2016-06-27
stream.c: strmNextFile, make sure strmNextFile modulo produces wanted…

kortemik:
… results. strmReadMultiLine, must append only lines longer than 0 after rsCStrGetSzStrNoNULL

### DIFF
--- a/runtime/stream.c
+++ b/runtime/stream.c
@@ -480,7 +480,7 @@ strmNextFile(strm_t *pThis)
 	 * make sure their iMaxFiles is large enough. But it is well-desired for other
 	 * use cases, e.g. a circular output log file. -- rgerhards, 2008-01-10
 	 */
-	if((pThis->iCurrFNum + 1) == 0 || pThis->iMaxFiles == 0) {
+	if((pThis->iCurrFNum + 1) == 0) {
 		pThis->iCurrFNum = 0;
 	}
 	else {
--- a/runtime/stream.c
+++ b/runtime/stream.c
@@ -480,7 +480,13 @@ strmNextFile(strm_t *pThis)
 	 * make sure their iMaxFiles is large enough. But it is well-desired for other
 	 * use cases, e.g. a circular output log file. -- rgerhards, 2008-01-10
 	 */
-	pThis->iCurrFNum = (pThis->iCurrFNum + 1) % pThis->iMaxFiles;
+	if((pThis->iCurrFNum + 1) == 0 || pThis->iMaxFiles == 0) {
+		pThis->iCurrFNum = 0;
+	}
+	else {
+		pThis->iCurrFNum = (pThis->iCurrFNum + 1) % pThis->iMaxFiles;
+	}
+
 
 finalize_it:
 	RETiRet;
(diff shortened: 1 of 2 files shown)


runtime/stream.c
```diff
@@ -480,7 +480,13 @@ strmNextFile(strm_t *pThis)
 	 * make sure their iMaxFiles is large enough. But it is well-desired for other
 	 * use cases, e.g. a circular output log file. -- rgerhards, 2008-01-10
 	 */
-	pThis->iCurrFNum = (pThis->iCurrFNum + 1) % pThis->iMaxFiles;
+	if((pThis->iCurrFNum + 1) == 0 || pThis->iMaxFiles == 0) {
```
portante: So on line 472 there is an assertion that `pThis->iMaxFiles` is not zero, so why do we still need this check in the conditional?

Without the need for checking for `pThis->iMaxFiles` is zero, won't that just revert back to what was originally there?

That is to say, we only need to change the sign of the variables used, right?

kortemik: besides extra check after the assertion, this is still a valid case. I will remove the extra

rgerhards: I think the question can also be reworded. looking at the assertion, iMaxfiles should never be zero. So I think the question really is why this invariant does not hold. Obviously there is a problem, but we should possibly try to identify the root cause.

runtime/stream.h
```diff
@@ -96,13 +96,13 @@ typedef struct strm_s {
 	BEGINobjInstance;	/* Data to implement generic object - MUST be the first data element! */
 	strmType_t sType;
 	/* descriptive properties */
-	int iCurrFNum;/* current file number (NOT descriptor, but the number in the file name!) */
+	unsigned int iCurrFNum;/* current file number (NOT descriptor, but the number in the file name!) */
```
rgerhards, line 99: Moving to unsigned is definitely the right thing to do, albeit int should not cause a segfault for negative numbers but rather somewhat strange file names (name-number).

runtime/stream.c
```diff
@@ -480,7 +480,7 @@ strmNextFile(strm_t *pThis)
 	 * make sure their iMaxFiles is large enough. But it is well-desired for other
 	 * use cases, e.g. a circular output log file. -- rgerhards, 2008-01-10
 	 */
-	if((pThis->iCurrFNum + 1) == 0 || pThis->iMaxFiles == 0) {
+	if((pThis->iCurrFNum + 1) == 0) {
```
rgerhards, line 483: As iCurrFNum now is unsigned, these two statements are not really needed - that's what module arithmetic does for us.

kortemik: yes you are totally right here

rgerhards:
I will take the data type change from this PR, but will not take the workaround for the segfault as it does not address the root cause.  I am trying to fix the root cause in https://github.com/rsyslog/rsyslog/pull/1173. Unfortunately, I have so far been unable to reproduce the issue. Any help with that would be appreciated.

kortemik:
We have reproduced it multiple times on a red hat and fedora based systems.
I will supply some example scenario as soon as possible.

rgerhards:
@kortemik repro would be great. If you can, it would also be good to see a stacktrace of a failure. What I do not understand is primarily why the function in question is called. This should not happen in those cases (and as I said, I can't reproduce it).

rgerhards:
side-note: the data type change in itself is not unproblematic. If the unsigned overflows MAX_INT, it still is passed as long to ```genFileName()``` which will result in the exact same failure as before. And, as before, it will corrupt disk queues in such cases.

PR #1077, opened 2016-07-04
bugfix omfile: handle chown() failure correctly

rgerhards:
If the file creation succeeds, but chown() failed, the file was
still writen, even if the user requested that this should be treated
as a failure case. This is corrected by this commit.

Alos, some refactoring was done to create better error messages.


portante:
Looks good from here.

PR #1085, opened 2016-07-12
build system: do not enable -Werror on release builds

rgerhards:


rgerhards:
We currently do not know how to disable AX_COMPILER_FLAGS on platfroms that do not have autoconf 2.64 or above. So putting this PR on hold until a solution is found.

Issue #985, opened 2016-04-24
use AX_COMPILER_FLAGS in rsyslog


rgerhards:
@mbiebl has added this to libfastjson, and it worked out very well. It ensures consistent warning messages. We should add this to rsyslog as well. Probably, this means we also need to do some cleanup (albeit we still have enabled many warnings).

rgerhards:
done, but we needed to exclude a very limited set of options.

rgerhards:
The macro should default to Warning only, IIRC. @mbiebl suggested it
exactly for this reason.

Sent from phone, thus brief.

Am 11.07.2016 22:40 schrieb "Thomas Deutschmann" <notifications@github.com>:

Whissi:
Mh, see https://www.gnu.org/software/autoconf-archive/ax_compiler_flags.html:

> Check for the presence of an –enable-compile-warnings option to configure, **defaulting to "error" in normal operation,** or "yes" if IS-RELEASE is equal to "yes". Return the value in the variable $ax_enable_compile_warnings.

[...]

> –enable-compile-warnings can take the values:
> 
>  * no:      Base compiler warnings only; not even -Wall.
>  * yes:     The above, plus a broad range of useful warnings.
>  * error:   The above, plus -Werror so that all warnings are fatal.
>             Use --disable-Werror to override this and disable fatal
>             warnings.

If you set IS-RELEASE in the release tarball I have nothing to do :)

PS: You should enable verbose build so you see which flags will be set.

Whissi:
*AX_IS_RELEASE* requires *>=autoconf-archive-20150224* (wily+).

But this shouldn't be a problem. Like I outlined in 2015 while doing 5f5eca03a09671b5df609f9ee6f885d63bbedaca we should depend on our special ".tarball-version" file for detection (i.e. move the detection to the beginning so we can use the variable not only for detecting the need of *flex*...). This will help people (I think Debian is doing that) who are storing releases in git so they don't create special development builds because they don't notice... At least this was my intention for the "hack".

I'll prepare a PR over the weekend.

rgerhards:
Some info: looks like I need to include the macros anyhow as i have seen
e.g. centos 6 did not provide autoconf-archive at all. That seems also to
account for some other errors...

Sent from phone, thus brief.

Am 14.07.2016 18:18 schrieb "Thomas Deutschmann" <notifications@github.com>:

mbiebl:
[...]

PR #1086, opened 2016-07-12
build system: remove AX_COMPILER_FLAGS macro, as it causes trouble

rgerhards:
On some platforms, warnings I don't see on any of my platforms are
triggered, and I don't get "is-relase" based on git directory working.
So I remove the macro for now and will re-try in the future.

see also https://github.com/rsyslog/rsyslog/issues/985


rgerhards:
applied to v8-stable, but master will continue to use the macro for now (we may remove it once again for the next release).

PR #1091, opened 2016-07-13
imfile: always set stream reopen flag

### LINKED ISSUE #1090: reopenOnTruncate is not obeyed unless a state file was loaded
`imfile` inputs with the `reopenOnTruncate` flag do not reopen when the file is truncated, unless rsyslogd had a state file present for the relevant file when it started.

This doesn't seem to be a problem with the actual usage of `reopenOnTruncate` flag inside the stream objects; instead, it seems that in these cases, `imfile` is not correctly constructing the streams, or passing the flag value to them, in this case. This was confirmed by patching in some extra debug logging in stream.c -- the stream obj sees a value of `false` for the relevant property within `strmHandleEOFMonitor`.

This has been observed on 8.16.0 and 8.20.0

**_Steps to reproduce:_**

Using a minimal `rsyslog.conf` file:

```
module(load="imfile")

template(name="a" type="list") {
         property(name="msg")
         constant(value="\n")
}

ruleset(name="a") {
    action(type="omfile" file="/var/log/rsyslogd-out.log" template="a")
}

input(type="imfile" file="/var/log/a.log" tag="info"  severity="info"  facility="local0" reopenOnTruncate="on" ruleset="a")
```

Run rsyslog, send content to the file then truncate it:

```sh
/usr/local/sbin/rsyslogd

echo 1 >> /var/log/a.log
echo 2 >> /var/log/a.log

truncate -s 0 /var/log/a.log

echo 3 >> /var/log/a.log
echo 4 >> /var/log/a.log
echo 5 >> /var/log/a.log

cat /var/log/rsyslogd-out.log
```

**_Expected Output_:**

`/var/log/rsyslogd-out.log` should contain:

```
1
2
3
4
5
```

**_Actual Output_:**
`/var/log/rsyslogd-out.log` contains:

```
1
2
5
```

From this, you can see that rsyslogd is behaving as if the reopenOnTruncate flag was never added. It looks exactly like #511 – even though this version includes #662 and the relevant config.

However... repeating the test _with state files present_ shows the expected behaviour. A simple way to demonstrate this is to stop and start rsyslogd before the test.

On a completely clean environment, run the following:

```sh
/usr/local/sbin/rsyslogd
kill $(cat /var/run/rsyslogd.pid)
/usr/local/sbin/rsyslogd

echo 1 >> /var/log/a.log
echo 2 >> /var/log/a.log

truncate -s 0 /var/log/a.log

echo 3 >> /var/log/a.log
echo 4 >> /var/log/a.log
echo 5 >> /var/log/a.log

cat /var/log/rsyslogd-out.log
```

This yields the expected output (numbers 1 to 5)

We can prove that it's related to state files by repeating that test, but cleaning up state files before the restart:

```sh
/usr/local/sbin/rsyslogd
kill $(cat /var/run/rsyslogd.pid)
sleep 1 # wait for state files to be written
rm /imfile*
/usr/local/sbin/rsyslogd

echo 1 >> /var/log/a.log
echo 2 >> /var/log/a.log

truncate -s 0 /var/log/a.log

echo 3 >> /var/log/a.log
echo 4 >> /var/log/a.log
echo 5 >> /var/log/a.log

cat /var/log/rsyslogd-out.log
```

This again shows the 'bad' output, missing `3` and `4`

orls:
Previously, it was set only when deserializing state files.

Fixes #1090 (see that issue for reproduction details)

### DIFF
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -538,7 +538,6 @@ openFile(lstn_t *pLstn)
 
 	/* read back in the object */
 	CHKiRet(obj.Deserialize(&pLstn->pStrm, (uchar*) "strm", psSF, NULL, pLstn));
-	CHKiRet(strm.SetbReopenOnTruncate(pLstn->pStrm, pLstn->reopenOnTruncate));
 	DBGPRINTF("imfile: deserialized state file, state file base name '%s', "
 		  "configured base name '%s'\n", pLstn->pStrm->pszFName,
 		  pLstn->pszFileName);
@@ -567,6 +566,7 @@ finalize_it:
 		if(pLstn->pStrm != NULL)
 			strm.Destruct(&pLstn->pStrm);
 		CHKiRet(strm.Construct(&pLstn->pStrm));
+		CHKiRet(strm.SetbReopenOnTruncate(pLstn->pStrm, pLstn->reopenOnTruncate));
 		CHKiRet(strm.SettOperationsMode(pLstn->pStrm, STREAMMODE_READ));
 		CHKiRet(strm.SetsType(pLstn->pStrm, STREAMTYPE_FILE_MONITOR));
 		CHKiRet(strm.SetFName(pLstn->pStrm, pLstn->pszFileName, strlen((char*) pLstn->pszFileName)));


orls:
I'm not sure if I can (should?) fix those Travis errors; they seem to be affecting recent 8.x stable builds too:

```configure: error: Package requirements (relp >= 1.2.12) were not met:
Requested 'relp >= 1.2.12' but version of relp is 1.2.11```

**Comment by davidelang** [comment]
librelp was updated last week, it sounds as if the build process isn't picking 
up the new version like it should.

David Lang

**Comment by rgerhards** [comment]
That's because we pulled the new Library versions yesterday. The 6th travis
run plus buildbot should be ok, anyhow. If it is just Travis 1-5 what
fails, you are ok.  I'll restart the tests once we have the libraries back
again.

Rainer

Sent from phone, thus brief.

Am 13.07.2016 18:58 schrieb "davidelang" <notifications@github.com>:

**Comment by rgerhards** [comment]
I am merging this under #1091 in order to get the new CI settings. Alternatively, you can rebase to master.

In any case, I think the patch is incomplete. When the experimental reopen feature was added, state file handling was not properly considered. The patch here IMHO will fail when a state file is present. I think we also need to persist the reopen flag to the state file and re-read it back in when the state file is processed. Alternatively, we can always pull it from the current rsyslog.conf (one may argue this is the right thing to do, as it is not really a file flag). In that case, it needs to be applied, too, when the state file is read.

Comments?

**Comment by orls** [comment]
> Alternatively, we can always pull it from the current rsyslog.conf [...] In that case, it needs to be applied, too, when the state file is read.

That's what this change does. It moves the flag setter from somewhere that is _only_ run when state file is loaded, to somewhere that is _always_ run, thus ensuring that the value from the config file is always used.

I'll rebase soon and re-test to confirm, but I've definitely locally tested the flag being set with and without state files.

File: plugins/imfile/imfile.c [hunk 1]
```diff
@@ -567,6 +566,7 @@ openFile(lstn_t *pLstn)
 		if(pLstn->pStrm != NULL)
 			strm.Destruct(&pLstn->pStrm);
 		CHKiRet(strm.Construct(&pLstn->pStrm));
+		CHKiRet(strm.SetbReopenOnTruncate(pLstn->pStrm, pLstn->reopenOnTruncate));
```
rgerhards, line 569: Have a look at the guarding `if`: this code is only executed if the state file did not exist. So if it exists, the flag is never applied.

rgerhards:
fixed via #1175 - thanks for the great help in finding and fixing the issue!

PR #1096, opened 2016-07-22
nsd_gtls: Handle chain of certificates

baloo:
Allow a server to serve a list of certificates. Much public CA now requires a server to serve a chain of certificates up to the root CA.

This allows to use valid public certificates for log-servers.


baloo:
Works with recent version of gnutls. I'd prefer to have a confirmation from a maintainer this is mergeable (from a feature point of view) before investing time on a compatibility for older versions of gnutls.

rgerhards:
Looking at the testbench runs, this seems to have a memory leak.

PR #1099, opened 2016-07-25
new maxminddb lookup message modify plugin

chenryn:


rgerhards:
It would be great if you could rebase to master. That would make the CI errors go away, at least if everything else is ok ;)

davidelang:
There are a lot of changes showing here that are not part of the new plugin, that may because you did a merge rather than a rebase.

the sparse table lookup functionality was designed specifically to be able to lookup maxmind database info. how is this better?

chenryn:
The table lookup function need to distribute the IP database as JSON file
which size would be much larger. And Maxmind.Inc only provider .mmdb and
.csv format.
I forgot use -f option for git push......I can send a new PR for this if
you think this module is needed.

2016-07-26 23:40 GMT+08:00 davidelang <notifications@github.com>:

davidelang:
csvjson --stream will do a wonderful job of converting from .csv to json

table lookup includes a mechanism to be able to re-load the table based on sending an appropriate log message, is there a mechanism for this module to re-load the table?

I'm not necessarily opposed to this module, but if we have a generic mechanism that can be used for a purpose, adding a specific mechanism for the same purpose had better have a pretty significant benefit to offset the additional maintenance and support that it's going to cost in the future. the size on disk of the data is not a big enough benefit (the maxmind database isn't _that_ large :-)

I doubt that there is anything you are doing that's going to end up being faster than a table lookup with sparse tables, so I'm trying to identify what other benefits adding a specific module for this database are going to add.

chenryn:
@davidelang I haven't use the new sparse lookup table so much, and don't find some complex examples in testsuites.
Can sparse lookup deal with different ip addr class in index like:
```
223.255.192.0/19
223.255.224.0/24
223.255.225.0/26
223.255.225.64/29
```
And can it return complex result rather than only string value?

davidelang:
sparse arrays work exactly like the maxmind database, you convert the IP address into a 32 bit int (there is a function to do this, I don't remember it right now), and then you look for the highest entry that isn't larger than your lookup value, so it can handle any arbitrary ranges, they don't need to be cidr blocks

the current version only returns one value, but that value could easily be a json string to be parsed by mmnormalize. I hope that eventually someone will be interested enough to enhance it to return an arbitrary structure (I suspect that this is mostly going to be a matter of changing the table load routine)

chenryn:
ok, need some ip2long convert, but already to cover the need. close this.

rgerhards:
@davidelang from a usability point of view, might it actually make sense to have this included as an easy way to get the functionality? I mean as long as it is a contributed module, we don't really care if it breaks. On the other hand, if it is not included in the doc set, it is unlikely that somebody will actually discover that it exists, so this might be a mood point (except when doc is added).

davidelang:
@chenryn when I was creating the spec for the table lookup, it was specifically with the plan to use it for the maxmind geoIP lookup use-case. (this is also why the table load function is async, because the geoIP database is large enough that you don't want to stop all log processing while it's happening)

Unfortunantly the company I was at when I was doing that never got around to making use of it, so a good doc example of using it would be great.

Also, an enhancement to let it return a structure instead of a simple string would be great.

hmm, it looks like the ip2long function (or whatever we were calling it) got dropped from the documentation. This probably means that it never got implemented.

rgerhards:
@chenryn I kept this closed because I thought you would not longer support merging it. However, I see that you refer to this PR. Any concerns if I merge it?

@davidelang I am with you with the lookup tables, but as I said above, I think it doesn't hurt to have this module inside the code base (as contrib), especially as it seems immediately usable.

davidelang:
On Sat, 26 Nov 2016, Rainer Gerhards wrote:

My concers are if we have too many ways of doing the same thing it can confuse 
people as to how to do it, and the ongoing maintinance effort (or reputational 
damage if it's not maintained and breaks for someone)

I'd like to have us do a comparison of the two, showing where each has an 
advantage (including performance comparisons for both small and large datasets)

I'd much rather have one solution that satisfies all needs than two, partially 
overlapping, solutions that each work in some cases, and have the situaion of 'U 
lets me do V, X lets me do Y, but I want to do both V and Y and there is no way.


do we possibly need an "experimental" category that is like contrib, but with 
more warnings that this may be dropped in a future release?

David Lang

rgerhards:
You have very valid concerns, but IMHO lookup tables cannot currently do
what this module does. It might not be missing much, but at least I do not
see how it would work. I admit I do not have time to work intensely on
this.

If you have time to try it out and document it, I am ready to implement a
missing function or two.

But as long as we do not have an easy alternative, I think this module
solves an actual need. Regarding "experimental", I think we clearly say
that we cannot guarantee anything for contributing, so I think there is not
a more experimental than this (but I may be wrong).

Rainer

Sent from phone, thus brief.

Am 26.11.2016 21:18 schrieb "davidelang" <notifications@github.com>:

davidelang:
two items that would be needed in table_lookup are #1279 and #1287 

once they are in place, the differences that I see are:

table_lookup:

- allows the table contents to be updated without requiring a full restart.
- can be used for any data type
- requires pre-processing of the maxminddb files before they are used

maxminddb_lookup

- allows you to fetch specific items instead of the entire record.
- only works for the maxminddb format
- uses the current maxminddb files as-is

I think a big factor would be if there is any significant difference in lookup performance between the two.

rgerhards:
given the fact that we need to do some development on lookup tables and this plugin here works out of the bug, I think I will merge it as experiental/may be removed later & is unsupported -- except if there is violent objection against this.

IMHO this gives us a feature many would like to have and some time to review what we do with lookup tables, write code and troubleshoot. So that we can make an informed decision in Q1/2017.

Anyone *violently* objecting this plan?

davidelang:
adding this as experimental, and having a plan for what to test and when to make 
a decision makes perfect sense

We aren't commiting to dealing with it long term, we aren't leaving it in limbo 
because we have a timeframe and specific criteria to evaluate.

David Lang

rgerhards:
@mostolog would you be willing to extract doc from README and add it to the regular doc set?

mostolog:
Sure, but probably won't be able before Dec 20th. Link, please!

rgerhards:
It's inside the README file of this PR. I'll also see if we can get it done earlier.

PR #1101, opened 2016-07-26
travis: re-enable all new component testing via special pkg repo

rgerhards:
This is an experiment and depends on whether or not we can setup a special PPA for travis use.


rgerhards:
PR #1113 superseeds this one, it restores the original functionality.

PR #1139, opened 2016-09-07
Update rsyslogd.8

feld:
Remove sentence from copy/paste of -n to create documentation for -N


rgerhards:
The opensuese buildbot failure is caused by newer version and new compiler warnings, so it can be safely ignored.

Issue #1115, opened 2016-08-12
various compiler warnings (with GCC 6)


mbiebl:
This is not a targetted issue. With GCC 6 I get a couple of compiler warnings which might be worth looking into. Attached as build.txt

[build.txt](https://github.com/rsyslog/rsyslog/files/414801/build.txt)

mostolog:
Today's build just raised a warning about _asprintf_, being discussed on mail list.

@radu-gheorghe  @rgerhards If you are working with a local copy, please consider using regex to **find and replace all places where an if without curly braces are being used** as a good practice. https://blog.codecentric.de/en/2014/02/curly-braces/

Please, close!

rgerhards:
This should be solved for a while, closing.

PR #1150, opened 2016-09-22
Always using curly braces makes your life much safer (some compile warnings fixed)

mostolog:
Related to https://github.com/rsyslog/rsyslog/issues/1115

### DIFF
--- a/runtime/debug.c
+++ b/runtime/debug.c
@@ -133,14 +133,18 @@ static pthread_key_t keyCallStack;
  * rgerhards, 2008-01-23
  */
 #define DLL_Del(type, pThis) \
-	if(pThis->pPrev != NULL) \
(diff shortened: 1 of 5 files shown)


mostolog:
Any idea why it's suse giving errors?

rgerhards:
If you can rebase to current master, the suse problems will go away. If you can't I'll later create a merge PR to get a full testbench run.

mostolog:
If you agree, I could try to regex all .c/.h files to "prettify" the curly braces this way:
Of course, I would've to successfully built before PR.

```
int main () {
    if (foo) {
        foo;
        bar;
    } else {
        foo;
    }
    do {
        bar;
    } while (foo);
}
```

mostolog:
Just issued a new/wide one, closing this

PR #1161, opened 2016-09-29
close/reopen file when write fails for specific error codes

### LINKED ISSUE #908: rsyslogd writing to shared storage file fails after almost concurrent reboot of storage servers
Hi,
 
1)      I have a setup where rsyslogd writes to shared storage file ( shared storage file which is provided by glusterfs with 2 servers ).
2)      I reboot glusterfs server 1 and then in less than a minute I reboot glusterfs server 2.
3)      Rsyslogd gets stack in the state below while 2 glusterfs servers seem to recover after their almost concurrent reboot.
 
Receive the following error from strace command of rsyslogd
strace -f -ewrite -f -p <rsyslog pid>
Process 17463 attached with 8 threads
[pid 4711] write(10, "Mar 8 15:27:05.592037 warning L"..., 92) = 92
[pid 4711] write(10, "Mar 8 15:27:05.633467 warning L"..., 92) = 92
[pid 4711] write(10, "Mar 8 15:27:05.669069 warning L"..., 92) = 92
[pid 4711] write(7, "Mar 8 15:27:11.077982 err LEI-0"..., 103) = 103
[pid 4711] write(10, "Mar 8 15:27:11.077982 err LEI-0"..., 103) = 103
[pid 4711] write(7, "Mar 8 15:27:11.078299 err LEI-0"..., 103) = -1 ENOTCONN (Transport endpoint is not connected)
[pid 4711] write(10, "Mar 8 15:27:11.078299 err LEI-0"..., 103) = -1 ENOTCONN (Transport endpoint is not connected)
 
Do you have any idea why this issue might happen ?
 
System versions in use :
glusterfs 3.6.9
rsyslogd 8.15.0

hese10:
Fixes #908. When writing fails for file with error codes ENOTCONN or EIO, then file probably locates in network mounted partition, and thus needs to be closed and reopened.

### DIFF
--- a/runtime/stream.c
+++ b/runtime/stream.c
@@ -1179,6 +1179,10 @@ doWriteCall(strm_t *pThis, uchar *pBuf, size_t *pLenBuf)
 			DBGPRINTF("log file (%d) write error %d: %s\n", pThis->fd, err, errStr);
 			if(err == EINTR) {
 				/*NO ERROR, just continue */;
+			} else if( !pThis->bIsTTY && ( err == ENOTCONN  || err == EIO )) {
+				/* Failure for network file system, thus file needs to be closed and reopened. */
+				close(pThis->fd);
+				CHKiRet(doPhysOpen(pThis));
 			} else {
 				if(pThis->bIsTTY) {
 					CHKiRet(tryTTYRecover(pThis, err));


runtime/stream.c
```diff
@@ -1179,6 +1179,10 @@ doWriteCall(strm_t *pThis, uchar *pBuf, size_t *pLenBuf)
 			DBGPRINTF("log file (%d) write error %d: %s\n", pThis->fd, err, errStr);
 			if(err == EINTR) {
 				/*NO ERROR, just continue */;
+			} else if( !pThis->bIsTTY && ( err == ENOTCONN  || err == EIO )) {
+				/* Failure for network file system, thus file needs to be closed and reopened. */
```
rgerhards, line 1183: This looks like a good work-around for this case.

Issue #1162, opened 2016-09-29
imfile plugin does not initialize iMaxFileSize which causes divide-by-zero exception


hubo1016:
It should also be the root cause of #1066 

imfile never set strm_t.iMaxFileSize field and leave it to 0. When uses multiline mode e.g. setting startmsg.regex, this causes a divide-by-zero exception in the following statements in strm

``` C
static rsRetVal
strmNextFile(strm_t *pThis)
{
    DEFiRet;

    ASSERT(pThis != NULL);
    ASSERT(pThis->iMaxFiles != 0);
    ASSERT(pThis->fd != -1);

    CHKiRet(strmCloseFile(pThis));

    /* we do modulo operation to ensure we obey the iMaxFile property. This will always
     * result in a file number lower than iMaxFile, so it if wraps, the name is back to
     * 0, which results in the first file being overwritten. Not desired for queues, so
     * make sure their iMaxFiles is large enough. But it is well-desired for other
     * use cases, e.g. a circular output log file. -- rgerhards, 2008-01-10
     */
    pThis->iCurrFNum = (pThis->iCurrFNum + 1) % pThis->iMaxFiles;

finalize_it:
    RETiRet;
}
```

This line of statement

``` C
pThis->iCurrFNum = (pThis->iCurrFNum + 1) % pThis->iMaxFiles;
```

causes the program to crash.

There are a few ways to fix this problem:
1. Set the iMaxFiles field in imfile module, but the same problem may appear again in other modules
2. Initialize iMaxFiles field to a reasonable non-zero value in strmConstruc to prevent it from crashing. Modules are free to initialize the field to preferred value after construction.
3. Add a judge in strmNextFile for iMaxFiles = 0. But the number may overflow after a very long time (though not quite possible)

Used configuration file:

```
module(load="imfile")
module(load="omkafka")

$WorkDirectory /data
$MainMsgQueueType LinkedList
$MainMsgQueueFileName local5.info
$MainMsgQueueHighWatermark 2400000
$MainMsgQueueLowWatermark 100000
$MainMsgQueueMaxDiskSpace 100g
$MainMsgQueueSize 4000000
$MainMsgQueueSaveOnShutdown on
$MainMsgQueueDiscardSeverity 8
$MainMsgQueueDiscardMark 4000000
$MainMsgQueueTimeoutEnqueue 3000
$MainMsgQueueDequeueBatchSize 1000
$MaxMessageSize 5000k

[...]

PR #1163, opened 2016-09-29
Initialize iMaxFiles to 1024 to prevent crash in strmNextFiles, #1162

hubo1016:
#1162


kortemik:
duplicate #1074

rgerhards:
closing this, as it cures the symptom, but not the root cause. #1173 will do that, once we are able to reproduce a failure.

PR #1169, opened 2016-10-04
Barakshechter/omelasticsearch max content length

barakshechter:
Adding the ability to specify the maxbytes to send to elasticsearch as part of a bulk request so as not to exceed the max_content_length specified in the elasticsearch configuration (/etc/elasticsearch/elasticsearch.yml: http.max_content_length). The default value for this is 100MB, which under heavy log aggregation load with batching can easily be reached.

Once maxbytes is reached, the messages that have been batched are flushed by POSTing to elasticsearch and a new buffer is started.


barakshechter:
Fixes #1170

rgerhards:
can you rebase this to master? That would get us rid of the suse build problem.

barakshechter:
Done

rgerhards:
Thx! It would be great if you could also update the doc at https://github.com/rsyslog/rsyslog-doc/blob/master/source/configuration/modules/omelasticsearch.rst

barakshechter:
https://github.com/rsyslog/rsyslog-doc/pull/261

rgerhards:
sorry to bother you again, I just noticed that during the final review and had overlooked it previously. If I am not mistaken, you have changed the existing testbench tests to include the new option. However, this leaves the case untested that the option is not given. I think the best path forward is to either duplicate all tests (may be overdone) or just select one or two out of them where the new option is used (and explicitly tested during that test). It would be great if you could have a look, else I'll do that later on during the merge.

barakshechter:
Makes sense; the default value is probably going to be the most used, so
perhaps it's better to revert all the tests and just duplicate the
es-basic-bulk with the override of maxbytes=1k? I think that should cover
what we need..

I can probably do it in a few days.

On Oct 11, 2016 3:22 PM, "Rainer Gerhards" <notifications@github.com> wrote:

sorry to bother you again, I just noticed that during the final review and
had overlooked it previously. If I am not mistaken, you have changed the
existing testbench tests to include the new option. However, this leaves
the case untested that the option is not given. I think the best path
forward is to either duplicate all tests (may be overdone) or just select
one or two out of them where the new option is used (and explicitly tested
during that test). It would be great if you could have a look, else I'll do
that later on during the merge.

—
You are receiving this because you authored the thread.
<https://github.com/rsyslog/rsyslog/pull/1169#issuecomment-252901727>, or mute
the thread

rgerhards:
sounds like a good plan!

barakshechter:
Reverted and test added.

--
Barak Shechter
T(443) 221-4647   F(509) 352-6941

rgerhards:
thx for all your good work and persistance!

PR #1173, opened 2016-10-05
imfile: abort in stream handler

rgerhards:
see #1162, #1163, #1074

### DIFF
--- a/runtime/stream.c
+++ b/runtime/stream.c
@@ -493,9 +493,10 @@ strmNextFile(strm_t *pThis)
 {
 	DEFiRet;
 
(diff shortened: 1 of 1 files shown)


rgerhards:
The main problem is that we cannot reproduce the issue, thus I put this PR on hold until we find some way to do so. Note that the current testbench runs succeed, but users report aborts (which we do not know how to reproduce).

rgerhards:
I am merging this as-is (mostly void) and will open a new PR once we get a way to reproduce. Tracking this via https://github.com/rsyslog/rsyslog/issues/1162

PR #1176, opened 2016-10-06
First prettify iteration

mostolog:
Run a few regex to prettify curly braces, break lines and so...lot of work still to be done.

### DIFF
--- a/action.c
+++ b/action.c
@@ -203,8 +203,7 @@ static struct cnfparamblk pblk =
 /* primarily a helper for debug purposes, get human-readble name of state */
 /* currently not needed, but may be useful in the future!
 static const char *
-batchState2String(const batch_state_t state)
-{
+batchState2String(const batch_state_t state) {
 	switch(state) {
 	case BATCH_STATE_RDY:
 		return "BATCH_STATE_RDY";
@@ -247,8 +246,7 @@ batchState2String(const batch_state_t state)
  * a lot. So we simply return the system time.
  */
 static inline time_t
-getActNow(action_t * const pThis)
-{
+getActNow(action_t * const pThis) {
 	assert(pThis != NULL);
 	if(pThis->tActNow == -1) {
 		pThis->tActNow = datetime.GetTime(NULL); /* good time call - the only one done */
@@ -269,8 +267,7 @@ getActNow(action_t * const pThis)
  * rgerhards, 2008-01-29
  */
 static rsRetVal
-actionResetQueueParams(void)
-{
+actionResetQueueParams(void) {
 	DEFiRet;
 
 	cs.ActionQueType = QUEUETYPE_DIRECT;		/* type of the main message queue above */
@@ -307,8 +304,7 @@ actionResetQueueParams(void)
 /* destructs an action descriptor object
  * rgerhards, 2007-08-01
  */
-rsRetVal actionDestruct(action_t * const pThis)
-{
+rsRetVal actionDestruct(action_t * const pThis) {
 	DEFiRet;
 	ASSERT(pThis != NULL);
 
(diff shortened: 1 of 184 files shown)


rgerhards:
we need to discuss this in more depth, pls don't invest too much time for the moment. While in general this kind of "fix" is great, it thrashes git history and hurts in troubleshooting if done very broadly. Also, we prefer
```
funct()
{
}
```
over
```
funct() {
}
```
@janmejay, @portante what do you think?

If we go for the large change, we would also need to automatically verify that PRs match in code style, else this is a constant struggle.

mostolog:
No problem.
I could do smaller commits/PR to make it easier, eg:
 - adding curly braces for single-line if
 - adding curly braces for single-line else if
 - adding curly braces for single-line else
...

Just saved the regex for a later re-usage.

janmejay:
Sounds alright to me.

I think we should setup automated style-checking and do a big-bang
style-change-only commit to fix all existing code.

I do prefer the later personally, but it's better to pick one style and
stick to it at a codebase level.

What do you think about moving away from tabs in favor of spaces?

rgerhards:
I think tabs is a better solution. Some folks want 4-space indention, some 8 some even somewhat else. With tab, you have one level and everybody can set his editor to what he wants. In most cases, this should NOT cause invalid indentions (I know it sometimes does, and I know it does more often in the current code base than necessary).

rgerhards:
also, with TABS converted to spaces, ALL code lines change and we have no usable history left at all...

janmejay:
Tab change will definitely hurt history quite bad.

What changes do you have in mind for next few iterations? May be its worth
folding some in and make them in few commits or single commit, because
it'll allow isolating style changes and functional changes.

mostolog:
IMHO there aren't so many contributions to project not to be "easy enough" to schedule this kind of changes. Changing from spaces to TABs can be done in one commit and -if you don't have several branches open- won't affect your work too much.

Same applies to scheduling adding curly braces to single line conditionals, splitting lines with several statements...whatever you decide

rgerhards:
spaces to tabs is probably acceptable as it doesn't affect the mass of files. In regard to style changes, we need to think how we can do this best and how we can enforce that with new PRs.

mostolog:
If there's a git expert around, I'm all ears...Otherwise, we could write a script to "regex all files" to camply with a style-guide and request to run it before PR. 2 cents.

rgerhards:
Well, that's a way forward. I am skeptic on the regex part, though. Looking at the testbench run, the regex seems not to handle many situations right. Probably a more specialised tool might be better up for the job. For example, I know that the clang suite has a code formatter, also gnu has, and we might use one of these tools with the proper options. I am not an expert here, that's for sure, but I think we should look in this direction. If we build something, we must ensure that it always works and code parsing is a very though job so we better use prooven tools IMHO.

rgerhards:
maybe it would also good if you could post your regex script so that we better see what you are actually proposing.

mostolog:
This is a draft I was using to try these changes:

#break multiple lines
^(( |\t)*)(((?!for).)*;) ?((.*);)( |\t)*\n       ->      $1$3\n$1$5\n
##review!

#split if(...) statement; into multiple lines
^(( |\t)*)if( |\t)*\((.*)\)( |\t)*(.*;)\n         ->      $1if \($4\) {\n$1\t$6\n$1}\n
##review, it may fail in some cases

#move curly braces to previous line
^(.*)\n( |\t)*\{( |\t)*\n         ->      $1 {\n

#join )\n...{ at first line, and check to always have an space in between
\)( |\t)*\n?( |\t)*\{( |\t)*\n           ->       \) \{\n
##lines having something but blank after { aren't be affected. review

##Merge all the following using better regex
#prettify single line if
^(( |\t)*)if( |\t)*\((.*)\)( |\t)*\n(.*;)\n
$1if \($4\) {\n$6\n$1}\n
#prettify single line else if
^(( |\t)*)else if( |\t)*\((.*)\)( |\t)*\n(.*;)\n
$1else if \($4\) {\n$6\n$1}\n
#prettify single line else
^(( |\t)*)else( |\t)*\n(.*;)\n
$1else {\n$4\n$1}\n

#pending do, while, for , switch, case, struct...

rgerhards:
I think we should turn this into an issue and also have a mailing list discussion on it. Closing the PR for now (trying to get rid of PR cluttering), please feel free to open an issue.

Issue #1179, opened 2016-10-07
Compile message 


mostolog:
When building on ubuntu16.04 latest 8.22 source I'm getting a few messages like:

```
make[2]: Entering directory '/rsyslog/compat'
  CC       compat_la-getifaddrs.lo
  CC       compat_la-strndup.lo
  CCLD     compat.la
ar: `u' modifier ignored since `D' is the default (see `U')
...
```

Is there anything I could do to sanitize the building script to avoid those messages?
Regards

rgerhards:
That's an autotools issues, see for example here: https://bugzilla.redhat.com/show_bug.cgi?id=1155273

If you find a way to fix it, I would appreciate a PR.

mostolog:
As stated on https://bugzilla.redhat.com/show_bug.cgi?id=1155273#c13, perhaps setting ARFLAGS &
AR_FLAGS to 'cr' fixes the warning, but I don't know what consequences it may have.
Seem little or none, so I'm going to try a local build and let you know

D3m0nKingx:
I did some digging into this, granted this issue is old. But the problem isn't with libtoolize, by default it recognizes the correct flags ('u' has been replaced by 'U'). The issue when executing `./autogen.sh` it reads the configure.ac file and sees that 'u' is the modifer being used, so by changing this to U, and re-running autogen.sh, the correct flag will be injected into the libtool script in the working directory. 

Hope this will be more clear for others that come across this in the future.

Furthermore AR is an archiving tool: https://sourceware.org/binutils/docs/binutils/ar-cmdline.html

From what I can tell, is that it uses the 'U' modifier to make sure it uses recently modified files (by timestamp) when creating the archives. using 'D' would disregard this and include any file (zero timestamp).

mbiebl:
Shouldn't this be rather fixed in libtool itself? Having individual projects work around this sounds like the wrong way to address this

D3m0nKingx:
Seems this is related more to automake: http://gnu-automake.7480.n7.nabble.com/bug-20082-new-warning-from-ar-on-rawhide-systems-td21769.html

rgerhards:
That mail thread IMHO clearly says it is nothing that rsyslog has to fix.

Rainer

PR #1183, opened 2016-10-10
Change AR_FLAGS defaults to avoid warning messages

mostolog:
#1179

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -22,6 +22,12 @@ AC_DISABLE_STATIC
 AC_PROG_LIBTOOL
 AC_CANONICAL_HOST
 
+# https://github.com/rsyslog/rsyslog/issues/1179
+# Seems libtool doesn't honors ARFLAGS, but statically set AR_FLAGS=cru
+# This causes a few warnings on build:
+#   "ar u modifier ignored since D is the default (see U)"
+m4_divert_text([DEFAULTS], [: "${ARFLAGS=Ucru} ${AR_FLAGS=Ucru}"])
+
 m4_ifdef([AX_IS_RELEASE], [
 	AX_IS_RELEASE([git-directory])
 	m4_ifdef([AX_COMPILER_FLAGS], [
--- a/configure.ac
+++ b/configure.ac
@@ -26,7 +26,7 @@ AC_CANONICAL_HOST
 # Seems libtool doesn't honors ARFLAGS, but statically set AR_FLAGS=cru
 # This causes a few warnings on build:
 #   "ar u modifier ignored since D is the default (see U)"
-m4_divert_text([DEFAULTS], [: "${ARFLAGS=Ucru} ${AR_FLAGS=Ucru}"])
+m4_divert_text([DEFAULTS], [: "${ARFLAGS=cr} ${AR_FLAGS=cr}"])
 
 m4_ifdef([AX_IS_RELEASE], [
 	AX_IS_RELEASE([git-directory])


configure.ac
```diff
@@ -22,6 +22,12 @@ AC_DISABLE_STATIC
 AC_PROG_LIBTOOL
 AC_CANONICAL_HOST
 
+# https://github.com/rsyslog/rsyslog/issues/1179
+# Seems libtool doesn't honors ARFLAGS, but statically set AR_FLAGS=cru
+# This causes a few warnings on build:
+#   "ar u modifier ignored since D is the default (see U)"
+m4_divert_text([DEFAULTS], [: "${ARFLAGS=Ucru} ${AR_FLAGS=Ucru}"])
```
rgerhards: Judging from the testbench failures, this does not look portable. @Whissi @mbiebl do you have an idea of how this could be addressed?

mostolog: As I don't have centos/freebsd to try, I wasn't able to test this.
Anyway, setting:

> m4_divert_text([DEFAULTS], [: "${ARFLAGS=cr} ${AR_FLAGS=cr}"])

instead could work.
May I retry?

rgerhards:
I think we need some solution that is thightly integrated into autotools. I don't have all possible platforms at hand, and risking to break some build platforms just to get rid of cosmetic warnings isn't the right thing to me. If some autotools expert suggests a way forward or a link with a defintive answer can be found, that's fine to me. But just experimenting with options isn't the right approach IMHO.

mostolog:
As far as I have read, the problem is libtool has hardcoded AR_FLAGS, instead of using ARFLAG (as seems it should). Of course, we could just wait to libtool being fixed...

rgerhards:
I suggest to take this discussion to the mailing list to get broader feedback. Right now, I would object any change that cannot be proven to be portable to all platforms autotools supports. And we don't have them all, so this cannot be done by purely experimental approach. 

Maybe you can add an --enable-my-ar-fix configure flag, which by default is off. That would be fine with me (and actually appreciated until a final clean solution).

Whissi:
@rgerhards: You mentioned some testbench failures caused by this change. Where do I find these? GitHub say "All checks have passed".

rgerhards:
@Whissi That's because the PR was changed  for freebsd/centos. But that's hardcoded and does not guarantee that e.g. Solaris or Gentoo will work.

ALL: Maybe we can model that path the other way around: for those platforms that we know to work with the changed definition, use it. For all unkonwn platforms, use whatever autotools provides as default. How does that sound?

mostolog:
To be honest, I wouldn't take so much worries with this. I'll dare to say it's all about making a cosmetic change to avoid a warning or not. I don't believe this will become a break change on any platform...but many errors along history are caused by this naïf way of thinking.

**libtool has hardcoded "ar cru" and not using ARFLAGS variable hence we're getting some warnings.** According to https://linux.die.net/man/1/ar and "Ubuntu --help"

{code}
  [c]          - do not warn if the library had to be created
  r[ab][f][u]  - replace existing or insert new file(s) into the archive
  [u]          - only replace files that are newer than current archive contents
  [U]          - use actual timestamps and uids/gids
{code}

So, setting **cr** instead of **cru** or even **Ucru** does little change.
First patch I proposed was using Ucru (update check based on timestamps/guid) but that option was unsupported on freebsd/centos6
Then I just update PR to use cr instead.

To my limited understanding that could lead only to "longer" building times (replacing all files instead of modified-only into archive), but as I said before, world is full of reckless people like me!!!

rgerhards:
Well, couldn't you re-write the patch so that it only changes the flags on those platforms where we know it works? It might be a small list, but nevertheless useful there. you obviously need to check platform name and version, of course and in theory we might get issues with future versions. I would be willing to take the risk for future versions, but am conservative in changing things, **for cosmetic reason** that might break builds (yes, **might** IMHO is sufficient, aka "cannot outrule for sure").

Whissi:
I wouldn't do anything regarding this warning in rsyslog. It isn't rsyslog which sets the AR FLAGS value, right? There's a warning because on some distributions which build *binutils* with `--enable-deterministic-archives` configure option which changes flags but are using a *libtool* and/or *automake* package which wasn't updated yet to reflect this change.

So this is a problem in the toolchain in these distributions. Nothing which should get fixed downstream in the actual software which get build using this toolchain. 

If you really want to address this in rsyslog than make sure you only change this for platforms which actually need this. For me this is something a package maintainer would do. I.e. on Gentoo for example we don't set `--enable-deterministic-archives` yet. So it wouldn't make sense to change flags for us.

And I guess you can't make the change just for Debian/Ubuntu/CentOS because rsyslog also supports older versions of these distributions which are still using *binutils* <2.26 (i.e. a version without these change).

mostolog:
I'll stick to the principle don't fix something on rsyslog which belongs to libtool just for cosmetic reasons. Open an issue on libtool instead.
Closing PR.

Whissi:
FYI: It is already fixed in libtool git however no version containing the fix was released yet. automake has nothing done yet AFAIK.

D3m0nKingx:
https://github.com/rsyslog/rsyslog/issues/1179#issuecomment-357524827 This should clarify for anybody that comes across this. The issue isn't with libtoolize. The correct flag that AR currently uses is 'U', the old 'u' is deprecated, and thus needs to be replaced to 'U' in configure.ac and then have `./autogen.sh' re-run.

rgerhards:
@D3m0nKingx thx for the post. I would appreciate if you could be a bit more specific where exactly the 'u' needs to be replaced to 'U' in configure.ac.

D3m0nKingx:
@rgerhards I think I've tracked down why this occurring: https://lists.gnu.org/archive/html/libtool-commit/2015-06/msg00002.html

Question is how to set it so that AR_FLAGS = cr in libtool.m4 because:

>m4/libtool.m4 (_LT_PROG_AR): Copy ARFLAGS value into AR_FLAGS
    variable if AR_FLAGS is not set.  Add new _LT_DECL'ed variable
    'lt_ar_flags' to keep the configure-time value of AR_FLAGS.  The
    new 'lt_ar_flags' is to be used as the default value for AR_FLAGS
    at libtool-runtime.

D3m0nKingx:
```
# _LT_PROG_AR
# -----------
m4_defun([_LT_PROG_AR],
[AC_CHECK_TOOLS(AR, [ar], false)
: ${AR=ar}
: ${AR_FLAGS=cr}
_LT_DECL([], [AR], [1], [The archiver])
_LT_DECL([], [AR_FLAGS], [1], [Flags to create an archive])
```

D3m0nKingx:
So I've basically come to the point that to make this work you have to modify two files and change "cru" to "cr" :
```bash
/usr/share/aclocal/libtool.m4
```
and
```bash
/usr/share/libtool/configure
```
then rerun `./autogen.sh and ./configure` 

You can additionally add to your command line  when executing make:

`make ARFLAGS="cr" AR_FLAGS="cr" `

Otherwise to have this be more clean, they would have to update the libtool repository to conform to AR's modifiers.

rgerhards:
@D3m0nKingx thanks for the info, but these are not rsyslog files. They are generated via autoconf/libtool. I suggest that you file an issue (or PR) with the libtool project. As far as rsyslog is concerned, we can unfortuantely not do anything against this.

D3m0nKingx:
@rgerhards I got a response from Libtool bug report I filed:

bug#30118
>Peter Rosin <peda@lysator.liu.se>
>On 2018-01-15 04:04, d3m0nkingx  . wrote:
>  Also I've looked into the libtool repository :  git://git.savannah.gnu.org/libtool.git <http://git.savannah.gnu.org/libtool.git>  and saw a NEWS file stating:
>
> * New features:
>
>   - Libtool script now supports (configure-time and runtime) ARFLAGS
>     variable, which obsoletes AR_FLAGS.  This is due to naming conventions
>     among other *FLAGS and to be consistent with Automake's ARFLAGS.
>
>   - Gnulib testsuite is enabled and run during 'make check'.
>
> ** Important incompatible changes:
>
>   - Libtool changed ARFLAGS/AR_FLAGS default from 'cru' to 'cr'.
>
>
> However, it's unclear how the configure-time and runtime ARFLAGS var should be set. Various options have been tried in the configure.ac <http://configure.ac> and Makefile.am files, but nothing worked. 
>
> Hope to hear from you soon on a viable option here. Thank you in advance for your attention to this matter. 

These NEWS entries (and corresponding changes) are not yet in any released version
of Libtool, they are only committed to the source repo. There seems to be little
progress on actually making a new Libtool release, so you're probably better off
not holding your breath...

Cheers,
Peter

rgerhards:
thanks, so it looks like we will continue to see the warnings for quite a while...

PR #1196, opened 2016-10-14
omelasticsearch high availability addressing of ElasticSearch cluster

### LINKED ISSUE #1171: omelasticsearch does not support HA addressing of elasticsearch cluster
ElasticSearch is highly available, where any node of the cluster is subject to failure at any given time. However, the ElasticSearch output module only accepts a single server url for the cluster. This requires ensuring dynamic resolution of the cluster servers, (for example, by using a reverse proxy). 

It would be nice to be able to specify the server url as a comma separated list, one of which is selected in round robin fashion when POSTing messages to ElasticSearch.

barakshechter:
Resolves #1171 
Changing server parameter to accept an array of servers/uris and port to
defaultPort if not specified on a given uri. Adding dedicated curl handle
for checkConn and refactoring the curl setup a bit, using ES health check
with timeout. Added checkConnFailure stat counting and HA test

### DIFF
--- a/plugins/omelasticsearch/omelasticsearch.c
+++ b/plugins/omelasticsearch/omelasticsearch.c
@@ -246,7 +246,7 @@ CODESTARTdbgPrintInstInfo
 	dbgprintf("omelasticsearch\n");
 	dbgprintf("\ttemplate='%s'\n", pData->tplName);
 	dbgprintf("\tnumServers=%d\n", pData->numServers);
-	dbgprintf("\thealthCheckTimeout=%lun", pData->healthCheckTimeout);
+	dbgprintf("\thealthCheckTimeout=%lu\n", pData->healthCheckTimeout);
 	dbgprintf("\tserverBaseUrls=");
 	for(i = 0 ; i < pData->numServers ; ++i)
 		dbgprintf("%c'%s'", i == 0 ? '[' : ' ', pData->serverBaseUrls[i]);
--- a/plugins/omelasticsearch/omelasticsearch.c
+++ b/plugins/omelasticsearch/omelasticsearch.c
@@ -308,8 +308,7 @@ computeBaseUrl(char* serverParam, int defaultPort, sbool useHttps, uchar **baseU
 	else if (strcasestr(serverParam, SCHEME_HTTPS))
 		host = serverParam + strlen(SCHEME_HTTPS);
 	else
-		r = useHttps ?
-			es_addBuf(&urlBuf, SCHEME_HTTPS, sizeof(SCHEME_HTTPS)-1) : 
+		r = useHttps ? es_addBuf(&urlBuf, SCHEME_HTTPS, sizeof(SCHEME_HTTPS)-1) : 
 			es_addBuf(&urlBuf, SCHEME_HTTP, sizeof(SCHEME_HTTP)-1);
 
 	if (r == 0) r = es_addBuf(&urlBuf, serverParam, strlen(serverParam));
--- a/plugins/omelasticsearch/omelasticsearch.c
+++ b/plugins/omelasticsearch/omelasticsearch.c
@@ -69,6 +69,7 @@ statsobj_t *indexStats;
 STATSCOUNTER_DEF(indexSubmit, mutIndexSubmit)
 STATSCOUNTER_DEF(indexHTTPFail, mutIndexHTTPFail)
 STATSCOUNTER_DEF(indexHTTPReqFail, mutIndexHTTPReqFail)
+STATSCOUNTER_DEF(checkConnFail, mutCheckConnFail)
 STATSCOUNTER_DEF(indexESFail, mutIndexESFail)
 
 
@@ -83,12 +84,15 @@ STATSCOUNTER_DEF(indexESFail, mutIndexESFail)
  */
 typedef struct curl_slist HEADER;
 typedef struct _instanceData {
-	int port;
+	int defaultPort;
 	int fdErrFile;		/* error file fd or -1 if not open */
 	pthread_mutex_t mutErrFile;
-	uchar *server;
+	uchar **serverBaseUrls;
+	int numServers;
+	long healthCheckTimeout;
 	uchar *uid;
 	uchar *pwd;
+	uchar *authBuf;
 	uchar *searchIndex;
 	uchar *searchType;
 	uchar *parent;
@@ -105,16 +109,18 @@ typedef struct _instanceData {
 	sbool bulkmode;
 	size_t maxbytes;
 	sbool asyncRepl;
-        sbool useHttps;
-        sbool allowUnsignedCerts;
+	sbool useHttps;
+	sbool allowUnsignedCerts;
 } instanceData;
 
 typedef struct wrkrInstanceData {
 	instanceData *pData;
+	int serverIndex;
 	int replyLen;
 	char *reply;
-	CURL	*curlHandle;	/* libcurl session handle */
-	HEADER	*postHeader;	/* json POST request info */
+	CURL	*curlCheckConnHandle;	/* libcurl session handle for checking the server connection */
+	CURL	*curlPostHandle;	/* libcurl session handle for posting data to the server */
+	HEADER	*curlHeader;	/* json POST request info */
 	uchar *restURL;		/* last used URL for error reporting */
 	struct {
 		es_str_t *data;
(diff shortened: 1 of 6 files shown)


plugins/omelasticsearch/omelasticsearch.c
```diff
@@ -127,8 +133,9 @@ typedef struct wrkrInstanceData {
 /* tables for interfacing with the v6 config system */
 /* action (instance) parameters */
 static struct cnfparamdescr actpdescr[] = {
-	{ "server", eCmdHdlrGetWord, 0 },
+	{ "server", eCmdHdlrArray, 0 },
```
barakshechter, line 136: I did not add a test where server is specified as just a string, not an array, since presumably this is handled by the grammar itself.

plugins/omelasticsearch/omelasticsearch.c
```diff
@@ -247,76 +276,119 @@ ENDdbgPrintInstInfo
...
+{
+	pWrkrData->serverIndex = (pWrkrData->serverIndex + 1) % pWrkrData->pData->numServers;
+}
 
 static rsRetVal
 checkConn(wrkrInstanceData_t *pWrkrData)
 {
-	es_str_t *url;
-	CURL *curl = NULL;
+#	define HEALTH_URI "_cat/health"
```
barakshechter, line 342: Testing _cat/health instead of just /, since if the server is up, but there is no master, '/' will respond successfully, but _cat/health will timeout indicating the cluster definitely cannot accept any messgaes. 

Not parsing the result of _cat/health, since if it can respond, we can defer to the cluster to decide whether or not it can accept messages

plugins/omelasticsearch/omelasticsearch.c
```diff
@@ -1206,50 +1274,93 @@ curlResult(void *ptr, size_t size, size_t nmemb, void *userdata)
...
 }
 
 static void
 setInstParamDefaults(instanceData *pData)
 {
-	pData->server = NULL;
-	pData->port = 9200;
+	pData->serverBaseUrls = NULL;
+	pData->defaultPort = 9200;
+	pData->healthCheckTimeout = 3500;
```
barakshechter, line 1359: curl's default is infinite and elastic's default is 30 seconds, both of which seemed way too long if we're trying to process messages quickly and fail fast. 3.5 seconds seems like a reasonable default, that can be extended if users need.

plugins/omelasticsearch/omelasticsearch.c
```diff
@@ -1084,14 +1151,14 @@ static rsRetVal
...
-	CURL *curl = pWrkrData->curlHandle;
+	CURL *curl = pWrkrData->curlPostHandle;
 	DEFiRet;
 
 	pWrkrData->reply = NULL;
 	pWrkrData->replyLen = 0;
 
-	if(pWrkrData->pData->dynSrchIdx || pWrkrData->pData->dynSrchType || pWrkrData->pData->dynParent)
-		CHKiRet(setCurlURL(pWrkrData, pWrkrData->pData, tpls));
+	CHKiRet(checkConn(pWrkrData));
```
barakshechter, line 1159: I went back and forth on whether this should be done only for bulk mode, or all the time, and finally settled on all the time since in either case, the ratio for the number of requests made is still ~2x.

By opting out of bulk mode, the user is already accepting significant overhead in the number of requests. Perhaps the default should be bulkmode=on?

barakshechter: Also, I opted for a very naive approach to health check, where it doesn't store status from one request to the next. 

As an enhancement, we could implement periodic health check (maintain state) and backoff (delay checking after failure), but I didn't think it was necessary now

plugins/omelasticsearch/omelasticsearch.c
rgerhards: yes, indeed it is. Also, if I understood correctly, the currently-existing tests should handle this, right?

plugins/omelasticsearch/omelasticsearch.c
```diff
@@ -247,76 +276,119 @@ ENDdbgPrintInstInfo
...
+	/* Find where the hostname/ip of the server starts. If the scheme is not specified
+	  in the uri, start the buffer with a scheme corresponding to the useHttps parameter.
+	*/
+	if (strcasestr(serverParam, SCHEME_HTTP))
+		host = serverParam + strlen(SCHEME_HTTP);
+	else if (strcasestr(serverParam, SCHEME_HTTPS))
+		host = serverParam + strlen(SCHEME_HTTPS);
+	else
+		r = useHttps ?
+			es_addBuf(&urlBuf, SCHEME_HTTPS, sizeof(SCHEME_HTTPS)-1) :
```
rgerhards: Is this indention intitional? If so, why? If now, can we correct it?

plugins/omelasticsearch/omelasticsearch.c
rgerhards: sounds like a very good solution

plugins/omelasticsearch/omelasticsearch.c
rgerhards: The rsyslog core does the automatic recheck for you. Also, it increases the wait time between tries.

plugins/omelasticsearch/omelasticsearch.c
rgerhards: sounds good to me

rgerhards:
Keeping the PR open till Tuesday to give other a chance to comment (as I am far from being an ES expert). @radu-gheorghe All looks good to me, what do you say?

plugins/omelasticsearch/omelasticsearch.c
barakshechter: The existing tests rely on the default. As far as the module is concerned, the two code paths are 1- array parameter or 2- default. 
Since the eCmdHdlrArray guarantees an array, I didn't create a test with the server specified as just a string, but i can add one in if you think it's worth it.

rgerhards:
2016-10-15 17:23 GMT+02:00 barakshechter <notifications@github.com>:

> *@barakshechter* commented on this pull request.
>
> In plugins/omelasticsearch/omelasticsearch.c
> <https://github.com/rsyslog/rsyslog/pull/1196>:
>
> > @@ -127,8 +133,9 @@ typedef struct wrkrInstanceData {
>  /* tables for interfacing with the v6 config system */
>  /* action (instance) parameters */
>  static struct cnfparamdescr actpdescr[] = {
> -	{ "server", eCmdHdlrGetWord, 0 },
> +	{ "server", eCmdHdlrArray, 0 },
>
> The existing tests rely on the default. As far as the module is concerned,
> the two code paths are 1- array parameter or 2- default.
> Since the eCmdHdlrArray guarantees an array, I didn't create a test with
> the server specified as just a string, but i can add one in if you think
> it's worth it.
>
>
It's not strictly needed, but I somehow would feel better if we had it ;-)

Rainer

plugins/omelasticsearch/omelasticsearch.c
barakshechter: The individual servers won't have backoff, only in the state that all servers are inaccessible will the core functionality happen. I have it calling checkConn before every POST to ES, which cycles through all servers until it finds one that works. Given how ES works, the worst case scenario (with partial cluster availability) is that we spend `(N-1)/2 * healthCheckTimeout` every time we submit to ES.

barakshechter: FWIW, I think it is probably fine and still an improvement on the current functionality, but can lead to degraded performance, especially if not using bulk mode (hence the recommendation for making it the default).

plugins/omelasticsearch/omelasticsearch.c
barakshechter: Added

plugins/omelasticsearch/omelasticsearch.c
barakshechter: Personal style? I changed the alignment, but if you prefer a different style- I am indifferent.

plugins/omelasticsearch/omelasticsearch.c
```diff
@@ -309,8 +309,8 @@ computeBaseUrl(char* serverParam, int defaultPort, sbool useHttps, uchar **baseU
 		host = serverParam + strlen(SCHEME_HTTPS);
 	else
 		r = useHttps ?
-			es_addBuf(&urlBuf, SCHEME_HTTPS, sizeof(SCHEME_HTTPS)-1) : 
-			es_addBuf(&urlBuf, SCHEME_HTTP, sizeof(SCHEME_HTTP)-1);
+		    es_addBuf(&urlBuf, SCHEME_HTTPS, sizeof(SCHEME_HTTPS)-1) :
```
rgerhards: The rule, unfortunately I think not yet written anywhere, is that if we indent, we ident by one TAB. So IMHO this should remain as it was.

barakshechter: no problem, chnaged to using a single tab.

Issue #1128, opened 2016-08-26
Update required dependencies to build on Ubuntu 14.04


mostolog:
Hi.

[README.md](https://github.com/rsyslog/rsyslog/blob/master/README.md) references a bunch of dependencies needed to build rsyslog, but they seem quite outdated.

I just compiled rsyslog **with omhiredis support** on **Ubuntu 14.04**, and needed to do:

> apt-get update && apt-get install -y software-properties-common

> add-apt-repository -y ppa:adiscon/v8-stable

> apt-get update && apt-get install -y build-essential pkg-config libestr-dev libfastjson-dev zlib1g-dev uuid-dev libgcrypt20-dev liblogging-stdlog-dev libhiredis-dev

Hope it helps others.

rgerhards:
closed via #1203

PR #1203, opened 2016-10-18
Update README.md

mostolog:
https://github.com/rsyslog/rsyslog/issues/1128

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,10 +59,20 @@ sudo yum install git valgrind autoconf automake flex bison python-docutils pytho
 
 #### Ubuntu
 
-Note: this list is (obviously) incomplete. We intend to complete it as we 
-have more information and install additional systems.
+Add Adiscon repository:
 ```
-sudo apt-get install libdbi-dev libmysqlclient-dev postgresql-client libpq-dev  libnet-dev   librdkafka-dev   libgrok-dev libgrok1 libgrok-dev libpcre3-dev libtokyocabinet-dev libglib2.0-dev  libmongo-client-dev  libhiredis-dev
+apt-get update && apt-get install -y software-properties-common
+add-apt-repository -y ppa:adiscon/v8-stable
+```
+
+Needed packages to build with omhiredis support:
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,10 +59,20 @@ sudo yum install git valgrind autoconf automake flex bison python-docutils pytho
 
 #### Ubuntu
 
-Note: this list is (obviously) incomplete. We intend to complete it as we 
-have more information and install additional systems.
+Add Adiscon repository:
```
rgerhards, line 62: It would be great if we could also add that the repro ist not necessarily needed. You can also build everything from scratch (this is useful to do for developers who use master branch, as it may require things newer than even in the v8-devel repo!).

mostolog: You mean downloading from git/zip all -dev packages ? I guess a bunch of repositories are needed, I think it will be better to reference then at the beginning of doc.

You self-assigned, so I won't do anything until you say so.

rgerhards:
2016-10-18 16:11 GMT+02:00 mostolog <notifications@github.com>:

> *@mostolog* commented on this pull request.
>
> In README.md <https://github.com/rsyslog/rsyslog/pull/1203>:
>
> > @@ -59,10 +59,20 @@ sudo yum install git valgrind autoconf automake flex bison python-docutils pytho
>
>  #### Ubuntu
>
> -Note: this list is (obviously) incomplete. We intend to complete it as we
> -have more information and install additional systems.
> +Add Adiscon repository:
>
> You mean downloading from git/zip and installing these packages? I guess a
> bunch of repositories are needed, I think it will be better to reference
> then at the beginning of doc.
>

more often via cloning from git[hub]. The projects are also in my buildbot
setup scripts:

https://github.com/rgerhards/buildbotsetup

It's unfortunately not yet scripted for  all platforms, but getting better
and probably still has good info.

You self-assigned, so I won't do anything until you say so.
>
For a PR, (self-)assigned means that  I handle the merge process, questions
etc. Doesn't mean I'll actually change the code (except for minor nits). I
think this is common workflow for PR assignments. Just FYI, appreciate all
your help.

Rainer

PR #1211, opened 2016-10-20
Escape \ for JSON_ESCAPE

### LINKED ISSUE #1126: Backslash (\) character is not properly escaped
It seems that the "\" character is not properly escaped when option.json="on" is used.

version used: rsyslogd 8.18.0

Configuration file:
```
template(name="template_test" type="list" option.json="on") {
        constant(value="{")
        constant(value="\"@timestamp\":\"")       property(name="timereported" dateFormat="rfc3339")
        constant(value="\",\"@src_event\":\"")    property(name="hostname")
        constant(value="\",\"@severity\":\"")     property(name="syslogseverity-text")
        constant(value="\",\"@facility\":\"")     property(name="syslogfacility-text")
        constant(value="\",\"@program\":\"")      property(name="programname")
        constant(value="\",\"message\":\"")       property(name="msg")
        constant(value="\"}")
}
```

**Message received:**
`<164>Aug 22 2016 09:30:52: %ASA: Deny udp src Internet:1.2.3.4/100(LOCAL\user) dst WAN:5.5.5.5/100 by access-group "access-group"
`

**After handling:**
```
{"@timestamp":"2016-08-22T09:30:52+02:00",
"@src_event":"src-asa.domain.lan",
"@severity":"warning",
"@facility":"local4",
"@program":"%ASA",
"message":" Deny udp src Internet:1.2.3.4/100(LOCAL\user) dst WAN:5.5.5.5/100 by access-group \"access-group\"
"}
```

The above message generates a "_jsonparsefailure" raised by logstash.
Let me know if there is a way to escape the "\" by increasing the template.

Thank you

mostolog:
I think this could be enough to fix #1126 @rgerhards 
Doing this from the phone, so don't get surprised if tests fail...

### DIFF
--- a/template.c
+++ b/template.c
@@ -467,7 +467,7 @@ doEscape(uchar **pp, rs_size_t *pLen, unsigned short *pbMustBeFreed, int mode)
 		for(p = *pp ; *p && *p != '\'' && *p != '\\' ; ++p)
 			;
 	else if(mode == JSON_ESCAPE)
-		for(p = *pp ; *p && *p != '"' ; ++p)
+		for(p = *pp ; *p && (*p != '"' || *p != '\' ) ; ++p)
 			;
 	/* when we get out of the loop, we are either at the
 	 * string terminator or the first character to escape */
@@ -485,7 +485,7 @@ doEscape(uchar **pp, rs_size_t *pLen, unsigned short *pbMustBeFreed, int mode)
 		} else if((mode == SQL_ESCAPE) && *p == '\\') {
 			CHKiRet(cstrAppendChar(pStrB, '\\'));
 			iLen++;	/* reflect the extra character */
-		} else if((mode == JSON_ESCAPE) && *p == '"') {
+		} else if((mode == JSON_ESCAPE) && (*p != '"' || *p != '\' )) {
 			CHKiRet(cstrAppendChar(pStrB, '\\'));
 			iLen++;	/* reflect the extra character */
 		}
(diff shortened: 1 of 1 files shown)


template.c
```diff
@@ -485,7 +485,7 @@ doEscape(uchar **pp, rs_size_t *pLen, unsigned short *pbMustBeFreed, int mode)
 		} else if((mode == SQL_ESCAPE) && *p == '\\') {
 			CHKiRet(cstrAppendChar(pStrB, '\\'));
 			iLen++;	/* reflect the extra character */
-		} else if((mode == JSON_ESCAPE) && (*p != '"' || *p != '\' )) {
+		} else if((mode == JSON_ESCAPE) && (*p != '"' || *p != '\\' )) {
```
rgerhards: IMHO this predicte is wrong and will lead to all chars being escaped. The testbench doesn't catch that, because it does not yet have a test for this functionality (will add one).

Think about it: when *p is '"', it is not '\', so the OR part will always be true. Same for *p not in { '"', '\'}.

rgerhards: looks like I picked the wrong commit, this is also true for the later one. Sorry for that. 

BTW: you should squash those commits to make the PR easier to follow. If you don't know how, you can also mail me privately and I'll help :-)

mostolog:
I copy-pasted and made the opposite as intended on that else...I don't know how to stash on github(neither using phone) 😝

rgerhards:
No problem, I can squash the branch if it's not locked against maintainer commits. I don't know how to squash from the phone (or github web interface) either.

rgerhards:
ok, can't do it. We cannot put all those tries into git history. Among others, it breaks git biset. So what I can do is squash it in my own branch, but that means your commits will be lost. The best thing would be if you squash it from your computer.

What shall I do?

PR #1212, opened 2016-10-20
Use unsigned to report rcvbuf sizes

### LINKED ISSUE #1146: Use unsigned int when reporting actual rcvbuf sizes in debug mode
rsyslog 8.21.0

Problem:

```
(ubuntu) ~ $ sudo rsyslogd  -n
rsyslogd: cannot set os socket rcvbuf size 1215752192 for socket 6, value now is -931731456: No such file or directory [v8.21.0]
rsyslogd: cannot set os socket rcvbuf size 1215752192 for socket 8, value now is -931731456 [v8.21.0]
```

Expected output: 
-931731456 treated as unsigned int should render 3363235840 (~3.3 GB) if I did my math good.

Steps to reproduce:

Set uberhigh rcvbuf size in UDP input module that won't be permitted by kernel.
You must do it on a system that will permit you setting it to >2GB so hat `int` will overflow.  I did it on Ubuntu 12.04 with kernel 3.13 and 7.2 GB of RAM (used 1100MB).

`input(type="imudp" port="514" rcvbufSize="10G")`

Run rsyslog in debug mode and notice overflown value.

The bug is somewhere here https://github.com/rsyslog/rsyslog/blob/master/runtime/net.c#L1343

mostolog:
Just a try to fix #1146 Probably many other int's should be unsigned

### DIFF
--- a/runtime/net.c
+++ b/runtime/net.c
@@ -1216,7 +1216,7 @@ closeUDPListenSockets(int *pSockArr)
  * param rcvbuf indicates desired rcvbuf size; 0 means OS default
  */
 static int *
-create_udp_socket(uchar *hostname, uchar *pszPort, int bIsServer, int rcvbuf, int ipfreebind)
+create_udp_socket(uchar *hostname, uchar *pszPort, int bIsServer, unsigned int rcvbuf, int ipfreebind)
 {
         struct addrinfo hints, *res, *r;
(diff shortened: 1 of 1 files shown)


mostolog:
Using casting doesn't seem very polite...although setting rcvbuf to unsigned will probably affect several files. I'll give a try tomorrow

PR #1215, opened 2016-10-21
Honor parseHostnameAndTag

### LINKED ISSUE #1190: global(parser.parseHostnameAndTag="off") not honored
Hi,

The problem happens with rsyslog 8.15, can't test it at the moment with latest release. Pardon me if it was already fixed. This is on Solaris, that's why I'm using the gegrep command.

/usr/lib/rsyslog/rsyslogd -d -n 2>&1 |
gegrep '(rfc3164 parser init called|Timezone information table)'
7672.075198900:main thread    : rfc3164 parser init called
7672.110824673:main thread    : Timezone information table (0 entries):

From the log it can be seen tht pmrfc3164 is initialized before global configuration is read (the "Timezone ..." message happens just before reading global configuration parameters). The module caches the value and never looks again whether it was already defined or not.

As a workaround I'm using this:

```
--- rsyslog-8.15.0/tools/pmrfc3164.c    2016-10-13 11:19:58.608653777 +0200
+++ rsyslog-8.15.0/tools/pmrfc3164.c    2016-10-13 11:19:49.091827128 +0200
@@ -208,7 +208,7 @@ CODESTARTparse
         * machine that we received the message from and the tag will be empty. This
         * is meant to be an interim solution, but for now it is in the code.
         */
-       if(bParseHOSTNAMEandTAG && !(pMsg->msgFlags & INTERNAL_MSG)) {
+       if(glbl.GetParseHOSTNAMEandTAG() && !(pMsg->msgFlags & INTERNAL_MSG)) {
                /* parse HOSTNAME - but only if this is network-received!
                 * rger, 2005-11-14: we still have a problem with BSD messages. These messages
                 * do NOT include a host name. In most cases, this leads to the TAG to be treated
```

Do you see the same issue in your setup?

Many thanks
__ 
  Vlad

mostolog:
try to fix #1190

### DIFF
--- a/tools/pmrfc3164.c
+++ b/tools/pmrfc3164.c
@@ -101,6 +101,7 @@ createInstance(instanceConf_t **pinst)
 	inst->bDetectYearAfterTimestamp = 0;
 	inst->bPermitSquareBracketsInHostname = 0;
 	inst->bPermitSlashesInHostname = 0;
+	bParseHOSTNAMEandTAG=glbl.GetParseHOSTNAMEandTAG();
 	*pinst = inst;
 finalize_it:
 	RETiRet;


mostolog:
I don't know if I did it well this time...

rgerhards:
The commit looks fine, and it is more the the philosophy used in other modules than the proposed fix in #1190. However, while reviewing I notice that the overall method is not well, because we use a single global for theoretically multiple config objects (which we today cannot have in practice). So I wonder if it would be smarter to change the overall approach. One way would actually to move towards the solution from #1190, but refactor part of the global object so that we can do this with higher performance.

PR #1217, opened 2016-10-21
Use unsigned to report rcvbuf sizes

### LINKED ISSUE #1146: Use unsigned int when reporting actual rcvbuf sizes in debug mode

mostolog:
Just a try to fix #1146 Probably many other int's should be unsigned

### DIFF
--- a/plugins/imudp/imudp.c
+++ b/plugins/imudp/imudp.c
@@ -115,7 +115,7 @@ struct instanceConf_s {
 	uchar *dfltTZ;
 	int ratelimitInterval;
 	int ratelimitBurst;
-	int rcvbuf;			/* 0 means: do not set, keep OS default */
+	unsigned int rcvbuf;			/* 0 means: do not set, keep OS default */
 	/*  0 means:  IP_FREEBIND is disabled
 	1 means:  IP_FREEBIND enabled + warning disabled
 	1+ means: IP+FREEBIND enabled + warning enabled */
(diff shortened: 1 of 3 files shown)


rgerhards:
Unfortunately, we cannot change the data type, as it stems back to the API. Sorry for that.

mostolog:
That's great anyway. IIUC limiting to 2GB seems a valid option. going back to original issue.

rgerhards:
actually I'd say 1GiB is also sufficient. Will do that and wait for
complaints ;)

2016-12-02 12:37 GMT+01:00 mostolog <notifications@github.com>:

PR #1218, opened 2016-10-21
Allow the key parameter of Omhiredis to point to a template

bobthemighty:
This PR introduces a new `dynakey` attribute.

If set to `on`, this causes omhiredis to treat the `key` attribute as the name of a template so that you can build a dynamic redis queue name or list key.

Absolutely _no_ warranty of quality or functionality is made here. It took me a while to get the thing to stop throwing Segfaults, so we're a way off production ready.

```
template(name="redis-key" type="list"){
    constant(value="logs-from-")
    property(name="syslogtag")
}


action(
        action.resumeretrycount="-1"
        action.resumeinterval="5"
        queue.maxDiskSpace="1G"
            queue.spoolDirectory="/var/spool/rsyslog"
            queue.filename="plain-docker" 
            queue.size="1000000" 
            queue.type="LinkedList"
            type="omhiredis"
            mode="publish"
        dynakey="on"
            key="redis-key"
            server="localhost"
            template="plain-syslog"
        )
```

```
bob@made-lt051:~|⇒  logger  -p local3.info "hello" -t "world"
bob@made-lt051:~|⇒  logger  -p local3.info "hello" -t "you"  
bob@made-lt051:~|⇒  logger  -p local3.info "hello" -t "badgers"
```

```
1477089483.314670 [0 172.17.0.1:54048] "PUBLISH" "logs-from-world:" "{\"@timestamp\":\"2016-10-21T23:38:03.312192+01:00\",\"host\":\"localhost\",\"severity\":\"info\",\"facility\":\"local3\",\"tag\":\"world:\",\"container\":\"world:\",\"@message\":\" hello\"}"
1477089486.936369 [0 172.17.0.1:54048] "PUBLISH" "logs-from-you:" "{\"@timestamp\":\"2016-10-21T23:38:06.935438+01:00\",\"host\":\"localhost\",\"severity\":\"info\",\"facility\":\"local3\",\"tag\":\"you:\",\"container\":\"you:\",\"@message\":\" hello\"}"
1477089490.932602 [0 172.17.0.1:54048] "PUBLISH" "logs-from-badgers:" "{\"@timestamp\":\"2016-10-21T23:38:10.931619+01:00\",\"host\":\"localhost\",\"severity\":\"info\",\"facility\":\"local3\",\"tag\":\"badgers:\",\"container\":\"badgers:\",\"@message\":\" hello\"}"
```

### DIFF
--- a/contrib/omhiredis/omhiredis.c
+++ b/contrib/omhiredis/omhiredis.c
@@ -40,6 +40,7 @@
 #include "module-template.h"
 #include "errmsg.h"
 #include "cfsysline.h"
+#include "unicode-helper.h"
 
 MODULE_TYPE_OUTPUT
 MODULE_TYPE_NOKEEP
@@ -53,9 +54,9 @@ DEFobjCurrIf(errmsg)
 #define OMHIREDIS_MODE_QUEUE 1
 #define OMHIREDIS_MODE_PUBLISH 2
 
-/*  our instance data.
- *  this will be accessable 
- *  via pData */
+/* our instance data.
+ * this will be accessable 
+ * via pData */
 typedef struct _instanceData {
 	uchar *server; /* redis server address */
 	int port; /* redis port */
@@ -63,7 +64,8 @@ typedef struct _instanceData {
 	uchar *tplName; /* template name */
 	char *modeDescription; /* mode description */
 	int mode; /* mode constant */
-	char *key; /* key for QUEUE and PUBLISH modes */
+	uchar *key; /* key for QUEUE and PUBLISH modes */
+	sbool dynaKey; /* Should we treat the key as a template? */
 } instanceData;
 
 typedef struct wrkrInstanceData {
@@ -79,6 +81,7 @@ static struct cnfparamdescr actpdescr[] = {
 	{ "template", eCmdHdlrGetWord, 0 },
 	{ "mode", eCmdHdlrGetWord, 0 },
 	{ "key", eCmdHdlrGetWord, 0 },
+	{ "dynakey", eCmdHdlrBinary, 0 }
 };
 
 static struct cnfparamblk actpblk = {
@@ -102,7 +105,7 @@ CODESTARTisCompatibleWithFeature
 		iRet = RS_RET_OK;
 ENDisCompatibleWithFeature
 
-/*  called when closing */
+/* called when closing */
 static void closeHiredis(wrkrInstanceData_t *pWrkrData)
 {
 	if(pWrkrData->conn != NULL) {
@@ -111,7 +114,7 @@ static void closeHiredis(wrkrInstanceData_t *pWrkrData)
 	}
 }
 
-/*  Free our instance data. */
+/* Free our instance data. */
 BEGINfreeInstance
 CODESTARTfreeInstance
 	if (pData->server != NULL) {
@@ -129,7 +132,7 @@ CODESTARTdbgPrintInstInfo
 	/* nothing special here */
 ENDdbgPrintInstInfo
 
-/*  establish our connection to redis */
+/* establish our connection to redis */
 static rsRetVal initHiredis(wrkrInstanceData_t *pWrkrData, int bSilent)
 {
 	char *server;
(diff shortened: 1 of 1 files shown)


rgerhards:
Note that some testbench failures are not related to your PR. I don't know why the happen only occasionally and are not always detected. Working on this on PR #1220 .

rgerhards:
@bobthemighty if you rebase to master now, the testbench failure should go away. Also, let me know when you consider this ready for merging.

taotetek:
I really like this - I'm going to adopt this technique for dynamic topics in the omczmq output.

rgerhards:
@bobthemighty just FYI: I will be doing final merges next week for the 8.23 release. So if you consider adding it to that release, it would be great if it is ready by mid-week.

bobthemighty:
Awesome - thanks Rainer,

I'll take a look at it this evening. When I thought again I remembered that
cases like "missing template" are already handled in the runtime, so I
probably just need to fix some indentation issues and rebase to squash.

-- Bob.

bobthemighty:
@rgerhards rebased to master, squashed, and cleaned up the inconsistent indentation.

I still haven't done any in-depth testing, but it sounded like @mostolog was going to put it through its paces. If he hasn't seen any issues, then I think I'm reasonably happy.

mostolog:
Thanks @bobthemighty 
What's the supposed configuration? 
setting dynakey="on" will use TAG as key?

Did you had time to have a look at https://github.com/rsyslog/rsyslog-doc/issues/271 ?
Perhaps it's a better approach to fix this.
Anyway, I'll test and let you know.

bobthemighty:
Set dynakey "on" and the value of "key" is expected to be a template name.
See the example.

My problem with the raw template mode is that I think it is much less
user-friendly, but let me know your thoughts.

mostolog:
> Set dynakey "on" and the value of "key" is expected to be a template name.

Ok

> My problem with the raw template mode is that I think it is much less user-friendly, but let me know your thoughts.

I could agree with you, but "_fixing_" raw template mode to be able to LPUSH an element containing spaces would've been enough as a workaround for this. Eg:

> Having msg = "this is a message"
> 
> LPUSH "%syslogtag% %msg%" should insert FOUR elements: "this", "is", "a", "message"
> 
> LPUSH \"%syslogtag% %msg%\" should insert ONE element: "this is a message" (notice quotes)

- Do you agree?
- Do you think that's a fix worths doing it?
- Will you be able to do it?
- Considering it's already done: will this patch still make sense?

rgerhards:
@bobthemighty I'll merge now, but it would be great if you could also add some doc to the rsyslog-doc project (so that the rest of the world knows about that feature and can actually use it ;-)).

bobthemighty:
I'll write some docs this morning, thanks rgerhards

mostolog:
I'll retry quoting myself:

    I could agree with you, but "fixing" raw template mode to be able to LPUSH an element containing spaces would've been enough as a workaround for this. Eg:

    Having msg = "this is a message"

    LPUSH "%syslogtag% %msg%" should insert FOUR elements: "this", "is", "a", "message"

    LPUSH \"%syslogtag% %msg%\" should insert ONE element: "this is a message" (notice quotes)

    Do you agree?
    Do you think that's a fix worths doing it?
    Will you be able to do it?
    Considering it's already done: will this patch still make sense?

rgerhards:
I may totally misunderstand this, but  isn't the introduction of a backslash not simply done in the template when the user defines it? Or are you talking about a rsyslog-doc patch? If so, I'd prefer to have the info added as it clarifies in general how things are supposed to work (at least I guess so ;-)).

mostolog:
I must be explaining myself quite wrong lately...
> redis-cli: LPUSH mykey this is a message
> omhiredis LPUSH mykey this is a message

do the same, inserting FOUR elements: "this", "is", "a", "message", while:

> redis-cli: LPUSH mykey "this is a message"
> omhiredis LPUSH mykey "this is a message"

doesn't do the same, as omhiredis is still adding 4 items, instead of 1 (as redis-cli does)

If you configure:
> template(name="redis" type="string" string="LPUSH %syslogtag% \"this does not work\"")

you wont see a "this does not work" item being added.
Am I clear this time?

rgerhards:
I understood that redis needs backslash-quote vs. quote, so I thought you meant this type of template definition would be needed:

```template(name="redis" type="string" string="LPUSH %syslogtag% \\\"this does not work\\\"")```

**Comment by mostolog** [comment]
> template(name="redis" type="string" string="LPUSH %syslogtag% \\\"this does not work\\\"")

Doesn't work. It's adding 4 items ("this", "does", "not", "work")

**Comment by rgerhards** [comment]
OK, then it's probably a question for @bobthemighty 
While the issue itself is obviously solved, I would still be interested to understand why it does not work.

**Comment by rgerhards** [comment]
Oh, I notice that you forgot one backslash each. It's three in a row! The first two to form a single backslash the thrid the escape for the quote char. you should even get an rsyslog error message with the sample you have posted.

**Comment by mostolog** [comment]
I'm using 3, but github showed 2 (spaced below)
> template(name="redis2" type="string" string="LPUSH %syslogtag% \ \ \ "this does not work \ \ \ "")

**Comment by mostolog** [comment]
@bobthemighty Before documenting this, and although this already has been merged, I still not sure if this change makes sense, considering _mode=template_ could be more than enough. Moreover, template mode allows to execute **any** REDIS command.

Instead of:
```
template(name="redis-key" type="list") {
    constant(constant="my-")
    property(name="syslogtag")
}
...
action(
        dynakey="on"
        key="redis-key"
        ...
)
```

It would be possible to use:
```
template(name="redis-template" type="string" string="LPUSH my-%syslogtag% \"this should add just one item but actually doesn't\"")
...
action(
        mode="template"
        template="redis-template"
        ...
)
```
What do you think?

bobthemighty:
Well that was my original question, but I agree with rgerhards that an
explicit dynakey mode is more user friendly than having to hand-craft Redis
commands.

I'm happy to take a look at template mode at some point, but it's the kind
of thing I do when I have a spare 15 minutes on the train :)

-- B

rgerhards:
Out of curiosity, I just had a look at the code. Whatever that means, in ```QUEUE``` mode the exact same command seems to be generated to redis. Just FYI.

rgerhards:
This is the exact location:h ttps://github.com/rsyslog/rsyslog/blob/master/contrib/omhiredis/omhiredis.c#L188
Maybe this helps the redis-literate.

rgerhards:
and looking at it, I wonder if

```rc = redisAppendCommand(pWrkrData->conn, (char*)message);```

should better be given as (what happens if ```%something``` is in message?):

```rc = redisAppendCommand(pWrkrData->conn, "%s", (char*)message);```

Just judging from printf-wisdom ;-)

mostolog:
LOL. Of course! Otherwise I'll be screwed.

- Queue mode works as designed, but won't allow a dynamic key
- On the other hand, template mode doesn't accept msg to contain spaces, and quotes doesn't solve this issue (that's the bug I'm trying to explain since https://github.com/rsyslog/rsyslog-doc/issues/271#issue-184772818)
- Finally, dynakey it's a workaround to solve this, but it's adding an unneeded parameter (again, IMHO), cause this could be done if template mode is fixed.

@rgerhards If you are interested, we can have an IRC chat about Redis ;)

mostolog:
In the first batch of tests (~200 dynamic keys), everything worked as expected.
I plan to heavy-test in the coming days. I'll keep you updated.
I still think fixing template mode will be a better option/all-in-one wonder tool.

Regards

PR #1224, opened 2016-10-26
Fix lookup find table

DavidLancelin:
Fixes the use of several lookup tables

### DIFF
--- a/runtime/lookup.c
+++ b/runtime/lookup.c
@@ -625,7 +625,7 @@ lookupFindTable(uchar *name)
 {
 	lookup_ref_t *curr;
 
-	for(curr = loadConf->lu_tabs.root ; curr != NULL ; curr = curr->next) {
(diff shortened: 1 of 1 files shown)


runtime/lookup.c
```diff
@@ -625,7 +625,7 @@ lookupFindTable(uchar *name)
 {
 	lookup_ref_t *curr;
 
-	for(curr = loadConf->lu_tabs.root ; curr != NULL ; curr = curr->next) {
+	for(curr = loadConf->lu_tabs.last ; curr != NULL ; curr = curr->next) {
```
rgerhards, line 628: Isn't `last` actually pointing at the last node, so that you can find at most that one? Without looking at the full code, `root` sounds right to me, which is also backed by the testbench failures. It would be great if you could provide additional information, best in the commit, why this change is needed.

ghost: It solved a problem with the configuration file in an earlier version, but I just noticed the lookupNew function has changed. So you are right, this change is not needed.

PR #1236, opened 2016-11-01
Removed the use of the deprecated hash function SHA-224.

vii5ard:
The SHA-224 hash algorithm has been removed from the future releases, thus removed the usage of it.


rgerhards:
I think we need to give an descriptive error message if a user tries to use the Algo. Thinking especially about cases with currently existing configs that would now surprisingly fail. I can apply a change if you like.

vii5ard:
It feels unlikely that sha-224 was used anyway, but adding the descriptive message would be polite.

rgerhards:
I have added a message and a general vehicle to retire hashes which are considered insecure, see #1244. I will merge your changes via #1244, **please do not close this PR**.

PR #1247, opened 2016-11-05
AIX Port

purnimam1:
Hi Rainer 

As discussed this is the PR for master branch with AIX code changes.

### DIFF
--- a/action.c
+++ b/action.c
@@ -92,6 +92,9 @@
 #include <strings.h>
 #include <time.h>
 #include <errno.h>
+#ifdef _AIX
+#include <pthread.h>
+#endif 
 #include <json.h>
 
 #include "dirty.h"
@@ -111,15 +114,21 @@
 #include "parserif.h"
 #include "statsobj.h"
 
+/* AIXPORT : cs renamed to legacy_cs as clashes with libpthreads variable in complete file*/
+#ifdef _AIX
+#define cs legacy_cs
+#endif
+#if !defined(_AIX)
 #pragma GCC diagnostic ignored "-Wswitch-enum"
+#endif
 
 #define NO_TIME_PROVIDED 0 /* indicate we do not provide any cached time */
 
 /* forward definitions */
 static rsRetVal processBatchMain(void *pVoid, batch_t *pBatch, wti_t * const pWti);
-static rsRetVal doSubmitToActionQ(action_t * const pAction, wti_t * const pWti, msg_t*);
-static rsRetVal doSubmitToActionQComplex(action_t * const pAction, wti_t * const pWti, msg_t*);
-static rsRetVal doSubmitToActionQNotAllMark(action_t * const pAction, wti_t * const pWti, msg_t*);
+static rsRetVal doSubmitToActionQ(action_t * const pAction, wti_t * const pWti, smsg_t*);
+static rsRetVal doSubmitToActionQComplex(action_t * const pAction, wti_t * const pWti, smsg_t*);
+static rsRetVal doSubmitToActionQNotAllMark(action_t * const pAction, wti_t * const pWti, smsg_t*);
 
 /* object static data (once for all instances) */
 DEFobjCurrIf(obj)
@@ -164,6 +173,7 @@ typedef struct configSettings_s {
 	int iActionQueueDeqtWinFromHr;			/* hour begin of time frame when queue is to be dequeued */
 	int iActionQueueDeqtWinToHr;			/* hour begin of time frame when queue is to be dequeued */
 } configSettings_t;
+
 
 configSettings_t cs;					/* our current config settings */
 configSettings_t cs_save;				/* our saved (scope!) config settings */
@@ -947,7 +957,7 @@ static rsRetVal actionDbgPrint(action_t *pThis)
 static rsRetVal
 prepareDoActionParams(action_t * __restrict__ const pAction,
 		      wti_t * __restrict__ const pWti,
-		      msg_t *__restrict__ const pMsg,
+		      smsg_t *__restrict__ const pMsg,
 		      struct syslogTime *ttNow)
 {
 	int i;
@@ -996,8 +1006,10 @@ finalize_it:
 
 
 /* the #pragmas can go away when we have disable array-passing mode */
+#if !defined(_AIX)
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wcast-align"
+#endif
 void
 releaseDoActionParams(action_t *__restrict__ const pAction, wti_t *__restrict__ const pWti, int action_destruct)
 {
@@ -1049,7 +1061,9 @@ releaseDoActionParams(action_t *__restrict__ const pAction, wti_t *__restrict__ 
 
 	return;
 }
+#if !defined(_AIX)
 #pragma GCC diagnostic pop
+#endif
 
 
 /* This is used in resume processing. We only finally know that a resume
@@ -1362,7 +1376,7 @@ actionCommitAllDirect(wti_t *__restrict__ const pWti)
 static rsRetVal
 processMsgMain(action_t *__restrict__ const pAction,
 	wti_t *__restrict__ const pWti,
-	msg_t *__restrict__ const pMsg,
+	smsg_t *__restrict__ const pMsg,
 	struct syslogTime *ttNow)
 {
 	DEFiRet;
@@ -1517,7 +1531,7 @@ static rsRetVal setActionQueType(void __attribute__((unused)) *pVal, uchar *pszT
  * rgerhards, 2010-06-08
  */
 static rsRetVal
-doSubmitToActionQ(action_t * const pAction, wti_t * const pWti, msg_t *pMsg)
+doSubmitToActionQ(action_t * const pAction, wti_t * const pWti, smsg_t *pMsg)
 {
 	struct syslogTime ttNow; // TODO: think if we can buffer this in pWti
 	DEFiRet;
@@ -1566,7 +1580,7 @@ finalize_it:
  * be filtered out before calling us (what is done currently!).
  */
 rsRetVal
-actionWriteToAction(action_t * const pAction, msg_t *pMsg, wti_t * const pWti)
+actionWriteToAction(action_t * const pAction, smsg_t *pMsg, wti_t * const pWti)
 {
 	DEFiRet;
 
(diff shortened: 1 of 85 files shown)


tools/smtradfile.c
```diff
@@ -40,6 +40,9 @@
 #include "msg.h"
 #include "module-template.h"
 #include "unicode-helper.h"
```
rgerhards: Please do these types of defines once and for all in the relevant headers where the base data type is defined. For msg_t, this is ./runtime/typedefs.c - this is the file where it usually happens.

tools/smtradfile.c
```diff
@@ -40,6 +40,9 @@
...
+#ifdef _AIX
+#define msg_t msg_tt
+#endif
```
rgerhards, line 43: Please do these types of defines once and for all in the relevant headers where the base data type is defined. For msg_t, this is `./runtime/typedefs.c` - this is the file where it usually happens.

purnimam1:
Hi Rainer
 
Yes we analysed typedefs.h as well before adding the #defines like the ones mentioned below. However, no matter which sequence of #defines/variables we tried it ended replacing the AIX header msg_t as well or did not resolve the issue in anyway.
 
Ex:
# grep -rn msg_tt *typedefs.h:40:#define msg_t msg_tttypedefs.h:72:typedef struct msg msg_tt;# make  CC       librsyslog_la-rsyslog.lo"./msg.h", line 253.41: 1506-022 (S) "iLenRawMsg" is not a member of "struct lc_messages_table"."./msg.h", line 254.15: 1506-022 (S) "iLenRawMsg" is not a member of "struct lc_messages_table"."./msg.h", line 255.15: 1506-022 (S) "pszRawMsg" is not a member of "struct lc_messages_table".make: *** [librsyslog_la-rsyslog.lo] Error 1#
 
So as a solution we can either chose to retain the AIX port changes for msg_t clash or the community can chose to replace msg_t in rsyslog code with something else as a generic fix to help with compatibility.
 
Please let us know your opinion. Meanwhile if you have any other comments let me know I will start working on them as we discuss this issue.
 
ThanksPurnima.MIndia Software & Technology LabsIBM India Pvt Ltd.Ph: 080-41777468Mail to : purnima.m@in.ibm.com
 
 
----- Original message -----From: Rainer Gerhards <notifications@github.com>To: rsyslog/rsyslog <rsyslog@noreply.github.com>Cc: Purnima Madanamanoharan/India/IBM@IBMIN, Author <author@noreply.github.com>Subject: Re: [rsyslog/rsyslog] AIX Port (#1247)Date: Sat, Nov 5, 2016 4:34 PM 
@rgerhards commented on this pull request.
In tools/smtradfile.c:
> @@ -40,6 +40,9 @@ #include "msg.h" #include "module-template.h" #include "unicode-helper.h"+#ifdef _AIX+#define msg_t msg_tt+#endif 
sorry, my comment actually refers to this part of the code ;-)

rgerhards:
@purnimam1 Thanks for the clarificaton, this makes indeed sense. I personally tend to change the rsyslog type name, but let me think a bit about this. Some community feedback would also be not bad. @davidelang @janmejay @bobthemighty @mbiebl @Whissi @taotetek @flicker581 @portante @theinric  (and those I haven't mentioned directly ;)) do you like to chime in?

@purnimam1 in the mean time, I would concentrate on getting the testbench failures settled. I'll probably will not go back to code review until we have a version without CI failures (due to time constraints).

rgerhards:
given the fact that we do not got any other feedback, I would suggest that you, @purnimam1, please go ahead and consistently change ```msg_t``` to ```smsg_t``` in all sources. Would that work for AIX?

rgerhards:
I also see that a rebase is required to ensure CI runs. Please do.

action.c
```diff
@@ -111,7 +114,14 @@
 #include "parserif.h"
 #include "statsobj.h"
 
+/* AIXPORT : cs renamed to cs as clashes with libpthreads variable in complete file*/
```
Whissi: cs renamed to **cs_t** as clashes with libpthreads variable...

Whissi:
Sorry for the late feedback: No problem with renaming `msg_t` in general. But could we please add a comment why we renamed for future reference (i.e. "renamed because it clashes with `msg_t` from...")?

action.c
rgerhards: as a side-note, `cs_t` is a bit misleading, because `cs` simply is a variable (but not a type). This would probably better be renamed to `legacy_cs` as it is only used for the legacy system.

configure.ac
```diff
@@ -1818,7 +1861,11 @@ AM_CONDITIONAL(ENABLE_DISTCHECK_WORKAROUND, test x$enable_distcheck_workaround =
 if test "x$in_git_src" = "xyes"; then
         AC_MSG_NOTICE([Running from git source])
 
+if test "$unamestr" = "AIX"; then
```
Whissi: This looks wrong. You basically tell `configure` to never ever re-generate man pages on AIX. Even if make detects that this is necessary... so you will run into a make error somewhere in future and can only fix it by hand.

grammar/lexer.l
```diff
@@ -100,9 +102,10 @@ extern int yydebug;
 /* somehow, I need these prototype even though the headers are 
  * included. I guess that's some autotools magic I don't understand...
  */
-#if !defined(__FreeBSD__)
+/*#if !defined(__FreeBSD__)
```
Whissi: Aren't you disabling this code block for all platforms? Is this problem fixed? If we really don't need this hack anymore, please get rid of the complete block.

janmejay:
Sorry for not expressing this earlier. I think renaming to avoid conflicts of any kind(portability, library-provided-type-name conflict etc) is the best way to handle such issues. Any magic adds to cognitive load.

grammar/rainerscript.h
```diff
@@ -26,6 +26,9 @@
 #include <regex.h>
 #include "typedefs.h"
 
+#ifdef _AIX
+#define var var_tt
```
Whissi: Is this another clash like the `msg_t` thing? In this case, shouldn't we take the same road and find a more generic name on all platforms?

rgerhards: how about `svar` or `rsvar` (s for syslog, rs for rsyslog)?

purnimam1: Yes I understand. So since renaming variables across rsyslog will be irrespective of OS check, I think it is better to open a separate issue/bug to drop these changes.

pduveau: Hi, I believe it's a better choice to rename concerned structs and typedefs as this make all future contributions compatible with AIX by design (except for unavoidable GCC pragmas).

purnimam1: yes working on it

purnimam1:
I had to rename Construct/Destruct/ClassInit functions of struct msg to make compilation error free. Please let me know if this suits the rsyslog coding standard or if I should handle this in any other way . Thanks

pduveau:
@purnimam1, Only destruct procedure has to be modified in msg.c.

My code is here 
`/* AIX port : impossible to use this MACRO because of renaming msg to smsg
 * BEGINobjDestruct(smsg)
 */
rsRetVal msgDestruct(smsg_t  **ppThis)
{
	rsRetVal iRet = RS_RET_OK;
	smsg_t *pThis;
	int currRefCount;
#	ifdef HAVE_MALLOC_TRIM
	int currCnt;
#	endif
  BEGINfunc;
`CODESTARTobjDestruct(msg)`

I thought you were more advanced in AIX port. And the imfile module needs a more deep refactor to get compiled. would you like me to publish my work in a fork ?

rgerhards:
@pduveau If you already have code (I thought you were thinking about starting), I think it would be very good to have a second PR, so that we can combine the best of both :-)

Also, it would be great if you could submit *your changes* under ASL 2.0. I have not yet checked if we have some GPL only code, but our intent is to move towards ASL 2.0 in the longer term (much work) and so anything that we do not need to dump later on is definitely preferrable (not saying this is with the current PR - as I said I have not yet checked as this time we want to take a "functionality first, problem solving later" approach).

purnimam1:
Thanks @pduveau I was getting centos6 errors after just modifying msgDestruct. Let me recheck and see if I have some mistakes after changing msg to smsg which is causing this. 
@rgerhards Let me know how I can combine @pduveau's changes so that we can have a faster port.

pduveau:
Hi,
During 3 years until today, we were working on a very big challenge: rehosting our mainframe application on AIX. And I worked on rsyslog AIX port one year ago for this project. But I got some problems with it: freeze of rsyslog after few hundred (400 to 700) events computed. I looked for the source of the problem several days without success. So I decided to suspend the port of rsyslog to find an alternate solution: a deprecated syslog-ng version but running.
Today was the milestone: during the night, we migrated to the rehosted application with success. So now, I can resume the work on rsyslog port.
@purnimam1 no problem, I will push a second PR but tomorrow (too tired for today).
@rgerhards I agree to change the license header. So could you send me the new one?
For the freeze, I suspect dead locks between threads but I can't explain why this can happen on AIX but not on Linux (POSIX threads in both OS) and the same code. Perhaps, it's due to atomics. The code compiled on AIX is using mutexes. I will try to provide a true AIX atomics implementation.

rgerhards:
I highly suggest to rebase this branch to master, so that we get a clean merge and valid CI integration testing.

purnimam1:
@rgerhards yes sure I was having git corruption so wanted to do clean commit hence I reset to older commit so that I can rebase again.

rgerhards:
Note: the Debian failure is unrelated and will go away

rgerhards:
The debian failure is unrelated, it goes away after rebase.

@pduveau I'd appreciate your comments (and maybe a second PR, if that still make sense). I'll check the licenses hopefully tomorrow and hope that we are more or less ready for merge, so I am looking for some final feedback,

rgerhards:
@pduveau sorry, I overlook the question on the license header. There are some files we currently cannot change, as we do not have authors permission. We intend to remove them and write the functionality from scratch as time permits. So we can unfortunately not yet attach a new license header.

What would be useful, for your conribution, is that you state you are OK with moving to ASL 2.0 (or that you rlease **your and only your** changes under ASL, which effectively is the same). Any of this will permit us to keep your code when those files need to be dumped. The files are indicated by having a GPL license header.

PR #1251, opened 2016-11-08
Ksi ignore service fail

allanpark:
The following improvements/bugfixes were introduced:

- In case requesting the KSI signing is unsuccessful (no access to gateway, incorrect credentials, etc.) the server will not give up and continue the log signing. The block are still useful even if lack KSI signature.
- Added log messages for KSI signing events.
- Fixed extending of the KSI signature in rsgtutil, added support for new (KSI) signature format and added warning in case the signature is missing.

Some fixes were also introduced to libksi and this patch should be merged together with the latest patch to libksi (to be sent separately).

Tested manually.

### DIFF
--- a/runtime/librsksi.c
+++ b/runtime/librsksi.c
@@ -247,7 +247,7 @@ hashID2AlgKSI(uint8_t hashID)
 	}
 }
 
-static void __attribute__ ((format (gnu_printf, 2, 3)))
+static void __attribute__ ((format (printf, 2, 3)))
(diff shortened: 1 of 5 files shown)


rgerhards:
@allanpark sorry, I accidently merged without having the libksi changes. Would it be possible to create this PR today? If not, no issue, I'll then revert this merge.

allanpark:
No problem, the libksi PR will come today too.

PR #1259, opened 2016-11-16
Bind to device option plus bugfix

dsahern:


dsahern:
Oops that bugfix patch should be != NULL. Will redo.

PR #1261, opened 2016-11-16
Bind to device and bugfix - v2

dsahern:


rgerhards:
I would be great if you could also add some doc to the rsyslog-doc project, so that others know about the new feature and can actually use it.

dsahern:
Will take care of the docs in the next week.

Issue #1262, opened 2016-11-16
Allow mmpstructdata to place fields under specific json node


mostolog:
Hi

Seems mmpstrucdata it's not working as expected and not exactly as explained in http://www.rsyslog.com/doc/v8-stable/configuration/modules/mmpstrucdata.html

https://github.com/rsyslog/rsyslog/blob/master/plugins/mmpstrucdata/mmpstrucdata.c#L129 shows _jsonRoot_ is initialized NULL, and _rfc5424-sd_ is always used as parent https://github.com/rsyslog/rsyslog/blob/master/plugins/mmpstrucdata/mmpstrucdata.c#L327

Review code to allow nesting structured-data fields under an specified node or as top level fields.

mostolog:
For backward compatibility, default value should be _rfc5424-sd_ 
Set _jsonRoot="foo"_ will nest fields under _foo_ field
Set _jsonRoot=""_ will place fields at top level.

mostolog:
Pending to confirm everything works as expected before updating https://github.com/rsyslog/rsyslog-doc/blob/master/source/configuration/modules/mmpstrucdata.rst

davidelang:
On Wed, 16 Nov 2016, mostolog wrote:
> For backward compatibility, default value should be _rfc5424-sd_
> Set _jsonRoot="foo"_ will nest fields under _foo_ field
> Set _jsonRoot=""_ will place fields at top level.

Yes, that's what I was rying to say :-)

David Lang

JPvRiel:
Just to nudge this along, when I try set `jsonRoot` to something and I fail to understand the nuances of how to set the value, I get this kind of error (v8.36):

```
rsyslogd: internal error:  getJSONRootAndMutex; invalid indicator char ...
```

E.g.
- `!$structured-data` is not permitted as a value. I get this error: "rsyslogd: internal error:  getJSONRootAndMutex; invalid indicator char $(24)"
- And `structured-data` is not permitted: "rsyslogd: internal error:  getJSONRootAndMutex; invalid indicator char s(73)"

So it must start with `!`?

What does work is `!structured-data`. However, instead of directly populating fields in $!structured-data, what it actually does is create `$!structured-data!rfc5424-sd` which for me isn't desired, since I prefer dynamically building into $! the exact field names I want and not needing to explicitly template out a lot of JSON...

Also `jsonRoot` is inconsistent with mmjsonparse's `container` option which in my view is better documented and works as expected.

[...]

PR #1263, opened 2016-11-16
Set mmpstrucdata jsonRoot default ("rfc5424-sd") and honor configuration parameter values

mostolog:
Closes https://github.com/rsyslog/rsyslog/issues/1262

### DIFF
--- a/plugins/mmpstrucdata/mmpstrucdata.c
+++ b/plugins/mmpstrucdata/mmpstrucdata.c
@@ -126,11 +126,12 @@ ENDfreeWrkrInstance
 static inline void
 setInstParamDefaults(instanceData *pData)
 {
-	pData->jsonRoot = NULL;
+	pData->jsonRoot = (uchar*) strdup("rfc5424-sd");
 }
 
 BEGINnewActInst
 	struct cnfparamvals *pvals;
+	char *cstr;
 	int i;
 CODESTARTnewActInst
 	DBGPRINTF("newActInst (mmpstrucdata)\n");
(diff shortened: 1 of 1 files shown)


mostolog:
@rgerhards why mmpstrucdata is using uchar, while mmnormalize char?

rgerhards:
We try to use uchar normally, in order to avoid problems with sign extension. However, if this becomes clumpsy (e.g. to many lib-calls with char) we tend to use char to not do sooo many casts. I think liblognorm uses char, because that's more generic (for non-rsyslog callers) and so this was the reason.

mostolog:
@rgerhards Could you relaunch tests? I think this patch + yesterday's github issues made this tests go crazy. Thanks

rgerhards:
could you do
```
$ git checkout <your-branch>
$ git commit --ammend
..no change..
$ git push -f <your branch>
```
This will trigger a new CI run. Otherwise I need to go into each of the systems individually.
Thanks!

rgerhards:
I have manually restarted the travis run, let's see how it works before doing more manual work.

rgerhards:
The travis failures are real ones.

mostolog:
@rgerhards I'm confused: It's a code issue or a testing issue?

rgerhards:
@mostolog looks like a code issue to me, as the test without the change works. have you run the test in your environment and did it work? You can do this quickly with ```make check TESTS="my_test.sh"```.

mostolog:
I didn't make any tests, just compiled.

make check TESTS="mmpstrucdata.sh" fails with:
```
rsyslogd: could not load module '../plugins/imdiag/.libs/imdiag.so', dlopen: ../plugins/imdiag/.libs/imdiag.so: cannot open shared object file: No such file or directory  [v8.23.0 try http://www.rsyslog.com/e/2066 ]
rsyslogd: invalid or yet-unknown config file command 'IMDiagServerRun' - have you forgotten to load a module? [v8.23.0 try http://www.rsyslog.com/e/3003 ]
rsyslogd started with pid  16786
./diag.sh: line 326: ./tcpflood: No such file or directory
error during tcpflood! see rsyslog.out.log.save for what was written
cp: cannot stat 'rsyslog.out.log': No such file or directory
FAIL mmpstrucdata.sh (exit status: 1)
```

tried building tcpflood diagtalker msleep chkseq but other problems appear:
```
rsyslogd started with pid  29477
starting run 1
Sending 100 messages.
00000100 messages sent
runtime: 0.000
End of tcpflood Run
imdiag[13500]: mainqueue empty
scanf error in index i=0
sequence error detected
FAIL mmpstrucdata.sh (exit status: 1)
```

I could revert+compile+test to confirm it's a code issue (probably, but I don't understand why)

rgerhards:
As general advise, without test development can never be thought of being completed ;-) Without testing, about 40% of my "trivial changes" turn out to be invalid.

To the failure you got under make check. It looks like configure is missing a dependency test. We need ```--enable-imdiag``` in order to do basic testing. So you need to add this to ./configure options.

You do not need to revert your change. You can simply do a local checkout of the last git commit (hash) before your change and then run make check on that code. If you need help with that, email my privately. I'll talk you through.

mostolog:
The following snippet shows one place where I have found issues

```
static inline void
setInstParamDefaults(instanceData *pData)
{
        //was
        //pData->jsonRoot = NULL;
        //changed to (it works)
        pData->jsonRoot = (uchar*) strdup("!");
        //it fails with
        //pData->jsonRoot = (uchar*) strdup("rfc5424-sd");
        //should it use CHK?
        //CHKmalloc(pData->jsonRoot = (uchar*) strdup("rfc5424-sd"));
}
```

I have tried to printf/debug which messages are being processed, but I'm not able to figure if this is a "comparing to expected result" error or what. Any hint?

rgerhards:
OK, I'll use this PR as a starting point to improve the testbench doc for contributors. So that you can interpret testbench failures better.

Issue #1267, opened 2016-11-18
Enchancement: lmsig_ksi module signature file owner/group configuration


allanpark:
Currently the owner of the signature files will default to the UID of the process. The lmsig_ksi should use the same owner/group/createmode as omfile.

rgerhards:
PR to implement this has been merged.

PR #1276, opened 2016-11-24
Fixing issue #1267: allow lmsig_ksi to configure file persmissions

allanpark:
Enabling lmsig_ksi module to use the same file permissions as omfile. Tested manually.

### DIFF
--- a/runtime/librsksi.c
+++ b/runtime/librsksi.c
@@ -665,9 +665,16 @@ writeStateFile(ksifile ksi)
 	struct rsksistatefile sf;
 
 	fd = open((char*)ksi->statefilename,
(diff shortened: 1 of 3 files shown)


rgerhards:
If I read correctly, you can use different  permissions for the KSI files and the log files. I assume this is intentional (makes sense to me)?

allanpark:
Yes, this is intentional.

rgerhards:
Note: the debian failure is unrelated and present as the PR code is not rebased to latest master branch.

rgerhards:
It would be good if you could also update the rsyslog-doc project, so that the new features are actually known.

PR #1277, opened 2016-11-24
mmrm1stspace: add module

PascalWithopf:
work in progress, do not merge


davidelang:
Please add a description of what this module is for and why it's needed.

I think I can guess from the name (removing a leading space if there is one from $msg so you don't need to do the %msg:::space-if-no-first-space% game in your template

But I wonder if having a mm module to do this is really the right thing, or if we should have the ability to add an option to the built-in parsers to remove the space?

rgerhards:
Let me chime in. This is a custom project and there is need for it (I don't know the specifics, but sufficient to know it makes sense). The intent is to remove the leading space, what later on simplifies a lot of rules.

rgerhards:
I should add: I was involved on the spec, and also thought about a parser option. I came to the (weak) conclusion that the mm is better, because it has some additional flexibility on where it is being used in process flow. In the sense of that you have the "original thing" up to a point where you then call the mm. Performance-wise it's not a big deal either, this kind of mm is very fast.

tests/Makefile.am
```diff
@@ -77,6 +77,7 @@ TESTS +=  \
 	timegenerated-utc-legacy.sh \
 	timereported-utc.sh \
 	timereported-utc-legacy.sh \
+	mmrm1stspace-basic.sh \
 	rawmsg-after-pri.sh \
```
rgerhards, line 80: the testbench failures indicate that we need to put our new test into a conditional, so that is it only executed when --enable-mmrm1stspace is given.

rgerhards:
We need to change CI definitions so that the test is run on most platforms (but at least on one not, so that we would find problems like the test not being properly guarded on --enable switch).

rgerhards:
failing check is due to issue in rsyslog core - will fix in merge-PR https://github.com/rsyslog/rsyslog/pull/1295

rgerhards:
Please add documentation to the rsyslog-doc project.

PR #1295, opened 2016-11-30
Merge PR #1277

rgerhards:
PR #1277 plus some fixes required for rsyslog core.


davidelang:
On Wed, 30 Nov 2016, Rainer Gerhards wrote:

>  * mmrm1stspace: add module
>  * travis: enable mmrm1stspace test
>  * mmrm1stspace: fix testbench failures

Just a note, we should probably try to reorder the commits and squash them so 
that we don't have points where bisect doesn't work.

'git rebase -i' lets you re-order commits and squash them together. In this case 
the 'add module' commit and the 'fix testbench failures' commit should have been 
squashed together.

the idea is that privately you do things in the order that you find them, but 
before you merge them, you clean them up to make it look like you had a perfect 
development process, and every step works.

David Lang

rgerhards:
relevant patch merged via https://github.com/rsyslog/rsyslog/pull/1299

Issue #1275, opened 2016-11-24
Enhancement request: implement pmnull parser


mostolog:
As exposed in https://github.com/rsyslog/rsyslog/issues/1191 Rsyslog consider messages have format.

When message is received it tries to match a set of parsers to get properties (syslogtag, priority...) populated.

If we consider using _rawmsg_ as input for mmnormalize (or any other module) it won't make sense to parse it first, wasting a few CPU cycles.

In order to avoid that, and considering it doesn't involve a significant effort, this issue may serve as a request to implement a dummy parser which set all attributes to "" but rawmsg.

Needless to say that this task has lowest priority, cause modules can still handle rawmsg, and it could be just a "performance" improvement (if any), so benchmarking is advised.

rgerhards:
Note that this module will not notably affect performance; parsers are way too fast so that this could make a difference. As of the discussion in #1191, I think it still makes sense to support use cases where messages without headers are to be processed.

As I said previously, we will not set all properties to "", we will use sensible defaults (e.g. set hostname to fromhost), so that the rest of the engine can work correctly.

davidelang:
It would also be extremely useful to be able to deal with raw json messages.

rather than doing a bunch of different new parsers, would it make sense to do a pmnormalize parser that let you use liblognorm and then moved variables named for the standard properties to those properties?

rgerhards:
ack on the pmnormalize, but I think pmnull is really a different beast. New issue?

davidelang:
Ok, I was thinking that pmnull was an uncommon enough use case that pmnormalize with the rule %msg:rest% would be simple enough to satisfy it.

rgerhards:
maybe we combine them later, let's see that in the devel process. My gut
says it's better to keep them separate, but I may be wrong. We'll see once
we do both. So better have two issue trackers, which we may ideally close
with a single commit :-)

2016-11-24 16:07 GMT+01:00 davidelang <notifications@github.com>:

rgerhards:
@davidelang After thinking about it, I really think we should go with a dedicated pmnull parser and one for lognorm. The reason is that there is a big performance difference between the two.

PR #1316, opened 2016-12-08
pmnull: add null parser module

PascalWithopf:
closes https://github.com/rsyslog/rsyslog/issues/1275


rgerhards:
@davidelang I would merge this first. When we look into liblognorm parser (I guess this will be January), we will decide if we make changes to this parser here or create a different one. My gut feeling is still "different one", but this needs to be seen.

Issue #1104, opened 2016-07-29
rsyslog stop working when imjournal.state is empty


haoren3696:
Hi, I met a problem after I reboot my system. Rsyslog does not log any logs after the reboot while the systemd-journald is working normally. It seems imjournal stop reading logs from journal files. Even when I restart rsyslog, It still can not work.
Here is the log:

```
Jul 28 14:51:28 localhost rsyslogd: [origin software="rsyslogd" swVersion="7.4.7" x-pid="41570" x-info="http://www.rsyslog.com"] exiting on signal 15.
Jul 28 14:51:28 localhost rsyslogd: [origin software="rsyslogd" swVersion="7.4.7" x-pid="41705" x-info="http://www.rsyslog.com"] start
Jul 28 14:51:28 localhost rsyslogd-2027: imjournal: fscanf on state file `/var/lib/rsyslog/imjournal.state' failed
 [try http://www.rsyslog.com/e/2027 ]
```

the imjournal.state is empty

```
[root@localhost ~]# ll /var/lib/rsyslog/imjournal.state 
-rw-r--r--. 1 root root 0 Jul 28 14:48 /var/lib/rsyslog/imjournal.state
```

It seems that the cursor did not write to imjournal.state successfully （Maybe there was no enough time for system to flush cache to disk）during the system powering down, 

for now, I change the state file path to /run/log/imjournal.state to avoid the problem, because in my system journal files store in /run which is a tmpfs. After reboot, journal files lost and rsyslog has no need to read outdated cursor in state file.

However, if I choose to store journal in /var, the problem still exist. Does anyone known howto fix the problem in this situation?

radosroka:
I would say that,  this PR #1317 resolves your problem.

rgerhards:
fixed in 8.24.0

avlaanderen:
My system wasn't logging to /var/log/messages anymore with "imjournal: fscanf on state file `/var/lib/rsyslog/imjournal.state' failed"

Tried changing properties for the file /var/lib/rsyslog/imjournal.state to 777 - no success
Then moved it to /var/lib/rsyslog/imjournal.state.old and restarted service rcsyslog.
Back on track

jgerhards:
This is a really old issue - what version of rsyslog are you currently using? If it is a newer one, please consider opening a new issue.

davidelang:
On Thu, 11 Oct 2018, avlaanderen wrote:

> Running CentOS 7.4, so very recent :)

[...]

PR #1317, opened 2016-12-08
Make state file handling in imjournal more robust

tomassykora:
There is a bug in rsyslog which is caused by not very atomic writes of imjournal statefile. It's hardly reproducible but there is a way. fscanf error appears only when rsyslog reads an empty statefile  which causes that imjournal is stopped so no logging from journal is performed. When the statefile contains random bytes error appears again but from journal and imjournal is stopped too.

In this patch Rsyslog writes imjournal statefile more atomically and secure. Reading the statefile is more robust and doesn't affect imjournal module so when corrupted statefile is read imjournal ignores statefile, continues with logging and it doesn't stop. Logger can be used as a test if it's logging or not.

Patch introduces a new option with both old and new config format "IgnoreNonValidStateFile" which is "on" by default and it can turn off ignorance of non valid statefile.

New config:

module(load="imjournal" 
    StateFile="imjournal.state" 
    IgnoreNonValidStatefile="on"/"off" #default on
    #IgnorePreviousMessages="on"
)

Old config:

$ModLoad imjournal
$IMJournalIgnoreNonValidStatefile on/off
#$IMJournalIgnorePreviousMessages on

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -793,8 +793,6 @@ CODEmodInit_QueryRegCFSLineHdlr
 		NULL, &cs.stateFile, STD_LOADABLE_MODULE_ID));
 	CHKiRet(omsdRegCFSLineHdlr((uchar *)"imjournalignorepreviousmessages", 0, eCmdHdlrBinary,
 		NULL, &cs.bIgnorePrevious, STD_LOADABLE_MODULE_ID));
-	CHKiRet(omsdRegCFSLineHdlr((uchar *)"imjournalignorenonvalidstatefile", 0, eCmdHdlrBinary,
-		NULL, &cs.bIgnoreNonValidStatefile, STD_LOADABLE_MODULE_ID)); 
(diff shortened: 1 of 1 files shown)


rgerhards:
Note that we strongly prefer not to provide legacy configuration statements for new functionality. As this is a bug fix, we can accept this, but it would be better if this would not be required. Please let me know your thoughts.

rgerhards:
Thanks! I would say that "on", albeit a change of behaviour, is more appropriate here, because the old behavious is basically a bug.

radosroka:
This is related with issue #1104.

rgerhards:
Also, it would be great if you could update the doc at rsyslog-doc project.

rgerhards:
@radosroka Thanks for the continous work. It would be great if you could also update the doc at rsyslog-doc project.

rgerhards:
@tomassykora Thanks for your continous work. It would be great if you could also update the doc at rsyslog-doc project.

tomassykora:
I will update the doc as soon as possible.

rgerhards:
available in 8.24.0

PR #1331, opened 2016-12-15
omelasticsearch: add write data fail counter

taotetek:
* Adds "failed.writeData" counter to stats for omelasticsearch.
* Each time writeDataError is called, increments the counter regardless of whether errorFile is configured.

### DIFF
--- a/plugins/omelasticsearch/omelasticsearch.c
+++ b/plugins/omelasticsearch/omelasticsearch.c
@@ -71,7 +71,7 @@ STATSCOUNTER_DEF(indexHTTPFail, mutIndexHTTPFail)
 STATSCOUNTER_DEF(indexHTTPReqFail, mutIndexHTTPReqFail)
 STATSCOUNTER_DEF(checkConnFail, mutCheckConnFail)
 STATSCOUNTER_DEF(indexESFail, mutIndexESFail)
-
+STATSCOUNTER_DEF(writeDataESFail, mutWriteDataESFail)
 
 #	define META_STRT "{\"index\":{\"_index\": \""
 #	define META_TYPE "\",\"_type\":\""
@@ -984,6 +984,8 @@ writeDataError(wrkrInstanceData_t *pWrkrData, instanceData *pData, cJSON **pRepl
 	ctx.errRoot=0;
 	DEFiRet;
 
+	STATSCOUNTER_INC(writeDataESFail, mutwriteDataESFail);
+	
 	if(pData->errorFile == NULL) {
 		DBGPRINTF("omelasticsearch: no local error logger defined - "
 		          "ignoring ES error information\n");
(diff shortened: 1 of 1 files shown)


davidelang:
how does this differ from the existing failure counter?

David Lang

taotetek:
@davidelang the existing failure counter does not count write errors to elasticsearch - "write error" in this case meaning "was able to submit the batch to elasticsearch successfully, but then elasticsearch, as part of its normal operation rejected the insert due to an issue with the log content itself".

I feel this scenario (rsyslog can write to ES but ES is rejecting it) is distinct from "failed because  rsyslog cannot write to ES at all) - so being able to distinguish the two is super useful in our monitoring.

wdauchy:
Looks like the second commit is fixing the first one; would have squashed the two

rgerhards:
+1 on the squash.

@taotetek could you quickly do that, then I'll finally merge. I also need to update the impstats counter doc at http://www.rsyslog.com/rsyslog-statistic-counter-plugin-omelasticsearch/ If you could suggest some wording, that would be great (else I use what was written in this thread).

wdauchy:
@rgerhards speaking about doc, is there a repo for this to send merge requests? http://www.rsyslog.com/rsyslog-statistic-counter-plugin-omelasticsearch/ does not seem to be up to date (s/fail\./failed\./g)

davidelang:
On Fri, 13 Jan 2017, William Dauchy wrote:

> @rgerhards speaking about doc, is there a repo for this to send merge requests? http://www.rsyslog.com/rsyslog-statistic-counter-plugin-omelasticsearch/ does not seem to be up to date (s/fail\./failed\./g)

rsyslog docs are at https://github.com/rsyslog/rsyslog-doc

David Lang

rgerhards:
2017-01-13 18:22 GMT+01:00 davidelang <notifications@github.com>:

> On Fri, 13 Jan 2017, William Dauchy wrote:
>
> > @rgerhards speaking about doc, is there a repo for this to send merge
> requests? http://www.rsyslog.com/rsyslog-statistic-counter-
> plugin-omelasticsearch/ does not seem to be up to date
> (s/fail\./failed\./g)
>
> rsyslog docs are at https://github.com/rsyslog/rsyslog-doc


Yes, but for some reason that slipped my mind, the counter docs are only
web page articles. I really can't remember the reason, but there was one
:-( Probably we should migrate these descriptions over to the regular doc
set...

Rainer

wdauchy:
ok that's why I did not found it; I agree we should move the content to this repo so it will be easier to maintain

wdauchy:
@taotetek any chance to squash the two commits as requested? I'd love to see this counter in the next release

rgerhards:
It would also be great if we get some text for the stats description (but as I said, I can add this myself from the text here if required).

taotetek:
I'll get the text for the stats description added today - I can merge and
squash at the same time (github supports it from the UI now, so handy).
I'll get to it after coffee.

Brian

taotetek:
Actually, through the miracle of a cup of coffee and code review, I realized that this:

https://github.com/rsyslog/rsyslog/blob/master/plugins/omelasticsearch/omelasticsearch.c#L1126

Gives the stat that was after. :man_facepalming: 

@rgerhards I'll close the PR, haha.

rgerhards:
okidoki, so that at leased proved the point that the review process (and inisiting on it) works :-)

Issue #1297, opened 2016-12-01
Broken flushOnTXEnd option in omfile


tomassykora:
When flushOnTXEnd option of omfile is off, disk writes should be done when the buffer is full. Now the behaviour in rsyslog is different and this is not true. The disk writes are now done after each transaction. It was probably broken with this commit  6de01034a. According to the current code, when asyncWriting is off, Flush function is called in both cases when flushOnTXEnd is on or off. I'm not sure if this is wanted behaviour.

tomassykora:
The else part of the mentioned commit should be executed only if flushOnTXEnd is off, otherwise the disk writes should be done in runtime/stream.c via strmWrite() function when the buffer if full. Am I right or do I missing something here?

tomassykora:
Any updates on this? Is this at least considered as a bug? @rgerhards

jvymazal:
Ok, investigated original bug when with flushOnTXEnd switch OFF omfile would flush buffer in middle of message, opened pull request #1353 to handle this while preserving buffering function, opinions on this? @rgerhards

davidelang:
I agree that commit looks odd, the description of what it's trying to do doesn't make sense, that's "works as designed"

I agree that this looks like it's doing far too many flushes.

rgerhards:
I have now had a deeper look into the issue.

Let's have a clear look of what was actually changed: From

```
	if(pData->bFlushOnTXEnd && pData->pStrm != NULL) {
		/* if we have an async writer, it controls the flush via
		 * a timeout. However, without it, we actually need to flush,
		 * else incomplete records are written.
		 */
		if(!pData->bUseAsyncWriter)
			CHKiRet(strm.Flush(pData->pStrm));
	}
```
which means flush the file if no async writer is used and flush on tx end is set; do **never** do this if the async writer is used
to
```
	if(pData->bUseAsyncWriter) {
		if(pData->bFlushOnTXEnd && pData->pStrm != NULL) {
			CHKiRet(strm.Flush(pData->pStrm));
		}
	} else {
		if(pData->pStrm != NULL) {
			CHKiRet(strm.Flush(pData->pStrm));
		}
	}
```
Which means flush the file even if async writer is used plus *flush always if not*.

[...]

PR #1353, opened 2017-01-10
omfile to always flush whole message when buffer limit reached

jvymazal:


rgerhards:
Can you please elaborate what is the intent of this patch and why you think it is needed? This smells like something that needs a dedicated config option. Also note that the code is buggy as can be seen by the testbench results.

davidelang:
On Tue, 10 Jan 2017, Rainer Gerhards wrote:

> Can you please elaborate what is the intent of this patch and why you think it is needed? This smells like something that needs a dedicated config option.

As I understand it, the intent is that a file on disk will not have any partial 
log messages, so when a buffer if flushed, the entire buffer isn't flushed, it's 
only flushed up to the last full log message.

That said (and I haven't looked at this code yet), I don't see how to know where 
the end of the last log message is, given that log messages can be multi-line 
and include variables in the template that can themselves be multi-line.

I also wonder at the performance of this. It seems like it would be very 
straightforward to pass a binary-sized block of data to the OS for it to write, 
but a lot more work to submit a portion of that block, and then copy the 
remainder to the beginning of the new buffer.

David Lang

rgerhards:
> That said (and I haven't looked at this code yet), I don't see how to know where
> the end of the last log message is, given that log messages can be multi-line
> and include variables in the template that can themselves be multi-line.
> 
> I also wonder at the performance of this. It seems like it would be very
> straightforward to pass a binary-sized block of data to the OS for it to write,
> but a lot more work to submit a portion of that block, and then copy the
> remainder to the beginning of the new buffer.

That's why I think it needs at least to be a dedicated option -- it is counter of what the buffered mode really is about. So this cannot become the default.

jvymazal:
Purpose of this is to solve issue #1297 where are commit 6de0103 omfile started flushing its buffer after each log message even with flushOnTXEnd off, regardless of ioBufferSize (current behavior in master). I believe this is not intended (at least omfile doc considering) so I proposed this patch to retain buffering with flushOnTXEnd off, yet retain that when actually flushing there would be no partial message in disk file (code is detecting by watching stream buffer pointer whether write actually happened and if so flushes the rest of a buffer to file as well). Another approach to this would be not to touch omfile.c but make changes to stream.c write() implementation but considering present [comment](https://github.com/rsyslog/rsyslog/blob/master/runtime/stream.c#L1827) I though this better option.

davidelang:
On Wed, 11 Jan 2017, jvymazal wrote:

> Purpose of this is to solve issue #1297 where are commit 6de0103 omfile 
> started flushing its buffer after each log message even with flushOnTXEnd off, 
> regardless of ioBufferSize (current behavior in master). I believe this is not 
> intended (at least omfile doc considering) so I proposed this patch to retain 
> buffering with flushOnTXEnd off, yet retain that when actually flushing there 
> would be no partial message in disk file (code is detecting by watching stream 
> buffer pointer whether write actually happened and if so flushes the rest of a 
> buffer to file as well). Another approach to this would be not to touch 
> omfile.c but make changes to stream.c write() implementation but considering 
> present 
> [comment](https://github.com/rsyslog/rsyslog/blob/master/runtime/stream.c#L1827) 
> I though this better option.

restating this to see if I understand it:

the problem is when the message is larger than the buffer (as opposed to the 
more normal case where multiple messages fit in one buffer), it gets partially 
written with the remainder in the next buffer.

you want to make it so the entire message gets written at once.

is this correct?

David Lang

jvymazal:
Yes, that is correct.

Problem is that while normally (flushOnTXEnd on - default) case when message is larger is indeed rare, however with this switch off, buffer contains all previous messages until buffer size is reached, causing this case to happen almost 100% of time.

rgerhards:
Yes, but it is intended behaviour that the code writes partial lines. This has to do with performance. I don't object adding a switch to change that behaviour, but the default should remain as-is. The buffered mode was created for highly demanding environments where buffers fill up very very quickly. So it really is irrelevant if for a short time a partial message is missing.

And, of course, I do not object fixing a bug. We just need to keep the semantics.

jvymazal:
Fine, in that case could you please somehow resolve mentioned #1297 issue (e.g. reverting/editing mentioned commit to enable for buffered mode to function) If writing of partial lines was intended than OK and we can close this. Thank you.

rgerhards:
replaced by https://github.com/rsyslog/rsyslog/pull/1364 as requested.

PR #1357, opened 2017-01-12
mmdblookup module cleanup

GrayTShirt:
this covers building, style, and readme

PR #1363, opened 2017-01-13
mmdblookup: build module correctly

PascalWithopf:


GrayTShirt:
@PascalWithopf, @rgerhards 
Given the traffic on the rsyslog-mailing-list there seems to be big a community want for geoip lookups. Thus, I think the module would get better long-term support if the syntax were fixed, along with compiler errors and warnings.

see #1357

Issue #1365, opened 2017-01-13
add option to permit slash in programname


rgerhards:
see https://github.com/rsyslog/rsyslog/issues/1361 for details

Issue #1361, opened 2017-01-13
imjournal programname name parsing problem when long programname


hese10:
It seems that rsyslog (imjournal plugin?) has bug that it does not get programname property at all in some case, but just empty string.
This was seen when outputting%programname% into file as separate field. The output format in this case was :
`template(name="FileFormat2" type="string" string="%timereported:::date-rfc3164%.%timereported:::date-subseconds% %syslogseverity-text% %HOSTNAME:F,46:1:uppercase% %programname% %syslogtag:R,ERE,1,FIELD:^(.*):--end%:%msg:::sp-if-no-1st-sp%%msg:::drop-cc%\n")
`
programname is correct in bash case, but empty in that long one "/opt/testi/SS_TstBash/bin/bash"
```
Jan 12 20:00:05.170717 info TMN-0 -bash -bash[2206]: HISTORY: PID=2206 UID=0 /opt/testi/SS_TstBash/bin/bash
Jan 12 20:00:07.071914 info TMN-0  /opt/testi/SS_TstBash/bin/bash[13629]: HISTORY: PID=13629 UID=0 pwd
```

so is there some limit in rsyslog for programname, and when over it, there comes empty programname?

davidelang:
On Thu, 12 Jan 2017, hese10 wrote:

> It seems that rsyslog has bug that it does not get programname property at all, but just empty string.
> This was seen when outputting%programname% into file as separate field. The output format in this case was :
> `template(name="FileFormat2" type="string" string="%timereported:::date-rfc3164%.%timereported:::date-subseconds% %syslogseverity-text% %HOSTNAME:F,46:1:uppercase% %programname% %syslogtag:R,ERE,1,FIELD:^(.*):--end%:%msg:::sp-if-no-1st-sp%%msg:::drop-cc%\n")
> `
> programname is correct in bash case, but empty in that long one "/opt/testi/SS_TstBash/bin/bash"
> ```
> Jan 12 20:00:05.170717 info TMN-0 -bash -bash[2206]: HISTORY: PID=2206 UID=0 /opt/testi/SS_TstBash/bin/bash
> Jan 12 20:00:07.071914 info TMN-0  /opt/testi/SS_TstBash/bin/bash[13629]: HISTORY: PID=13629 UID=0 pwd
> ```
>
> so is there some limit in rsyslog for programname, and when over it, there comes empty programname?

Yes, the RFC3184 parser has lots of heuristics to try and guess what to do when 
[...]

PR #1366, opened 2017-01-13
add option "parser.PermitSlashInProgramname"

rgerhards:
The new option permits to have slashes inside the programname
property. This is counter the BSD syslog defintitions, but
sometimes useful (e.g. the journal records path names, which
obviously contain slashes).

closes https://github.com/rsyslog/rsyslog/issues/1365
see also https://github.com/rsyslog/rsyslog/issues/1361

### DIFF
--- a/runtime/glbl.c
+++ b/runtime/glbl.c
@@ -7,7 +7,7 @@
  *
  * Module begun 2008-04-16 by Rainer Gerhards
  *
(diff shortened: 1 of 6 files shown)


rgerhards:
new CI tests still missing; wanted to see how it works cross-platform with existing testbench before digging deeper.

PR #1380, opened 2017-01-25
mmnormalize: add possibility to get rule from string

PascalWithopf:


rgerhards:
FYI: Testbench machines need to be updated to new liblognorm version

tests/mmnormalize_rule_from_string.sh
```diff
@@ -0,0 +1,33 @@
...
+. $srcdir/diag.sh add-conf '
+module(load="../plugins/imtcp/.libs/imtcp")
+module(load="../plugins/mmnormalize/.libs/mmnormalize")
+
+input(type="imtcp" port="13514" ruleset="norm")
+
+template(name="outfmt" type="string" string="%hostname% %syslogtag%\n")
+
+ruleset(name="norm") {
+	action(type="mmnormalize" rule="rule=:%host:word% %tag:char-to:\\x3a%: no longer listening on %ip:ipv4%#%port:number%")
```
rgerhards: We also need one test where the rules are populated from an array. This is actually a case where problems are more likely to happen.

PR #1381, opened 2017-01-25
validate mmdblookup tests

GrayTShirt:


GrayTShirt:
ah I see now, I forgot to add the new rule file to the test dist

GrayTShirt:
awesome we validated the new test.

tests/mmdb.sh
```diff
@@ -2,13 +2,16 @@
...
+. $srcdir/diag.sh startup mmdb.conf
+. $srcdir/diag.sh tcpflood -m 1 -j "202.106.0.20\ "
 echo doing shutdown 
-. $srcdir/diag.sh shutdown-when-empty 
+. $srcdir/diag.sh shutdown-when-empty
 echo wait on shutdown 
-. $srcdir/diag.sh wait-shutdown  
-. $srcdir/diag.sh content-check '"city" : "Beijing"' 
+. $srcdir/diag.sh wait-shutdown
+. $srcdir/diag.sh content-check '{ "city": "Beijing" }'
```
rgerhards, line 16: can we be sure this IP will always remain valid for "Beijing"? In other words: is there a dedicated testing entry inside maxmindb? If so, we should use it. My fear is that in the distant future, this test might fail simply because the address assignment has changed. If we can't do it differently, we should put some comments into the test that tell us how we can find a new testing address. So that someone without knowledge of maxmindb has a chance to update the test.

dmolik: the original committer added a custom maxmind database for testing. Located at tests/test.mmdb, it's referenced in the test's config file (tesst/testsuites/mmdb.conf)

rgerhards: excellent!

PR #1392, opened 2017-02-03
Fixed situation when time goes backwards

radosroka:
This is resolving the situation when system is after reboot and boot_id
doesn't match so cursor pointing into "future". Usually sd_journal_next
jump to head of journal due to journal aproximation, but when system
time goes backwards and cursor is still invalid, rsyslog stops logging.
We use sd_journal_get_cursor to validate our cursor. When cursor is
invalid we are trying to jump to the head of journal. This problem with
time is not affecting persistent journal.

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -527,7 +527,28 @@ loadJournalState(void)
 						"couldn't seek to cursor `%s'\n", readCursor);
 					iRet = RS_RET_ERR;
 				} else {
+					char * tmp_cursor = NULL;
 					sd_journal_next(j);
+					/*
+					* This is resolving the situation when system is after reboot and boot_id doesn't match
+					* so cursor pointing into "future".
+					* Usually sd_journal_next jump to head of journal due to journal aproximation,
+					* but when system time goes backwards and cursor is still invalid, rsyslog stops logging.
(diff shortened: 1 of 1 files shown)


rgerhards:
@radosroka just in case you have not seen: CI tests err out.

radosroka:
@rgerhards I'm guessing that sd_journal_has_persistent_files isn't available on that ubuntu system. I'm not very familiar with ubuntu but I will look at that.

rgerhards:
@radosroka then I think it is necessary to add a configure check for that function, and either avoid its user or mimic its functionality via another method. I can't merge the change if it break build on some platforms.

radosroka:
@rgerhards that function wasn't in that older systemd. I found out that this function isn't necessary. It is good to provide that check whether is journal persistent or not. It is more secure.

This problem could be tested in VM:
* turn of NTP
* disable rsyslog.service
* stop rsyslog.service
* set time to future(e.g. year or two)
* run rsyslog and log something(update statefile)
* stop rsyslog
* set time back or into past
* reboot VM
* run rsyslogd -n
* without this commit rsyslog doesn't logging anymore
* with this commit there is an message about that and rsyslog works

This problem appears in infrastructure when VMs synchronize time with host and host changes time backwards.

PR #1393, opened 2017-02-04
Support SNI when connecting as a client

artbristol:
I only tested the happy path, but it seems to work OK.

Not sure what happens if it's a bare IP address.


rgerhards:
Could you test the bare IP Address case?

We probably should also add a testbench test for this, will hold merge until we have.

runtime/nsd_gtls.c
```diff
@@ -1709,6 +1709,8 @@ Connect(nsd_t *pNsd, int family, uchar *port, uchar *host, char *device)
 	pThis->bHaveSess = 1;
 	pThis->bIsInitiator = 1;
 
+	gnutls_server_name_set(pThis->sess, GNUTLS_NAME_DNS, host, ustrlen(host));
```
rgerhards: requires testbench coverage, best with hostname and IP

artbristol:
I would like to try to add a testbench test but I can't work out how to run the test suite.

The README in the `tests` directory doesn't say what to do and `make check` runs suspiciously quickly.

rgerhards:
I guess you did not do --enable-testbench?

artbristol:
I included that flag - how do I run a single test?

rgerhards:
make check TESTS="testname"

2017-02-19 18:51 GMT+01:00 Art O Cathain <notifications@github.com>:

artbristol:
I'm trying to track down a test I can adapt/copy for this use case. Currently it seems that `sndrcv_tls_anon.sh` and `sndrcv_tls_anon_rebind.sh` are the only ones that fail if I sabotage the TLS sending code, so I think they might be a good starting point.

However I saw this comment in `tests/sndrcv_drvr_noexit.sh`

```
# A note on TLS testing: the current testsuite (in git!) already contains
# TLS test cases. However, getting these test cases correct is not simple.
# That's not a problem with the code itself, but rater a problem with
# synchronization in the test environment. So I have deciced to keep the
# TLS tests in, but not yet actually utilize them. This is most probably
# left as an excercise for future (devel) releases. -- rgerhards, 2009-11-11
```

It it still accurate?

rgerhards:
I think this is solved by now

artbristol:
I added tests including the bare IPv4 and IPv6 cases.

Note, because the rsyslog server doesn't do anything with the SNI information, the tests still pass even without the main-code changes. However they're still useful for coverage.

rgerhards:
can you please rebase to current master. It looks like the testbench failures are due to unrelated issues, but I am not 100% sure. If you cannot rebase, I'll later create a merge PR for it.

rgerhards:
I have done the rebase myself as I had permissions.

artbristol:
I thought I'd done the rebase, but OK.

Looks like tests still failing for some reason. I'll have a look, might be socket binding problems or something.

rgerhards:
sorry, I did the rebase as I thought this would fix the remaining issue. I think you did not have the latest version of master. Unfortunately, it looks like the failure on Freebsd is related to the patch. If you think this is not the case, let me know and I'll re-run the CI check on Freebsd.

rgerhards:
@artbristol any interest working on a final version of the patch?

artbristol:
yes, sorry for long delay. i'll take a look in the next week

rgerhards:
Thanks! looking forward to merge it :-)

artbristol:
It's passing now but I can't see any FreeBSD info in the Travis CI build - is FreeBSD still part of CI?

rgerhards:
give the testbench a bit of time - status will (unfortunately) only show up when the CI run begins.

Full status: http://build.rsyslog.com:8010/waterfall?tag=rsyslog-rsyslog

Also note: https://twitter.com/rgerhards/status/927215776213463040

rgerhards:
looks good! (Y) I'll review once more and hope to merge next week. Thanks a ton!

rgerhards:
In order for folks to use that feature, they need to know it exists and how to use it. Could you please ammend the doc at https://github.com/rsyslog/rsyslog-doc

rgerhards:
@artbristol can you give me a little bit of description that I could add myself to the doc? I would really like to merge this ASAP and the missing doc is blocker.

artbristol:
Hi

Sorry for delay. The feature will be enabled automatically. If you specify
a hostname  (which you usually would), then SNI will be used during the TLS
connection. I'm not sure it really needs documentation other than a line in
the release note, since SNI is pretty standard when using TLS these days?

Art

rgerhards:
OK, I'll see that I craft a little change. Is there any way to disable it (I haven't looked up the code once again)?

artbristol:
Not specifically, but you can specify a bare IP then it won't use SNI.

On 11 Nov 2017 13:57, "Rainer Gerhards" <notifications@github.com> wrote:

OK, I'll see that I craft a little change. Is there any way to disable it
(I haven't looked up the code once again)?

—
You are receiving this because you were mentioned.
<https://github.com/rsyslog/rsyslog/pull/1393#issuecomment-343666529>, or mute
the thread

rgerhards:
@davidelang 
> Not specifically, but you can specify a bare IP then it won't use SNI.
Just a double-check: do you see any issue with this?

davidelang:
just to make sure I'm understanding this correctly, this is an added header that 
tells the server what name you are intending to connect to, so that you can do 
multi-site hosting with encryption (so that the server can respond with the 
correct cert)

I can't think of any server side software that would be safe to use that doesn't 
implement this reliably. As such, just note that we now support this and move 
on.

No need to have a way to disable it.

PR #1398, opened 2017-02-07
core: fix sequence error in msg object deserializer

### LINKED ISSUE #1404: core: disk queue corruption (error -2308)
Corruption always happens if the "json" property (message variables) is present and "structured-data" property is also present. This causes rsyslog to serialize to the queue in wrong property sequence, which will lead to error -2308 on deserialization. Seems to be a long-standing bug. Depending on version used, some or all messages in disk queue may be lost.

rgerhards:
closes #1404


rgerhards:
note: contains PR #1402 so that we can ensure the bug is actually fixed. When this is merged, #1402 also will be.

PR #1403, opened 2017-02-08
testbench: add test for config optimizer

rgerhards:
check if optimization to prifilt() works properly


rgerhards:
**DO NOT MERGE**

It turned out this test case was already covered by a number of  other tests. I just keep it open for the testbench run for a few more moments to check the testbench environment. After that, this will be closed unmerged.

rgerhards:
redundant test

Issue #1406, opened 2017-02-09
Rate limiting not logging process name


jvymazal:
This would generally help to quickly understand which process is sending more logs than the rsyslog default limit. And user does not need to go to the terminal to grep that process. 
Also assume we are analysing previous logs. Since they were collected pid could change or process could be terminated so it would not help to see pid but if name is logged it can help.

davidelang:
how would you record this? the rate limiting makes rsyslog stop processing logs 
for a while, so you either already have the data in the logs that are processed, 
or rsyslog doesn't have the data to do the reporting you are asking for.

David Lang

rgerhards:
merged via pr in 8.26.0

PR #1407, opened 2017-02-09
added logging name of source of rate-limited messages

jvymazal:
attempt to solve #1406

### DIFF
--- a/runtime/ratelimit.c
+++ b/runtime/ratelimit.c
@@ -132,7 +132,7 @@ tellLostCnt(ratelimit_t *ratelimit)
  * be called concurrently.
  */
 static int
-withinRatelimit(ratelimit_t *ratelimit, time_t tt)
+withinRatelimit(ratelimit_t *ratelimit, time_t tt, char* appname)
 {
 	int ret;
 	uchar msgbuf[1024];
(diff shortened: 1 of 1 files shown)


runtime/ratelimit.c
```diff
@@ -215,7 +215,7 @@ ratelimitMsg(ratelimit_t *ratelimit, smsg_t *pMsg, smsg_t **ppRepMsg)
 	/* Only the messages having severity level at or below the
 	 * treshold (the value is >=) are subject to ratelimiting. */
 	if(ratelimit->interval && (pMsg->iSeverity >= ratelimit->severity)) {
-		if(withinRatelimit(ratelimit, pMsg->ttGenTime) == 0) {
+        if(withinRatelimit(ratelimit, pMsg->ttGenTime, getAPPNAME(pMsg, 0)) == 0) {
```
rgerhards: I am not sure if it is a good idea to just pull the APPNAME. This is generic code here. So if e.g. a message is received from a remote system which is under ratelimiting, we get that remote system's APPNAME. I think that would be pretty confusing. Maybe it would be a good idea to add a source indication?

jvymazal: Does not the `ratelimit->name` contain enough information to distinguish that? Or do you think something more should be added in name construction in `ratelimitNew()`?

rgerhards:
please squash the commits, as they are just corrections. It would be great if you could also add a testbench test (but not vital).

jvymazal:
Unfortunately I fail to devise any test case for usage I mentioned in #1406 , can this be merged to make it to 8.25.0 release?

rgerhards:
I would love to have a test case. I'll try to merge in 8.25 timeframe, but not sure if I meet it. Freeze is Friday and I have tight meeting deadlines this week.

rgerhards:
finally merging without a test case, we can always add this later. I think it does not make sense to hold merge any longer. @PascalWithopf maybe you are up to crafting a test case?

PR #1414, opened 2017-02-13
add new test

jgerhards:
to check if es_str2num (in libestr) handles empty strings
correctly. Not yet in use, but will be useful later on.


rgerhards:
We need to look into the testbench environment. We could reproduce an issue when run on local development machine. Question is why this is not seen on ASAN-enabled CI tests.

Issue #1419, opened 2017-02-15
Add chdir to chroot call


jvymazal:
Rsyslog has (with the `T` cmd-line option) option to perform `chroot` immediately after starting. It is generally recommend practice to do a `chdir` call ASAP after `chroot`. This prevents easy escape outside of chroot'ed environment. Could this call be added?

rgerhards:
Just so that this is said here again for anyone coming in from Google: -T is a **testing aid** not a security aid. For security needs, you **should not** rely on it. more inside the PR.

rgerhards:
done in 8.26 via separate pr

PR #1420, opened 2017-02-15
added chdir call after successful chroot

jvymazal:
attempt to solve #1419

### DIFF
--- a/tools/rsyslogd.c
+++ b/tools/rsyslogd.c
@@ -1350,6 +1350,10 @@ initAll(int argc, char **argv)
 				perror("chroot");
 				exit(1);
 			}
+            		if(chdir("/") != 0) {
+                		perror("chdir");
+		                exit(1);
+		            }
 			break;
 		case 'u':		/* misc user settings */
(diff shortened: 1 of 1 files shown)


tools/rsyslogd.c
```diff
@@ -1350,7 +1350,7 @@ initAll(int argc, char **argv)
 				perror("chroot");
 				exit(1);
 			}
-            if(chdir("\\") != 0) {
```
rgerhards: please fix indenting to comply to the style used in the rest of the switch

rgerhards:
It must be noted that the -T option is a testing option, and not meant for secure operations (as per the comment). So this is not an option for general use. I would suspect that the change could even cause some scenarios to fail, but I would accept that until I get feedback from tests that this poses a real issue.

rgerhards:
sorry, I forgot: could you squash the changes, as it really is just one change with two fixes to clean up problems in the change itself. If you do not know how to do this, I can do it for you (assuming you have given appropriate permissions).

jvymazal:
Ok, I squashed the commits, even if -T is presently testing option it would be nice to have it there (for example if it would become security option one day). Thanks.

PR #1428, opened 2017-02-21
EXPERIMENTAL imuxsock: add option to never wait on queue enqueue

rgerhards:
**DO NOT MERGE**


rgerhards:
This option most probably is NOT a good idea. I still prefer to keep the PR open for a bit more of discussion.

**DO NOT MERGE**

PR #1429, opened 2017-02-22
Add ability to set owner/group/access perms for imptcp uds

nbrownus:
Adds similar functionality to `omfile` for setting file ownership and access permissions


rgerhards:
@nbrownus just in case you didn't notice: this is unmergable due to segfaults in CI testing.

nbrownus:
Finally had the time to properly fix this. Should be good to go now.

Issue #272, opened 2015-03-17
Setting net.aclResolveHostname/net.acladdhostnameonfail segfaults


nbartos:
When compiling using hardned gcc (gentoo), specifying net.aclResolveHostname or net.acladdhostnameonfail results in rsyslogd segfaulting on startup.

I did a quick look at the code, but my C is fairly rusty.  It seems that net.pACLAddHostnameOnFail/net.pACLDontResolve are not being initialized properly before trying to be used in runtime/glbl.c.  Is there a reason net.pACLAddHostnameOnFail and net.pACLDontResolve are integer pointers instead of just being integers?

radosroka:
This is still an issue even in fedora machines. (rsyslog 8.24.0)

radosroka:
@rgerhards after small research I think this is caused by uninitialized net static structure in glbl objUse(net) is not called after start of glbl module. net.pACLAddHostnameOnFail/net.pACLDontResolve are zeros and they are segfaulting inside of glblDoneCnf() when derefencing *(net.pACLDontResolve).

radosroka:
I'm not sure if it even possible to call objUse for net in BEGINAbstractObjClassInit, but it is the first option.

The second option is that instead of dereferencing *(net.pACLDontResolve) we can use global variable which it points to.

Third option can be initialization inside BEGINAbstractObjClassInit line net.pACLDontResolve = ACLDontResolve.

rgerhards:
fixed via pr

PR #1433, opened 2017-02-23
Fixed segfault in glblDoneLoadCnf()

radosroka:
This is caused by uninitialized net static structure in glbl objUse(net)
is not called after start of glbl module.

It looks like it is not possible to use objUse(net) when glbl module
starts. So we call it inside of glblDoneLoadCnf().

### DIFF
--- a/runtime/glbl.c
+++ b/runtime/glbl.c
@@ -1076,11 +1076,13 @@ finalize_it:
 /* This processes the "regular" parameters which are to be set after the
  * config has been fully loaded.
  */
-void
+rsRetVal
 glblDoneLoadCnf(void)
 {
 	int i;
 	unsigned char *cstr;
+	DEFiRet;
+	CHKiRet(objUse(net, CORE_COMPONENT));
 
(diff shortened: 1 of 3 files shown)


radosroka:
This is resolving the issue  #272

radosroka:
@rgerhards What do you think about such solution?

rgerhards:
I think it looks good. I was thinking if we could use a more straightforward way that you do not yet know about ;-) The objUse() and helpers stem back to support for systems that do not support dynamic loading of modules that refer back to global functions in already loaded code. At the time we designed this, there were a couple of such systems, but in the meantime all platforms support this. So I plan to refactor the whole system in order to simplify it and make it more linux-like. That would probably solve the root problem.

HOWEVER.I have the plan for quite a while, and have occasionally used the new approach. But full refactor requires quite some time, so it is probably best to merge as-is.

I'd already merged if I were not working on a very strange but very high priority task right now and the next release is far away. You've already seen the release goal attached ;-)

radosroka:
Thanks for sharing your opinion. 👍

PR #1441, opened 2017-03-01
core/file handling: add new error messages

rgerhards:
Those messages previously went only to the debug log, as due to design limitations inside the error message subsystem they could not be emitted at the related code locations. This has changed now, and so we can improve error reporting.


rgerhards:
re-issued as #1460 (due to rebase issues)

PR #1457, opened 2017-03-09
test for CI integration - why do some unrelated tests fail in PR?

rgerhards:
**DO NOT MERGE**

Test for CI system, needs to be deleted after testing (sorry, due to being CI-related, this **must** be done on the real project git...)


rgerhards:
seems to have been transient, no repro even though made several tries -- good!

PR #1458, opened 2017-03-09
omfwd: add support for network namespaces

BastianStender:
Network namespaces are part of the linux kernel since v2.6.24. This
enables forwarding output into a configurable network namespace. E.g.,

action(type="omfwd" networknamespace="my_ns" target="192.168.1.23"
port="10514")

Signed-off-by: Bastian Stender <bst@pengutronix.de>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -179,6 +179,7 @@ AC_FUNC_STAT
 AC_FUNC_STRERROR_R
 AC_FUNC_VPRINTF
 AC_CHECK_FUNCS([flock inotify_init recvmmsg basename alarm clock_gettime gethostbyname gethostname gettimeofday localtime_r memset mkdir regcomp select setsid socket strcasecmp strchr strdup strerror strndup strnlen strrchr strstr strtol strtoul uname ttyname_r getline malloc_trim prctl epoll_create epoll_create1 fdatasync syscall lseek64])
+AC_CHECK_FUNC([setns], [AC_DEFINE([HAVE_SETNS], [1], [Define if setns exists.])])
 AC_CHECK_TYPES([off64_t])
 
 # getifaddrs is in libc (mostly) or in libsocket (eg Solaris 11) or not defined (eg Solaris 10)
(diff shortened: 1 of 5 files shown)


BastianStender:
Hi,

do you have any idea how to fix this on CentOS 6 and FreeBSD?

I suspect that CentOS 6 has a kernel < 3.0 that does not know setns(). I thought that my [#ifdef](https://github.com/rsyslog/rsyslog/pull/1458/files#diff-c099b5b1da959592a560a0970ebbe645R760) would exclude BSD.

Regards,
Bastian Stender

rgerhards:
how about a configure check for setns()?

configure.ac
```diff
@@ -1316,6 +1316,10 @@ AC_ARG_ENABLE(testbench,
 )
 AM_CONDITIONAL(ENABLE_TESTBENCH, test x$enable_testbench = xyes)
 if test "x$enable_testbench" = "xyes"; then
+	AC_CHECK_PROG(IP, [ip], [yes], [no])
```
rgerhards: I think we should make this optional, as only one test needs it. The test should skip out if ip is not found. As it is now, the testbench does not work at all on any platform that does not have "ip".

BastianStender:
setns() and ip availability are now checked in configure.ac

rgerhards:
one final thing: would you be willing to contribute a doc update to https://github.com/rsyslog/rsyslog-doc/blob/master/source/configuration/modules/omfwd.rst ? Otherwise, nobody knows about the new feature. It would be great if you could also add an example use case...

BastianStender:
Sure, will do.

BastianStender:
See https://github.com/rsyslog/rsyslog-doc/pull/321 for the documentation pull request.

BastianStender:
I cannot reproduce the error in the test suite. I guess that has nothing to do with my commit really.

Any comments to the pull request or is anything missing?

rgerhards:
This seems to be a race on Travis, I have restarted the test. Note that, as advised on the mailing list, all Solaris tests currently fail as we are working on getting this up and running.

BastianStender:
The tests now look good. Anything left to do?

rgerhards:
all fine as it looks, but I do not merge every day due to other work. Probably today,.

PR #1465, opened 2017-03-15
pmnormalize: new parser module

PascalWithopf:
Parser module that uses liblognorm to parse incoming messages.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -99,6 +99,10 @@ if ENABLE_PMNULL
 SUBDIRS += plugins/pmnull
 endif
 
+if ENABLE_PMNORMALIZE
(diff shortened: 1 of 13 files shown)


davidelang:
I may be just missing it, but will this set the standard properties? fromhost-ip and timereceived should be set normally.

I see config parameters to set tag, facility, severity defaults, but there should be a way to set these from within the json (something along the lines of special variable names that, if they exist, get copied to the standard properties)

PascalWithopf:
@davidelang 
The module takes the RawMsg, normalizes it with ln_normalize and sets the properties with the methode MsgSetPropsViaJSON_Object from the json object.
Could you test the module to see if it works?

PR #1467, opened 2017-03-15
Add option for disable reuse of connection to elastic

iavael:
In some setups several elastic cluster nodes are located behind HTTP load balancers. In case when some of these nodes go down, connections from rsyslog daemons group on healthy nodes and after failed nodes go up, rsyslogs doesn't connect to them for some time. This patch allows sysadmins of such setups to disable connection reuse and improve load balancing.

### DIFF
--- a/plugins/omelasticsearch/omelasticsearch.c
+++ b/plugins/omelasticsearch/omelasticsearch.c
@@ -108,6 +108,7 @@ typedef struct _instanceData {
 	sbool dynBulkId;
 	sbool bulkmode;
 	size_t maxbytes;
+	sbool reuseconn;
 	sbool useHttps;
 	sbool allowUnsignedCerts;
 } instanceData;
@@ -145,6 +146,7 @@ static struct cnfparamdescr actpdescr[] = {
 	{ "dynparent", eCmdHdlrBinary, 0 },
 	{ "bulkmode", eCmdHdlrBinary, 0 },
 	{ "maxbytes", eCmdHdlrSize, 0 },
+	{ "reuseconn", eCmdHdlrBinary, 0 },
 	{ "asyncrepl", eCmdHdlrGoneAway, 0 },
         { "usehttps", eCmdHdlrBinary, 0 },
 	{ "timeout", eCmdHdlrGetWord, 0 },
@@ -263,6 +265,7 @@ CODESTARTdbgPrintInstInfo
 	dbgprintf("\tuse https=%d\n", pData->useHttps);
 	dbgprintf("\tbulkmode=%d\n", pData->bulkmode);
 	dbgprintf("\tmaxbytes=%zu\n", pData->maxbytes);
+	dbgprintf("\treuseconn=%d\n", pData->reuseconn);
 	dbgprintf("\tallowUnsignedCerts=%d\n", pData->allowUnsignedCerts);
 	dbgprintf("\terrorfile='%s'\n", pData->errorFile == NULL ?
 		(uchar*)"(not configured)" : pData->errorFile);
@@ -395,7 +398,9 @@ finalize_it:
 BEGINtryResume
 CODESTARTtryResume
 	DBGPRINTF("omelasticsearch: tryResume called\n");
-	iRet = checkConn(pWrkrData);
+	if(pWrkrData->pData->reuseconn) {
+		iRet = checkConn(pWrkrData);
+	}
 ENDtryResume
 
 
@@ -1315,7 +1320,7 @@ curlCheckConnSetup(CURL *handle, HEADER *header, long timeout, sbool allowUnsign
 }
 
 static void
-curlPostSetup(CURL *handle, HEADER *header, uchar* authBuf)
+curlPostSetup(CURL *handle, HEADER *header, uchar* authBuf, sbool reuseconn)
 {
 	curl_easy_setopt(handle, CURLOPT_HTTPHEADER, header);
 	curl_easy_setopt(handle, CURLOPT_WRITEFUNCTION, curlResult);
@@ -1325,6 +1330,10 @@ curlPostSetup(CURL *handle, HEADER *header, uchar* authBuf)
 	if(authBuf != NULL) {
 		curl_easy_setopt(handle, CURLOPT_USERPWD, authBuf);
 		curl_easy_setopt(handle, CURLOPT_PROXYAUTH, CURLAUTH_ANY);
+	}
+	if(!reuseconn) {
+		curl_easy_setopt(handle, CURLOPT_FORBID_REUSE, 1);
+		curl_easy_setopt(handle, CURLOPT_FRESH_CONNECT, 1);
 	}
 }
 
@@ -1336,7 +1345,7 @@ curlSetup(wrkrInstanceData_t *pWrkrData, instanceData *pData)
 	if (pWrkrData->curlPostHandle == NULL) {
 		return RS_RET_OBJ_CREATION_FAILED;
 	}
-	curlPostSetup(pWrkrData->curlPostHandle, pWrkrData->curlHeader, pData->authBuf);
+	curlPostSetup(pWrkrData->curlPostHandle, pWrkrData->curlHeader, pData->authBuf, pData->reuseconn);
 
 	pWrkrData->curlCheckConnHandle = curl_easy_init();
 	if (pWrkrData->curlCheckConnHandle == NULL) {
@@ -1369,6 +1378,7 @@ setInstParamDefaults(instanceData *pData)
 	pData->useHttps = 0;
 	pData->bulkmode = 0;
 	pData->maxbytes = 104857600; //100 MB Is the default max message size that ships with ElasticSearch
+	pData->reuseconn = 1;
 	pData->allowUnsignedCerts = 0;
 	pData->tplName = NULL;
 	pData->errorFile = NULL;
@@ -1427,6 +1437,8 @@ CODESTARTnewActInst
 			pData->bulkmode = pvals[i].val.d.n;
 		} else if(!strcmp(actpblk.descr[i].name, "maxbytes")) {
 			pData->maxbytes = (size_t) pvals[i].val.d.n;
+		} else if(!strcmp(actpblk.descr[i].name, "reuseconn")) {
+			pData->reuseconn = pvals[i].val.d.n;
 		} else if(!strcmp(actpblk.descr[i].name, "allowunsignedcerts")) {
 			pData->allowUnsignedCerts = pvals[i].val.d.n;
 		} else if(!strcmp(actpblk.descr[i].name, "timeout")) {


davidelang:
rather than disabling all connection reuse, can you implement something like the 
rebindinterval that we hvae on TCP and UDP syslog where instead of creating a 
new connection for every message, it does so only every X messages.

The overhead of establishing a new connection all the time can be quite nasty, 
but doing so every 100 or every 1000 messages is down in the noise and still 
works quite well for load balancing purposes.

iavael:
@davidelang 

> Even a connection per batch could be significant.

But hasn't you just proposed this in your previous message quoted down here?

> but doing so every 100 or every 1000 messages is down in the noise and still works quite well for load balancing purposes.

P.S. JFYI recommended batch size for elastic is between 1000 and 5000 messages

davidelang:
On Wed, 15 Mar 2017, Iavael wrote:
> @davidelang
>
>> Even a connection per batch could be significant.
>
> But hasn't you just proposed this in your previous message quoted down here?

a batch in rsyslog processing can be as small as a single message. The 
configured batch size is the max size allowed, rsyslog will not delay messages 
to fill a batch, it sends all pending messages up to the max batch size.

So it's not quite the same thing as what ES is thinking of when they talk about 
batches and assume that the sender will delay messages to fill up a batch.

>> but doing so every 100 or every 1000 messages is down in the noise and still works quite well for load balancing purposes.
>
> P.S. JFYI recommended batch size for elastic is between 1000 and 5000 messages
>
>

iavael:
@davidelang 
Then it's also needed to add cutoff timer to provide maximum delay of messages delivery to elastic, if log stream is weak and it's needed to get logs in elastic not less than in determined amount of time.

davidelang:
On Wed, 15 Mar 2017, Iavael wrote:

> @davidelang
> Then it's also needed to add cutoff timer to provide maximum delay of messages delivery to elastic, if log stream is weak and it's needed to get logs in elastic not less than in determined amount of time.

If you have a batch system that delays messages, then yes you do need to setup a 
timeout to keep messages from being delayed 'too long' (however that is defined 
in your case)

but the rsyslog batch system is setup to minimize the latency in getting 
messages delivered, even at the cost of using more (otherwise idle) resources. 
If there is a low volume of traffic and no resource contention, messages will 
get delivered immediatly and individually. As resource contention happens (which 
could be network contention or the recipient having a high per-batch overhead), 
rsyslog will start sending larger batches.

ES is not a real-time system, rsyslog is (or as near to it as it can afford to 
be given resource contraints)

There has been talk of having omelasticsearch implement it's own queue/batch 
mechanism so that it can force larger batches at the cost of delaying delivery, 
but afaik, that has not yet been implemented.

David Lang

rgerhards:
@davidelang  do you violently object merging this change? I think it is good to have albeit the performance may not be as good as possible. This should probably be mentioned in doc.

davidelang:
I would absolutly support this if it was on a counter (like the retryinterval we 
have for other protocols)

I'm pretty sure that if we add this version, we are going to end up adding 
retryinterval later when people find this performance crippling, adn I really 
hate to see us supporting two options long term (we have enough of this sort of 
cruft around as is, I hate to see us add more). It's not just the cruft in the 
code, but the confusion in the docs and explaining which option to use.

the concept is useful, it's not an "over my dead body" objection. let me think 
about it and look at how hard it would be to change it to retryinterval to match 
the other outputs.

retryinterval=1 would match the current behavior

taotetek:
As a heavy user of omelasticsearch, I would strongly prefer the default
behavior of the module not change. If it does change, I'd like this to be
called out very loudly in the release notes. Otherwise I'm going to forget
about this email thread and shoot myself in the foot  during and upgrade,
haha.

Cheers,
Brian

davidelang:
by "match the current behavior", I was meaning match what this patch does. In 
any case the behavior if this parameter is not set would not change.

taotetek:
Thanks for the clarification!

Brian

rgerhards:
I added the work-in-progress label as there is no consense  to merge as-is. So we need someone to implement the counter functionality @davidelang requested.

rgerhards:
tenatively set milestone to 8.31, but it really depends on magnitude of work (I think low) and time available.

bayaro:
As mentioned "rsyslogs doesn't connect to them for some time". As I understand rsyslog returns to woke up node. But @iavael wants to decrease the wait for wakeup time? But this time is regulated by system. This is a unused connection system timeout .
If anybody has so high loaded logging machine, and cluster in which nodes are jumping from in to out and vise versa, that he needs rebalance ES cluster by rsyslog. Imho it should be done by decreasing unused connection timeout system value. But not with additional complexity of rsyslog.

bayaro:
I can imagine how many connections will be open on really loaded machine with such feature.....
Does anybody really need it?

iavael:
@bayaro 
> As I understand rsyslog returns to woke up node.

No, rsyslogs continue to reuse already established connections to other nodes behind LB. And it won't establish connections to fresh nodes until it somehow loses it's tcp session.

> But @iavael wants to decrease the wait for wakeup time?

I want an option for rsyslog not to stick to elastic nodes and establish new connection to random node for every messages batch. So that in configuration with load balancer a load will actually get balanced

> But this time is regulated by system.

It's a feature of libcurl.

> But not with additional complexity of rsyslog.

I propose just a knob to disable an feature in libcurl.

> I can imagine how many connections will be open on really loaded machine with such feature.....

This change should not increase simultaneous connections count.

bayaro:
@iavael 
> In some setups several elastic cluster nodes are located behind HTTP load balancers. In case when some of these nodes go down, connections from rsyslog daemons group on healthy nodes and after failed nodes go up, rsyslogs doesn't connect to them for some time.
> No, rsyslog continues to reuse already established connections to other nodes behind LB. And it won't establish connections to fresh nodes until it somehow loses it's tcp session.

Could you please describe in more details. What is the problem?

long time before switching to working node of cluster (if current gone away)
OR
rsyslog does not switch between nodes so often as you want?

iavael:
@bayaro 

> rsyslog does not switch between nodes so often as you want?

It doesn't switch at all until tcp connection somehow tears (network problem, elastic node goes down, LB failure, rsyslog restart). It's not a problem when everythin is OK (law of big numbers is on our side). But if some of elastic nodes are restarted (so that all rsyslogs would reestablish their connections with other nodes) or added, they would be underloaded for significant amount of time.

Simply, long sessions are hard to be balanced evenly, because they tend to stick to elastic nodes. This options make these sessions short living so that load balancer can do it's job well.

iavael:
This problem can be solved by restarting all rsyslogs after any network issue, but that's kinda kludgy.

davidelang:
This is why we have the rebindinterval option on many outputs. It sends that 
many messages and then disconnects and reconnects so that load balancers can 
have a chance to do their job.

It sounds like what you need is for rebindinterval to work on ES output

Personally, I would probably send to the local machine via syslog protocol, and 
then have a copy of rsyslog on the ES box deliver it via localhost to ES on that 
box

iavael:
@davidelang 

> It sounds like what you need is for rebindinterval to work on ES output

Yes, something like that, but with taking in accont that libcurl reuses connections from it's internal connection pool and option in this PR disables such behaviour.

> Personally, I would probably send to the local machine via syslog protocol, and
then have a copy of rsyslog on the ES box deliver it via localhost to ES on that
box

This can be a solution, but it's not always applicable or optimal: no authentication/authorization, requirement for specialized network service with listened port in multipurpose ES infrastructure etc.

bayaro:
how often do you restart or add nodes?
once time per year? and you want to add additional open/close connection for every logline/bulk?

do you restart/add/remove cluster nodes automatically? imho, it is always non-usual issue. That's why the script or any other approach can be used (called manually) for
> restarting all rsyslogs after any network issue, but that's kinda kludgy.
and I don't think that it will be a kludge.

if you have dozens of nodes and clusters then `rebindinterval` should be used (as @davidelang wrote). Because of with your suggestion the huge portion of delivery time will be spending for an opening connections instead of transferring data. Please read carefully https://curl.haxx.se/libcurl/c/CURLOPT_FORBID_REUSE.html
But curl does not have api for flashing a connection pool. So maybe curl session maybe recreated, or we can manipulate with CURLOPT_MAXCONNECTS, or something other.

Other solution is to use array of ES servers for your endpoint (put node names in this array, and use aws route53 for auto cnaming). So rsyslog send data to them one by one. Look at `server` parameter http://www.rsyslog.com/doc/v8-stable/configuration/modules/omelasticsearch.html

Anyway "opening/closing connection for every logline/bulk" is **incredibly expensive**.

rgerhards:
Let me jump in:

* ```rebindinterval``` is **not** available for omelasticsearch -- in essence, I think this is what @iavael actually would like to have
* the current PR is not merged because it does disable reuse globally. Thus it is not merged. It should be upgraded to do it only every ```n```batches, what actually is the ```rebindinterval``` equivalent. I was thinking of doing this for the next milestone, but it is not clear if I can mange so. Currently does not look to good.

bayaro:
how are you going to clear curl connections pool?

rgerhards:
> how are you going to clear curl connections pool?

good question -- we should get an answer to it before we continue ;-)

iavael:
@bayaro 

> how often do you restart or add nodes?

I disabled this libcurl feature when I used to work in one company. There were 3 datacenters with 14-15 elastic nodes per DC in different racks (more racks = higher probability to meet TOR-switch failure). Also there were regular planned DC network outages every couple of weeks. So local network outage itself wasn't something uncommon.
Also ES bugs (e.g. deadlocked batch-processing threads) added some entropy.

> Because of with your suggestion the huge portion of delivery time will be spending for an opening connections instead of transferring data

Opening new connection is relatively nothing compared with sending hundreds of messages 1-2KB each in a batch. Especially in local datacenter network. At least it's not that expensive to sacrifice load distribution.

> and I don't think that it will be a kludge.

If applications send logs in local rsyslog via udp (to avoid blocking), then restart of rsyslog will lead to loss of some logs. Also it's quite inconviniet if there are dozens of rsyslogs.
Using 2 layers of rsyslogs (local + aggregating) with tcp syslog protocol between them isn't always an option.

iavael:
@rgerhards 
then maybe force rebind with setting CURLOPT_FRESH_CONNECT option on timer?

rgerhards:
> then maybe force rebind with setting CURLOPT_FRESH_CONNECT option on timer?

I think a message (or batch)-counting feature is good, we do not need to get into the complexities of timers. But we must be sure that libcurl properly cleans up the old connections. I do not know how to do that. I can learn, of course, and it probabyl is quick. But look at my TODO. So this is a blocker for me. Actually, TBH I just wanted to finally get rid of the PR by implementing the missing bits, but personally I have zero need nor interest for this feature. So it would probably be better if someone who has finishes the work ;-)

davidelang:
On Thu, 4 Jan 2018, Yaroslav Bo wrote:

> how often do you restart or add nodes?
> once time per year? and you want to add additional open/close connection for every logline/bulk?

it's also a load balancing issue, you want to load balance traffic across 
systems, but the load balancing algorithm only has a chance to work when the TCP 
connection is established. If you reconnect for every message (or every batch, 
or better yet every N batches), then the load balancing works pretty well

> do you restart/add/remove cluster nodes automatically? imho, it is always non-usual issue. That's why the script or any other approach can be used (called manually) for
>> restarting all rsyslogs after any network issue, but that's kinda kludgy.
> and I don't think that it will be a kludge.

having to restart every system sending logs when the destination cluster has a 
problem is a bad approach


> if you have dozens of nodes and clusters then `rebindinterval` should be used 
> (as @davidelang wrote). Because of with your suggestion the huge portion of 
> delivery time will be spending for an opening connections instead of 
> transferring data. Please read carefully 
> https://curl.haxx.se/libcurl/c/CURLOPT_FORBID_REUSE.html

> But curl does not have api for flashing a connection pool. So maybe curl 
> session maybe recreated, or we can manipulate with CURLOPT_MAXCONNECTS, or 
> something other.
>
> Other solution is to use array of ES servers for your endpoint (put node names 
> in this array, and use aws route53 for auto cnaming). So rsyslog send data to 
> them one by one. Look at `server` parameter 
> http://www.rsyslog.com/doc/v8-stable/configuration/modules/omelasticsearch.html

not everyone runs inside of AWS, and in any case, it's good to have a high 
volume server move between destination nodes on a regular basis instead of being 
stuck to one system forever

> Anyway "opening/closing connection for every logline/bulk" is **incredibly expensive**.

no disagreement there.

davidelang:
worst case, kill and restart curl (or stop/reinitialize the lib)

rgerhards:
> worst case, kill and restart curl (or stop/reinitialize the lib)

we cannot do this. There are other parts of code inside rsyslog that use libcurl. Whatever is done, I think it must be clean from a libcurl PoV.

bayaro:
@davidelang it does not matter, what you are using (aws was an example).
Do you need LB? Please read carefully section about option `server` http://www.rsyslog.com/doc/v8-stable/configuration/modules/omelasticsearch.html
And this is a clear & simple solution (round-robin, pool cache, all is working). A well-adjusted resolve will help you.

Anybody don't need to restart all servers, just wait a usual timeouts. But if somebody wants to switch to new node more quickly he needs to restart all of them. Of course the server array also can be used for (and it is preferable solution, imho).

Suggestion to reopen connection for every transfer is a rude kludge. I want to repeat: - such technic can be used only for debugging.

rgerhards:
@radu-gheorghe I would appreciate your thought on this issue

iavael:
@rgerhards 

To sum everything up: it should be ok to reestablish connection every `<rebindinterval>` messages.
Then I gonna modifiy this PR to set CURLOPT_FRESH_CONNECT and CURLOPT_FORBID_REUSE not for every connection, but if messages counter of worker instance is greater than rebindinterval (if rebindinterval > 0). Then reset counter and unset curl opts.

rgerhards:
@iavael Thanks for working towards ability to merge this PR, much appreciated. Your descriptions sounds good. The ```rebindinterval``` should be a configurable parameter with a default of 0 (disabled).

I know there are different views if this capability is useful or not. I think it is simple to implement and thus easy to maintain and if there is some need for it, then it is useful to have it. After all, an explicit opt-in is required to activate it. We should clearly document both the use case and the concerns, the related doc update should mention that. Please make sure you understand the curl options being used. We must avoid to accidentally:
* keep connections hanging
* spawn connections for each request

And if we really see issues with all that, we can always disable the code again (but I honestly don't see why this should be needed given the opt-in nature).

All: does that sound acceptable? Or is there hard objection?

davidelang:
On Fri, 5 Jan 2018, Yaroslav Bo wrote:

> 
> @davidelang it does not matter, what you are using (aws was an example).
> Do you need LB? Please read carefully section about option `server` http://www.rsyslog.com/doc/v8-stable/configuration/modules/omelasticsearch.html
> And this is a clear & simple solution (round-robin, pool cache, all is working). A well-adjusted resolve will help you.

round-robin doesn't help if the connections last for days and different sending 
systems have vastly different volumes.

In this situation you have a very real risk that several high-volume senders 
will happen to hit one ES server and overload that server. To prevent this, you 
need to have something that forces the senders to disconnect and reconnect 
periodically so that round-robin (or any other load balancing option) has a 
chance to make a decision.

It's not good to do this on the receiving side (cutting a TCP connection 
unexpectedly forces the sender to wait for timeouts, wasting bandwidth, and in 
worst cases duplicating or loosing messages depending on the protocol specifics)

> Anybody don't need to restart all servers, just wait a usual timeouts. But if 
> somebody wants to switch to new node more quickly he needs to restart all of 
> them. Of course the server array also can be used for (and it is preferable 
> solution, imho).

I don't think the elasticsearch module that was contributed to rsyslog supports 
the server array mode of operation

> Suggestion to reopen connection for every transfer is a rude kludge. I want to repeat: - such technic can be used only for debugging.

Sometimes it's appropriate, most of the time it's not, which is why it's a 
config option, not a change to the default.

bayaro:
@davidelang  `server` http://www.rsyslog.com/doc/v8-stable/configuration/modules/omelasticsearch.html

bayaro:
@davidelang 
> I don't think the elasticsearch module that was contributed to rsyslog supports
the server array mode of operation
Please read carefully section about option `server` http://www.rsyslog.com/doc/v8-stable/configuration/modules/omelasticsearch.html
Also you can look at code https://github.com/rsyslog/rsyslog/blob/master/plugins/omelasticsearch/omelasticsearch.c#L380
For what do you think this function?

rgerhards:
closing as there seems to be no more interest

PR #1472, opened 2017-03-17
Honor imrelp input name if provided when submitting to impstats

### LINKED ISSUE #1470: impstats module not logging correct name for imrelp inputs
Running rsyslog 8.18 on CentOS 6.8. 

I have the following in my rsyslog config:

```
module(load="impstats"
    interval="60"
    resetCounters="on"
    format="json-elasticsearch"
    ruleset="stats"
)

*other stuff*
input(type="imrelp" port="20514" name="jhenry2-imrelp")
input(type="imudp" port="514" name="jhenry2-imudp")
```
When I look at what is being emitted to ElasticSearch, I see inputs named like jhenry2-imudp(*:514). However, the RELP input gets logged as simply "imrelp[20514]". Seems like imrelp is not supplying the correct input name to impstats. And also formatting the port number differently, which is mildly annoying but not a deal breaker.

I am not at all familiar with the rsyslog codebase, but this line seems to be responsible? https://github.com/rsyslog/rsyslog/blob/master/plugins/imrelp/imrelp.c#L346. The imudp module for example seems to take care to set the metric name to the name of the input. https://github.com/rsyslog/rsyslog/blob/master/plugins/imudp/imudp.c#L336

The reason I care about this is I have servers writing to multiple remote rsyslog hosts using RELP. They all listen on the same port number, so the metric is logged with the same name regardless of which server handled it. I would like to monitor that they're all performing similarly. But if all of them log with the same input  name, this is difficult.

jhenry82:
Fixes #1470

### DIFF
--- a/plugins/imrelp/imrelp.c
+++ b/plugins/imrelp/imrelp.c
@@ -343,8 +343,8 @@ addListner(modConfData_t __attribute__((unused)) *modConf, instanceConf_t *inst)
 	CHKiRet(prop.ConstructFinalize(inst->pInputName));
 	/* support statistics gathering */
 	CHKiRet(statsobj.Construct(&(inst->data.stats)));
-	snprintf((char*)statname, sizeof(statname), "imrelp[%s]",
-		 inst->pszBindPort);
+	snprintf((char*)statname, sizeof(statname), "%s(%s)",
+		 inst->pszInputName, inst->pszBindPort);
 	statname[sizeof(statname)-1] = '\0'; /* just to be on the save side... */
 	CHKiRet(statsobj.SetName(inst->data.stats, statname));
 	CHKiRet(statsobj.SetOrigin(inst->data.stats, (uchar*)"imrelp"));


jhenry82:
I see there's a test failure, but it seems unrelated to my change (Ubuntu cannot find some package). 

```
if [ "x$ESTEST" == "xYES" ]; then sudo apt-get install -qq elasticsearch ; fi
if [ "$DISTRIB_CODENAME" == "trusty" ]; then sudo apt-get install -qq libhiredis-dev; export HIREDIS_OPT="--enable-omhiredis"; fi
if [ "$DISTRIB_CODENAME" == "trusty" ]; then sudo apt-get install -qq libsystemd-journal-dev; export JOURNAL_OPT="--enable-imjournal --enable-omjournal"; fi
sudo apt-get install -qq --force-yes libqpid-proton3-dev
E: Unable to locate package libqpid-proton3-dev
```

rgerhards:
This looks like something upstream has changed yesterday. Checking if I can get hold of the package any other way...

rgerhards:
FYI: https://github.com/rsyslog/rsyslog/pull/1473 addresses this. Took a while to find out what happened...

I'll hopefully merge this PR later today. It would be great if you could rebase your branch when this is done. I'll let you know.

jhenry82:
@rgerhards thanks for fixing the test problem. I've rebased my branch and all tests are now passing.

Issue #1476, opened 2017-03-21
str2num() breakes condition evaluation


jvymazal:
After solving #1412  by #1413 , safeguard added to str2num() causes, in some cases conditional expression to be incorrectly evaluated. Silly minimal example with mmcount:

relevant config statements:
`module(load="imuxsock") `
`module(load="mmcount")`
`action(type="mmcount" appname="fo0")`
`$template tpl,"cnt: %$!mmcount%, msg: %msg%.\n"`
`*.* /var/log/my.log;tpl`
`if $!mmcount <> 0 and $!mmcount % 5 == 0 then /var/log/mmcount.log;tpl`

then run
`for ((i=0; i<10; i++)) ; do`
`        logger -t fo0 "bar[$i]"`
`        logger -t foX "baz[$i]"`
`      done`

after this I would expect all lines to be in `/var/log/my.log` and only the  `cnt: 5` and `cnt: 10` lines from `fo0` source be in `/var/log/mmcount.log`. This is the case with rsyslog  8.24 . However with rsyslog 8.25 (with the aforementioned commit) all "baz" lines pass the condition as well although `foX` source is not registered with mmcount (so the first condition should stop them).

jvymazal:
I do not know how readable is it from the example, the issue itself is that when checking for JSON variable in message with condition like `if $var <> 0 then ...` before #1413 this condition would be false when no `var` was present in massage, however now the condition is true.

Issue #1412, opened 2017-02-13
str2num() function mishandles empty strings


rgerhards:
We noticed this while developing new functionality (so no bug report from practice): If str2num() receives an empty string, misadressing happens. Under extreme conditions, this can lead to a segfault.

PR #1477, opened 2017-03-22
reverted str2num() bSuccess return to be consistent with past behavior

jvymazal:
attempt to solve #1476

### DIFF
--- a/grammar/rainerscript.c
+++ b/grammar/rainerscript.c
@@ -1279,7 +1279,7 @@ str2num(es_str_t *s, int *bSuccess)
 	if(s->lenStr == 0) {
 		DBGPRINTF("rainerscript: str2num: strlen == 0; invalid input (no string)\n");
 		if(bSuccess != NULL) {
-			*bSuccess = 0;
+			*bSuccess = 1;
 		}
 		goto done;
 	}
--- a/tests/rscript_num2ipv4.sh
+++ b/tests/rscript_num2ipv4.sh
@@ -6,6 +6,7 @@
 module(load="../plugins/imtcp/.libs/imtcp")
 input(type="imtcp" port="13514")
 
+set $!ip!v0 = num2ipv4("");
 set $!ip!v1 = num2ipv4("0");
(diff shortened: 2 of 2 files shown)


rgerhards:
I am not sure if this is the right path to take. That fixed an actual error, albeit I see that this fix causes change to existing behavior. Still, I think the new way of doing things is correct... Let's see what @davidelang means...

davidelang:
In other languages, sending anything (including a null string) other than a valid number to the str2num() equivalent function returns 0, not -1 as is happening here (looking at the num2ipv4 tests in #1413 )

Since we have allowed $!var == 0 to be the same as $!var == "" in the past, we need to support that going forward (it's arguably wrong, but not insanely wrong, so backwards compatibility should win)

looking at the str2num function in rainerscript.c, we convert strings that contain text by bailing out at the first non-digit value, so str2num("a") = 0, with a failure code, which would be an argument against str2num("") = 0 with a success code.

I feel like I'm missing some background here just looking at the issues and pull requests. what was the misaddressing that could happen that was noticed in #1412 ?

but I would expect str2num("") to return 0 and num2ipv4("0") to return "0.0.0.0" so I don't see a problem with num2ipv4("") returning "0.0.0.0"

so I think the right thing to do here is to make the change listed here. But I would like more info on the potential problems this could cause.

jvymazal:
I would agree with @davidelang , to be consistent I would say that if $!var == 0 is same as $!var == "" it should go like num2ipv4("0") is same as num2ipv4(""), hence the testbench commit.

rgerhards:
I agree with you and @davidelang Just let me check if we need to look at some related code.

On the misadressing: I don't remember all details, but @jgerhards  found out that in some cases it happend and could even lead to a segfault. I think @jgerhards found this as a side-effect of running tests for the ip42num function. It had something to do with empty string (or so). So the fix is actually important. As @jvymazal recommened and @davidelang agreed, we should keep old semantics, but need to make sure the original problem is fixed.

PR #1479, opened 2017-03-24
enforce consistent code style; apply this style

rgerhards:
*This PR is for community review purposes. It must NOT be merged before community consensus is reached!*

**COMMUNITY REVIEW PROCESS FINISHED** So this PR may be merged.

This PR sets a consistent standard for rsyslog project coding style. Note that this requires changes to many files and thus makes diff'ing before and after very hard. We think it is worth that burden, but we should be sure that we do not need to repeat this procedure.

**AS SUCH, PLEASE PROVIDE FEEDBACK**, especially if you do not like something. I will merge this PR April 5th at earliest, so please provide comments until then. Of course, if there is disagreement, we will merge later and go through multiple iterations.

The code style is changed with the clang-format tool, which provides great capabilities and also a more or less self-explanatory control file describing the style. Please have a look at https://github.com/rsyslog/rsyslog/pull/1479/commits/84a78580be3e997d0a830af38cfa6ae0b861e9fb#diff-69a68f7ce1ea3687a38d8e4705e6fa48 to see the settings. In-depth doc for them is available at https://clang.llvm.org/docs/ClangFormatStyleOptions.html

A good feel for the new style can be gained by simply looking at the diff that comes with this PR.

The style used tries to pick up most common formatting rules. At some places (e.g. "if ()" vs. "if()") it departs from traditional rsyslog style, because we assume the traditional style is only used by a minority. Note that the control file specified above also tells about some restrictions of the current style settings and how we want to work around them. Finally, there are still some issues with the handling of some rsyslog macros. They merge with their predecessor lines. We can fix this, but that is labor-intense, so we want to do this only after full agreement has been reached.

Note that once we have fully agreed on the style, we will also add CI tests to enforce the style for all new contributions. As such, it will never again be needed to do a large-scale reformatting (except, of course, if there is a very good reason for a future style change inside the community).

Please post your comments to this PR as I would like to have a single source of reference.


davidelang:
I'm in the camp that doesn't care what the standards are, as long as there are standards :-)

I work on too many different projects that have their own style to be dogmatic about it.

The one thing I would say is about line length, I would make any check on that be a warning, not an error. Far too much of the time, folding the line adds confusion rather than clarity. Warn on long lines and leave it up to your judgement (in accepting a PR) if they will be allowed in that case

rgerhards:
on line length: if you do not enforce it, you'll get the warning ever and ever again, this gets unmanagable. And I would really like to have this as I had PRs with lines over 300 chars long, which are really unreadable. We currently set the limit to 120 chars, which really should pose no problems at all. Also, we enforce this limit since around 4 month (via a different method) and only had a single case or two where it really was violated.

mostolog:
If I understood properly the clang-format, here are my two cents:
- **IndentWidth: 4** Shorter lines, fewer "long lines" viewing problems
- **BreakBeforeBraces: Attach** One style, fewer problems
- **AllowShortCaseLabelsOnASingleLine: false** As AllowShortIfStatementsOnASingleLine is false.
- **IndentCaseLabels: true** Personal preference...
- **SpaceBeforeAssignmentOperators: false** Personal pref.
- **SpaceBeforeParens: Never** As _BreakBeforeBraces_: one style, fewer problems

rgerhards:
> IndentWidth: 4 Shorter lines, fewer "long lines" viewing problems

This cannot be set if we use tabs. I **very strongly** favor tabs (as does the Linux kernel team) because with tabs, everyone can set his or her own indention width preference. Also, going away from tabs would require an almost 100% reformatting of the code.

> AllowShortCaseLabelsOnASingleLine: false As AllowShortIfStatementsOnASingleLine is false.
If you feel strong on this, we can do it, but IMHO it makes things less readable. e.g. 
```
case 10:    do_something()
``` 
will become
```
case 10: 
             do_something()
``` 
Now envision this for 20 of such small cases. The same is true for if, but with if() the text is usualy much longer.

> IndentCaseLabels: true Personal preference...
I really don't like this as it makes for very long lines quickly.

> BreakBeforeBraces: Attach One style, fewer problems
> SpaceBeforeAssignmentOperators: false Personal pref.
> SpaceBeforeParens: Never As BreakBeforeBraces: one style, fewer problems

I don't like these three, as (for me) they make reading the code pretty hard. I often rely on the spaces as visual cues to where things begin and end. I know this is debatable, but I would like to see at least one more vote pro this behaviour to change it. 

Just FYI, I think about cases like this here:

```
if((c=getchar())!=NULL)&&(c!='\n')){
```
It's a complex statement without any spaces inside it and I admit such beasts always require me to think much more about what's going on compared to when it is written like this:
```
if( (c=getchar()) !=NULL) && (c!='\n') ) {
```
At least I use the spaces here as guiding how to read the sentence. Itisthesameproblemwhyinnaturallanguageswehavespacesbecauseitmakesthemeasiertoread (the Romans didn't have them).

There are many places inside the code which will be affected.

davidelang:
On Fri, 31 Mar 2017, Rainer Gerhards wrote:

>> IndentWidth: 4 Shorter lines, fewer "long lines" viewing problems
>
> This cannot be set if we use tabs. I **very strongly** favor tabs (as does the Linux kernel team) because with tabs, everyone can set his or her own indention width preference. Also, going away from tabs would require an almost 100% reformatting of the code.

I thought the style defined 8 space indentation, not tabs. I would agree with 
tabs, but if we're doing spaces, I like smaller indentations as well.

>> AllowShortCaseLabelsOnASingleLine: false As AllowShortIfStatementsOnASingleLine is false.
> If you feel strong on this, we can do it, but IMHO it makes things less readable. e.g.
> ```
> case 10:    do_something()
> ```
> will become
> ```
> case 10:
>             do_something()
> ```
> Now envision this for 20 of such small cases. The same is true for if, but with if() the text is usualy much longer.

no strong opinion

>> IndentCaseLabels: true Personal preference...
> I really don't like this as it makes for very long lines quickly.
>
>> BreakBeforeBraces: Attach One style, fewer problems
>> SpaceBeforeAssignmentOperators: false Personal pref.
>> SpaceBeforeParens: Never As BreakBeforeBraces: one style, fewer problems
>
> I don't like these three, as (for me) they make reading the code pretty hard. I often rely on the spaces as visual cues to where things begin and end. I know this is debatable, but I would like to see at least one more vote pro this behaviour to change it.
>
> Just FYI, I think about cases like this here:
>
> ```
> if((c=getchar())!=NULL)&&(c!='\n')){
> ```
> It's a complex statement without any spaces inside it and I admit such beasts always require me to think much more about what's going on compared to when it is written like this:
> ```
> if( (c=getchar()) !=NULL) && (c!='\n') ) {
> ```
> At least I use the spaces here as guiding how to read the sentence. Itisthesameproblemwhyinnaturallanguageswehavespacesbecauseitmakesthemeasiertoread (the Romans didn't have them).
>
> There are many places inside the code which will be affected.

I agree the spaces make it more readable

mostolog:
> IndentWidth: 4 Shorter lines, fewer "long lines" viewing problems

Seems we all agree about tabs, why have you set it to 8 ?
IMHO tab = 4 spaces. I would set it on my editor, but suggest to change that to 4 if needed.

AllowShortIfStatementsOnASingleLine is recommended to be false in order to avoid human mistakes like https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/
Same may apply to AllowShortCaseLabelsOnASingleLine

IndentCaseLabels: false then ;)

Let's see if anyone else comments on BreakBeforeBraces, SpaceBeforeAssignmentOperators,SpaceBeforeParens

> the Romans didn't have them
+1 for historical reference

Glad to see we agree on most things.

rgerhards:
> Seems we all agree about tabs, why have you set IndentWidth: 8 then?
> IMHO tab = 4 spaces. I would set it on my editor, but suggest to change that to 4 if needed.

I have re-checked why I had done this and tried removinb IndentWidth. Doesn't work, there now is a new comment:
```
# Note: we use TAB, but we need to set IndenWidth to 8 because
# this is the clang-format assumed default tab width. If we do
# not set IndentWidth to 8, no tabs will be used in most case
# and instead everything is indented by 2 spaces. Our settings
# do NOT interfer with the user's tab expansion preferrencies
# as set in her or his editor.
```
> 
> AllowShortIfStatementsOnASingleLine is recommended to be false in order to avoid human mistakes like https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/
> Same may apply to AllowShortCaseLabelsOnASingleLine

I also checked this once again. The affected body of code is very small, and my argument seems to be pretty weak looking at the existing code. With ```false``` everything looks better readable. So I have set this for case as well.

Updated PR is pushed.

rgerhards:
ALL: last call, if no more objections, I'll apply the style changes as soon as this makes sense (tring to cut down the number of open PRs first, as they will fail with the style change).

mostolog:
LGTM.
I guess _SortIncludes: true_ it's the only non-critical-at-all objection I could make.
Just let me know if you need help formatting (regex to the rescue!)

rgerhards:
> I guess SortIncludes: true it's the only non-critical-at-all objection I could make.

You mean you would like to see it set to true (because it defaults to false, so it currently is not active)? I had this initially, but it results in many more changes, which I do not consider worth the effort. No strong opinion on my side either, but I would prefer not to sort them

mostolog:
Me neither. ;)

rgerhards:
So I keep it as is until further comments...

rgerhards:
In the final stages of applying this, I notice that clang-format sometimes generates too-long lines. I now need to prevent this before being able to merge.

rgerhards:
another round of manual review found more issues with macros :-(

rgerhards:
This did not work out; better leave current state. Trying to change does more harm than it does good.

Issue #1491, opened 2017-03-31
main_queue directives RainerScript syntax


jvymazal:
While researching (undocumented) new syntax for main queue options, I found out that config validator ignores whether config directives in main_queue() block are actually valid. For example putting 
`main_queue(totalNonsense="whatever")`
into config file produces no error whatsoever.

jvymazal:
Are those directives supposed to work @rgerhards ? Thanks for info.

rgerhards:
they are

PR #1492, opened 2017-03-31
fixed MainqObj destruction

jvymazal:
not a solution but connected to #1491, and is nice to fix

### DIFF
--- a/runtime/glbl.c
+++ b/runtime/glbl.c
@@ -1021,9 +1021,9 @@ glblDestructMainqCnfObj(void)
 	/* Only destruct if not NULL! */
 	if (mainqCnfObj != NULL) {
 		nvlstChkUnused(mainqCnfObj->nvlst);
+		cnfobjDestruct(mainqCnfObj);
+		mainqCnfObj = NULL;
 	}
-	cnfobjDestruct(mainqCnfObj);
-	mainqCnfObj = NULL;
 }
 
 /* comparison function for qsort() and string array compare


rgerhards:
This is just fixing the indention, right? Or am overlooking something?

jvymazal:
Actually bit more than indentation, if you look closely on that diff you will see that the DestroyObj() and NULL assignment lines are now _inside_ the if(NULL) block, previously the were executed always. I did not do some extensive checking as to what problems could it cause, however according to both logic and present comment (in code) I would suggest we fix it.

runtime/glbl.c
```diff
@@ -1021,9 +1021,9 @@ glblDestructMainqCnfObj(void)
 	/* Only destruct if not NULL! */
 	if (mainqCnfObj != NULL) {
 		nvlstChkUnused(mainqCnfObj->nvlst);
-	}
-	cnfobjDestruct(mainqCnfObj);
-	mainqCnfObj = NULL;
+        cnfobjDestruct(mainqCnfObj);
```
rgerhards: please fix indention (note that we use tabs for indention)

rgerhards:
Ah, you are right, I overlooked this. Sorry for that. And, yes, I think this will make things work like intended.

rgerhards:
Can you please squash the second commit (it's just a fix of the first) and update the first commit message to be a bit more up to the point. I tried to do this, but I do not have permissions to push my result back to the PR. Thanks.

PR #1502, opened 2017-04-05
Add IPv4 UDP multicast support to omfwd and imudp plugins.

blakebrain:
These changes add IPv4 UDP multicast support to the omfwd and imudp plugins.

On the imudp listener there is now:
  - $UDPServerMulticastInterface specifies the IGMP join interface by IP or interface name.
    - $UDPServerMulticastInterface 5.6.7.8
    - $UDPServerMulticastInterface eth4000
    - In the new config syntax it is: udp.multicastinterface
  - The multicast destination group is specified in the existing '$UDPServerAddress' option just as a specific unicast interface address would be.

On the omfwd actions (the in-built legacy '@9.6.7.8' remote destination directive):
  - The multicast group to public to is specified just like a unicast UDP destination: @239.1.1.4
  - $ActionSendUDPMulticastTtl specifies the IP TTL to use on UDP sends to multicast groups. By default this is 1.
    - In the new config syntax it is: udp.multicastttl

### DIFF
--- a/plugins/imudp/imudp.c
+++ b/plugins/imudp/imudp.c
@@ -101,6 +101,7 @@ static time_t ttLastDiscard = 0;	/* timestamp when a message from a non-permitte
 static struct configSettings_s {
 	uchar *pszBindAddr;		/* IP to bind socket to */
 	char  *pszBindDevice;		/* Device to bind socket to */
+	uchar *pszBindIntf;		/* Interface to join multicast socket to */
 	uchar *pszSchedPolicy;		/* scheduling policy string */
 	uchar *pszBindRuleset;		/* name of Ruleset to bind to */
 	int iSchedPrio;			/* scheduling priority */
@@ -110,6 +111,7 @@ static struct configSettings_s {
 struct instanceConf_s {
 	uchar *pszBindAddr;		/* IP to bind socket to */
 	char  *pszBindDevice;		/* Device to bind socket to */
+	uchar *pszBindIntf;		/* Interface to join multicast socket to */
 	uchar *pszBindPort;		/* Port to bind socket to */
 	uchar *pszBindRuleset;		/* name of ruleset to bind to */
 	uchar *inputname;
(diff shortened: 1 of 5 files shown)


rgerhards:
would you like to fix the signedness problem? A rebase would also be helpful. 
Also, it would be good if you could update the doc at rsyslog-doc.

plugins/imudp/imudp.c
```diff
@@ -1355,6 +1372,8 @@ CODEmodInit_QueryRegCFSLineHdlr
 		addInstance, NULL, STD_LOADABLE_MODULE_ID));
 	CHKiRet(omsdRegCFSLineHdlr((uchar *)"udpserveraddress", 0, eCmdHdlrGetWord,
 		NULL, &cs.pszBindAddr, STD_LOADABLE_MODULE_ID));
+	CHKiRet(omsdRegCFSLineHdlr((uchar *)"udpservermulticastinterface", 0, eCmdHdlrGetWord,
```
rgerhards, line 1375: Please remove the old-style config directives. We do not want to add any more of them.

tools/omfwd.c
```diff
@@ -1484,6 +1494,8 @@ CODEmodInit_QueryRegCFSLineHdlr
 		NULL, &cs.iTCPRebindInterval, NULL));
 	CHKiRet(regCfSysLineHdlr((uchar *)"actionsendudprebindinterval", 0, eCmdHdlrInt,
 		NULL, &cs.iUDPRebindInterval, NULL));
+	CHKiRet(regCfSysLineHdlr((uchar *)"actionsendudpmulticastttl", 0, eCmdHdlrInt,
```
rgerhards, line 1497: Please remove the old-style config directives. We do not want to add any more of them.

rgerhards:
@blakebrain are you still interested in getting a final mergable version of this PR?

blakebrain:
Yes.  We have an environment that still utilizes the legacy configuration style, which is why those were included.  I believe work halted on fixing the build errors when I needed to come up with a Solaris host.

rgerhards:
> Yes. We have an environment that still utilizes the legacy configuration style, which is why those were included. 

We as a project can unfortunatley not support this. But I can delete them after merge if you prefer.

Is there a hard requirement to not use new style? You can mix and match both, so there should not be any need to add new legacy conf statement.

> I believe work halted on fixing the build errors when I needed to come up with a Solaris host.

You can ask the folks at https://www.opencsw.org/ for access to their build farm. They are usually very supportive and the environment has all the dev tools already set up. Alternatively, you can use the testbench runs to do some debugging. I often use the latter approach.

rgerhards:
@blakebrain are you still interested in getting a final mergable version of this PR?

rgerhards:
assigning the 8.33 milestone. If we can't merge it by then, I think the best thing to do is close it unmerged.

PR #1510, opened 2017-04-11
replaced gethostbyname() with getaddrinfo() call

jvymazal:
Gethostbyname() is generally considered obsolete, is not reentrant and cannot really work with IPv6. This is only place in rsyslog where this call remained.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,6 @@
 ------------------------------------------------------------------------------
+Version 8.27.0 [v8-stable] 2017-05-16
+------------------------------------------------------------------------------
 Version 8.26.0 [v8-stable] 2017-04-04
 - NEW BUILD REQUIREMENT: liblognorm 2.0.3 is required for mmnormalize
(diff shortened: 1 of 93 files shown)


rgerhards:
I would suggest to squash the commits as they seem to only correct the first one. A rebase would make the solaris tests pass (if they don't hurt compilation there).

rgerhards:
something seem to have gone wrong with the merge... Even though it was no rebase, it should not see the commits as separate. Did you rewrite history on the merged commits?

rgerhards:
In other words: which is the commit that you would like to have applied? I'll cherry-pick it and create a new PR for it...

jvymazal:
I tried not to rewrite history, and I agree that something seems off with the merge, the commit is here: https://github.com/jvymazal/rsyslog/commit/9b8f9476466d00a1ed8296c680ea4dd209766d0f

rgerhards:
replaced by #1518

PR #1520, opened 2017-04-18
Rate limit proc name

jvymazal:
This is sort of addition to recently merged changes, first commit addresses real-world use-case finding that hostname and message source are better split by some formatting char (colon seems ok for this purpose) and second commit adds much better source logging for those using imuxsock which does not rely on APPNAME mesage field which is, more often than not, blank. What do you think of this @rgerhards ?

### DIFF
--- a/plugins/imuxsock/imuxsock.c
+++ b/plugins/imuxsock/imuxsock.c
@@ -594,7 +594,7 @@ findRatelimiter(lstn_t *pLstn, struct ucred *cred, ratelimit_t **prl)
 	ratelimit_t *rl = NULL;
 	int r;
 	pid_t *keybuf;
-	char pidbuf[256];
+	char pinfobuf[512];
 	DEFiRet;
 
 	if(cred == NULL)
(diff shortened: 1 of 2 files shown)


plugins/imuxsock/imuxsock.c
```diff
@@ -618,7 +618,7 @@ findRatelimiter(lstn_t *pLstn, struct ucred *cred, ratelimit_t **prl)
 		STATSCOUNTER_INC(ctrNumRatelimiters, mutCtrNumRatelimiters);
 		/* read process name from system  */
 		char procName[256]; /* enough for any sane process name  */
-		sprintf(procName, sizeof(procName), "/proc/%lu/cmdline", (unsigned long) cred->pid);
+		snprintf(procName, sizeof(procName), "/proc/%lu/cmdline", (unsigned long) cred->pid);
```
rgerhards: I would prefer if you could squash this commit into the erroneous one.

rgerhards:
@jvymazal FYI: Unfortunately, I now found a regression in this PR that I did not notice at the time of merge: https://github.com/rsyslog/rsyslog/issues/1827 - please be aware that we need to change the defaults.

rgerhards:
@jvymazal hold on - I might have taken the wrong path. I'll post more as soon as I have definite answer (most probably not before Tuesday or Friday)

rgerhards:
@jvymazal sorry for the noise - my initial analysis was wrong, all is well with the PR

PR #1518, opened 2017-04-18
replaced gethostbyname() with getaddrinfo() call

rgerhards:
replaces #1510 

@jvymazal please subscribe to this tracker; note that some Solaris slaves seem to be out of space (this is a different admin domain, I have notified the owner).

PR #1521, opened 2017-04-18
replaced gethostbyname() with getaddrinfo() call

jvymazal:
replaces #1518

@rgerhards  I cloned it into my repo and will continue on it to find out why it hangs in some environments, I will close the original PR (from your repo)

### DIFF
--- a/runtime/net.c
+++ b/runtime/net.c
@@ -1181,24 +1181,22 @@ getLocalHostname(uchar **ppName)
 	}
 
 	char *dot = strstr(hnbuf, ".");
+	struct addrinfo *res = NULL;
 	if(!empty_hostname && dot == NULL) {
 		/* we need to (try) to find the real name via resolver */
-		struct hostent *hent = gethostbyname((char*)hnbuf);
-		if(hent) {
-			int i = 0;
-			if(hent->h_aliases) {
-				const size_t hnlen = strlen(hnbuf);
-				for(i = 0; hent->h_aliases[i]; i++) {
-					if(!strncmp(hent->h_aliases[i], hnbuf, hnlen)
-					   && hent->h_aliases[i][hnlen] == '.') {
-						break; /* match! */
-					}
-				}
-			}
-			if(hent->h_aliases && hent->h_aliases[i]) {
-				CHKmalloc(fqdn = (uchar*)strdup(hent->h_aliases[i]));
-			} else {
-				CHKmalloc(fqdn = (uchar*)strdup(hent->h_name));
+		struct addrinfo flags;
+		memset(&flags, 0, sizeof(flags));
+		flags.ai_flags = AI_CANONNAME;
+		int error = getaddrinfo((char*)hnbuf, NULL, &flags, &res);
+		if (error != 0) {
+			dbgprintf("getaddrinfo: %s\n", gai_strerror(error));
+			ABORT_FINALIZE(RS_RET_IO_ERROR);
+		}
+		if (res != NULL) {
+			/* When AI_CANONNAME is set first member of res linked-list */
+			/* should contain what we need */
+			if (res->ai_canonname != NULL && res->ai_canonname[0] != '\0') {
+				CHKmalloc(fqdn = (uchar*)strdup(res->ai_canonname));
 			}
 			dot = strstr((char*)fqdn, ".");
 		}
@@ -1215,6 +1213,9 @@ getLocalHostname(uchar **ppName)
 
 	*ppName = fqdn;
 finalize_it:
+	if (res != NULL) {
+		freeaddrinfo(res);
+	}
 	RETiRet;
 }
 


Whissi:
I am not sure if the impact of this change is clear for everyone and I am also not sure if I like this change:

Now that *rsyslogd* uses *getaddrinfo()* the system's DNS lookup must be working **before** *rsyslogd* starts. I.e. in most cases this requires a working network.

I am not sure if I like that change because the *syslogger* is normally a process which should start at the very beginning. Due to this change, *rsyslogd* must be started **after** your network is up or *rsyslog* will use *localhost* as hostname for its lifetime...

davidelang:
I'm pretty sure that both calls go through the system infrastructure that uses 
/etc/nsswitch.conf and by default references /etc/hosts prior to making a call 
to DNS

jvymazal:
I agree with @davidelang , only thing that was done was exchange gethostbyname with getaddrinfo call, rest of code logic stays the same, functionality-wise these two calls are not different, gethostbyname is basically just obsolete version without IPv6 support and with some cosmetic differences.

Whissi:
That's not correct.

I cannot talk about *systemd* systems but before *systemd* (which is still the case for systems running without *systemd*) you had a "hostname" service which set the hostname very early on boot. When set, *gethostbyname()* returned the system's hostname (no entry in */etc/hosts* needed). Like you know I am using Gentoo, but this is also true for Debian which still can be used without *systemd*, see https://anonscm.debian.org/cgit/collab-maint/sysvinit.git/tree/debian/src/initscripts/etc/init.d/hostname.sh#n9).

Used *nsswitch.conf* is similar to https://sources.debian.net/src/base-files/8%2Bdeb8u8/share/nsswitch.conf/

davidelang:
gethostbyname(<what>) returned the hostname?

AFAIK, gethostbyname() has never had any interaction with hostname, just with 
/etc/hosts and whatever else was defined in nsswitch (DNS,ldap,yp,nis,etc)

This has nothing to do with systemd.

Whissi:
> This has nothing to do with systemd.

Right, I just wanted to make sure that I haven't tested on systems using *systemd* yet.

I cannot tell you exactly at the moment what's going on. I'll have to prepare a short PoC to test.
I only know for sure that once we published rsyslog-8.27.0 we are receiving bug reports because *rsyslogd* is now using "localhost" as hostname. To fix that, we had to change start dependencies, i.e. we now have to depend on "dns" which wasn't needed before.

davidelang:
the hostname in rsyslog is not determined by gethostbyname() or getaddinfo(), 
instead it uses gethostname() and only falls back to trying gethostbyname() if 
that fails.

so why is gethostname() failing on your system?

Whissi:
I have no indication that the problem is related to my system and or *gethostname()* (and it isn't just my system, as said, we are receiving bug reports from other users). But I run bisect:

```
git bisect start
# good: [0626157464aa8f07fcdf4a864979ee7d34a2a8cc] Merge branch 'master' into v8-stable
git bisect good 0626157464aa8f07fcdf4a864979ee7d34a2a8cc
# bad: [8533fc1fbaae6a33e537dbb54cbeea9330b68e6e] Merge pull request #1577 from rgerhards/testbench-silence-kafka
git bisect bad 8533fc1fbaae6a33e537dbb54cbeea9330b68e6e
# bad: [55e9ee87fc7069099498fbd2f2157b5948074e9b] Merge pull request #1529 from PascalWithopf/imptcp-discardTruncatedMsg
git bisect bad 55e9ee87fc7069099498fbd2f2157b5948074e9b
# good: [0db021cc3a9149d351fb4de4707c623403160b67] Merge pull request #1513 from jvymazal/man_discard
git bisect good 0db021cc3a9149d351fb4de4707c623403160b67
# good: [a39361ccfa55c6d341cf897c52e3ec74b4909bd5] maintain ChangeLog
git bisect good a39361ccfa55c6d341cf897c52e3ec74b4909bd5
# bad: [4a34ba78c7670293db7d7cbb1a0d0d9a961c772f] Merge pull request #1521 from jvymazal/getaddrinfo_replace
git bisect bad 4a34ba78c7670293db7d7cbb1a0d0d9a961c772f
# good: [6b07d0701a1b2f7d065121938d82009979d430c5] Merge pull request #1523 from PascalWithopf/maxFrameSize
git bisect good 6b07d0701a1b2f7d065121938d82009979d430c5
# good: [84c8e7a4fda71914bfee68aee97a4fa80b8237ed] Merge branch 'master' of https://github.com/rsyslog/rsyslog
git bisect good 84c8e7a4fda71914bfee68aee97a4fa80b8237ed
# bad: [9593cd95a5de92df23b11a9a1526d3709c3030ee] replaced gethostbyname() with getaddrinfo() call
git bisect bad 9593cd95a5de92df23b11a9a1526d3709c3030ee
# first bad commit: [9593cd95a5de92df23b11a9a1526d3709c3030ee] replaced gethostbyname() with getaddrinfo() call
```
This is how I test:

I disabled the service which provides "dns" on my system. In my case I disabled the network interface because I am just using another resolver through my */etc/resolv.conf*.

I build *rsyslog* and restarted *rsyslogd*. When I saw
```
May 23 13:30:15 [localhost] rsyslogd:  [origin software="rsyslogd" swVersion="8.27.0.master" x-pid="67528" x-info="http://www.rsyslog.com"] start
```
in */var/log/messages* I knew there was a problem, when I saw
```
May 23 13:25:36 vm-gentoo-x64 rsyslogd:  [origin software="rsyslogd" swVersion="8.27.0.master" x-pid="30946" x-info="http://www.rsyslog.com"] start
```
instead I knew it was working like all the years before.

davidelang:
ok, what sets the name vm-gentoo-x86 on your system? is that set in /etc/hosts?, 
in DNS, or is that what you get if you type 'hostname' at the prompt

kkkk

Whissi:
On Gentoo OpenRC-based system the value is set in */etc/conf.d/hostname* and will be set by https://github.com/OpenRC/openrc/blob/master/init.d/hostname.in#L35

This is equal to Debian's https://anonscm.debian.org/cgit/collab-maint/sysvinit.git/tree/debian/src/initscripts/etc/init.d/hostname.sh#n9

davidelang:
change the script that starts rsyslog on your system to run hostname and save 
the result to a file somewhere.

Rsyslog only does name resolution to try and find a value for hostname if there 
is no hostname set on the system when it starts.

see the function getLocalHostname in runtime/net.c around line 1160

Whissi:
> change the script that starts rsyslog on your system to run hostname and save the result to a file somewhere.

Sorry, I don't understand. There's nothing wrong with the runscript. When I manually call */usr/sbin/rsyslogd* I get the same results: Before commit 9593cd95a5de92df23b11a9a1526d3709c3030ee *rsyslogd* will use the system's hostname, after the commit, *rsyslogd* will use *localhost* as hostname.

(As said only when name resolution isn't possible, like on system's start or now simulated. That's said, changing rsyslog dependency to start on boot **after** DNS will *solve* the problem. However, I am not sure if a syslogger shouldn't be able to use the system's hostname **without** name resolution. That's why I am not sure if I like that change).

jvymazal:
I do not know how your `/etc/nsswitch.conf` looks, however according to all docs and info I could find `getaddrinfo` should behave same as `gethostbyname` in this regard (e.g. follow configured priorities in resolution). If it is not my first guess would be that problem may lie outside of rsyslog completely.

Whissi:
Like said, my  nsswitch.conf isn't special.
```
# cat /etc/nsswitch.conf
# /etc/nsswitch.conf:
# $Header: /var/cvsroot/gentoo/src/patchsets/glibc/extra/etc/nsswitch.conf,v 1.1 2006/09/29 23:52:23 vapier Exp $

passwd:      compat
shadow:      compat
group:       compat

# passwd:    db files nis
# shadow:    db files nis
# group:     db files nis

hosts:       files dns
networks:    files dns

services:    db files
protocols:   db files
rpc:         db files
ethers:      db files
netmasks:    files
netgroup:    files
bootparams:  files

automount:   files
aliases:     files
```
On Debian it is https://sources.debian.net/src/base-files/8%2Bdeb8u8/share/nsswitch.conf/

davidelang:
I'm trying to figure out if your system has always been falling back to a name 
lookup because hostname has not been set yet, or if there is some other behavior 
change that happened here

davidelang:
@jvymazal, you changed more in this file than just the function call, it's 
possible that you broke something if a hostname was already set

Please test (with no entries in /etc/host and no DNS set) if a simple hostname 
like "myname" works, or if it insists on trying to find the FQDN

also check if a hostname of "myname.mydomain" works or not.

rgerhards:
I haven't followed very closely, and have not in-depth checked the API doc, but still some possibly useful info. As @davidelang says, the new code is not called if gethostname() succeeds ... almost ;-) It is also called when the hostname returned from gethostname() does not contain a dot. Have a look at the sequence here:

https://github.com/rsyslog/rsyslog/pull/1521/commits/9593cd95a5de92df23b11a9a1526d3709c3030ee#diff-3939442df366b522ac130e847343d773R1169

So for my understanding it looks like a hostname is returned, but it does not contain a dot. But that unfortunately does not explain the core issue...

davidelang:
requiring a dot in a hostname seems like a new requirement, was it there before?

I know there are times when you want a fqdn, but if you are given a simple 
hostname, you should use that rather than jumping though hoops to try and guess 
what the fqdn is going to be.

jvymazal:
The condition before resolution function was there, you can see in a diff that the line was unchanged. I tried to make the change as small as possible.

rgerhards:
Pretty old.

Sent from phone, thus brief.

Am 23.05.2017 15:29 schrieb "David Lang" <notifications@github.com>:

Whissi:
I was able to reproduce the problem on a default Debian Jessie system (with *systemd*).

```
# cat /etc/rsyslog.conf
$ModLoad imuxsock # provides support for local system logging
*.=info;*.=notice;*.=warn;\
	auth,authpriv.none;\
	cron,daemon.none;\
	mail,news.none		-/var/log/messages
```
```
# cat /etc/hosts
127.0.0.1	localhost

# The following lines are desirable for IPv6 capable hosts
::1     localhost ip6-localhost ip6-loopback
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
```
Name resolution happens via network.

For testing I compiled a default *rsyslogd* (just "configure" without any options).

I disabled network and started *rsyslogd*.

rsyslog-8.26.0: Prints a line like
```
6760.927572895:main thread    : GenerateLocalHostName uses 'vm-debian8-x64'
```
and */var/log/messages/* contains
```
2017-05-23T15:38:38.670676+02:00 vm-debian8-x64 rsyslogd:  [origin software="rsyslogd" swVersion="8.26.0" x-pid="32218" x-info="http://www.rsyslog.com"] start
```

rsyslog-8.27.0: Prints a line like
```
6438.445699990:main thread    : getaddrinfo: Name or service not known
```
and */var/log/messages/* contains
```
2017-05-23T15:33:58.449628+02:00 [localhost] rsyslogd:  [origin software="rsyslogd" swVersion="8.27.0" x-pid="32183" x-info="http://www.rsyslog.com"] start
```

https://www.dropbox.com/s/c88lxw2q4vn7q4h/rsyslog-8.26.0-debug.log?dl=0
https://www.dropbox.com/s/gakkies3jirgbu0/rsyslog-8.27.0-debug.log?dl=0

davidelang:
what is the result of running the hostname command?

does your hostname have a . in it or not?

if not, what happens if you change it to a fqdn?

Whissi:
The hostname doesn't contain a dot.

`hostname` returns the hostname as expected (even without working name resolution due to disabled network). `hostname -f` returns the FQDN as expected (but only with working name resolution, i.e. enabled network in my case).

When I do `hostname my.new.hostname`, `hostname` will return `my.new.hostname`, `hostname -f` still not working due to disabled network, i.e. returns just "hostname: Name or service not known" but rsyslog-8.27.0 will now use "my" (=the first part before the first dot) as hostname.

rgerhards:
@jvymazal @davidelang In the light of what @Whissi reports, I think we should revert the patch, albeit we do not know why it happens. Or anyone with a better idea. A compromise might be to add a configure option which defaults to the old code...

jvymazal:
I am investigating sources for both glibc/kernel calls at the moment, from what I found so far it seems that `gethostbyname` has some fallback mechanism which, in case of network unavailability, copies supplied name (provided it is not address) and puts it back to returned `hostent` structure and,  from some reason mysterious to me, this piece of code was removed from `getaddrinfo`. When I am sure about it (hopefully tomorrow) I will open new PR to retain this functionality on rsyslog level.

bilihor:
How do we notify RHEL7 /  CentOS7 to use 8.28 ?
Maybe this is not critical enough so it is ignored

jvymazal:
@bilihor in Fedora/RHEL/CntOS rsyslog is delayed to start after network so the problem should not arise there.

bilihor:
I have custom distribution based on CentOS with no dns and rsyslog need to be starting before network.

Issue #1519, opened 2017-04-18
test dynstats_prevent_premature_eviction.sh fails on very slow machines


rgerhards:
This test fails fairly consistently on the very slow new solaris 10 sparc buildbot slave. However, it also fails occasionally on the other testbench machines as well. This looks like there is still some race condition in the test setup. I do **NOT** think that the code itself has a problem.

@janmejay if you have some time, it would be great if you could have a look at it.

I have worked around the really problematic Solaris issue by disabling make check on that slave. This is probably needed in any case, as the salve is so slow it would only permit us to run 10..15 CI tests per day. Once this issue here is solved we can re-think what we do with that slave.

janmejay:
Looked at this one: http://build.rsyslog.com:8010/builders/rsyslog%20solaris10sparc%20rsyslog/builds/510/steps/shell_8/logs/test-suite.log

Basically, randomly hunted for solaris10 sparc failures and found it.

Here is the diagnosis:
This line appears in the post-assert-failure log-dump: 
`2017-04-18T11:56:13.556549+02:00 unstable10s.bo.opencsw.org rsyslogd-pstats: msg_stats: origin=dynstats.bucket foo=2 bar=1 baz=1`

Let us call this flush **k**

It was supposed to be more like:
`foo=1 bar=1`
and then 
`foo=2 baz=1`

Because we injected input_1 set before the flush _k-1_ and injected input_2 (containing one foo and one baz message) just before flush _k_, we expected bar to be evicted (input_1 was the only one that has bar), foo to be retained (since it was added by input_1 and incremented by input_2) and baz to be added (since input_2 had foo and baz).

I suspect the flush _k-1_ happened before input_1 could be processed completely (or may be before it was even injected).
And then input_1 and input_2 were both injected before flush _k_.

Solution is trivial, I just need to space out impstat reporting interval a little (make it 2s perhaps), which will give the test more tolerance.

Shall I do this?

rgerhards:
yes, please do. Once I see the PR, I'll re-enable the slow Solaris slave for testing (it really is too slow to be always kept in the loop). Changing this still makes sense as I see the failure occasionally on the regular salves as well, so it would be good if we can fix this once and forever. Thanks again for your time!

janmejay:
PR increases tolerance to 2 seconds. The valgrind test uses a coordination mechanism (because valgrind slowness was adding to chaos), but I didn't want to add it if unnecessary. Timing based coordination is more "realistic" test.

[...]

PR #1524, opened 2017-04-19
Double impstats flush-interval because 1-flush-per-second is not enough tolerance on very slow machines

janmejay:
dynstats_prevent_premature_eviction.sh always fails on Solaris10 sparc buildbot due to this tolerance issue (https://github.com/rsyslog/rsyslog/issues/1519).


rgerhards:
This looks like it does not base on current master branch?

rgerhards:
@janmejay  FYI: I have just changed the buildbot conf to do a make check on the slow Solaris slave, and I have also re-initiated the build. It'll probably take 90 to 120 minutes until we see a result.

rgerhards:
looks good so far, I have just given it another shot. Two success in a row should be good enough to says it's done ;-)

rgerhards:
Thanks, this seems to have done the trick!

PR #1525, opened 2017-04-20
Expose configuration for max data size in librelp

nbrownus:
depeds on https://github.com/rsyslog/librelp/pull/63

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1249,7 +1249,7 @@ AC_ARG_ENABLE(relp,
         [enable_relp=no]
 )
 if test "x$enable_relp" = "xyes"; then
-	PKG_CHECK_MODULES(RELP, relp >= 1.2.12)
+	PKG_CHECK_MODULES(RELP, relp >= 1.2.14)
 	AC_DEFINE([ENABLE_RELP], [1], [Indicator that RELP is present])
 fi
 AM_CONDITIONAL(ENABLE_RELP, test x$enable_relp = xyes)
--- a/plugins/imrelp/imrelp.c
+++ b/plugins/imrelp/imrelp.c
@@ -86,6 +86,7 @@ struct instanceConf_s {
 	sbool bEnableTLS;
 	sbool bEnableTLSZip;
 	int dhBits;
+	size_t maxDataSize;
 	uchar *pristring;		/* GnuTLS priority string (NULL if not to be provided) */
 	uchar *authmode;		/* TLS auth mode */
 	uchar *caCertFile;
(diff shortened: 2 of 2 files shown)


nbrownus:
The build won't succeed until a new version of `librelp` is released

rgerhards:
@nbrownus could you please adjust the configure.ac check for librelp to  >= 1.2.14, which we release today. After this change and when the librelp release is fully through, CI should run OK and I can merge (note that if there are CI errors, I'll take care of the failing systems in question).

rgerhards:
@nbrownus I forgot: if possible, a rebase to current master may also be very useful. master has some changes Centos6 needs.

PascalWithopf:
@nbrownus we are having several issues with librelp and max message sizes at the moment.

- https://github.com/rsyslog/rsyslog/issues/1741
- https://github.com/rsyslog/rsyslog/issues/1782#issuecomment-381687403

@rgerhards and I discussed the parameter maxDataSize and we came to the conclusion that it is not needed. The global parameter maxMessageSize should be used to configure the maximal size.

Do you see any reason to keep the parameter and when it would be of use?

Otherwise we plan to deprecate it.

nbrownus:
Yeah I think that is the right thing to do as well. This parameter was just the easiest path forward to fix issues we were seeing at the time.

nbrownus:
Only a portion of our infrastructure used `rsyslog` client -> `rsyslog` aggregator -> `logstash` using `omrelp` and `imrelp`. The problem was that `rsyslog` clients would happily send `maxMessageSize` using `omrelp` but the aggregator side would either split or reject messages since `imrelp` was hard coded for a smaller size. This change, and the librelp change, allowed us to configure the max size `imrelp` would accept to match the max size of what the clients were sending.

I can see a use case for the two parameters to continue to exist but `imrelp` should certainly default to `maxMessageSize`, which is enough to keep our deployment happy.

rgerhards:
> I can see a use case for the two parameters to continue to exist but imrelp should certainly default to maxMessageSize, which is enough to keep our deployment happy

What is that use case? I am very interested to learn about it, because I can not envision any reason why this parameter actually might be needed.

Sorry for these late questions. I should really have noticed at the time we merged, but unfortunately I did not.

rgerhards:
To be clear: I mean I don't see why it might be needed if the configured max message size is properly passed down to omrelp. Previously, with the hardcoded limit, it was obviously wrong.

rgerhards:
Actually, I got myself a use case. Let's say someone permits a max message size of 512KiB. If this is set as max relp data size, every message above that needs to be truncated or accepted for error logging. With error logging, we run into the issue that this may be misused for attacks with extremely large messages. So it could make sense to add a max data size of e.g. 50 MiB, which should be really, really never reached and accept everything within that limit for error logging. However, over 50 MiB, we consider the connection malicious and so close it (as specified in the relp spec) or truncate. This actually warrants having both settings with the following rules:

* max data size shall default to rsyslog's max message size
* max data size, if configured, cannot be lower than max message size - if so, emit an error message

davidelang:
that makes a lot of sense.

Issue #745, opened 2016-01-20
im[p]tcp: add option to truncate oversize message


rgerhards:
Currenty, the oversize part is treated as a separate message, but that part is malformed. So this causes more trouble in most cases than it is worth.

ymattw:
Do you mean the feature implemented in the pending PR #664 ?

rgerhards:
No, sorry for the confusion. Changed the title. It's in regard to receiving oversize messages via TCP.

PR #1528, opened 2017-04-20
imtcp: add new parameter discardTruncatedMsg

PascalWithopf:
Parameter discardTruncatedMsg makes it possible to discard the rest of the message after it has been truncated.
closes https://github.com/rsyslog/rsyslog/issues/745

### DIFF
--- a/plugins/imtcp/imtcp.c
+++ b/plugins/imtcp/imtcp.c
@@ -98,6 +98,7 @@ static struct configSettings_s {
 	int bEmitMsgOnClose;
 	int iAddtlFrameDelim;
 	int bDisableLFDelim;
+	int discardTruncatedMsg;
 	int bUseFlowControl;
 	uchar *pszStrmDrvrAuthMode;
 	uchar *pszInputName;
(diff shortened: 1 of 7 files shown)


plugins/imtcp/imtcp.c
```diff
@@ -549,6 +556,7 @@ CODESTARTendCnfLoad
 		pModConf->bSuppOctetFram = cs.bSuppOctetFram;
 		pModConf->iAddtlFrameDelim = cs.iAddtlFrameDelim;
 		pModConf->bDisableLFDelim = cs.bDisableLFDelim;
+		pModConf->discardTruncatedMsg = cs.discardTruncatedMsg;
```
rgerhards: cs. is only for old-style config params, for which we should not add any new ones. So no changes to the cs. part should be necessary.

runtime/tcps_sess.c
```diff
@@ -413,6 +413,9 @@ processDataRcvd(tcps_sess_t *pThis,
 			DBGPRINTF("error: message received is larger than max msg size, we split it\n");
 			defaultDoSubmitMessage(pThis, stTime, ttGenTime, pMultiSub);
 			++(*pnMsgs);
+			if(pThis->pSrv->discardTruncatedMsg == 1) {
+				pThis->pSrv->discardingOn = 1;
```
rgerhards: why do we need a second variable here? Can't we use discardTruncatedMsg?

Issue #1535, opened 2017-04-28
imfile: add option not to give error messages on "file not found"


rgerhards:
Some users configure file monitors for non-existing files as an aid for mass deployments (see #1531 for an example). Prior to 8.26.0, file-not-found error messages were missing. They have been added in 8.26.0 and above. For those users, this generates unnecessary error messages. So we should provide an option not to report these type of messages (albeit this is a bad idea in the general case).

This needs to be done in both operating modes:
* inotify
* polling

PR #1536, opened 2017-04-28
imfile: add error message when file not found and parameter to disable the message

PascalWithopf:
As default imfile returns an error if file is not found, but if the parameter is set off then no error is returned.
Closes https://github.com/rsyslog/rsyslog/issues/1535

### DIFF
--- a/tests/imfile-file-not-found-error.sh
+++ b/tests/imfile-file-not-found-error.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+# add 2017-04-28 by Pascal Withopf, released under ASL 2.0
+uname
+if [ `uname` = "SunOS" ] ; then
+   echo "Solaris: inotify isn't supported on Solaris"
+   exit 77
+fi
+
+. $srcdir/diag.sh init
+. $srcdir/diag.sh generate-conf
+. $srcdir/diag.sh add-conf '
+module(load="../plugins/imfile/.libs/imfile")
+input(type="imfile" File="./rsyslog.input" Tag="tag1" ruleset="ruleset1")
+
+template(name="tmpl1" type="string" string="%msg%\n")
+ruleset(name="ruleset1") {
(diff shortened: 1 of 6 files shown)


tests/imfile-file-not-found-error.sh
```diff
@@ -0,0 +1,24 @@
...
+
+input(type="imfile" File="testsuites/NotExistingInputFile" Tag="tag1")
+
+action(type="omfile" file="rsyslog.out.log")
+'
+. $srcdir/diag.sh startup
+. $srcdir/diag.sh shutdown-when-empty
+. $srcdir/diag.sh wait-shutdown
+
+grep "imfile: error.*No such file or directory" rsyslog.out.log > /dev/null
```
rgerhards: The file name should also be checked in the test. I think it is mandatory that it is given, else the user gets confused.

tests/imfile-fileNotFoundError-parameter.sh
```diff
@@ -0,0 +1,24 @@
...
+input(type="imfile" File="testsuites/NotExistingInputFile" Tag="tag1" fileNotFoundError="off")
+
+action(type="omfile" file="rsyslog.out.log")
+'
+. $srcdir/diag.sh startup
+. $srcdir/diag.sh shutdown-when-empty
+. $srcdir/diag.sh wait-shutdown
+
+grep "imfile: error.*No such file or directory" rsyslog.out.log > /dev/null
+if [ $? -eq 0 ]; then
```
rgerhards, line 17: mhh... isn't it an error if the error message *is* emitted in this case?

ghost: yes, that's why I'm testing if the result of grep search equals 0

rgerhards: lol, you got me ;-)

PR #1537, opened 2017-04-28
Merge KSI LS12 signature plugin

allanpark:
New version of KSI log signature module that enables efficient signing of log records with Guardtime KSI blockchain. This version produces log signatures of version 1.2
https://guardtime.com/technology/blockchain-developers


runtime/lmsig_ksi-ls12.c
```diff
@@ -0,0 +1,329 @@
...
+			rsksiSetDirCreateMode(pThis->ctx, pvals[i].val.d.n);
+		} else if (!strcmp(pblk.descr[i].name, "filecreatemode")) {
+			rsksiSetCreateMode(pThis->ctx, pvals[i].val.d.n);
+		} else {
+			DBGPRINTF("lmsig_ksi: program error, non-handled "
+			  "param '%s'\n", pblk.descr[i].name);
+		}
+	}
+
+	if (!ag_uri)
```
rgerhards: You should set the parameter flag (in pblk) to mandatory, then these missing parameters can not happen. You do this by setting the third parameter to CNFPARAM_REQUIRED.

runtime/lib_ksils12.h
```diff
@@ -0,0 +1,197 @@
+/* librsksi.h - rsyslog's KSI support library
+ *
+ * Copyright 2013-2015 Adiscon GmbH.
```
rgerhards: I suggest to modify the copyright notice accordingly.

runtime/lmsig_ksi-ls12.c
```diff
@@ -0,0 +1,329 @@
...
+
+	RETiRet;
+}
+
+BEGINobjQueryInterface(lmsig_ksi_ls12)
+CODESTARTobjQueryInterface(lmsig_ksi_ls12)
+	 if(pIf->ifVersion != sigprovCURR_IF_VERSION) {/* check for current version, increment on each change */
+		ABORT_FINALIZE(RS_RET_INTERFACE_NOT_SUPPORTED);
+}
+pIf->Construct = (rsRetVal(*)(void*)) lmsig_ksi_ls12Construct;
```
rgerhards: It would be good to check indention level.

runtime/lmsig_ksi-ls12.c
```diff
@@ -0,0 +1,329 @@
...
+		REPORT_PARAM_MISSING("sig.aggregator.key");
+
+	if (!bHaveLevelLimit)
+		REPORT_PARAM_MISSING("sig.aggregator.levelLimit");
+
+	if (!bHaveHashFunction)
+		rsksiSetHashFunction(pThis->ctx, "SHA2-256");
+
+	if (rsksiSetAggregator(pThis->ctx, ag_uri, ag_loginid, ag_key) != KSI_OK) {
+		pThis->ctx->disabled = true;
```
rgerhards: Would it make sense to emit an error message here?

allanpark:
Hi again,

On 09.05.2017 13:13, Rainer Gerhards wrote:
> ------------------------------------------------------------------------
>
> In runtime/lmsig_ksi-ls12.c 
> <https://github.com/rsyslog/rsyslog/pull/1537#discussion_r115451502>:
>
> > +			rsksiSetFileUID(pThis->ctx, pvals[i].val.d.n);
> +		} else if (!strcmp(pblk.descr[i].name, "filegroup")) {
> +			rsksiSetFileGID(pThis->ctx, pvals[i].val.d.n);
> +		} else if (!strcmp(pblk.descr[i].name, "filegroupnum")) {
> +			rsksiSetFileGID(pThis->ctx, pvals[i].val.d.n);
> +		} else if (!strcmp(pblk.descr[i].name, "dircreatemode")) {
> +			rsksiSetDirCreateMode(pThis->ctx, pvals[i].val.d.n);
> +		} else if (!strcmp(pblk.descr[i].name, "filecreatemode")) {
> +			rsksiSetCreateMode(pThis->ctx, pvals[i].val.d.n);
> +		} else {
> +			DBGPRINTF("lmsig_ksi: program error, non-handled "
> +			  "param '%s'\n", pblk.descr[i].name);
> +		}
> +	}
> +
> +	if (!ag_uri)
>
> You should set the parameter flag (in pblk) to mandatory, then these 
> missing parameters can not happen. You do this by setting the third 
> parameter to CNFPARAM_REQUIRED.
>

Actualy I found out that there is a problem using CNFPARAM_REQUIRED with 
sigprovider and omfile modules: in tools/omfile.c:1175 the return value 
of SetCnfParam (of the signature provider) is not checked and this flag 
doesn't have any effect. Actually it is even worse: the potentially 
uninitialized plugin is kept and used. I didn't want to add it to review 
notes and used e-mail instead. I can add an issue with detailed 
information to github. I cannot fix it according to your suggestion but  
the ksi-ls12 plugin works even without  using the flag.

I shall fix the other things you pointed out too.

A.

runtime/lmsig_ksi-ls12.c
allanpark: Refactored the code so that if nvlstGetParams() fails the plugin will be disabled

runtime/lmsig_ksi-ls12.c
allanpark: The error message is emitted inside rsksiSetAggregator where proper error code is available

runtime/lmsig_ksi-ls12.c
allanpark: Fixed.

runtime/lib_ksils12.h
allanpark: Fixed.

rgerhards:
> Actualy I found out that there is a problem using CNFPARAM_REQUIRED with
> sigprovider and omfile modules: in tools/omfile.c:1175 the return value
> of SetCnfParam (of the signature provider) is not checked and this flag
> doesn't have any effect. Actually it is even worse: the potentially
> uninitialized plugin is kept and used. I didn't want to add it to review
> notes and used e-mail instead. I can add an issue with detailed
> information to github. I cannot fix it according to your suggestion but
> the ksi-ls12 plugin works even without  using the flag.

OK, we need to craft a test and then fix for what you describre. It would be good if you could create a dedicated issue. I think we should merge as-is, and once we fix the problem in rsyslog core, we can streamline the module.

It would be great if you could squash the non-essential commits so that I can do the merge tomorrow.

rgerhards:
We currently experience some issues with the Centos6 buildslave. I am looking into it. The failure here is NOT caused by the PR.

PR #1553, opened 2017-05-11
testbench: enable some more tests on solaris

rgerhards:
I guess they were accidently disabled.


rgerhards:
looks like a more problematic issue than I thought. Waits for sponsor.

PR #1555, opened 2017-05-11
Added new module: KSI log signing ver. 1.2 (lmsig_ksi_ls12)

allanpark:
New version of KSI log signature module that enables efficient signing of log records with Guardtime KSI blockchain. This version produces log signatures of version 1.2
https://guardtime.com/technology/blockchain-developers


rgerhards:
I wonder if this is the same as https://github.com/rsyslog/rsyslog/pull/1537 ?

allanpark:
Yes, this is the same as #1537 but all the commits squashed into single commit to avoid polluting the rsyslog history with irrelevant stuff.

Issue #11, opened 2014-01-17
omprog: add capability to specify environment variables


rgerhards:
This is required for a number of use cases, with JAVA and it's required environment settings being a prime example. Note that currently this can be worked around easily by using a script, so it's not a hard or pressing requirement.

rgerhards:
we had no requests for this, so I am closing the bug. Anyone who actually needs this, please re-open and mention use case.

PR #1561, opened 2017-05-17
Fix mangled pmsnare messages when tabs not escaped

shane-lawrence:
When `escapecontrolcharacters="off"`, if an incoming Snare event's fields are only separated by tabs, pmsnare will mangle the message because an actual tab character is smaller than `"#011"`.

Sample message:
```
<14>May 17 02:25:10 SL-Win2008R2DC.lab.home MSWinEventLog	1	Security	606129	Wed May 17 02:25:10 2017	4634	Microsoft-Windows-Security-Auditing	N/A	N/A Success Audit	SL-Win2008R2DC.lab.home Logoff	An account was logged off.	Subject:	Security ID:	S-1-5-21-1004336348-1177238915-682003330-512	Account Name:	SL-WIN2008R2IIS$	Account Domain:	LAB	Logon ID:	0x9648a78	Logon Type:	3	This event is generated when a logon session is destroyed. It may be positively correlated with a logon event using the Logon ID value. Logon IDs are only unique between reboots on the same computer. 984658
```

Resultant message without the fix:
```
FROMHOST: 'localhost', fromhost-ip: '::1', HOSTNAME: 'SL-Win2008R2DC.lab.home', PRI: 14,
syslogtag 'MSWinEventLog', programname: 'MSWinEventLog', APP-NAME: 'MSWinEventLog', PROCID: '-', MSGID: '-',
TIMESTAMP: 'May 17 02:25:10', STRUCTURED-DATA: '-',
msg: ' ecurity	606129	Wed May 17 02:25:10 2017	4634	Microsoft-Windows-Security-Auditing	N/A	N/A Success Audit	SL-Win2008R2DC.lab.home Logoff	An account was logged off.	Subject:	Security ID:	S-1-5-21-1004336348-1177238915-682003330-512	Account Name:	SL-WIN2008R2IIS$	Account Domain:	LAB	Logon ID:	0x9648a78	Logon Type:	3	This event is generated when a logon session is destroyed. It may be positively correlated with a logon event using the Logon ID value. Logon IDs are only unique between reboots on the same computer. 984658'
escaped msg: ' ecurity606129Wed May 17 02:25:10 20174634Microsoft-Windows-Security-AuditingN/AN/A Success AuditSL-Win2008R2DC.lab.home LogoffAn account was logged off.Subject:Security ID:S-1-5-21-1004336348-1177238915-682003330-512Account Name:SL-WIN2008R2IIS$Account Domain:labLogon ID:0x9648a78Logon Type:3This event is generated when a logon session is destroyed. It may be positively correlated with a logon event using the Logon ID value. Logon IDs are only unique between reboots on the same computer. 984658'
inputname: imptcp rawmsg: '<14>May 17 02:25:10 SL-Win2008R2DC.lab.home MSWinEventLog ecurity	606129	Wed May 17 02:25:10 2017	4634	Microsoft-Windows-Security-Auditing	N/A	N/A Success Audit	SL-Win2008R2DC.lab.home Logoff	An account was logged off.	Subject:	Security ID:	S-1-5-21-1004336348-1177238915-682003330-512	Account Name:	SL-WIN2008R2IIS$	Account Domain:	LAB	Logon ID:	0x9648a78	Logon Type:	3	This event is generated when a logon session is destroyed. It may be positively correlated with a logon event using the Logon ID value. Logon IDs are only unique between reboots on the same computer. 984658'
```


contrib/pmsnare/pmsnare.c
```diff
@@ -95,79 +97,97 @@ CODESTARTparse
...
-			lenMsg -=(tablength-2);
-			memmove(p2parse, p2parse + (tablength-2), lenMsg);
-			*(p2parse + lenMsg) = '\n';
-			*(p2parse + lenMsg + 1)  = '\0';
-			pMsg->iLenRawMsg -=(tablength-2);
-			pMsg->iLenMSG -=(tablength-2);
-			dbgprintf("found a Snare message with snare not set to send syslog messages\n");
+		
+		/* Found a Snare event (not syslog). */
+		if(len_snare_tag) {
```
rgerhards: I think the orginal variable name made more sense here.

contrib/pmsnare/pmsnare.c
```diff
@@ -70,8 +73,8 @@ DEFobjCurrIf(datetime)
 
 
 /* static data */
-static int bParseHOSTNAMEandTAG;	/* cache for the equally-named global param - performance enhancement */
-
+// cache for the equally-named global param - performance enhancement
```
rgerhards: We try to avoid C++ comments as some platforms have shown issues with them.

contrib/pmsnare/pmsnare.c
```diff
@@ -227,16 +261,17 @@ ENDqueryEtryPt
 
 BEGINmodInit()
 CODESTARTmodInit
-	*ipIFVersProvided = CURR_MOD_IF_VERSION; /* we only support the current interface specification */
+	/* we only support the current interface specification */
```
rgerhards: IMHO this should be as it was.

davidelang:
We have needed the ability to not escape tabs for years (I had the same problem)

I've been hoping that we would get a generic way to do this rather than having 
to do something special for snare messages.

contrib/pmsnare/pmsnare.c
```diff
@@ -95,79 +97,97 @@ CODESTARTparse
...
-	  else, wrong message, abort
-	 else, assume that we have a valid timestamp, move over to the syslog tag
-	  if that is tab separated from the rest of the message and one of our expected tags
-	   if so, replace the tab with a space so that it will be parsed properly
-	    optionally replace the hostname at the beginning of the message withone from later in the message 
-
-	*/
-	snaremessage=0;
-	lenMsg = pMsg->iLenRawMsg - pMsg->offAfterPRI;
+	/* Check if this message is of the type we handle in this parser. */
```
davidelang: I think it's useful to have the comments in place to explain what we are intending to do, especially in a case like this where the code can tell you what is happening, but not the why (the thinking behind the code)

contrib/pmsnare/pmsnare.c
```diff
@@ -95,79 +97,97 @@ CODESTARTparse
...
+				*p2parse = ' ';
+				p2parse++;
+				lenMsg--;
+				/* Length of tab character minus char already replaced and
+				 * null terminator */
+				lenMsg -= (tablength-2);	
+				memmove(p2parse, p2parse+(tablength-2), lenMsg);
+				*(p2parse + lenMsg) = '\n';
+				*(p2parse + lenMsg + 1)  = '\0';
+				pMsg->iLenRawMsg -=(tablength-2);
```
davidelang: the whole tablength stuff was planning to try and handle unescaped tabs as well as escaped ones. I never did get around to handling the case where we disable control character escaping

contrib/pmsnare/pmsnare.c
```diff
@@ -177,31 +197,45 @@ CODESTARTparse
...
+				"hostname: [%d]'%s'\n", 
+				tablength,TabRepresentation,lenMsg, p2parse);
+		if(lenMsg > 13 && strncasecmp( \
+				(char*) p2parse, "MSWinEventLog", 13) == 0) {
+			len_snare_tag=13;
 		}
-		if(lenMsg > 11 && strncasecmp((char*) p2parse, "LinuxKAudit", 11) == 0) {
-			snaremessage=11; /* 0 means not a snare message, a number is how long the tag is */
+		if(lenMsg > 11 && strncasecmp( \
+				(char*) p2parse, "LinuxKAudit", 11) == 0) {
```
davidelang: This is an example of a case where I think line wrapping to stay under an arbitrary limit makes things less clear rather than more clear.

rgerhards: I would like to say that this change has nothing to do with our line length limits. Looks like a change for personal style. We permit 120 chars and this is *a lot*.

contrib/pmsnare/pmsnare.c
```diff
@@ -81,11 +84,10 @@ CODESTARTisCompatibleWithFeature
...
 
-
 BEGINparse
 	uchar *p2parse;
 	int lenMsg;
-	int snaremessage;
+	int len_snare_tag;
 	int tablength;
 
 CODESTARTparse
```
davidelang: The entire reason for having TabRepresentation and tablength was to try and provide a clean way of handling tabs that are represented in different ways.

The "normal" case is tabs escaped (because we should be escaping control characters and there isn't a way to escape all control characters except for tab)

The idea was that by changing TabRepresentation to match the escaping (or lack of escaping), the rest of the code would not need to have tests for the various possible representations.

you should be able to just change TabRepresentation to "\t" and have it match unescaped tabs and 'do the right thing'

I never went back to actually implement and test it.

If you are not going to make use of this (and instead add separate tests for \t vs #011) then you should simplify the code and rip out all the tabrepresentation and tablength stuff.

But I think this entire commit could be replaced by some code that detected if tabs are escaped or not (and if they are escaped into octal or decimal) and sets tabrepresentation appropriately

contrib/pmsnare/pmsnare.c
davidelang: ignore this comment, see the more detailed explanation I made earlier in the file (but later in time)

contrib/pmsnare/pmsnare.c
davidelang: @rgerhards How can we find out what sort of escaping is set?

contrib/pmsnare/pmsnare.c
davidelang: In that case, I think we should not wrap individual statements (like these if conditions) unless they hit 120 characters (which these don't)

contrib/pmsnare/pmsnare.c
davidelang: hmm, thinking out loud, can we just define a variable to contain the string "\t" and then have the escaping do whatever conversion it is doing to the message to this variable as well? if we can do that, then it should be almost as easy as changing TabRepresentation from being an #define to being this variable (there is the checking that a variable works everywhere that currently has a macro defining a constant string, and changing things to dereference as needed)

contrib/pmsnare/pmsnare.c
shane-lawrence: I saw from recent commit messages that you had started enforcing a line length limit, but I thought that it was 80 characters instead of 120. I'll remove the wrapping.

contrib/pmsnare/pmsnare.c
shane-lawrence: Fixed.

shane-lawrence:
Thanks for the prompt response and suggestions! I've edited the patch. I couldn't find any other parser modules that care how tabs are represented. Snare seems to be an outlier (though you already know that!). For this reason, I kept the check in pmsnare. If you think it would be more useful elsewhere, I can move the check and store the result in a global config variable.

I did improve the check to account for many different tab representations, fixed the style, and tried to simplify a bit by removing some redundancy. Please let me know what you think and I'll update accordingly.

contrib/pmsnare/pmsnare.c
```diff
@@ -81,94 +84,93 @@ CODESTARTisCompatibleWithFeature
...
+			tabLength, tabRepresentation, lenMsg, p2parse);
+
+		/* Look for the Snare tag. */
+		if(strncasecmp((char*)(p2parse + tabLength), "MSWinEventLog", 13) == 0) {
+			dbgprintf("Found a non-syslog Windows Snare message.\n");
+			snaremessage = p2parse - pMsg->pszRawMsg + tabLength + 13;
 		}
-		if(strncasecmp((char*) (p2parse + tablength - 1), "LinuxKAudit", 11) == 0) {
-			snaremessage=11; /* 0 means not a snare message, a number is how long the tag is */
+		else if(strncasecmp((char*) (p2parse + tabLength), "LinuxKAudit", 11) == 0) {
```
davidelang: tablength is a string with a terminating \0 on it, so "#011" is a length 5. "\t" is length 2

This is why the old code used tablength -1 for comparison and tablength -2 for shrinking the message

contrib/pmsnare/pmsnare.c
```diff
@@ -81,94 +84,93 @@ CODESTARTisCompatibleWithFeature
...
 	if((unsigned) lenMsg < 30) {
 		/* too short, can not be "our" message */
-		dbgprintf("msg too short!\n");
+		dbgprintf("Message is too short to be Snare!\n");
 		ABORT_FINALIZE(RS_RET_COULD_NOT_PARSE);
 	}
 
-	while(lenMsg && *p2parse != ' ' && *p2parse != '\t' && *p2parse != '#') {
+	/* Check for non-syslog snare format. */
+	while(lenMsg && *p2parse != ' ' && *p2parse != tabRepresentation[0]) {
```
davidelang: since you don't check for both \t and an escaped control character (something starting with #) I don't think this will work if escapecontrolcharacters is enabled

contrib/pmsnare/pmsnare.c
```diff
@@ -81,94 +84,93 @@ CODESTARTisCompatibleWithFeature
...
 	}
-	dbgprintf("pmsnare: separator [%d]'%s'  msg after the first separator: [%d]'%s'\n", tablength,
-	TabRepresentation,lenMsg, p2parse);
-	if ((lenMsg > tablength) && (*p2parse == '\t' || strncasecmp((char*) p2parse, TabRepresentation ,
-	tablength-1) == 0)) {
-	//if ((lenMsg > tablength) && (*p2parse == '\t' || *p2parse == '#')) {
-	dbgprintf("pmsnare: tab separated message\n");
-		if(strncasecmp((char*) (p2parse + tablength - 1), "MSWinEventLog", 13) == 0) {
-			snaremessage=13; /* 0 means not a snare message, a number is how long the tag is */
+	if ((lenMsg > tabLength) && (strncasecmp((char *)p2parse, tabRepresentation, tabLength) == 0)) {
```
davidelang: you have removed the definition of tagRepresentation, but use it here

contrib/pmsnare/pmsnare.c
```diff
@@ -70,8 +72,9 @@ DEFobjCurrIf(datetime)
 
 
 /* static data */
-static int bParseHOSTNAMEandTAG;	/* cache for the equally-named global param - performance enhancement */
-
+static int bParseHOSTNAMEandTAG;  /* cache for the equally-named global param - performance enhancement */
+static char tabRepresentation[5];
```
davidelang: where is the content of this array defined?

davidelang:
Let's back up and restate the problem here.

Snare cannot send a valid syslog message, it has two formats it sends, one with something close to a syslog header and one without.

In both cases, instead of using spaces to separate the hostname and programname fields, it uses tabs

Depending on what escaping of control characters that is configured in rsyslog, at the time that the parser module is looking at the message, a tab can look like many different things, most commonly
it can be the single character '\t'
it can be the octal escaped '#011'

this module does quick-and-dirty surgery on the rawmsg to do just enough work on the message to make it so that it can be parsed with the rfc3164 parser. It then lies to the rsyslog core and says that it couldn't parse the message, so that the rest of the parser stack can then run against the modified message (it's an ugly hack, but it works)

There is the further problem of dealing with escaped tabs when parsing the message, but that is something that really needs a modified version of the escapecontrolcharacters option

Since I always escape control characters (bad things can happen to you if you cat a logfile to a terminal that has just the right set of control characters in it, including arbitrary command execution), I wrote the module to deal with tabs that are escaped, but I started setting up the framework to have it work when they are anything else.

Since in your case, you are not escaping control characters, if you simply change the #define of tabRepresentation from "#011" to "\t" (note that this needs to be double quotes to represent a string, not single quotes to represent a character), the module should work for your case.

I would suggest going back to the stock code and trying this before working on this patch any further.

The tricky thing (and what I never figured out) is how to tell what escaping is enabled in a given config at runtime.

If modifying the #define works for you, then let's do the following.

1. change the #define to a char array and populate it with \t\0 or #011\0 (null terminated string) and make sure that works.

2. figure out a way to tell what the escaping is configured to be and set the string to the appropriate value. 

one way this could be done is to wait until line 122 in the original version where we are looking for the first separator value (space, tab, or escaped control character) and check for each possible representation of tab and then set tabRepresentation and tablength at that point.

another way could be to check config variables (if we can get at them, @rgerhards ??)

contrib/pmsnare/pmsnare.c
davidelang: never mind, this is what I get for reviewing code too soon after I wake up :-)

contrib/pmsnare/pmsnare.c
davidelang: even if tabRepresentation is set correctly, it's still worth checking for both \t and # as well as space. the reason for this check is that we don't yet know if this is a snare message, so we are looking for the first separator in the message and then deciding if it's a space (in which case this isn't a snare formatted message, it's a legitimate syslog message) or if it's a tab, in which case this probably is a snare message (or it's probably not something that's going to be corrupted any more than it already is by replacing the first two tabs with spaces)

shane-lawrence:
That's very similar to what I've done. It checks to see what the escaping is supposed to be based on settings in the config file (using the global getter functions) and uses that to set tabRepresentation. This was added to CodeModInit in my commit just a couple hours ago.

contrib/pmsnare/pmsnare.c
shane-lawrence: This will match a space or a valid tab or an escaped character. It uses only the first character of the tabRepresentation string, which will either be an actual tab or the escape prefix (most likely '#' but if it's changed in the config file then it should still work).

contrib/pmsnare/pmsnare.c
shane-lawrence: I set tablength to strlen instead of sizeof so that the sizing would be consistent.

contrib/pmsnare/pmsnare.c
shane-lawrence: Perhaps I should move the CodeModInit to the top? It is a bit counterintuitive to have it at the bottom of the file when it will be called first chronologically.

contrib/pmsnare/pmsnare.c
davidelang: Even if the tabRepresentation is \t, we should still search for # because if we find it, that will mean that this isn't a snare message (avoiding false positives)

that will mean that we ran into something else 'odd' (or some other control character) and so it doesn't fit our pattern.

contrib/pmsnare/pmsnare.c
davidelang: That would probably reduce confusion

contrib/pmsnare/pmsnare.c
```diff
@@ -1,38 +1,41 @@
...
  * You may obtain a copy of the License at
- * 
- *       http://www.apache.org/licenses/LICENSE-2.0
- *       -or-
- *       see COPYING.ASL20 in the source distribution
- * 
+ *
+ *		http://www.apache.org/licenses/LICENSE-2.0
+ *		-or-
+ *		see COPYING.ASL20 in the source distribution
```
davidelang: I can't tell through the github interface, is this replacing a tab with spaces? The correct indentation here should be one tab.

davidelang:
Ok, this looks good to me, it would probably be good to update the tests to test both versions of this (I'm not seeing anything related to this as to why the travis build is failing)

given the changes that have happened to this PR, it's probably a good idea to squash the commits.

I have no ability to test this any longer, so please validate that both versions work with real snare logs (I now prefer nxlog as a sender from windows as it can send more than just the eventlog and it can send the data json formatted, without the truncation that snare sometimes does)

shane-lawrence:
I'll figure out why CI is failing and fix it, then add test cases and then squash my commits and get back to you. Thanks!

contrib/pmsnare/pmsnare.c
shane-lawrence: Fixed.

contrib/pmsnare/pmsnare.c
shane-lawrence: Fixed.

contrib/pmsnare/pmsnare.c
shane-lawrence: Fixed.

contrib/pmsnare/pmsnare.c
shane-lawrence: Yes, it seems to work well for that specific use case and it is likely the most common. I hope that my changes will expand the functionality so that it can be used in many other scenarios regardless of other config settings.

contrib/pmsnare/pmsnare.c
shane-lawrence: Fixed.

contrib/pmsnare/pmsnare.c
shane-lawrence: These changes should effectively detect what escaping is set and set the tabRepresentation accordingly. They can be overridden, if desired, when configuring the module.

contrib/pmsnare/pmsnare.c
shane-lawrence: This functionality has been moved to endCnfLoad, since the config has not loaded before that point. I looked into laying out the code to match the chronological order the sections are called in, but it's non-trivial and would require many unnecessary changes to other files.

contrib/pmsnare/pmsnare.c
shane-lawrence: It's set in endCnfLoad. Once the rsyslog config has been fully loaded, it will use the global settings and override them as needed with instance-specific settings, then set this.

shane-lawrence:
I have added several test cases to validate all of the changes. They will ensure that the module correctly fixes Snare messages without impacting non-Snare messages. There were some test failures due to missing execute permission on the new test file, but I fixed it quickly.

The module now supports the newer config system and also has the ability to override global settings with instance-specific settings when configuring the parser. This may be useful in case some inputs are escaped but others are not (e.g. relayed messages were already escaped by a different rsyslog). 

I've added responses to your previous questions and squashed my commits. Please let me know if there's anything else.

rgerhards:
Just in case you didn't notice, we have some real CI failures here. Note that the Centos7 buildbot slave fails do to problems with the previous master branch. A rebase to current master will make that go away (but I can work around that manually if you do not like to rebase).

rgerhards:
It looks like you have rewritten history and so some old commits appear multiple times. I'll probably cherry-pick your commit and merge it via a dedicated pull request to get a cleaner git history. Do I need to cherry-pick some other than https://github.com/rsyslog/rsyslog/pull/1561/commits/ebe5271b86c19d1960ef765caf2f36cb3d63ef15?

shane-lawrence:
I ended up with all of the upstream commits in my PR when I rebased. ebe5271
<https://github.com/rsyslog/rsyslog/commit/ebe5271b86c19d1960ef765caf2f36cb3d63ef15>
contains
all of the changes that I intended to make.

rgerhards:
thanks -- maybe this was due to the problems github had. normally, after a ```git rebase``` you should not see duplicated master commits.

rgerhards:
@shane-lawrence just in case you did not notice, there are testbench failures so the PR is currently not mergable. I may be a good idea to create a new PR based on current master and a cherry-picked version of your patch plus the necessary bug fixes.

shane-lawrence:
Thanks! I'll close this PR and create a new one with just the up-to-date master and my changes.

Issue #1538, opened 2017-05-01
rsyslog 8.26 on macOS 10.12.4


hdatma:
```
>./configure --prefix=/opt/rsyslogd --enable-liblogging-stdlog=no --enable-inet=no --enable-regexp=no
...
> make
...
2 warnings generated.
  CC       librsyslog_la-obj.lo
  CC       librsyslog_la-modules.lo
  CC       librsyslog_la-statsobj.lo
  CC       librsyslog_la-dynstats.lo
  CC       librsyslog_la-stream.lo
stream.c:1452:8: error: implicit declaration of function 'fdatasync' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
        ret = SYNCCALL(pThis->fd);
              ^
stream.c:1438:22: note: expanded from macro 'SYNCCALL'
#       define SYNCCALL(x) fdatasync(x)
                           ^
1 error generated.
make[2]: *** [librsyslog_la-stream.lo] Error 1
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2
```
Looking into the configurator...

```
configure:16711: checking for fdatasync
...
configure:16711: $? = 0
configure:16711: result: yes
```

This is what we have:
/usr/include/sys/syscall.h:227:#define	SYS_fdatasync      187
However, it is not the same function. 
Forcing fsync instead of fdatasync at line 1437 of runtime/stream.c flings the bug. 

This is the rest of the compilation. 

```
...
Making all in runtime
  CC       librsyslog_la-stream.lo
  CC       librsyslog_la-var.lo
  CC       librsyslog_la-wtp.lo
  CC       librsyslog_la-wti.lo
  CC       librsyslog_la-queue.lo
  CC       librsyslog_la-ruleset.lo
  CC       librsyslog_la-prop.lo
  CC       librsyslog_la-ratelimit.lo
  CC       librsyslog_la-lookup.lo
lookup.c:961:35: error: too many arguments to function call, expected 1, have 2
        pthread_setname_np(lu->reloader, reloader_thd_name);
        ~~~~~~~~~~~~~~~~~~               ^~~~~~~~~~~~~~~~~
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/pthread.h:499:1: note: 'pthread_setname_np' declared here
__OSX_AVAILABLE_STARTING(__MAC_10_6, __IPHONE_3_2)
^
[...]

PR #1563, opened 2017-05-17
build system: some changes for osx

rgerhards:
manual PR for patch provided in #1538


rgerhards:
breaks solaris,so cannot be used as is

PR #1566, opened 2017-05-17
Improve librdkafka detection

Whissi:
I was wondering about the message
```
librdkafka is missing but library present, using -lrdkafka
```
so I looked into the code and improved librdkafka detection.


alorbach:
Looks like a better librdkafka detection to me ;)
From my point it can be merged (Y).

PR #1587, opened 2017-05-30
Fix #1588: Handle inotify event `IN_MOVED_TO` as `IN_CREATE`

### LINKED ISSUE #1588: imfile: Unhandled inotify event -- `IN_MOVED_TO`
In our case, we use `logrotate` to rotate log file. It moves `foo.log` to `foo.log.1`, then to `foo.log.2`, etc. At last , `logrotate` cleans `foo.log.9` and moves the empty file to `foo.log` (we just keep 9 log dump files), and this action causes a mistake: rsyslog does not handle inotify's `IN_MOVED_TO` event, so  `foo.log` will never be monitored.

In order to solve this problem, we update `logrotate`'s configuration to avoid moving file back. It works, but I hope rsyslog can do right thing with files which were moved to directory.

My PR: #1587

zacharyzhao:
This PR is submitted for solving the problem that rsyslog will not monitor files which were moved to monitored directory.

In our case, we use `logrotate` to rotate log file. It moves `foo.log` to `foo.log.1`, then to `foo.log.2`, etc. At last , `logrotate` cleans `foo.log.9` and moves the empty file to `foo.log` (we just keep 9 log dump files), and this action causes a mistake: rsyslog does not handle inotify's `IN_MOVED_TO` event, so  `foo.log` will never be monitored.

In order to solve this problem, we update `logrotate`'s configuration to avoid moving file back. It works, but I hope rsyslog can do right thing with files which were moved to directory.

### DIFF
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -1677,7 +1677,7 @@ in_setupDirWatch(const int dirIdx)
 	int dirnamelen = 0;
 	char* psztmp;
 
-	wd = inotify_add_watch(ino_fd, (char*)dirs[dirIdx].dirName, IN_CREATE|IN_DELETE|IN_MOVED_FROM);
+	wd = inotify_add_watch(ino_fd, (char*)dirs[dirIdx].dirName, IN_CREATE|IN_DELETE|IN_MOVED_FROM|IN_MOVED_TO);
 	if(wd < 0) {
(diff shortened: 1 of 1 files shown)


alorbach:
Can be merged (Y)

zacharyzhao:
Thanks for @alorbach 's review. :)

Issue #607, opened 2015-11-19
Enhancement for imjournal - use SYSLOG_PID, then fall back to _PID


portante:
By default, `imjournal` will pull the systemd user property, `SYSLOG_PID`, value for the `pid` message property.  If the `SYSLOG_PID` property does not exist, the `pid` message property ends up not being defined.

Can we consider having `imjournal` fallback to the `_PID` property if `SYSLOG_PID` is not found?

radosroka:
I'll try to make a PR. There is a problem with "usepidfromsystem" option because it has binary value. I suggest that we should have one option e.g. "usepid" with three states like "system", "syslog" and "both"(default). In Ideal situation we should depricate "usepidfromsystem".

rgerhards:
> I suggest that we should have one option e.g. "usepid" with three states like "system", "syslog" and "both"(default). In Ideal situation we should depricate "usepidfromsystem".

I think this approach makes sense. We should probably support the depricated ```usepidfromsystem``` if only it is given. This keeps old configs working.

radosroka:
> I think this approach makes sense. We should probably support the depricated usepidfromsystem if only it is given. This keeps old configs working.

I agree.

rgerhards:
done via #1591

PR #1591, opened 2017-05-31
imjournal: _PID fallback

radosroka:
- added fallback for _PID proprety when SYSLOG_PID is not available
- introduced new option "usepid" which sets which property should
  rsyslog use, it has 3 states system|syslog|both, default is both
- deprecated "usepidfromsystem" option
- it is possible to revert previous default with usepid="syslog"


radosroka:
Resolves #607

radosroka:
Added backward compatibility for "usepidfromsystem" which can override the "usepid" and it prints deprecation warning.

rgerhards:
FYI: I currently work on the centos7 salve, thus the fail. We re-initiate test once I am done. Sorry for the hassle.

plugins/imjournal/imjournal.c
```diff
@@ -93,6 +94,7 @@ static struct cnfparamdescr modpdescr[] = {
 	{ "defaultseverity", eCmdHdlrSeverity, 0 },
 	{ "defaultfacility", eCmdHdlrString, 0 },
 	{ "usepidfromsystem", eCmdHdlrBinary, 0 },
+	{ "usepid", eCmdHdlrString, 0 },
```
rgerhards: I notice that was wrong previously as well, but it would be great if you could remove the extra/invalid comma.

plugins/imjournal/imjournal.c
```diff
@@ -761,6 +799,10 @@ CODESTARTsetModCnf
 			free(fac);
 		} else if (!strcmp(modpblk.descr[i].name, "usepidfromsystem")) {
 			cs.bUseJnlPID = (int) pvals[i].val.d.n;
+			errmsg.LogError(0, RS_RET_DEPRECATED,
```
rgerhards: parser_errmsg() might be more appropriate here, as it automatically outputs the error location

radosroka:
> As a side-note, I notice that the module in general seems to use the config system in an non-standard way (cs. is only for old style), but that's a different issue ;-)

So what is right approach? Every variable is parsed from new config and stored into "cs" structure,

rgerhards:
> So what is right approach? Every variable is parsed from new config and stored into "cs" structure,

"cs" is a variable for the old config system. All config data is supposed to go into the _instanceData struct. The majority of all other modules use it that way. Content from "cs" is usually shuffled over to instanceData. A good example to see that use is omfile (but also most other modules).

There are two reasons for that:
* old-style statements must not be able to modify new-style config, which we cannot guarantee if the same "cs" item is used for both old- and new-style
* a currently theoretical one: the new system is designed so that it will work with dynamic config reload, as it keeps different config objects. As long as we do not support dynamic reloads, that's of course not a strong argument ;-)

All in all, I think "cs" usage can stay as is without harm, but at some point we should consider moving to the new code philosophy.

rgerhards:
side-note: if you could do a rebase to master, the centos7 issue would go away without need for me to manually intervene.

plugins/imjournal/imjournal.c
radosroka: Is it still appropriate? I've moved it into run input.

rgerhards: looking right now, it seems to me that the error message is not moved (and I also think the current place is appropriate, I was just suggesting a different function).

plugins/imjournal/imjournal.c
```diff
@@ -599,12 +618,28 @@ CODESTARTrunInput
...
-		/* Use the PIDs obtained by journald. */
-		pid_field_name = "_PID";
-	} else {
-		/* Use the PIDs obtained from the message. */
-		pid_field_name = "SYSLOG_PID";
+	/* handling old "usepidfromsystem" option */
+	if (cs.bUseJnlPID != -1) {
+		free(cs.usePid);
+		cs.usePid = strdup("system");
+		errmsg.LogError(0, RS_RET_DEPRECATED,
```
radosroka, line 625: This is the place where I can cover new and old syntax with error message. Also I wanted to know if it is possible to use parse_errmsg because I think that this code runs after config parsing. So line number will be inconsistent or not? @rgerhards

radosroka:
> 

    So what is right approach? Every variable is parsed from new config and stored into "cs" structure,

"cs" is a variable for the old config system. All config data is supposed to go into the _instanceData struct. The majority of all other modules use it that way. Content from "cs" is usually shuffled over to instanceData. A good example to see that use is omfile (but also most other modules).

There are two reasons for that:

    old-style statements must not be able to modify new-style config, which we cannot guarantee if the same "cs" item is used for both old- and new-style
    a currently theoretical one: the new system is designed so that it will work with dynamic config reload, as it keeps different config objects. As long as we do not support dynamic reloads, that's of course not a strong argument ;-)

All in all, I think "cs" usage can stay as is without harm, but at some point we should consider moving to the new code philosophy.

Yes I agree, this re-factoring can be provided in some other PR in future.

plugins/imjournal/imjournal.c
rgerhards: yeah, line number will be misleading. So better leave LogError.

radosroka:
This is the final version I hope.

radosroka:
Is this failure libfastjson problem, isn't it? @rgerhards

rgerhards:
I would be great if you could also update the doc at https://github.com/rsyslog/rsyslog-doc

radosroka:
Doc PR: https://github.com/rsyslog/rsyslog-doc/pull/339

PR #1599, opened 2017-06-01
use cloexec in a few more places

vapier:


rgerhards:
@vapier in case you did not notice: CI failed, so this is unmergeble for the time being

rgerhards:
SOCK_CLOEXEC is not a define on linux.

PR #1609, opened 2017-06-08
Kafka extension omkafka

alorbach:
- Added support for timestamps
- Implemented support for Action suspend/resume based on kafka events like broker down. 
- Implemented new feature to resubmit messages on delivery failure (resubmitOnFailure) as well as persisting failed messages to disk (keepFailedMessages & failedMsgFile). 
- Disabled unstable kafka tests for now. 
- Added configure option to statically link librdkafka (disabled by default) into kafka plugins.


davidelang:
does this solve the problem where omkafka was not detecting that the kafka 
server was unreachable and therefor the libkafka code was just queuing messages 
in RAM (to be lost on shutdown)?

alorbach:
Yes, the new code detects when kafka brokers become unavailable and the actino is set into suspend mode which lets rsyslog queue messages internally. And even after kafka failure, messages that fail can be saved amd resend automatically when kafka broker come back online again. 

However this may cause message duplication here and there caused by librdkafka, or kafka.

Issue #165, opened 2014-11-26
Deprecate legacy command line configuration options


rgerhards:
There is still a small set of command line options that affect configuration. The really bad thing of them is set setting command line options is distro-specific and usually far from trivial for most users. We should get rid of these and, when actually still needed, migrate them to real config options.

See also: http://lists.adiscon.net/pipermail/rsyslog/2014-November/038995.html

mostolog:
Can I start updating random doc? (ie: removing legacy format reference and redirecting to a legacy-format.rst configuration-example file)

rgerhards:
Not sure if I understand you correctly, but all doc updates are **very welcome**. In regard to legacy, we keep a short reference to the legacy syntax inside the current doc, as this aids in migration AND understanding existing rsyslog.conf, which one must often work with.

mostolog:
This was my proposal:

Create a large legacy-configuration-example file somewhere under https://github.com/rsyslog/rsyslog-doc (please, tell me where), and quote "An example using legacy syntax (deprecated) can be found at <link>"

rgerhards:
That's probably more for the rsyslog-doc project as it not directly relates to the command line options we talk here. Would you mind opening the topic there?

mostolog:
Would you mind to give me the web link (rsyslog.com, and I'll look for matching .rst) for command line documentation?

mostolog:
@davidelang I won't do anything related to command line documentation, then.

rgerhards:
@PascalWithopf we should just remove the already-depricated options and then we are ready to close this bug.

topical:
Please don't remove "-s" command line option!

I know I'm late but I have been using this option for years with CentOS 6, which contains an older version that didn't show the warning messages that this option is deprecated. So I discovered this problem this week when I upgraded rsyslog and hurried to github.

I need it for a syslog relay that receives messages from FreeBSD machines. FreeBSD's syslog daemon never includes its hostname in sent messages, so the relay has to add the hostname by using RDNS. I don't want the domain name in every message, so I strip it using "-s". So, if you remove this option, I really need a config file option instead.

Could you please undo the merge and/or add a configuration option for "-s", like "$StripDomains"? This would be great.

Thank you

PR #1613, opened 2017-06-13
rsyslogd: delete unused command line options

PascalWithopf:
closes https://github.com/rsyslog/rsyslog/issues/165

### DIFF
--- a/tools/rsyslogd.c
+++ b/tools/rsyslogd.c
@@ -1188,9 +1188,9 @@ initAll(int argc, char **argv)
 	 * rgerhards, 2008-04-04
 	 */
 #if defined(_AIX)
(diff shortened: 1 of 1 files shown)


topical:
Please don't remove "-s" command line option!

I know I'm late but I have been using this option for years with CentOS 6, which contains an older version that didn't show the warning messages that this option is deprecated. So I discovered this problem this week when I upgraded rsyslog and hurried to github.

I need it for a syslog relay that receives messages from FreeBSD machines. FreeBSD's syslog daemon never includes its hostname in sent messages, so the relay has to add the hostname by using RDNS. I don't want the domain name in every message, so I strip it using "-s". So, if you remove this option, I really need a config file option instead.

Could you please undo the merge and/or add a configuration option for "-s", like "$StripDomains"? This would be great.

Thank you

Issue #1621, opened 2017-06-19
ksi-ls12 plugin: calculated tree height is wrong (too high)


allanpark:
There is a bug in the tree height calculation code.

PR #1622, opened 2017-06-20
Fixing issue #1621: fixing ksi ls12 tree height calculation

allanpark:
Fixed the tree height calculation in two places.


rgerhards:
Note: The travis failures seem to be related to travis image update, we need to look into that. However, the Centos7 error is based on the patch and needs to be fixed.

Issue #1629, opened 2017-06-24
implement 'fileoffset' metadata for imfile


derekjsonar:
Something like this:

```
diff --git a/plugins/imfile/imfile.c b/plugins/imfile/imfile.c
index 758d8273..b5b2ce64 100644
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -547,6 +547,7 @@ static rsRetVal enqLine(lstn_t *const __restrict__pLstn,
        MsgSetRuleset(pMsg, pLstn->pRuleset);
        if(pLstn->addMetadata)
                msgAddMetadata(pMsg, (uchar*)"filename", pLstn->pszFileName);
+               msgAddMetadata(pMsg, (uchar*)"fileoffset", pLstn->pStrm->iCurrOffs);
        ratelimitAddMsg(pLstn->ratelimiter, &pLstn->multiSub, pMsg);

```

To enable a template that can reference `$!metadata!fileoffset%` like this:

```
$template fwdtemplate,"%hostname%^%$!metadata!filename%^$!metadata!fileoffset%^%msg%"
```

rgerhards:
this is implementag. Definitely in 8.30.0. maybe earlier (see ChangeLog)

PR #1633, opened 2017-06-26
Issue #1629: implement 'fileoffset' as imfile metadata 

derekjsonar:
As per the discussion in the mailing list here and issue #1629:
http://lists.adiscon.net/pipermail/rsyslog/2017-June/044498.html

### DIFF
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -533,11 +533,15 @@ getStateFileName(lstn_t *const __restrict__ pLstn,
 /* enqueue the read file line as a message. The provided string is
  * not freed - thuis must be done by the caller.
  */
+#define MAX_OFFSET_REPRESENTATION_NUM_BYTES 20 
 static rsRetVal enqLine(lstn_t *const __restrict__ pLstn,
 			cstr_t *const __restrict__ cstrLine)
 {
 	DEFiRet;
 	smsg_t *pMsg;
+ 	uchar file_offset[MAX_OFFSET_REPRESENTATION_NUM_BYTES+1];
+ 	const uchar *metadata_names[2] = {(uchar *)"filename",(uchar *)"fileoffset"} ;
+ 	const uchar *metadata_values[2] ;
 
 	if(rsCStrLen(cstrLine) == 0) {
 		/* we do not process empty lines */
@@ -565,8 +569,12 @@ static rsRetVal enqLine(lstn_t *const __restrict__ pLstn,
 	MsgSetTAG(pMsg, pLstn->pszTag, pLstn->lenTag);
 	msgSetPRI(pMsg, pLstn->iFacility | pLstn->iSeverity);
 	MsgSetRuleset(pMsg, pLstn->pRuleset);
-	if(pLstn->addMetadata)
-		msgAddMetadata(pMsg, (uchar*)"filename", pLstn->pszFileName);
+	if(pLstn->addMetadata) {
+ 		metadata_values[0] = pLstn->pszFileName;
+ 		snprintf((char *)file_offset,MAX_OFFSET_REPRESENTATION_NUM_BYTES+1, "%lld",pLstn->pStrm->iCurrOffs);
+ 		metadata_values[1] = file_offset ;
+ 		msgAddMultiMetadata(pMsg, metadata_names, metadata_values ,2);
+	}
 	ratelimitAddMsg(pLstn->ratelimiter, &pLstn->multiSub, pMsg);
 finalize_it:
 	RETiRet;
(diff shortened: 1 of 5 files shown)


derekjsonar:
It would be great for this PR to make the v8.28 milestone, if possible. 

As per the CONTRIBUTING guidlines I have added an rsyslog-doc PR here: 
https://github.com/rsyslog/rsyslog-doc/pull/348

If there's something else I can do to make the v8.28 milestone please advise.

rgerhards:
We are already past the 8.28 merge frame; however I have some more late merge, so it might be possible...

rgerhards:
The PR basically looks pretty good, but I would like to have a testbench test to ensure the new functionality works as expected.

rgerhards:
@davidelang do you think this should be optional so that we save the write in those (potentially frequent) cases the file offset is not needed? I tend to think "no", because in the imfile context this is still pretty lightweight, and may affect only very heavy users... which would make sense to wait until somone complains.

davidelang:
I think there is far more of a win working to improve all variable performance than in adding options to enable and disable different pieces of metadata. There is a cost to every option added, and I don't see a lot of value in disabling this.

derekjsonar:
If I add a script alongside one of the tests/imfile-wildcards* that verifies some of the fileoffset functionality is that automagically registered as a testbench test? I'm happy to give this a shot but I'm unfamiliar with your testbench system -- if you have any pointers for me or more documentation, please let me know. Thanks!

rgerhards:
That would be great. Actually all you need to do is copy such a test as proposed, be sure it checks the offset. It does not auto-register, but it's a simple to do: open ./testbench/Makefile.am and search for existing test. Add new test to the same spots. That's all. The CI system will then automatically apply the new tests. It is advised to use ```make check TESTS="new-test.sh"``` to verify the new test locally before pushing to the PR. Any questions, just let me know.

rgerhards:
One more question as I have not been able to dig into the details: what exactly is the offset showing? I would assume the file offset where the message started. Is that right? If so, that should probably be explicitly mentioned in the doc.

Is it obeyed correctly in all cases, e.g. if imfile is doing block reads because the unprocessed data is large?

derekjsonar:
Ok -- I'll work on adding some tests.

As for the behaviour, if you had 
```
$template offsets,"%$!metadata!filename%^$!metadata!fileoffset%^%msg%"
```

And we had an imfile input watching `test.txt` with the following lines:
```
hello
rsyslog
fileoffsets
```

The resulting output messages would be:
```
test.txt^6^hello
test.txt^14^rsyslog
test.txt^26^fileoffsets
```

So the fileoffset is the number of bytes including the new line of all preceding input messages including the current one.

I'm not sure how to trigger block reads -- if that's one of the existing testbench examples, I can incorporate that.

rgerhards:
Block reads happen if the file contains more than one line since the last read. The simplest way to trigger it is to create a file with a couple of lines before startup. I think the after-the-msg offset is a bit unusual, though I understand it's simplicity. All in all, I get the impression we should not rush merging this after the merge window has closed but give some room for more comments.

rgerhards:
@kortemik I'd really love to get your feedback on the handling of offset here.

rgerhards:
@derekjsonar sorry, too late for 8.28. Anyways, 8.29 will be out 8/8 ;-)

davidelang:
I understand that at the time the metadata is created, the line has been read 
in, so the pointer address you are reading is at the end of the line.

But if we have trivial access to the line length, I think we should subtract it 
so that the offset points at the beginning the line we are reporting.

By trivial access, I mean if we have the length in a variable already. If we 
have to run a loop to count the bytes in some string to get the length, it's 
probably not worth it.

Or, if we can move the creation of the metadata variable up to the beginning of 
the read rather than the end, so that it grabs the data before reading the 
line...

I know that when using this to troubleshoot what's going wrong when reading a 
file, having it 'point at the start of the next line' instead of pointing at the 
start of the current line would confuse me for a bit.

rgerhards:
@davidelang I tend to merge this patch as experimental with the doc explicitly saying the offset calculation might change in the future. I guess I won't be able to look into that before 8.29 release window close (tomorrow) but I think the patch is useful as is.

Comments anyone?

davidelang:
sounds reasonable.
kk

kortemik:
I think we can test this for now, but need to see into the other chunking
feature later!

-Mikko

rgerhards:
Do **NOT** close #1633, it will automatically be merged and closed when this PR here is merged.
I have based new PR #1758 on this PR; however it contains the change to make fileoffset point to the begin of the message, not the end.

PR #1686, opened 2017-07-21
omfwd: Use compression level from config

CodeHarsh:
omfwd reads compression level from config but deflateInit has hardcoded level 9


rgerhards:
good catch! Please squash the commits, as one just correct the other.

CodeHarsh:
Done

rgerhards:
thx!

PR #1695, opened 2017-07-25
imtcp: ratelimit based on a key string parameter

wdauchy:
This patch is a small proof of concept as a starting point for discussion.

This would add ratelimit.key parameter, e.g:
 ratelimit.key="procid|appname"
in order to ratelimit on the given key.

The key could be, as in this code,  a given parameter (e.g appname); but ultimately, a combinaison of multiple parameters (example: hostname + application) and allow us to define limits per group. 

the patch is not finished yet but gives an overview of how the code would look like. The same approach could be applied to udp.

How do you feel about it? Do you think this feature would be acceptable?

### DIFF
--- a/plugins/imtcp/imtcp.c
+++ b/plugins/imtcp/imtcp.c
@@ -118,6 +118,7 @@ struct instanceConf_s {
 	sbool bSPFramingFix;
 	int ratelimitInterval;
 	int ratelimitBurst;
+	uchar *ratelimitKey;
 	int bSuppOctetFram;
 	struct instanceConf_s *next;
 };
@@ -188,7 +189,8 @@ static struct cnfparamdescr inppdescr[] = {
 	{ "supportoctetcountedframing", eCmdHdlrBinary, 0 },
 	{ "ratelimit.interval", eCmdHdlrInt, 0 },
 	{ "framingfix.cisco.asa", eCmdHdlrBinary, 0 },
-	{ "ratelimit.burst", eCmdHdlrInt, 0 }
+	{ "ratelimit.burst", eCmdHdlrInt, 0 },
+	{ "ratelimit.key", eCmdHdlrString, 0 },
 };
 static struct cnfparamblk inppblk =
 	{ CNFPARAMBLK_VERSION,
@@ -285,6 +287,7 @@ createInstance(instanceConf_t **pinst)
 	inst->bSPFramingFix = 0;
 	inst->ratelimitInterval = 0;
 	inst->ratelimitBurst = 10000;
+	inst->ratelimitKey = NULL;
 
 	/* node created, let's add to config */
 	if(loadModConf->tail == NULL) {
@@ -386,7 +389,8 @@ addListner(modConfData_t *modConf, instanceConf_t *inst)
 	CHKiRet(tcpsrv.SetOrigin(pOurTcpsrv, (uchar*)"imtcp"));
 	CHKiRet(tcpsrv.SetDfltTZ(pOurTcpsrv, (inst->dfltTZ == NULL) ? (uchar*)"" : inst->dfltTZ));
 	CHKiRet(tcpsrv.SetbSPFramingFix(pOurTcpsrv, inst->bSPFramingFix));
-	CHKiRet(tcpsrv.SetLinuxLikeRatelimiters(pOurTcpsrv, inst->ratelimitInterval, inst->ratelimitBurst));
+	CHKiRet(tcpsrv.SetLinuxLikeRatelimiters(pOurTcpsrv, inst->ratelimitInterval, inst->ratelimitBurst,
+		inst->ratelimitKey));
 	tcpsrv.configureTCPListen(pOurTcpsrv, inst->pszBindPort, inst->bSuppOctetFram, inst->pszBindAddr);
 
 finalize_it:
@@ -439,6 +443,8 @@ CODESTARTnewInpInst
 			inst->ratelimitBurst = (int) pvals[i].val.d.n;
 		} else if(!strcmp(inppblk.descr[i].name, "ratelimit.interval")) {
 			inst->ratelimitInterval = (int) pvals[i].val.d.n;
+		} else if(!strcmp(inppblk.descr[i].name, "ratelimit.key")) {
+			inst->ratelimitKey = (uchar*)es_str2cstr(pvals[i].val.d.estr, NULL);
 		} else {
 			dbgprintf("imtcp: program error, non-handled "
 			  "param '%s'\n", inppblk.descr[i].name);
--- a/runtime/msg.c
+++ b/runtime/msg.c
@@ -2335,7 +2335,7 @@ msgGetJSONMESG(smsg_t *__restrict__ const pMsg)
 	jval = json_object_new_string((char*)pRes);
 	json_object_object_add(json, "timereported", jval);
 
-	jval = json_object_new_string(getHOSTNAME(pMsg));
+	jval = json_object_new_string(getHOSTNAME(pMsg, 0));
 	json_object_object_add(json, "hostname", jval);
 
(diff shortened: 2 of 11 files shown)


rgerhards:
Sorry for the long silence, I've been out of office for a couple of days. The PoC looks good to me, and the feature seems very interesting. The only detail problem (not necessarily a small one) is that it needs a fully parsed message to handle what it does. I have not checked in depth if that works with the proposed patch (but can if you would like). This also has some performance implications, as parsing usually works off the input thread.

Let me know how you would like to proceed.

davidelang:
would it make sense to have this based off of a dyn_stats() setting? that still 
requires processing the message to that point first, but it gives more 
flexibility to do different things than just throwing the message away (counting 
how many are thrown away, writing to a local file instead of sending remote, 
implementing random drop after a given number have shown up, etc)

wdauchy:
@rgerhards thanks for your answer.
I indeed moved de `ParseMsg` call inside the input module; I understand the performance implications of it but I don't think it is fundamentally different from the current default ratelimiting in place.
currently we have:
`defaultDoSubmitMessage -> ratelimitAddMsg -> ParseMsg`
what I propose is:
`defaultDoSubmitMessage -> ParseMsg -> ratelimitAddMsg` (which does not call `ParseMsg` later)

The same thread is handling the same thing; but maybe I am missing something.

@davidelang thank you for the `dyn_stats` suggestion; I will have a look to see how we can improve things.

davidelang:
On Tue, 1 Aug 2017, William Dauchy wrote:

> @davidelang thank you for the `dyn_stats` suggestion; I will have a look to see how we can improve things.

The reason I suggested not doing this during the input module is that a lot of 
the time people want to filter based on something in the content of the message 
(frequently something that is parsed out of the message via mmnormalize), not 
just filtering on the standard properties.

once you start looking into the body of the message, just matching a simple 
string really isn't good enough (too high a chance of a false positive)

rgerhards:
@wdauchy @davidelang do you (still) feel comfortable with me merging the PR? Or should we discuss some more? I ask because the merge window for 8.29 closes tomorrow and I prefer to avoid late merges on  Mondays. So if I do not merge yet, it'll be 8.30.

From my personal perspective, I like the proposed patch because it is easy to use. The performance implications could also be addressed by doing parsing on receiver thread only when the option is set. We could add this now or in 8.30 (if we merge as-is).

Please let me know your thoughts.

rgerhards:
I tenatively attach the 8.29 milestone just so that I do not forget to check the PR in my final release step.  I can always move to 8.30.

davidelang:
First off, my apologies for not looking at this earlier and raising issues then.

My concerns are less about this implementation than they are about the long term approach.

This only works with imtcp, not with imptcp, imrelp, etc

This currently only works with appname/procid (which it treats as hostname, which is not correct), but even the discussion talks about wanting to match other things (appname + hostname)

When it rate-limits, does it drop messages? or just slow them down? Is this configurable? (should this be configurable??)

other than debug level messages, is there any way to know that rate limiting took place?

Even as fully expanded, this isn't going to be able to rate limit during input based on fields that need mmnormalize to parse from the body of the message, but people are going to want to rate limit based on those things.

What is the final code going to end up looking like when it has lots of different possible keys and combinations of keys? How much work is it going to be to make this a general thing across all inputs?

I hate to be a naysayer, but I think we are probably better holding off one release rather than adding a feature that may change drastically going forward (just from a maintainability point of view), unless there is a way to clearly mark this as "Experimental, may change or disappear"

davidelang:
Thinking out loud about the big picture around this a bit more.

the reason for wanting to rate limit at input time instead of at log processing time is to prevent a flood of messages from overwelming the main queue and blocking other, more useful messages.

At input time, we have separate connections for different sources (at least for TCP or imfile based methods, UDP messages are mixed before we have a chance to do anything)

My knee jerk reaction to this (as I noted above) is to do this as part of normal queue processing, you could setup a ruleset (with it's own queue) for a given input, and in that ruleset, do mmnormalize or other processing to extract the data you want to rate limit on, and then use a mechanism like dyn_stats to count the messages in each bucket that you want to rate limit on, and have the flexibility to take lots of actions after you have decided to rate limit on a message (but one thing you cannot do is slow down that message and let others past it, which may be an important use case). After this decision, you can then call() the main ruleset to move the message into the main queue.

So per-source (i.e. per fromhost-ip) throttling (i.e. only process X messages/min from any one host, but don't discard messages faster than that, just slow them down to let other hosts have a fair chance) needs to be done in the input stage, but I'm not sure any rate limiting that discards messages or looks at any other data belongs in the input module.

rgerhards:
@davidelang Good points. Even more so as @wdauchy flagged this is a discussion starter. I'll move the release goal to 8.30 and then let's see what we discuss. I just want to start the actual part of the discussion a bit later, as I am busy with getting release-ready and also setting up some new work over here.

wdauchy:
@rgerhards @davidelang indeed it was flagged as a discussion starter, so definitively not for v8.29; also the current patch is not finished (only supporting procid and appname as an example); also only imtcp as a POC (but could be seen as experimental from the protocol point of view)


@davidelang
thanks for all your feedbacks, there is a lot to discuss.

My original plan to address the two subjects in two different patch sets.
- rate limit for a given key (as an example, a cluster sending lots of logs from different machines)
- improve queue consuming per-source (instead of consuming everything possible in a given socket, read part of it, and continue with the next one, in order to give more chances to other talkers to be able to send logs). We currently have the wrong behaviour with a probe testing the log system, not able to send any message while another machine is taking all the traffic.

currently messages are dropped, as it is handled in the current ratelimiting code; so no change of behaviour; debug messages are from the current ratelimit code as well.

I disagree about being able to ratelimit based on a key from the body (through mmnormalize); that was not the original plan and I fear this would make the code too complicated.
the combinaison of keys was a proposition, but I did not thought much about it; complexity should be determined I agree. I was thinking about a use case with different application using "httpd" as the appname and so being able to ratelimit on appname + hostip
maybe we should stick to defined syslog fields, as a first version.
I however did not found my mistake about "appname/procid (which it treats as hostname, which is not correct)"; could you point me this out?

This patch is trying to improve the current ratelimiting in place as a first step by adding a key. I also agree the current situation is not ideal since each input module is calling this ratelimit function. We may also think about completely moving the current ratelimiting to another design, but I am not quite sure this is doable regarding the back compatibility. If I understood correctly you were mentioning the possibility to be implemented inside rulesets, am I wrong?
At the end I am not sure if you are completely against this patch set, or if may continue to improve it. 
As a second step I still want to address the second point, i.e improve queue consuming per-source.

wdauchy:
@davidelang any thoughts about my previous message?

wdauchy:
@rgerhards what do we do with that? I was hoping to target 8.32 this time.
since we talked about it, I added support for appname/programname/hostname/procid

I think that's a good start to see there is an echo about that functionality, what do you think?

rgerhards:
give me a bit time to really think it over and have a good look at it - I am preping for next week's release and also fiddeling with the CI system at the moment, so I would not be as concentrated on the issues right now as it deserves. Sounds OK to you?

wdauchy:
yes, sounds good, thanks @rgerhards

wdauchy:
@rgerhards could we take some time for this patch?

wdauchy:
ping @rgerhards

rgerhards:
Sorry, for this release time frame I am more than swamped :-( containers,imfile rewrite, packaging, bug fixes ... I guess I actually don't manage to get to all of it...

InformatiQ:
@rgerhards does abandonned mean you are not agreeing to the work done here? 
@wdauchy

rgerhards:
no it means nobody has been working on this PR for quite a while. That means sombody needs to adopt it. I didn't want to close it right now. Will do so in a while if nobody adopts it.

Issue #1546, opened 2017-05-09
[omelasticsearch] Use GET instead of HEAD for health check


Nayar:
Hi, 

According to my HAProxy graphs, I'm getting lots of 404 HTTP status on my elasticsearch. It turns out rsyslog's module omelestaic does lots of `HEAD` on `/_cat/health`

![spectacle j24314](https://cloud.githubusercontent.com/assets/454815/25839303/280ad7ee-34a8-11e7-93cf-db5386d55f0d.png)

![spectacle m24314](https://cloud.githubusercontent.com/assets/454815/25839315/378ddd74-34a8-11e7-930c-96a9ffc02a78.png)

A manual curl on the elasticsearch give me this
```
nayar@sys-asus-nayarj:~$ curl -u xxx:xxx http://es1:9200/_cat/health -X HEAD -v
Warning: Setting custom HTTP method to HEAD with -X/--request may not work the 
Warning: way you want. Consider using -I/--head instead.
*   Trying 100.64.111.41...
* Connected to es1 (100.64.111.41) port 9200 (#0)
* Server auth using Basic with user 'xxx'
> HEAD /_cat/health HTTP/1.1
> Host: es1:9200
> Authorization: Basic xxxxxxxx
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 404 Not Found
< Warning: 299 Elasticsearch-5.3.0-3adb13b "[HEAD /{index}/{type}] is deprecated! Use [HEAD /{index}/_mapping/{type}] instead." "Tue, 09 May 2017 07:12:20 GMT"
< content-type: text/plain; charset=UTF-8
< content-length: 0
< Connection: close
< 
* Closing connection 0
```

ES Version: 5.3.0
RSYSLOG: 8.24.0

jgerhards:
i have tried to fix this now, but since i don't have access to an adequate testing environment i can't test it. Could you give it a try and report back?

wdauchy:
@jgerhards this would break support for ES2 clusters, you need to introduce a version check or parameter

wdauchy:
@jgerhards I had a closer look and I don't know where you found out about "_cat/_mapping/health". "{index}/_mapping/{type}" is used for checking a specific index.

"_cat/health" still exists and is still the reference https://www.elastic.co/guide/en/elasticsearch/reference/master/cat-health.html

tested on a local ES5.4.3 version

The issue here is about switching from HEAD to GET request

[...]

PR #1697, opened 2017-07-26
omelasticsearch: avoid 404 during health check

wdauchy:
For http(s), CURLOPT_NOBODY makes libcurl do a HEAD request.
this procudes a "HTTP/1.1 404 Not Found" instead of an expected
"HTTP/1.1 200 OK".
see https://curl.haxx.se/libcurl/c/CURLOPT_NOBODY.html

remove this option to produce a GET request

Fixes: issue #1546

### DIFF
--- a/plugins/omelasticsearch/omelasticsearch.c
+++ b/plugins/omelasticsearch/omelasticsearch.c
@@ -1297,7 +1297,6 @@ static void
 curlCheckConnSetup(CURL *handle, HEADER *header, long timeout, sbool allowUnsignedCerts)
 {
 	curl_easy_setopt(handle, CURLOPT_HTTPHEADER, header);
-	curl_easy_setopt(handle, CURLOPT_NOBODY, TRUE);
 	curl_easy_setopt(handle, CURLOPT_TIMEOUT_MS, timeout);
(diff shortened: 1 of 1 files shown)


rgerhards:
just out of curiosity: does ES really indicated a 404 for the same ressource when doing a HEAD request vs. when doing a GET request? I would assume that is against the HTTP RFCs.

Nevertheless, thanks for the patch series!

wdauchy:
@rgerhards indeed. tested on both ES2 and ES5:

```
$ curl "http://elasticsearch5/_cat/health" -I -X GET
HTTP/1.1 200 OK
content-type: text/plain; charset=UTF-8
content-length: 60
```

```
$ curl "http://elasticsearch5/_cat/health" -I -X HEAD
HTTP/1.1 404 Not Found
Warning: 299 Elasticsearch-5.4.3-eed30a8 "[HEAD /{index}/{type}] is deprecated! Use [HEAD /{index}/_mapping/{type}] instead." "Tue, 01 Aug 2017 20:04:29 GMT"
content-type: text/plain; charset=UTF-8
content-length: 0
```

I may have a look in ES code to understand this difference and see if it's fixable for the long term

Issue #1668, opened 2017-07-14
omelasticsearch "errorfile" and "bulkId" memory leak


GUI:
After upgrading from 8.27.0 to 8.28.0, we started to see rsyslogd's memory use to grow indefinitely. The more messages rsyslogd processes, the more it's memory grew. I've tracked the issue down to our usage of omelasticsearch, but I think there might be 2 separate issues going on:

- When omelasticsearch's `bulkId` setting is used, rsyslogd 8.28.0 considers successful responses from Elasticsearch to be errors. I guess this setting isn't documented, but we had been using it in 8.27.0 and previous versions successfully.
- If omelasticsearch's `errorfile` is defined, then the errors generated seem to trigger a memory leak in rsyslogd. It seems like rsyslogd's memory growth corresponds to the amount of error data logged in the error file. I'm not sure if this is a new problem in 8.28.0, or whether we just hit it since the `bulkId` issue caused a lot more error data to be generated.

So it seems like the memory leak issue might really just be specific to `errorfile`, but using `bulkId` is now logging all responses as errors in 8.28.0 (despite the indexing actually being successful).

The `bulkId` issue seems to stem from this check for the `create` attribute on the response: https://github.com/rsyslog/rsyslog/blob/v8.28.0/plugins/omelasticsearch/omelasticsearch.c#L746 When the `bulkId` option is used to explicitly set the `_id` of the document being sent to elasticsearch, then Elasticsearch uses the `"index"` bulk mechanism, instead of `"create"` (see https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html). This in turn means the response will not be nested under the `create` attribute, but instead it's nested under an `index` attribute.

I put together this little repo with some scripts to demonstrate the memory leak with the different combinations of settings and versions: https://github.com/GUI/rsyslog-omelasticsearch-leak

wdauchy:
I proposed a possible fix in #1699

ajrpayne:
Not using omelasticsearch's bulkId setting, but rsyslogd 8.28.0 seems to be considering successful responses from Elasticsearch as errors. No memory leak though for me. downgraded to 8.27.0 to fix the issue.

rgerhards:
fixed via https://github.com/rsyslog/rsyslog/pull/1699

PR #1699, opened 2017-07-27
omelasticsearch: fixing leak while writing error file

wdauchy:
since we switched to libfastjson, the default error function is
duplicating the string from fjson_object_to_json_string in order to free
the fjson object.
we however don't free the duplicated string later.

fix that by adding a free
also add a second duplication for the second case of rendered string;
this makes the code easier to read instead of adding another condition.

Fixes: issue #1668

### DIFF
--- a/plugins/omelasticsearch/omelasticsearch.c
+++ b/plugins/omelasticsearch/omelasticsearch.c
@@ -1039,7 +1039,7 @@ writeDataError(wrkrInstanceData_t *pWrkrData, instanceData *pData, fjson_object 
 			DBGPRINTF("omelasticsearch: error creating file content.\n");
 			ABORT_FINALIZE(RS_RET_ERR);
 		}
-		rendered = (char*)fjson_object_to_json_string(ctx.errRoot);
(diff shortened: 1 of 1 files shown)


plugins/omelasticsearch/omelasticsearch.c
```diff
@@ -1039,7 +1039,7 @@ writeDataError(wrkrInstanceData_t *pWrkrData, instanceData *pData, fjson_object
 			DBGPRINTF("omelasticsearch: error creating file content.\n");
 			ABORT_FINALIZE(RS_RET_ERR);
 		}
-		rendered = (char*)fjson_object_to_json_string(ctx.errRoot);
+		rendered = strdup((char*)fjson_object_to_json_string(ctx.errRoot));
```
rgerhards, line 1042: This is sub-optimal from a performance point of view; however, we should not run into this situation too frequently. So IMHO code simplicity wins (as suggested).

wdauchy: indeed. I spent a few moments asking myself what would be the best solution; but for now I think that the best we can do to fix the issue quickly. We might think about some code re organisation but I believe this can wait.

PR #1705, opened 2017-08-07
add debug.file and debug.whitelist parameters

jgerhards:
add parameters to configure debug-output to only be shown if it is generated by specific files.


rgerhards:
debian fails with environment induced issues (too old librelp as it seems). So merging anyways.

rgerhards:
merged in 8.29.0 - unfortunately selected wrong milestone tag

PR #1709, opened 2017-08-09
Fix cross-compile errors checking for atomic ops

jer-gentoo:


jer-gentoo:
I think AC_COMPILE_IFELSE requires and AC_LANG_PROGRAM definition as first argument.

PR #1713, opened 2017-08-10
Allow cross-compiling by using AC_COMPILE_IFELSE

jer-gentoo:


jer-gentoo:
Perhaps AC_LINK_IFELSE works better in some cases?

rgerhards:
This unfortunately looks like it does not work...

jer-gentoo:
What makes you say that?

rgerhards:
> What makes you say that?

The failed CI tests ;-)

jer-gentoo:
Which one of those are related directly to this patch? I see a missing symbol in imrelp, and there is some weird stuff on Solaris probably attributable to choosing AC_COMPILE_IFELSE instead of AC_LINK_IFELSE (as noted above - does Solaris need a specific library to implement atomic operations?).

rgerhards:
Sorry, I forgot to restart the debian buildbot slave. This one had a problem. Other than that, I am not an autotools expert. All I know that without the patch, everything passes CI, with it, that's not the case. So we need someone who makes it CI-clean again in order to accept the patch.

jer-gentoo:
It is not apparent what further contributions you require. Closing the PR makes it less visible to people who might work on the Solaris problem.

rgerhards:

Issue #1722, opened 2017-08-16
imtcp: priorityString does not work


PascalWithopf:
When setting the priorityString for GnuTLS in imtcp, set in the parameter gnutlsPriorityString, it is never used.

PR #1724, opened 2017-08-16
imtcp: fix usage of priorityString

PascalWithopf:
fixes https://github.com/rsyslog/rsyslog/issues/1722


rgerhards:
@PascalWithopf could you please elaborate a bit more in the commit message which bug this fixes

Issue #1719, opened 2017-08-15
Compilation of 8.29.0 fails in imptcp.c


Hoshpak:
While trying to cross-compile rsyslog 8.29.0 using gcc 6.3.0 on Void Linux for the aarch64 architecture, I encountered the following error in imptcp.c:

```
Making all in plugins/imptcp
make[2]: Entering directory '/builddir/rsyslog-8.29.0/plugins/imptcp'
  CC       imptcp_la-imptcp.lo
In file included from ../../runtime/prop.h:25:0,
                 from ../../runtime/glbl.h:38,
                 from ../../runtime/stream.h:73,
                 from ../../runtime/obj.h:48,
                 from ../../runtime/rsyslog.h:634,
                 from imptcp.c:61:
imptcp.c: In function 'AcceptConnReq':
imptcp.c:831:39: error: 'pThis' undeclared (first use in this function)
  STATSCOUNTER_INC(pLstn->ctrSessOpen, pThis->pLstn->mutCtrSessOpen);
                                       ^
../../runtime/atomic.h:206:22: note: in definition of macro 'ATOMIC_INC_uint64'
   pthread_mutex_lock(phlpmut); \
                      ^~~~~~~
imptcp.c:831:2: note: in expansion of macro 'STATSCOUNTER_INC'
  STATSCOUNTER_INC(pLstn->ctrSessOpen, pThis->pLstn->mutCtrSessOpen);
  ^~~~~~~~~~~~~~~~
imptcp.c:831:39: note: each undeclared identifier is reported only once for each function it appears in
  STATSCOUNTER_INC(pLstn->ctrSessOpen, pThis->pLstn->mutCtrSessOpen);
                                       ^
../../runtime/atomic.h:206:22: note: in definition of macro 'ATOMIC_INC_uint64'
   pthread_mutex_lock(phlpmut); \
                      ^~~~~~~
imptcp.c:831:2: note: in expansion of macro 'STATSCOUNTER_INC'
  STATSCOUNTER_INC(pLstn->ctrSessOpen, pThis->pLstn->mutCtrSessOpen);
  ^~~~~~~~~~~~~~~~
make[2]: *** [Makefile:527: imptcp_la-imptcp.lo] Error 1
make[2]: Leaving directory '/builddir/rsyslog-8.29.0/plugins/imptcp'
make[1]: *** [Makefile:815: all-recursive] Error 1
make[1]: Leaving directory '/builddir/rsyslog-8.29.0'
[...]

PR #1725, opened 2017-08-16
correct mutex arg in imptcp

petiepooo:
Remove the 'pThis->' from the second argument to STATSCOUNTER_INC macro.
This only affects architectures without atomic uint64 actions, like aarch64.
Others do not use the mut argument, and it gets compiled out.

Resolves issue #1719.  Thank you @Hoshpak for reporting.

### DIFF
--- a/plugins/imptcp/imptcp.c
+++ b/plugins/imptcp/imptcp.c
@@ -828,7 +828,7 @@ AcceptConnReq(ptcplstn_t *pLstn, int *newSock, prop_t **peerName, prop_t **peerI
 		LogMsg(0, RS_RET_NO_ERRCODE, LOG_INFO, "imptcp: connection established with host: %s", propGetSzStr(*peerName));
 	}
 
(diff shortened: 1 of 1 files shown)


petiepooo:
I don't know why those build checks failed.  Should I be targeting this pull request to a specific build or commit level?  Or can we just trigger the tests a second time?

rgerhards:
I am puzzled, too. I've restarted Centos7. Let's see how it goes...

petiepooo:
Thanks. Looks like that worked. 
Unfortunately, I suspect all the systems tested have atomic uint64 instructions so would not hit the fault.

petiepooo:
Weird.. The "Details" link shows the ubuntu build passed, but it still shows pending below.  Perhaps a glitch from github's outage this morning..

rgerhards:
> Perhaps a glitch from github's outage this morning..

Guess so. The buildbot integration doesn't really like outages. I've re-started the test just to make sure the status shows correctly.

Issue #1718, opened 2017-08-15
[Improvement] It's impossible to guess by error message from which input large message was received


jay7x:
I've found those messages today in log:
```
rsyslogd[12306]: error: message received is at least 1429 byte larger than max msg size; message will be split starting at: " <149>2017-08-15T08:56:36.779453"  [v8.27.0]
rsyslogd[12306]: error: message received is at least 837 byte larger than max msg size; message will be split starting at: " <150>2017-08-15T08:56:43.599000"  [v8.27.0]
```

I have multiple inputs configured on this instance of rsyslog. So I'm unable to guess on which input rsyslog got those large messages. Is it possible to add input name to error message?

PascalWithopf:
These messages probably come from imptcp.
I will change the message to include the module name.

rgerhards:
well, it's already done ;-)

* https://github.com/rsyslog/rsyslog/pull/1678
* https://github.com/rsyslog/rsyslog/pull/1677

Part of 8.29.0.

rgerhards:
sorry, my bad. I misread the request.

@PascalWithopf please go ahead, makes a lot of sense

jay7x:
Hi @PascalWithopf! I've checked your MR. There is a bit more. I know those messages come from imptcp module. But I have 6 imptcp inputs on those machine. They have different names (name="xxx"). This is the name I'd like to see in error message to know which input to debug.

PascalWithopf:
@jay7x I changed the message so it includes now at the beginning first the module name and then the parameter name.

PR #1726, opened 2017-08-17
imptcp/imtcp: include module name in error msg

PascalWithopf:
closes https://github.com/rsyslog/rsyslog/issues/1718


rgerhards:
@PascalWithopf This looks good, but we should also change the equivalent message in imtcp. Can you do a PR for this,too?

PR #1732, opened 2017-08-20
Fix Elasticsearch response handling

### LINKED ISSUE #1731: omelasticsearch: Every request is logged into the error file when bulkmode is enabled
When I enable "_bulkmode_" in the omelasticsearch plugin, Rsyslog seems to consider all responses from Elasticsearch as errors, even the successful ones. As a consequence, every single request ends up being logged into the error file.

If I'm not mistaken, this issue has also been reported by @GUI and @ajrpayne in #1668.

NB: I'm using version 8.29.0.

vquemener:
Fixes #1731

### DIFF
--- a/plugins/omelasticsearch/omelasticsearch.c
+++ b/plugins/omelasticsearch/omelasticsearch.c
@@ -746,9 +746,12 @@ parseRequestAndResponseForContext(wrkrInstanceData_t *pWrkrData,fjson_object **p
 		}
 		fjson_object_object_get_ex(item, "create", &result);
 		if(result == NULL || !fjson_object_is_type(result, fjson_type_object)) {
-			DBGPRINTF("omelasticsearch: error in elasticsearch reply: "
-				  "cannot obtain 'result' item for #%d\n", i);
-			ABORT_FINALIZE(RS_RET_DATAFAIL);
+			fjson_object_object_get_ex(item, "index", &result);
+			if(result == NULL || !fjson_object_is_type(result, fjson_type_object)) {
+				DBGPRINTF("omelasticsearch: error in elasticsearch reply: "
+					  "cannot obtain 'result' item for #%d\n", i);
+				ABORT_FINALIZE(RS_RET_DATAFAIL);
+			}
 		}
 
 		fjson_object_object_get_ex(result, "status", &ok);


rgerhards:
just FYI: there is an issue with CI. I don't see why, but I tried to restart the test three times now, so I think it is a valid failure.

vquemener:
I will look into this.

I wonder if this might be linked to the version of Elasticsearch. So far I ran some tests using v5.5.1; do you know which version is used by Travis?

rgerhards:
I admit I don't know. It's whatever they have in the OS package (which we consider to be default). We could consider to download a more current version (like we do with kafaka), but then someone would need to help us do that.

vquemener:
OK, it turns out that Travis is using v1.4 by default on Ubuntu 12.04, and Elasticsearch responses for bulk requests changed with version 5.

I adapted my commit to handle both cases. Let's see how the build goes.

rgerhards:
> I adapted my commit to handle both cases. Let's see how the build goes.

thanks, much appreciated!

rgerhards:
> I adapted my commit to handle both cases.

Can you let me know what you changed? I am asking because we could potentially use that info for further work ;-)

vquemener:
Elasticsearch responses for bulk requests used to look like this for version 2 :
```json
'{
    "took":1205,
    "errors":false,
    "items":[{
        "create":{
            "_index":"log-2017.08.25",
            "_type":"log",
            "_id":"AV4JLGyt91GzIgdyx6s7",
            "_version":1,
            "_shards":{
                "total":2,
                "successful":1,
                "failed":0
            },
            "status":201
        }
    }]
}'
```
Whereas they now look like this, starting from version 5 (there was no version between 2.4 and 5):
```json
'{
    "took":1829,
    "errors":false,
    "items":[{
        "index":{
            "_index":"log-2017.08.25",
            "_type":"log",
            "_id":"AV4JPZ_fquWCvYqrs7tN",
            "_version":1,
            "result":"created",
            "_shards":{
                "total":2,
                "successful":1,
                "failed":0
            },
            "created":true,
            "status":201
        }
    }]
}'
```

Note that the main difference is that the object "create" was renamed as "index".

The problem was precisely that the code was looking for the key "status" of this object "create". With my change, it now also tries to find the object "index" if the object "create" does not exist.

wdauchy:
@vquemener I would add that information in the commit message to better understand the context of it

vquemener:
Good idea! I have updated the pull request.

vquemener:
The logs seem to indicate that Buildbot did not succeed in fetching the pull request.

@rgerhards could you please restart the builds?

wdauchy:
@vquemener try to repush with a small change in your commit message, it will restart the build

rgerhards:
I can re-init the buildbot run.

Just as general advise on how to restart: just do a ```git commit --amend``` without doing any changes. That'll still change the git hash, and thus will re-init a new CI run.

rgerhards:
mhhh - I restarted one of the slaves and got the same error. Checking if there are connectivity issue to github (but doesn't look so...)

vquemener:
It's weird because Travis can fetch the code without any error.

Do you want me to amend the commit to restart all the builds?

rgerhards:
> It's weird because Travis can fetch the code without any error.
yup, I am also surprised. I even restarted one Travis test, and it worked...

> Do you want me to amend the commit to restart all the builds?
yes, please do. At least that'll may clean out anything that might have gone wild... Let's see what then happens.

vquemener:
That sounds better now!

rgerhards:
great! So I guess there was indeed a glitch in the buildbot database (which was cleared by the new commit).

dharrop:
I'm having a problem that may be related to this change.  rsyslogd prints elasticsearch status lines on stdout, I assume each time it writes batch records to elasticsearch.  Like this:

`1509080969 22:09:29 server1 green 1 1 600 600 0 0 0 0 - 100.0%`

My rsyslogd is started with start-stop-daemon so I initially didn't see that output, but rsyslogd would crash after a brief period of time.  I assume some buffer was filling with rsyslogd output and caused the crash.  I finally saw this output after running it from the command line in the foreground.

I worked around the problem by changing the parameters to start-stop-daemon to use **-1 /dev/null -2 /dev/null -b** to redirect stdout and stderr to /dev/null and force rsyslogd into the background while passing the **-n** flag to rsyslogd so those redirections can happen properly.

In addition to suppressing the elasticsearch status messages it might be a good idea to add configuration options to direct rsyslogd's stdout somewhere.

This problem happened after upgrading from rsyslog 8.28 and elasticsearch 5.0 to rsyslog 8.30 and elasticsearch 5.6.2.

rgerhards:
@dharrop Thankx for the bug report. This is already known as https://github.com/rsyslog/rsyslog/issues/1909

PR #1740, opened 2017-08-25
omprog: added error handling and transaction support for external plugins

jsiwrk:
I've implemented this planned feature myself as a personal challenge, to have some fun and learn. What's your opinion? Did you intend to implement it this way?

I have manually tested the patch and it seems to work very well. Also, the CI tests show no regression (although I'm not sure if the omprog module was sufficiently covered).

For my part the patch would be ready to merge, EXCEPT that I have not implemented any automated test. Sorry for that. I would need some guidance on how to do it. On the other hand I'm not sure if I will be able to dedicate enough time to this personal project in the short term, so it would be great if someone else could implement at least a first basic test.

Other pending things apart from the tests:
* Update the rest of the documentation. At least, add reference doc about the new omprog settings. I the patch is finally accepted I could contribute to this, I think it'll be easy.
* The clang analyzer gives a (presumably false) warning I don't know how to eliminate.
* I'm pretty confident the Python skeleton is sufficiently pythonic. However I'm a Python newbie, so it would be great if someone with more knowledge could review it.
* It seems rsyslog does not require the standard getline function to be available, and I didn't find any quality implementation of it within the rsyslog code. So I implemented my own getline within omprog.c, but probably should be placed somewhere else, since it is reusable.

Note: I had to reorder functions in omprog.c due to new dependencies between them. I realise this will make comparing changes a bit difficult, sorry for that. I've not attempted to minimize this with forward declarations since I haven't seen them used elsewhere.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -10,6 +10,13 @@ Version 8.30.0 [v8-stable] 2017-10-10
     closes https://github.com/rsyslog/rsyslog/issues/1717
 - RainerScript: add ltrim and rtrim functions
   closes https://github.com/rsyslog/rsyslog/issues/1711
+- omelasticsearch bugfix: incompatibility with newer ElasticSearch version
+  ElasticSearch changed its API in newer versions. When "bulkmode" is enabled
+  in omelasticsearch, rsyslog seems to consider all responses from Elasticsearch
+  as errors, even the successful ones. As a consequence, every single request
+  ends up being logged into the error file.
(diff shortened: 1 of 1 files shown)


rgerhards:
hi, can you check for the CI errors, please? Also, it would be good to squash the commits.

jsiwrk:
Submitting this PR again in #1753; I have cleaned up my fork.

Issue #1642, opened 2017-07-02
warnings when forwarding logs to Elastic Search 5


jostyee:
`rsyslogd 8.27.0/elasticsearch 5.4.2`

my rsyslog.conf snippet as below:
```
# parse json log
*.*    :mmjsonparse:

if ($parsesuccess == "OK") then {
    action(type="mmutf8fix")
    action(type="omelasticsearch"
        template="cee-log"
        server="elasticsearch_server"
        searchIndex="logstash-index"
        dynSearchIndex="on"
        bulkmode="on"
        queue.dequeuebatchsize="100"
        action.resumeretrycount="-1"
        errorfile="/var/log/omelasticsearch.log")
}
```

get warning logs when elasticsearch receives requests:

```log
[2017-07-02T07:46:04,937][WARN ][org.elasticsearch.deprecation.rest.action.admin.indices.RestTypesExistsAction] [HEAD /{index}/{type}] is deprecated! Use [HEAD /{index}/_mapping/{type}] instead.
[2017-07-02T07:46:04,939][WARN ][org.elasticsearch.deprecation.rest.RestController] Content type detection for rest requests is deprecated. Specify the content type using the [Content-Type] header.
```

seems es has changed their APis but rsyslog still using the deprecated ones.

rgerhards:
duplicate of https://github.com/rsyslog/rsyslog/issues/1546

jostyee:
@rgerhards thanks for your QUICK reply, close this now.

jostyee:
@rgerhards I'm sorry, #1546 issued the `[HEAD /{index}/_mapping/{type}]` API change, but there is another `[Content-Type]` warning, I'm going to reopen this one.

rgerhards:
I see, it just looked similar. Looks like ES is having a pretty unstable API. Do you have a link to doc where the changes can be seen (and ideally how to best query the feature set)?

jostyee:
@rgerhards It's been change since 5.3: https://www.elastic.co/guide/en/elasticsearch/reference/5.3/docs-bulk.html

> When sending requests to this endpoint the Content-Type header should be set to application/x-ndjson.

jimmypw:
I have just encountered this issue, I believe the fix is simply to change the content type of the requests from text/json to application/json. I have made the change locally, I will test tonight.

wdauchy:
I've sent a proposal patch in #1698

jostyee:
[...]

PR #1743, opened 2017-08-30
omelasticsearch: avoid ES5 warnings while sending json in bulkmode

wdauchy:
my original patch attented to fix an deprecation warnings while sending
json related to content type.
I however wanted to improve my commit and support the mentioned x-ndjson
content type in case of bulk mode.
see
https://www.elastic.co/guide/en/elasticsearch/reference/master/docs-bulk.html

However, firstly my patch was wrong, since, in bulk mode, the x-ndjson was
set even for connection checks. Secondly I discovered that ES community
kind of disagree of what is the correct behavior:
 in RFC 6648 the x- prefix is already decided as "deprecated"
see https://github.com/elastic/elasticsearch/issues/25718

So I decided to put only application/json as content type; it is
supported for both simple request or bulk requests containing multiple
json. We can improve it when there is a final decision about it.

Re-Fixes: #1642
Fixes: commit 462be8b2ee4 (omelasticsearch: avoid ES5 warnings while sending json)

### DIFF
--- a/plugins/omelasticsearch/omelasticsearch.c
+++ b/plugins/omelasticsearch/omelasticsearch.c
@@ -1323,16 +1323,11 @@ curlPostSetup(CURL *handle, HEADER *header, uchar* authBuf)
 }
 
 #define CONTENT_JSON "Content-Type: application/json; charset=utf-8"
-#define CONTENT_JSON_BULK "Content-Type: application/x-ndjson; charset=utf-8"
 
 static rsRetVal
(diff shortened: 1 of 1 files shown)


rgerhards:
To the ES-literate folks here: any concerns merging this PR? For me it looks good to go...

taotetek:
@rgerhards - this looks ok and reasonable to me. application/json type works in bulk mode, given it sounds like there's some debates / unknowns about x-ndjson I think this is a reasonable solution.

mostolog:
LGTM

PR #1747, opened 2017-09-01
switching to persistent journal possible without rsyslog restart, 

jvymazal:
bumped corresponding systemd version dependency

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -115,6 +115,25 @@ static int bPidFallBack;
 static ratelimit_t *ratelimiter = NULL;
 static sd_journal *j;
 
+static rsRetVal persistJournalState(void);
+static rsRetVal loadJournalState(void);
+
+static rsRetVal openJournal(sd_journal* jj) {
+	DEFiRet;
+
+	if (sd_journal_open(&jj, SD_JOURNAL_LOCAL_ONLY) < 0)
+		iRet = RS_RET_IO_ERROR;
+	RETiRet;
+}
+
+static void closeJournal(sd_journal* jj) {
+
+	if (cs.stateFile) { /* can't persist without a state file */
+		persistJournalState();
+	}
+	sd_journal_close(jj);
+}
+
 
 /* ugly workaround to handle facility numbers; values
  * derived from names need to be eight times smaller,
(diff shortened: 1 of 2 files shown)


jvymazal:
@rgerhards  please, would this change be viable?

rgerhards:
I just had a look, but need to have a closer one. The change looks good on first view. However, you should squash the commits to get a clean commit history. I guess it's actually just a single commit from a logical PoV.

jvymazal:
I will squash the commits once the change is approved, thanks for your review.

plugins/imjournal/imjournal.c
```diff
@@ -472,12 +496,28 @@ pollJournal(void)
...
+		char* tmp = cs.stateFile;
+		cs.stateFile = NULL;
+		closeJournal(j);
+		cs.stateFile = tmp;
+
+		iRet = openJournal(j);
+		if (iRet != RS_RET_OK) {
+			char errStr[256];
+			rs_strerror_r(errno, errStr, sizeof(errStr));
+			errmsg.LogError(0, RS_RET_IO_ERROR,
```
rgerhards: just a small way to improve: we made ```LogError``` available without the need to prefix ```errmsg.``` - it's a little bit more efficient. Of course, all is well, but the new method is the preferred one.

plugins/imjournal/imjournal.c
```diff
@@ -472,12 +496,28 @@ pollJournal(void)
...
+				"sd_journal_open() failed: '%s'", errStr);
+			ABORT_FINALIZE(RS_RET_ERR);
+		}
 
+		iRet = loadJournalState();
+		errmsg.LogError(0, RS_RET_OK, "imjournal: journal reloaded...");
```
rgerhards: That's probably better to do as informational: ```LogMsg(0, RS_RET_OK, "imjournal: journal reloaded...");```

rgerhards:
@jvymazal oh, one thing: it would be good if you could amend the commit text with a good description of the actual motivation for the change and what it means to the average user.

rgerhards:
@jvymazal it looks like the journal version now requested is too recent for almost all distros. Can you work on graceful degradation if that version cannot be found?

jvymazal:
Ok, I will try to make it easily conditionable (at least temporarily before 234 systemd gets more widespread) and open a new PR with it.

rgerhards:
That's very good. I'll keep the current PR until the weekend. If the problems become to big, I'll temporarily revert it.

PR #1751, opened 2017-09-04
TEST ONLY, DO NOT MERGE

rgerhards:
This triggers a change against master


rgerhards:
misson acomplished, this proved the point we had some problems in plain master as well. Lead to #1754.

PR #1757, opened 2017-09-06
Fixed possible segfault. 

tammar96:
Fixed possible segfault.


rgerhards:
Note: there is a potential that I need to rever https://github.com/rsyslog/rsyslog/pull/1747 -- so I will only merge this patch if I can keep https://github.com/rsyslog/rsyslog/pull/1747

rgerhards:
Note: this was a regression of #1747, which was never released and merge only a couple of hours ago.

Issue #170, opened 2014-12-02
No paclages for epel-7 for latest 8.6.0 version


cristifalcas:
It seems that the packages for epel-7 are not build for latest versions

PR #1759, opened 2017-09-07
queue bugfix: file write error message was incorrect

rgerhards:
when a queue was restarted from disk file, it almost always
emitted a message claiming
"file opened for non-append write, but already contains xxx bytes"
This message was wrong and did not indicate a real error condition.
The predicate check was incorrect.

closes https://github.com/rsyslog/rsyslog/issues/170 (kind of)


sourcejedi:
If you wanted GitHub to auto-close the issue, I think you need to do it manually because you accidentally the last digit :).  (170 v.s. 1704).

rgerhards:
thx!

PR #1760, opened 2017-09-07
enabling falback to older systemd version without restart-less

jvymazal:
persistent journal switch.

This addresses a point raised in #1747


rgerhards:
Thanks a lot!

PR #1768, opened 2017-09-11
Support static linking `libmaxminddb` to plugin `mmdblookup`

zacharyzhao:
Update `configure.ac` and `mmdblookup`'s `Makefile.am` to support static link `libmaxminddb`'s static library.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -21,6 +21,8 @@ if test "$unamestr" = "AIX"; then
 	GNUTLS_CFLAGS="-I/usr/include"
 	RELP_LIBS="-L/usr/lib/ -lrelp"
 	RELP_CFLAGS="-I/usr/include"
+	LIBMAXMINDDB_LIBS="-L/usr/lib/ -lmaxminddb"
+	LIBMAXMINDDB_CFLAGS="-I/usr/include"
 	CC="xlc_r"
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -21,6 +21,8 @@ if test "$unamestr" = "AIX"; then
 	GNUTLS_CFLAGS="-I/usr/include"
 	RELP_LIBS="-L/usr/lib/ -lrelp"
 	RELP_CFLAGS="-I/usr/include"
+	LIBMAXMINDDB_LIBS="-L/usr/lib/ -lmaxminddb"
+	LIBMAXMINDDB_CFLAGS="-I/usr/include"
```
rgerhards, line 25: we can't hardcode this - see CI results. Doesn't it have a pkgconfig file?

rgerhards:
@zacharyzhao Do you still have interest in working on the PR?

rgerhards:
@Whissi Would you be willing to lend a helping hand? I think it would be desirable to have static linking capability. As of my (weak) understanding, I also think this should be possible just by proper configure options. Is this right? If so, how? ;-)

rgerhards:
nobody seems to have interest to move this forward ... so closing it.

PR #1772, opened 2017-09-12
mmanon: bugfix: IPv6 addresses anonymized by too few bits in random mode

jgerhards:


rgerhards:
Note: this did not affect pre-8.30 as IP6 is first introduced in 8.30.

PR #1798, opened 2017-09-27
mmanon: fix fall through warning

wdauchy:
mmanon.c:454:6: warning: this statement may fall through [-Wimplicit-fallthrough=]
    if(cyc == 0) {
      ^
mmanon.c:458:3: note: here
   default:
   ^~~~~~~

### DIFF
--- a/plugins/mmanon/mmanon.c
+++ b/plugins/mmanon/mmanon.c
@@ -455,6 +455,7 @@ isValidHexNum(const uchar *const __restrict__ buf,
 				(*nprocessed)++;
 				return -1;
 			}
+			return cyc;
 		default:
 			return cyc;
 		}


rgerhards:
I am not 100% sure what is the proper way to address this - maybe a bit more refactoring would make sense. I have asked @jgerhards for his opinion.

jgerhards:
Fallthrough would be better in my opinion, since if the ```if``` does not apply, the ':' just gets handled as any other character. Perhaps it would be better to use the ```/*FALLTHROUGH*/``` comment in this case (provided the compiler accepts it)?

wdauchy:
ok I will have a look

rgerhards:
@wdauchy I am also after this one, together with the others. If you'd like to fix it, I suggest you wait until I have added a new macro for this case -- the old style comment is not really a good solution looking forward. I first need to get a CI env up and running where I can use gcc 7, it's unfortunately rather tricky (e.g. Travis does not support it).

wdauchy:
ok I'll wait

rgerhards:
I had quite some trouble with the new gcc 7 and testbench. As such, I think it is best if I merge your PR as-is (so the credits go to the right one) and I'll update the patch later on. Let you know of the final version. Thanks again for looking into this.

wdauchy:
ok thanks

PR #1799, opened 2017-09-27
imptcp: simplify parsing in bind port

wdauchy:
I don't see in which case it could be useful to compare a pointer
a \0 char since we are already comparing it to NULL value just before

imptcp.c: In function ‘addInstance’:
imptcp.c:1519:35: warning: comparison between pointer and zero character constant [-Wpointer-compare]
  if((pNewVal == NULL) || (pNewVal == '\0')) {
                                   ^~
imptcp.c:1519:27: note: did you mean to dereference the pointer?
  if((pNewVal == NULL) || (pNewVal == '\0')) {
                           ^

### DIFF
--- a/plugins/imptcp/imptcp.c
+++ b/plugins/imptcp/imptcp.c
@@ -1515,8 +1515,6 @@ static rsRetVal addInstance(void __attribute__((unused)) *pVal, uchar *pNewVal)
 	CHKiRet(createInstance(&inst));
 	if(pNewVal == NULL || *pNewVal == '\0') {
 		errmsg.LogError(0, NO_ERRCODE, "imptcp: port number must be specified, listener ignored");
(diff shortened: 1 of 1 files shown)


wdauchy:
WDYT @rgerhards?

rgerhards:
Thanks for catching this. But the whole thing looks fishy... At least the error message does not do what it says, as it does not ignore the listener (it would need to err out). Not sure yet if we can really work without a listener port. So I need to dig somewhat further before I can merge.

rgerhards:
@wdauchy I'll merge this, but after that I'll do some more corrections. Review shows it's indeed a bit strange, but it may be the result of some backward compatibility issue...

wdauchy:
thanks!

PR #1800, opened 2017-09-28
ompgsql: Added DB server port parameter

yurayko:
This patch permit to specify a server db port. This parameter is required, existing configurations stop working and users must add a port number into their config files.

### DIFF
--- a/plugins/ompgsql/ompgsql.c
+++ b/plugins/ompgsql/ompgsql.c
@@ -343,7 +343,7 @@ CODE_STD_STRING_REQUESTparseSelectorAct(1)
 
 
 	/* sur5r 2007-10-18: added support for PgSQL
-	 * :ompgsql:server,port,dbname,userid,password
+	 * :ompgsql:server,dbname,userid,password,server_port
 	 * Now we read the PgSQL connection properties 
 	 * and verify that the properties are valid.
 	 */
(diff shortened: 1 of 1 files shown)


davidelang:
we really,really don't like to break existing configs, is there any way to have 
the parameter be optional and default to the old value?

yurayko:
I move parameter to the end and make this optional, as proposed.

plugins/ompgsql/ompgsql.c
```diff
@@ -364,8 +360,14 @@ CODE_STD_STRING_REQUESTparseSelectorAct(1)
 		iPgSQLPropErr++;
 	if(*pData->f_dbuid == '\0')
 		iPgSQLPropErr++;
-	if(getSubString(&p, pData->f_dbpwd, _DB_MAXPWDLEN+1, ';'))
+	if(getSubString(&p, pData->f_dbpwd, _DB_MAXPWDLEN+1, ','))
 		iPgSQLPropErr++;
+	if(*pData->f_dbpwd == '\0')
+		iPgSQLPropErr++;
+	if(getSubString(&p, pData->f_dbsrvport, _DB_MAXDBSRVPORTLEN+1, ';'))
```
rgerhards: This looks like it still requires the parameter and does not gracefully handle it's absence. So it still breaks backward compatibility.

yurayko: ping?

rgerhards: see my previous comment - I think it still applies. Did you also test without the new parameter? Unfortunately we have no testbench test for this.

rgerhards:
now by done by PR #2172, which includes much better functionality.

Issue #971, opened 2016-04-20
add ability to format unix date


davidelang:
Please add the ability to turn numeric values into timestamps

i.e. if you have $!foo = "1454527741.918"

be able to treat it as a timestamp and extract month, day, year, convert it to rfc3164 date, rfc3339 date, etc

either a rainerscript function to convert it or make the template date formatting code work on unix timestamps.

It would probably also be useful to have something like an isdate() function that will return true if the variable can be treated as a timestamp.

mrworkman:
@rgerhards, I was reviewing this and some other RainerScript "enhancement" issues, and I'd like to implement it/them (if that's ok).

For this particular one, I thought it'd be best if I got your input since I imagine you already have some idea of how you'd like to see it implemented.

I see it as something like this (returning a string):

* `to_timestamp(unix_time, dformat);`

e.g.
* `to_timestamp(unix_time, 'rfc3164')`; or
* `to_timestamp(unix_time, 'rfc3339')`; or
* `to_timestamp(unix_time, '%Y-%m-%d %H:%M:%S')`.

Or would you prefer some kind of conversion function to convert from any 'timestamp' to another, such as: `convert_timestamp(stimestamp, sformat, dformat)`?

davidelang:
I think it's best to run everything explicitly through unix timestamp (possibly 
a modified unix timestamp to allow for subsecond resolution)

instead of to_timestamp(), how about format_time()?

and then there could be a function

parse_time(format,dformat)

I'm not sure if the source should be a format or a variable. I expect that most 
of the time this is needed (especially with the liblognorm support to put the 
result in unix format) is when the parts of the timestamp end up in different 
variables and need to be combined and then parsed. The parseformat would be the 
normal date standards

mrworkman:
`format_time()` sounds reasonable. 
<br/>

RE: `parse_time()`
> I'm not sure if the source should be a format or a variable.

Wouldn't a variable be required, regardless?

davidelang:
On Tue, 3 Oct 2017, Stephen Workman wrote:

> RE: `parse_time()`
>> I'm not sure if the source should be a format or a variable.
>
> Wouldn't a variable be required, regardless?

the thing is that it may be many variables. If date is in one place and time in 
another place, to get a timestamp, you will need to combine them.

So you would need to have a template and then do

set $.x = exec_template(template)
followed by
parse_time($.x,dformat)

I think this is going to be more common than having a timestamp in a single 
[...]

PR #1814, opened 2017-10-05
WIP: Add RainerScript format_time() funcion to format UNIX timestamps.

mrworkman:
Work in progress.

Per issue #971. This PR is to add support to RainerScript for converting UNIX timestamps to RFC 3164/3339 date/time strings.

Adds:
```c
format_time(UNIX_time, "date-rfc3164")
```
```c
format_time(UNIX_time, "date-rfc3339")
```


grammar/rainerscript.c
```diff
@@ -2122,6 +2122,50 @@ doFuncCall(struct cnffunc *__restrict__ const func, struct svar *__restrict__ co
...
+		cnfexprEval(func->expr[1], &r[1], usrptr);
+
+		str = (char*) var2CString(&r[0], &bMustFree);
+		formatstr = (char*) es_str2cstr(r[1].d.estr, NULL);
+
+
+		// TODO: Error checking, and account for different
+		//       possible sizes of time_t 32/64 bit.
+		errno = 0;
+		unixtime = (time_t) strtoll(str, &endptr, 10);
```
rgerhards: There is a specific function to obtain a number from the script processor. This should be used.

grammar/rainerscript.c
```diff
@@ -2122,6 +2122,50 @@ doFuncCall(struct cnffunc *__restrict__ const func, struct svar *__restrict__ co
...
+
+		printf("STR: %s, FORMATSTR: %s, NUM: %ld\n", str, formatstr, unixtime);
+
+		localtime_r(&unixtime, &lt);
```
rgerhards: Is it for sure a good idea to use localtime? For example, a RFC3164 timestamp does not contain the TZ information. So we can simply convert it as if it were UTC. When that value is used with localtime, the TZ offset is applied once again. Unfortunately there are many subtle issues with timezones...

Maybe it's OK to do this, and we only need to document the problem.

grammar/rainerscript.c
```diff
@@ -2122,6 +2122,50 @@ doFuncCall(struct cnffunc *__restrict__ const func, struct svar *__restrict__ co
...
+
+		if (strncmp(formatstr, "rfc3164", 7) == 0 || strncmp(formatstr, "rfc3164", 7) == 0) {
+			strftime(result, 64, "%b %e %H:%M:%S", &lt);
+		} else if (strncmp(formatstr, "rfc3339", 7) == 0 || strncmp(formatstr, "RFC3339", 7) == 0) {
+			strftime(result, 64, "%Y-%m-%dT%H:%M:%S%z", &lt);
```
rgerhards: Same here: I am a bit concerned on the use of strftime. The man page says it consults TZ and LC_TIME. However, the timestamp may have orginated in many different timezones. Are we sure the result is what we want?

grammar/rainerscript.c
```diff
@@ -2122,6 +2122,50 @@ doFuncCall(struct cnffunc *__restrict__ const func, struct svar *__restrict__ co
...
+			strcpy(result + 22, ":00"); // <-- FIXME: Testing only, does not account for all TZs
+		} else {
+			strcpy(result, "NOT IMPLEMENTED");
+		}
+		
+		printf("%ld -> '%s'\n", unixtime, result);
```
rgerhards: dbgprintf is your friend ;-) [I know that's probably to be removed, but using the debug log system is actually much easier]

rgerhards:
set DO_NOT_MERGE just because it's WiP - so that nobody accidently merges ;-)

grammar/rainerscript.c
mrworkman: There have been a number of changes since this commit, including the printfs being removed ;)

grammar/rainerscript.c
mrworkman: Riiiggt, I'm still a bit of n00b on the RainerScript internals. Let me review. My intent is to allow fractional seconds to be passed in where I split off of the decimal (if present). I see `var2Number()` and `str2num()`, and they look like they'd parse up to the decimal and stop, but they'd not tell me where the decimal is, where strto?l? would...

rgerhards: Ah, I see. We do **not** have fractional numbers in a unix timestamp. The ElasticSearch ms timestamp is as follows: let t be the regular unix timestamp and f be the fractional part. Then you will see timestamp = t*1000+f, so it always is an integer.

I would strongly suggest to stick with that definition. As an alternative, the right thing would be to use the syslog-internal timestamp format, which has all fields nicely split of (including fractional part, precisio, TZ). But shuffling this is **a lot** of work and it is not compatible with anything (including incompatibility with current liblognorm). I think the ES approach is indeed a good way to do it.

grammar/rainerscript.c
mrworkman: Still considering the implications of using 'localtime'. Converting to syslogTimestamp and using the existing formatting methods could still end up being the way I go.

I admit strftime's use of TZ is annoying.

mrworkman:
Yes, please don't merge yet! :)

As for squashing, I was going to do that prior to merging. But I can squash every time I commit if you like.

rgerhards:
@mrworkman on squashing: it really depends: I, too, prefer to use multiple commits during devel, as it simplifies going back. On the other hand, multiple commits make it hard to review the code. Or is there a review tool on github I have overlooked? Ohhhhh yes, there is and it is pretty obvious. Something new learnt ;-) OK, so let's agree on squashing when everything is done and I try to use the "all commits" view...

grammar/rainerscript.c
rgerhards: IMHO the whole TZ system on Linux is unfortunately broken... How nice it would be just to be able to query the TZ database and tell the system what you want ;-) But complaining doesn't help and I am not yet up to propose a big clib patch ;-)

mrworkman:
Sounds good!

rgerhards:
I have tenatively set the release goal to 8.31.0. But if we keep it need an small, we may make it for this PR in the 8.30.0 frame - which would also be good given the new capabilities of liblognorm. Just so that you know. We should still avoid rush, better get things solid.

grammar/rainerscript.c
mrworkman: True, but I have seen unix-like timestamps with thousandths after a decimal which is what I was trying to accommodate. Maybe I'm getting carried away :)

I have no problem saying UNIX timestamps are non-fractional, and that's that. It's easier to support. I'll forget about it for now - if someone in the future asks for that sort of thing, we can review it then.

rgerhards: Let's say it's non fractional. There are a myriad of different formats. Once this all works, I would suggest to review this again. I think we can also probably apply the same heuristics ES uses (at least I assume so). For any recent timestamp, you can detect if it is with or without fractional seconds just by looking at the value (as the ms timestamp is 1000 times as high). Doesn't work for timestamp back in the 1970s, but that should no longer be a concern in our context.

But for now I'd just use standard format and get that solidly done.

davidelang:
On Thu, 5 Oct 2017, Rainer Gerhards wrote:

> Ah, I see. We do **not** have fractional numbers in a unix timestamp. The 
> ElasticSearch ms timestamp is as follows: let t be the regular unix timestamp 
> and f the fractional part. Then you will see timestamp = t*1000+f, so it 
> always is an integer.
>
> I would strongly suggest to stick with that definition. As an alternative, the 
> right thing would be to use the syslog-internal timestamp format, which has 
> all fields nicely split of (including fractional part, precisio, TZ). But 
> shuffling this is **a lot** of work and it is not compatible with anything 
> (including incompatibility with current liblognorm). I think the ES approach 
> is indeed a good way to do it.

unfortunantly this will run into grief if you have more than 3 digits past the 
decimal (some people will want to deal with more precise times)

davidelang:
On Thu, 5 Oct 2017, Stephen Workman wrote:

> True, but I have seen unix-like timestamps with thousandths after a decimal which is what I was trying to accommodate. Maybe I'm getting carried away :)
>
> I have no problem saying UNIX timestamps are non-fractional, and that's that. It's easier to support. I'll forget about it for now - if someone in the future asks for that sort of thing, we can review it then.

I really think we should support sub-second times, otherwise we are just going 
to have to come back and re-implement almost the exact same thing to deal with 
sub-second times.

I'm not a fan of limiting things to three decimal places though.

davidelang:
On Thu, 5 Oct 2017, Rainer Gerhards wrote:

> Doesn't work for timestamp back in the 1970s, but that should no longer be a 
> concern in our context.

not neccessarily correct. If you are only thinking of the timestamp of the 
message, you would be correct. But if we are talking about dates and times in 
the body of the  messages that we are parsing, it's not as safe to say that.

rgerhards:
I still would find it surprising to find 1970s dates in current log
content. Do you have an example for such?

Anyhow... I was after a simple and useful starting point that could
possibly materialize within a couple of days. So I would still prefer to
support the classical Unix timestamp and after that build on it. I don't
think that requires much code to be changed later.

Rainer

Sent from phone, thus brief.

Am 05.10.2017 22:01 schrieb "David Lang" <notifications@github.com>:

mrworkman:
> I was after a simple and useful starting point that could possibly materialize within a couple of days. So I would still prefer to support the classical Unix timestamp and after that build on it. I don't think that requires much code to be changed later.
I think the incremental approach is best here. Let's get _something_ working well, then we can extend it later. We'll still need to define/determine how to handle the different representations of fractional seconds anyway (and that will take time - no pun intended).

WRT fractional seconds, I don't think RFC 3164 even supports this, and RFC 3339 seems to support infinite resolution (it's defined as 1*DIGIT), but I'd say we should draw the line somewhere by setting a maximum resolution -  a discussion for later.

grammar/rainerscript.c
mrworkman: I did some quick testing on Linux (I'll do more), and I think the man page for `strftime()` is just misleading (no surprise there).

The glibc version of `strftime()` doesn't actually manipulate the time/date it's passed before producing its output. In fact, according to a comment in a recent copy of `strftime_l.c`:

```c
/* The POSIX test suite assumes that setting
   the environment variable TZ to a new value before calling strftime()
   will influence the result (the %Z format) even if the information in
   TP is computed with a totally different time zone.
   This is bogus: though POSIX allows bad behavior like this,
   POSIX does not require it.  Do the right thing instead.  */
```
So, at least in  GNU, it has no effect. `strftime()` itself _does_ indeed call `tzset()`, but it's only to set a flag, and to satisfy POSIX. 

I've tested this with the following simple program:

```c
#include <time.h>
#include <stdio.h>

int main(void) {

   time_t t = 0xdeadbeef;
   struct tm tm;

   int max = 64;
   char result[max];

   localtime_r(&t, &tm);


   strftime(result, max, "%Y-%m-%d %H:%M:%S %z", &tm);
   printf("RESULT: %s\n", result);
   printf("DST?: %d\n", tm.tm_isdst);

   tzset();
   gmtime_r(&t, &tm);

   strftime(result, max, "%Y-%m-%d %H:%M:%S %z", &tm);
   printf("RESULT: %s\n", result);
   printf("DST?: %d\n", tm.tm_isdst);

   return 0;
}
```

And, indeed, changing `TZ` has no effect on the time produced by `gmtime()`/`strftime()` (including `%Z`).

That's GNU. The C library on Solaris seems to be a bit of a different story (according to the latest CIs). The Solaris man 
 page `strftime()` doesn't mention `TZ` like the GNU one does. Solaris seems to respect the same rules as far as ignoring the `TZ` value with respect to `%Z`, but for some reason it preserves the `tm_isdst` flag of `struct tm` resulting in UTC/GMT times with a `+0100` value for `%Z`. 🙄 

I think that can be addressed by simply setting `tm_isdst` to zero prior to calling `strftime()` (I still have to test this). This should be safe as the test output for Solaris matches the output for GNU, as far as the date/time parameters are concerned -- as in, the dates and times are the same, it's just the `+01:00` that's different.

<hr />

Since UNIX timestamps are technically supposed to be expressed in terms of UTC, then `format_time()` should probably represent them this way (when `date-rfc3339` or `%Z` are used). Perhaps later versions could allow the user to manipulate the `%Z` value to match the actual timezone of the remote log source. What do you think?

mrworkman:
I'll put this here too, because I don't know what will happen to the change request comments when the commits are squashed:
<hr />
I did some quick testing on Linux (I'll do more), and I think the man page for `strftime()` is just misleading (no surprise there).

The glibc version of `strftime()` doesn't actually manipulate the time/date it's passed before producing its output. In fact, according to a comment in a recent copy of `strftime_l.c`:

```c
/* The POSIX test suite assumes that setting
   the environment variable TZ to a new value before calling strftime()
   will influence the result (the %Z format) even if the information in
   TP is computed with a totally different time zone.
   This is bogus: though POSIX allows bad behavior like this,
   POSIX does not require it.  Do the right thing instead.  */
```
So, at least in  GNU, it has no effect. `strftime()` itself _does_ indeed call `tzset()`, but it's only to set a flag, and to satisfy POSIX. 

I've tested this with the following simple program:

```c
#include <time.h>
#include <stdio.h>

int main(void) {

   time_t t = 0xdeadbeef;
   struct tm tm;

   int max = 64;
   char result[max];

   localtime_r(&t, &tm);


   strftime(result, max, "%Y-%m-%d %H:%M:%S %z", &tm);
   printf("RESULT: %s\n", result);
   printf("DST?: %d\n", tm.tm_isdst);

   tzset();
   gmtime_r(&t, &tm);

   strftime(result, max, "%Y-%m-%d %H:%M:%S %z", &tm);
   printf("RESULT: %s\n", result);
   printf("DST?: %d\n", tm.tm_isdst);

   return 0;
}
```

And, indeed, changing `TZ` has no effect on the time produced by `gmtime()`/`strftime()` (including `%Z`).

That's GNU. The C library on Solaris seems to be a bit of a different story (according to the latest CIs). The Solaris man 
 page `strftime()` doesn't mention `TZ` like the GNU one does. Solaris seems to respect the same rules as far as ignoring the `TZ` value with respect to `%Z`, but for some reason it preserves the `tm_isdst` flag of `struct tm` resulting in UTC/GMT times with a `+0100` value for `%Z`. 🙄 

I think that can be addressed by simply setting `tm_isdst` to zero prior to calling `strftime()` (I still have to test this). This should be safe as the test output for Solaris matches the output for GNU, as far as the date/time parameters are concerned -- as in, the dates and times are the same, it's just the `+01:00` that's different.

<hr />

Since UNIX timestamps are technically supposed to be expressed in terms of UTC, then `format_time()` should probably represent them this way (when `date-rfc3339` or `%Z` are used). Perhaps later versions could allow the user to manipulate the `%Z` value to match the actual timezone of the remote log source. What do you think?

grammar/rainerscript.c
rgerhards: Great analysis!

And: yeah, I was always very skeptic of strftime(). Actually the portability problems are one reason (I now remember in more detail) of staying away from it. There may even be such a discussion here or in the mailing list archive.

That said, I think it would be best to implement our own version of strftime(), at least for the common formats. Otherwise, we'll probably end up with the need to implement it anyhow. This is not uncommon to happen years later, which means usually I am personally than in trouble -- thus I have begun to point out these things very early.

rgerhards: Oh, and I forgot: testing these things never gives the definite answer. Does, for example, AIX handle it the same way? How about BSD? What in regard to old versions of Solaris? And ultimately: what about upcoming OS vendor patches?

That's also the reason why rsyslog has it's own rs_strerror() which works around all the bad things you find in the distros. Unfortunately, I don't think it is so easy to do with strftime() as well.

mrworkman:
Ok, I'll go by your experience on this one.

In that case, I think I should forgo the support for custom `strftime()`
formatting for the first implementation because that's going to be a pain
to implement, and will only cause delay. The "standard" syslog time formats
should be good enough for most cases anyway.

On Oct 6, 2017 12:33 PM, "Rainer Gerhards" <notifications@github.com> wrote:

*@rgerhards* commented on this pull request.
------------------------------

In grammar/rainerscript.c
<https://github.com/rsyslog/rsyslog/pull/1814#discussion_r143235937>:

> +		formatstr = (char*) es_str2cstr(r[1].d.estr, NULL);
+
+
+		// TODO: Error checking, and account for different
+		//       possible sizes of time_t 32/64 bit.
+		errno = 0;
+		unixtime = (time_t) strtoll(str, &endptr, 10);
+
+		printf("STR: %s, FORMATSTR: %s, NUM: %ld\n", str, formatstr, unixtime);
+
+		localtime_r(&unixtime, &lt);
+
+		if (strncmp(formatstr, "rfc3164", 7) == 0 || strncmp(formatstr,
"rfc3164", 7) == 0) {
+			strftime(result, 64, "%b %e %H:%M:%S", &lt);
+		} else if (strncmp(formatstr, "rfc3339", 7) == 0 ||
strncmp(formatstr, "RFC3339", 7) == 0) {
+			strftime(result, 64, "%Y-%m-%dT%H:%M:%S%z", &lt);

Great analysis!

And: yeah, I was always very skeptic of strftime(). Actually the
portability problems are one reason (I now remember in more detail) of
staying away from it. There may even be such a discussion here or in the
mailing list archive.

That said, I think it would be best to implement our own version of
strftime(), at least for the common formats. Otherwise, we'll probably end
up with the need to implement it anyhow. This is not uncommon to happen
years later, which means usually I am personally than in trouble -- thus I
have begun to point out these things very early.

—
You are receiving this because you were mentioned.
<https://github.com/rsyslog/rsyslog/pull/1814#discussion_r143235937>, or mute
the thread

mrworkman:
AIX is Thunderdome as far as I'm concerned. I wasn't even aware rsyslog
even targeted it.


On Oct 6, 2017 12:35 PM, "Rainer Gerhards" <notifications@github.com> wrote:

*@rgerhards* commented on this pull request.
------------------------------

In grammar/rainerscript.c
<https://github.com/rsyslog/rsyslog/pull/1814#discussion_r143236319>:

> +		formatstr = (char*) es_str2cstr(r[1].d.estr, NULL);
+
+
+		// TODO: Error checking, and account for different
+		//       possible sizes of time_t 32/64 bit.
+		errno = 0;
+		unixtime = (time_t) strtoll(str, &endptr, 10);
+
+		printf("STR: %s, FORMATSTR: %s, NUM: %ld\n", str, formatstr, unixtime);
+
+		localtime_r(&unixtime, &lt);
+
+		if (strncmp(formatstr, "rfc3164", 7) == 0 || strncmp(formatstr,
"rfc3164", 7) == 0) {
+			strftime(result, 64, "%b %e %H:%M:%S", &lt);
+		} else if (strncmp(formatstr, "rfc3339", 7) == 0 ||
strncmp(formatstr, "RFC3339", 7) == 0) {
+			strftime(result, 64, "%Y-%m-%dT%H:%M:%S%z", &lt);

Oh, and I forgot: testing these things never gives the definite answer.
Does, for example, AIX handle it the same way? How about BSD? What in
regard to old versions of Solaris? And ultimately: what about upcoming OS
vendor patches?

That's also the reason why rsyslog has it's own rs_strerror() which works
around all the bad things you find in the distros. Unfortunately, I don't
think it is so easy to do with strftime() as well.

—
You are receiving this because you were mentioned.
<https://github.com/rsyslog/rsyslog/pull/1814#discussion_r143236319>, or mute
the thread

mrworkman:
Date/time calculations are notoriously difficult especially with leap
years, leap seconds, etc.. do you want me to steer away from gmtime() as
well?

davidelang:
Assuming we are only worried about supporting glibc, the strftime that uses the 
environment variable calls another function where the timezone is passed 
explicitly. If those latter functions are available, they should be used 
instead.

mrworkman:
@rgerhards Travis wasn't working earlier. Can you restart the build? Also, let me know how you feel about the current implementation (with `strftime` removed).

rgerhards:
> Date/time calculations are notoriously difficult especially with leap
> years, leap seconds, etc.. do you want me to steer away from gmtime() as
> well?

One thing: for now, I would stay with gmtime(). But we can discuss this in more depth with the next steps.

grammar/rainerscript.c
```diff
@@ -2122,6 +2124,53 @@ doFuncCall(struct cnffunc *__restrict__ const func, struct svar *__restrict__ co
...
+		cnfexprEval(func->expr[0], &r[0], usrptr);
+		cnfexprEval(func->expr[1], &r[1], usrptr);
+
+		unixtime = var2Number(&r[0], &retval);
+
+		// Make sure that the timestamp we got can fit into
+		// time_t on older systems.
+		if (sizeof(time_t) == sizeof(int)) {
+			if (unixtime < INT_MIN || unixtime > INT_MAX) {
+				DBGPRINTF(
```
rgerhards: I tend to think this would best be message that is emitted to the user. Maybe as LOG_WARN. Search for ```LogMsg(``` inside the code to see how these kinds of messages are emitted.

grammar/rainerscript.c
```diff
@@ -2122,6 +2124,53 @@ doFuncCall(struct cnffunc *__restrict__ const func, struct svar *__restrict__ co
...
+					"Timestamp value %lld is out of range for this system (time_t is 32bits)!", unixtime
+				);
+				retval = 0;
+			}
+		}
+
+		// We want the string form too so we can return it as the
```
rgerhards, line 2151: I think we should (on the main issue) describe the semantics for the call. I don't think it makes sense to return the number if a date is expected, but that probably needs some discussion.

runtime/datetime.c
```diff
@@ -1258,6 +1260,64 @@ timeConvertToUTC(const struct syslogTime *const __restrict__ local,
...
+ * Format a UNIX timestamp.
+ */
+static int
+formatUnixTimeFromTime_t(time_t unixtime, const char *format, char *pBuf, uint pBufMax) {
+	struct tm lt;
+
+	assert(format != NULL);
+	assert(pBuf != NULL);
+
+	// Let's not depend on debug asserts to save the day in the wild!
```
rgerhards: I would **not** do this in non-debug build. After all, that's a calling sequence error which should never happen. If we do checks all over the block, this is pretty time consuming.

runtime/datetime.c
```diff
@@ -1258,6 +1260,64 @@ timeConvertToUTC(const struct syslogTime *const __restrict__ local,
...
+
+	// Convert to struct tm
+	if (gmtime_r(&unixtime, &lt) == NULL) {
+		DBGPRINTF("Unexpected error calling gmtime_r().\n");
+		return -1;
+	}
+
+	// Do our conversions
+	if (strcmp(format, "date-rfc3164") == 0) {
+		if (pBufMax < 16) {
```
rgerhards: I would tend to do the next two statements as assert() as well, but am open to discussion. @davidelang what do you think?

grammar/rainerscript.c
mrworkman: Makes sense.

grammar/rainerscript.c
mrworkman: Ok. Maybe I'm making assumptions about what the user would want to happen - rather than leaving it up to them to control the end result. I'm just loathe to discard the "real" value if something unexpected comes in.

rgerhards:
2017-10-10 14:20 GMT+02:00 Stephen Workman <notifications@github.com>:

> *@mrworkman* commented on this pull request.
>
> In grammar/rainerscript.c
> <https://github.com/rsyslog/rsyslog/pull/1814#discussion_r143709483>:
>
> > +		cnfexprEval(func->expr[1], &r[1], usrptr);
> +
> +		unixtime = var2Number(&r[0], &retval);
> +
> +		// Make sure that the timestamp we got can fit into
> +		// time_t on older systems.
> +		if (sizeof(time_t) == sizeof(int)) {
> +			if (unixtime < INT_MIN || unixtime > INT_MAX) {
> +				DBGPRINTF(
> +					"Timestamp value %lld is out of range for this system (time_t is 32bits)!", unixtime
> +				);
> +				retval = 0;
> +			}
> +		}
> +
> +		// We want the string form too so we can return it as the
>
> Ok. Maybe I'm making assumptions about what the user would want to happen
> - rather than leaving it up to them to control the end result. I'm just
> loathe to discard the "real" value if something unexpected comes in.
>
I am actually not sure what makes most sense, so we should discuss. For the
initial experimental version, we could use your proposal and see how it
works. I would just expect that putting the unix timestamp instead of the
requested one will almost always lead to wrong results.

runtime/datetime.c
mrworkman: I come from a security background, and worry about CVEs. Where things that should never happen do happen :)

This situation would never occur in the existing calling sequence, but what about if/when this function is called from somewhere else in the future?

rgerhards: I see the point. But if we introduce all those checks consistently, we will loose quite a lot of performance. I think there needs to be made a compromise. Maybe no longer, but than the concept of high performance also needs to be redefined.

If the calling sequence is violated, it's a bug in any case. Again, I see the dilemma here. But if we really want to handle that, than we need to emit a very big user error message, because a calling sequence violation is a very strong indication that something got really, really wrong. Doing it just one place doesn't make sense IMHO, so we would need to do it at the other 2,000 or so places as well (not exaggegrating). @davidelang what do you think?

mrworkman:
> I am actually not sure what makes most sense, so we should discuss. For the
initial experimental version, we could use your proposal and see how it
works. I would just expect that putting the unix timestamp instead of the
requested one will almost always lead to wrong results.

There are a few ways we could handle this: 
1) As is, return the original value unmodified. 
* **Pro**: If we preserve the original unconvertible value, then at least it still gets logged instead of being thrown away.
* **Con**: Not an expected date/time string.

2) Return something like -1.
* **Pro**: This makes no assumptions about what the configurer/user wants, so they are able to define their own behaviour.
* **Con**: If its return value is not checked correctly, then '-1' would appear in any logs produced instead of a date/time (or value that could be represented as such). Original value is lost.

3) Return the current date/time.
* **Pro**: Always get a valid date/time.
* **Con**: Most definitely incorrect. Original value is lost.
* **Con**: No way to tell if the conversion *didn't* work properly.

No. 3 is probably the worst choice - but it returns a date in the expected format (if that's the concern). 
No. 2 can be "handled", but like returning the original UNIX timestamp, is not an expected date/time string.

runtime/datetime.c
mrworkman: I suppose there's really no way the configurer/user could somehow use a value that would result in a NULL being passed to the function. Also, NULL checks are rather superficial anyway - If the pointer were somehow set to an address of 0x00000001 (just as likely as being accidentally set to 0x00000000) rsyslog would still segfault.

rgerhards: The NULL checks actually date back to pre-valgrind and pre-ASAN. I often remove them now nowadays as either of the two will catch them even better as assert(). I do use assert still for some invariant checks that point to logic errors. And this is my point: if we intend to handle all this with more checks and workarounds, I think in the long term we get "needless" complex code that probably results in more CVEs (due to logic errors) than when we accept that we cannot cover everything (and let the hardening done by the compiler as much as possible). Of course, proper testing is important here as well. Thankfully we've come a long way on the CI front compared to e.g. 5 years ago. Still more work to be done (I don't get unit tests going due to problems with linking that nobody ever managed to tell me how to do it right :-( ).

davidelang:
On Tue, 10 Oct 2017, Rainer Gerhards wrote:

> if we intend to handle all this with more checks and workarounds, I think in 
> the long term we get "needless" complex code that probably results in more 
> CVEs (due to logic errors) than when we accept that we cannot cover everything 
> (and let the hardening done by the compiler as much as possible).

I agree with this.

David Lang

rgerhards:
No. 3 is I think a very bad choice as it would permit everything to continue to work, "just with wrong data". I'd say let's keep as-is for the experimental version, so that we can merge tomorrow. Does that sound OK?

mrworkman:
Works for me.

What about documentation? Do you want me to do a PR to update that?

rgerhards:
yeah, doc update would be very good :-)

mrworkman:
FYI, I'll be on planes all day tomorrow, so if you need any more changes, let me know today 😉

davidelang:
especially since we have had at least two other people asking about this while 
you have been writing it :-)

rgerhards:
rsyslog.pid and rsyslog.started have accidently be added to git -- will remove them after merge

PR #1837, opened 2017-10-10
do not call fjson_object_put from jsonMerge

### LINKED ISSUE #1822: rsyslog (8.29.0) segfaults when using mmexternal plugin
I'm running rsyslog 8.24.0 on CentOS 7 (so, `rsyslog-8.24.0-12.el7.x86_64`).  I'm trying to write a filtering plugin in Python.  Starting with [the example code](https://github.com/rsyslog/rsyslog/blob/master/plugins/external/skeletons/python/mm-python.py), I modified `onReceive` so that it looks like:

```python
def onReceive(msg):
        data = json.loads(msg)
        print json.dumps({"$!": {"sometag": "somevalue"}})
```

That causes a segfault as soon as any message comes through. 

Responses that don't attempt to modify `$!`  don't cause a segfault, but the updated keys do not end up in `$!all-json`.

larsks:
MsgSetPropsViaJSON_Object will call fjson_object_put on an ancestor of
the tree seen by jsonMerge.

I don't know if this is the correct way of solving this issue, but this at least lets me get back to working on an external filter rather than trying to figure out why rsyslog keeps segfaulting.

Closes #1822

### DIFF
--- a/runtime/msg.c
+++ b/runtime/msg.c
@@ -4692,11 +4692,6 @@ jsonMerge(struct json_object *existing, struct json_object *json)
 			json_object_get(json_object_iter_peek_value(&it)));
 		json_object_iter_next(&it);
 	}
-	/* note: json-c does ref counting. We added all descandants refcounts
-	 * in the loop above. So when we now free(_put) the root object, only
-	 * root gets freed().
-	 */
-	json_object_put(json);
 	RETiRet;
 }
 


rgerhards:
I am in the middle of complex other work so I cannot help analyze this at the moment. The PR is appreciated. I would suggest that you keep an eye on a new memory leak. That's the most probable cause if the patch is not valid.

JAORMX:
@rgerhards have you had a chance to look at this? would sure be useful 😃

rgerhards:
>  have you had a chance to look at this?

yeah, definitely, but it's still on the list. Working on 6 things in parallel right now...

rgerhards:
@JAORMX To be a bit more precise:  I tried to reproduce this problem with a simple test case, but that did not show it. So I need to do a more elaborate test. This was when I pushed the issue back because there are more pressing issues I need to work on. To accelerate this, it would be extremely beneficial if you could craft a testbench test that reproduces this issue. With that, I can probably much quicker find the final solution. I am **very skeptic** that the patch proposed here is valid in all cases and so would like to know the root cause before I merge.

rgerhards:
I went through the code once again, starting with mmexternal. Everything looks fine, So we **really** need a repro to find the root cause. Aplying this patch is 99.9999% the wrong thing to do.

larsks:
The reproducer in #1822 works reliably for me.  Is there a way I can make that more consumable for you?  Both @bdolez and I thought, looking at the code, that it looked like a double-free situation (as described [here](https://github.com/rsyslog/rsyslog/issues/1822#issuecomment-335352833)). Do you disagree that this is a problem?

rgerhards:
I have just asked @PascalWithopf to create a testbench test. Let's see when has time, hopefully next week. I have followed the code path once more. It's a double free issue, I agree on that (hard to beat ;-)). But we need to be very careful that we fix the right spot. Thus I would like to have a testbench test.

rgerhards:
I mean we need to trace all pathes and should make sure we have tests for everything that is relevant. Too often we ran into regressions when I merged too quickly just because "it looked so right".

larsks:
Oh, I totally agree with all that, and I was the first to suggest that this patch may not be the correct fix :). I thought you were saying you didn't see a problem.  In any case, thanks for looking!

rgerhards:
Yeah... I admit I have been a bit hard to work with... Tons of updates
going on, many in regard to further improve QA. Getting a bit lost in the
myriad of commits and things to do ;-)

I appreciate your help and hope we can finally fix this within the 8.31
time frame.

rgerhards:
Manual review brings up at least one case where a memleak occurs with this patch: call msgAdd[Multi]Metadata() -> msgAddJSON -> jsonMerge() leaks json object created in msgAddMetadata().

rgerhards:
Path 1 described in https://github.com/rsyslog/rsyslog/issues/1822#issuecomment-343422217 is not addressed by this PR.

rgerhards:
correct fix #2014 merged, closing this w/o merge

PR #1844, opened 2017-10-13
testbench: kill hanging rsyslog instances from past tests

rgerhards:
**EXPERIMENTAL: may not make it into release** - depends on if it really works out ;-)

this sometimes happens on buildbot when a previous test instance
was not able to be shut down.


rgerhards:
does not make sense, even problem was differently

PR #1851, opened 2017-10-17
imgssapi: Incooperate changes from tcpsrv subsystem

Whissi:
Commit 5c5a7a3ed509dfb39b451c806646ab67782f6eda introduced an additional parameter "oserr".

Fixes: https://github.com/rsyslog/rsyslog/issues/1838

PR #1855, opened 2017-10-18
fix imgssapi

rgerhards:
merge after https://github.com/rsyslog/rsyslog/pull/1851


plugins/omgssapi/omgssapi.c
```diff
@@ -57,7 +57,6 @@
 
 MODULE_TYPE_OUTPUT
 MODULE_TYPE_NOKEEP
-MODULE_CNFNAME("omgssapi")
```
mbiebl, line 60: Why does imgssapi have ```MODULE_CNFNAME("imgssapi")```?

mbiebl: nvm, I see that this was removed as well in an unrelated commit 8d793eafdde0b74d0b4424f0d194b5dc6801a5d1

rgerhards: iIRC I was working on removing compile errors...

Issue #1869, opened 2017-10-20
imrelp/omrelp: Segmentation fault when certificate is NULL


PascalWithopf:
When a certificate is specified in an imrelp or omrelp action, the module checks if the certificate can be accessed. If this isn't possible, e.g. because the file does not exist or you don't have permission to do this, a Segmentation Fault will appear.

PR #1871, opened 2017-10-20
omrelp/imrelp: fix certificate check

PascalWithopf:
Fix Segmentation Fault when certificate is NULL
closes https://github.com/rsyslog/rsyslog/issues/1869


rgerhards:
please rebase to current master. That will make the fedora26x64 CI failure go away.

rgerhards:
Also, the commit needs more information on the triggering condition. One needs to evaluate when this is a risk.

Issue #1876, opened 2017-10-20
Enhance "queue.fullDelayMark is set below high water mark" message to explicitly indicate it is an error condition


deoren:
From #1865 :

```
Oct 19 12:13:19 mww-allsight rsyslogd-2358: queue "action 1 queue": queue.fullDelayMark is set below high water mark. This will result in DA mode  NOT being activated for full delayable messages [v8.16.0 try http://www.rsyslog.com/e/2358 ]
Oct 19 12:13:19 mww-allsight rsyslogd-2358: queue "action 3 queue": queue.fullDelayMark is set below high water mark. This will result in DA mode  NOT being activated for full delayable messages [v8.16.0 try http://www.rsyslog.com/e/2358 ]
```

Request:

Modify the message to explicitly indicate it is an error condition that requires user interaction to fix.

jgerhards:
I have enhanced the warning message. However, it is not an error since this configuration might be intended, for example when using UDP and TCP as input for the same queue and only wanting UDP to use the high watermark config.

deoren:
@jgerhards Thanks! I suspect that @andrewthad and @tmahany419 will find that useful.

andrewthad:
Thanks! Yes, we will find this useful.

PR #1878, opened 2017-10-21
queue: enhance error message

jgerhards:
closes https://github.com/rsyslog/rsyslog/issues/1876


rgerhards:
Note to myself: we should add a more elaborate description to the message landing page on what the user actually should do -- best before merging.

rgerhards:
Added the description to the web ressource: http://kb.monitorware.com/post27515.html

Issue #1879, opened 2017-10-21
testbench: build mmcount


rgerhards:
looks like it is not being build in CI (https://github.com/rsyslog/rsyslog/issues/1790)

PR #1882, opened 2017-10-22
travis: enable building mmcount

rgerhards:
**EXPERIMENTAL** - need early access to full buildbot slaves

while it is "just" a contributed module, we should make sure it
builds at least

closes https://github.com/rsyslog/rsyslog/issues/1879


rgerhards:
solaris 10 is out of space

PR #1912, opened 2017-10-27
testbench: make sure try_merge does not fail because of git settings

rgerhards:
also fix issue with user name/email not set on merge


rgerhards:
The original issue is corrected, but it is probably not a good idea to make a merge fail a test fail ... what if we simply need a rebase?

rgerhards:
changed commit so that it does not fail CI test for reasons given above. Now ready to merge if runs OK.

Issue #1624, opened 2017-06-20
add rainerscript function is_timestamp()


rgerhards:
It would be useful to add a function

```is_timestamp(property, format)```

where property is any (string) property and format is a set of known formats (e.g. rfc3339, unix timestamp, ...). The function shall return 1 if this is a timestamp in specified format and 0 if it is invalid. This could also be used in conjunction with conversion functions (common core functionality).

see also https://github.com/rsyslog/rsyslog/issues/971

mrworkman:
@mrworkman

rgerhards:
@mrworkman any plans to implement this one here? ;-) I wonder if I should shuffle the release goal, thus the question.

mrworkman:
Yes, but probably not by the 24th. I wish the days were longer!

rgerhards:
> I wish the days were longer!

haha, we are in the same camp. I'll move the release goal to 8.32 - we can always move it back. But so it doesn't pollute the list of "this could be done" the next couple of days ;-)

jgerhards:
I have just seen this issue on the tracker and I have noticed I have already worked on an isDate function, though it does not work. You can find it here: https://github.com/rsyslog/rsyslog/commit/52bbf957e7bc9f56d9be2d7a94b1143979f35a09
Would you like to work together on date related functions? However, i will probably be short on time next week.

mrworkman:
I'm ok working on them together, though I had a different approach in mind for this function. I'm open to sharing thoughts, etc...

jgerhards:
great. I personally have not worked much on open source yet and would appreciate your advice. What is the direction you wanted to take?

mrworkman:
@jgerhards, sorry for the delay, guess I didn't have much time this week either :) 

I think, from reading your code, you're trying to parse numerically formatted dates only. No time parsing. Rainer already has some functions internally for parsing standard syslog date/time strings that can be used (see the recently added `parse_time()` function which uses them). I was going to start by using those, with future iterations supporting strptime-like options.

mrworkman:
Hoping to work on this this weekend, but my wife wants me to do Christmassy stuff, so we'll see ;)

mrworkman:
Suggest going with `is_time()` rather than `is_timestamp()` for consistency with `parse_time()` and `format_time()`.

rgerhards:
implemented via #450

PR #1925, opened 2017-10-28
Add RainerScript parse_time() function to convert date/time strings to UNIX time.

mrworkman:
Per issue #971. This PR is to add support to RainerScript for converting RFC 3164/3339 date/time strings to UNIX timestamps.

Adds:

```c
parse_time(datetime_string)
```


mrworkman:
This one presents a bit of an irksome problem... converting RFC 3164 date/time strings - because they don't have a year.  In my code, right now, I assume (and apply) the current year (based on UTC time). But, this can be a problem particularly near the beginning of a new year - where logs come in late from the previous year. They'd get the new year rather than the one they "should"... there are other problems too, of course. Any ideas?

Also, the test I wrote will only work *this year* 😁. I'll fix that once we figure out the expected behaviour.

davidelang:
give the function an optional parameter to specify the year.

if you wanted to be really fancy, you could add an options to assume that you 
aren't reading dates more than 6 months old (so in Jan, when you read a Dec date 
you use the prior year instead of this year)

but the simple thing is just to allow a year to be specified, and if not, use 
the current year.

rgerhards:
> if you wanted to be really fancy, you could add an options to assume that you
> aren't reading dates more than 6 months old (so in Jan, when you read a Dec date
> you use the prior year instead of this year)

I would say that even without any explicit option, I would treat Dec timestamps in early Jan as being the past year. It's probably something that really often will happen around midnight new years eve. In general the question really is if a timestamp far in the future couple days? should be treated as old year.

davidelang:
On Sun, 29 Oct 2017, Rainer Gerhards wrote:

> In general the 
> question really is if a timestamp far in the future couple days? should be 
> treated as old year.

I wouldn't, it's too common to see times off, even by a few day. I would give a 
good long guard period (which is why I made the 6 month comment)

rgerhards:
Side-note: this might also be implemented via this code and/or an addition? https://github.com/rsyslog/rsyslog/issues/1624

mrworkman:
> give the function an optional parameter to specify the year.

Like using a fixed value: `parse_time($.x, 2017)`? 

> if you wanted to be really fancy, you could add an options to assume that you
aren't reading dates more than 6 months old (so in Jan, when you read a Dec date
you use the prior year instead of this year)

Some variant of this might be the best option. But it'd have to account for some timestamps from the "future", and some from the "past". Some kind of sliding window would represent *the current year*, and whatever falls outside of that would be the previous/next year... let me brainstorm.

mrworkman:
Sorry for the delay, busy week.

How about something like this:

Base it on the relationship between the current and "incoming" months. If the incoming month is within one month in the future (relative to the current month), it's from the same year - unless the current month is December, then it is for the next year.

|Current Month|Incoming Month|Result|
|-|-|-|
|Dec|Jan|2018|
|Nov|Dec|2017|
|Oct|Nov|2017|

If the incoming month appears to be earlier than the current month, or *more than* one month in the future, it must actually be from the *past*.

|Current Month|Incoming Month|Result|
|-|-|-|
|Dec|Feb|2017|
|Dec|Mar|2017|
|Apr|May|2017|
|Jul|Aug|2017|
|Jul|Feb|2017|
|Jul|Jan|2017|
|Jul|Dec|2016|
|Jul|Sep|2016|
|Jan|Nov|2016|
|Apr|Jun|2016|
|Apr|Oct|2016|

**NOTE**: Tables assume current year is 2017 

I'd expect that most logs that come in with a month that does not match the current one are probably from the past (back-logged) - especially when it diverges a lot from the current month. If they're coming in in the "future", then maybe it's more likely that the clock is wrong on the sending host. Accepting one month "in the future" allows for reasonably incorrect clocks to get what is *probably* the correct year.

We could reduce this from one month in the future, but then we're dealing with day offsets which are more difficult to get right.

rgerhards:
LGTM - @davidelang what do you think?

davidelang:
we need this sort of logic, exactly where in the window we put the current time 
is a heuristic. My knee jerk approach was to put current time in the middle of 
the window instead of one month from the end, but a month in the future is 
probably good enough.

mrworkman:
@rgerhards, this should be good to go, so please review when you get a chance.

I'll submit a separate PR for #1624 soon.

tests/rscript_parse_time.sh
```diff
@@ -0,0 +1,104 @@
...
+set $!datetime!inval2 = parse_time("2017-10-05T01:10:11");
+set $!datetime!inval3 = parse_time("2017-SOMETHING: 42");
+
+template(name="outfmt" type="string" string="%!datetime%\n")
+local4.* action(type="omfile" file="rsyslog.out.log" template="outfmt")
+local4.* :omstdout:;outfmt
+'
+
+. $srcdir/diag.sh startup
+. $srcdir/diag.sh tcpflood -m1 -y | sed 's|\r||'
```
rgerhards: what is the intent here? (in line 87)

tests/rscript_parse_time_get-ts.py
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env python
```
rgerhards, line 1: for the non-pythonic (and as an aid in any case), could you add a little doc of what the script actually does - same for the test. I think I get the gist - do an eval that the test driver itself is OK, then inject the data, but it would be much easier to spell it out IMHO. I think this is important as the method is quite a bit different from the regular way tests operate.

tests/rscript_parse_time.sh
mrworkman: You mean the `sed 's|\r||'`? `tcpflood` produces a carriage return in its output (see below), it was bugging me :)

```
^M00000001 messages sent
```

the `sed` changes it to

```
00000001 messages sent
```

tests/rscript_parse_time_get-ts.py
mrworkman: Done. Let me know if that seems clear enough.

tests/rscript_parse_time.sh
rgerhards: I would suggest to remove that as it may make other stumble over it as well. The problem in travis is that it mimics a tty, so that tcpflood's supression logic for non-ttys does not kick in.

mrworkman: You're da boss ;)

Done.

rgerhards:
thx a ton!

PR #1948, opened 2017-10-31
imczmq: remove unallocated and unused pointer

NachiP:
Fix a SIGSEGV in a call to
371: zcert_destroy(&serverCert) called from rcvData().

- remove serverCert as it is only used uninitialized in a call
to zcert_destroy() - this causes a core dump at rsyslogd exit.
- also initialize authActor at definition.

Signed-off-by: Nachiketa Prachanda <nachiketa.prachanda@att.com>

### DIFF
--- a/contrib/imczmq/imczmq.c
+++ b/contrib/imczmq/imczmq.c
@@ -278,8 +278,7 @@ static rsRetVal rcvData(void){
 		}
 	}
 
-	zactor_t *authActor;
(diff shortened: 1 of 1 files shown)


rgerhards:
LGTM

taotetek:
LGTM

NachiP:
Thanks.

rgerhards:
Thanks for the patch! :-)

Issue #1920, opened 2017-10-27
Segfault when attempting to check length of non-existent JSON message property


deoren:
As noted on #1071, when I attempt to check the length of a JSON message property (whether at the root or part of a subtree) rsyslog quickly segfaults and crashes. 

## Configuration that can be used to repro the problem

Here is a trimmed configuration that allows me to reproduce the problem:

```
$ActionFileDefaultTemplate RSYSLOG_FileFormat
$RepeatedMsgReduction off
$WorkDirectory /var/spool/rsyslog

module(
    load="builtin:omfile"
    fileOwner="syslog"
    fileGroup="adm"
    dirOwner="syslog"
    dirGroup="adm"
    fileCreateMode="0640"
    dirCreateMode="0755"
)

module(load="imudp")
module(load="imptcp")

input(type="imudp" port="514" ruleset="receiver-standard-rules")
input(type="imptcp" port="514" KeepAlive="on" ruleset="receiver-standard-rules")

ruleset(name="receiver-standard-rules") {

    if ($!origin!hostname == "") then {
        continue
    }

    action(
        name="test-messages"
        type="omfile"
        file="/var/log/rsyslog_testing.log"
    )
}
```

## Systems I used to reproduce the problem

I believe I can trim the configuration further to still replicate the segfault and will do if/when requested.

I've reproduced the segfault on:

* Ubuntu 16.04 + rsyslog v8.30.0 (Adiscon PPA)
* Ubuntu 14.04 + rsyslog v8.30.0 (Adiscon PPA)
* Ubuntu 16.04, stock rsyslog 8.16.0-1ubuntu3

## Steps to reproduce

- `logger --tcp --port 514 --server TESTBOX_FQDN "hello!"`
- `logger --udp --port 514 --server TESTBOX_FQDN "hello!"`

This is whether I'm testing against a remote instance of rsyslog v8.30.0 running on an Ubuntu 16.04 or an Ubuntu 14.04 box using the latest v8.30.0 version from the Adiscon Ubuntu PPA. 

If testing the rsyslog instance from the same box you run logger, be aware (I wasn't until today) that the `logger` command from Ubuntu 14.04 box fails to work properly when using the `-n` option. The [Sending UDP packets with `logger` command](https://askubuntu.com/a/416480) askubuntu.com Q/A entry has more info.

## Debug/Valgrind output

[...]

PR #1957, opened 2017-11-02
Segfault when variable path consists of variables

PascalWithopf:
When no variables exist and a variable with at least one variable in its path was used, then rsyslog was searching for that non existant variable
Rsyslog now aborts it when no variables exist.
closes https://github.com/rsyslog/rsyslog/issues/1920


rgerhards:
#1920 indicates that this indeed fixed the issue.

deoren:
@PascalWithopf @rgerhards thanks!

Issue #1964, opened 2017-11-03
omelasticsearch: improve curl reply buffer handling


rgerhards:
As seen https://github.com/rsyslog/rsyslog/issues/1962 in currently the buffer (```pWrkrData->reply```) is allocated, realloced and freed with each request. This can be reduced to once per module. Or every now and then, to free up memory buffer... I have not changed that in the PR for #1962 because we are still fighting with a segfault issue and refactoring this has potential to complicate things. Should be done in the not so distant future after the segfault has been fixed.

PR #1968, opened 2017-11-04
! omelasticsearch: pWrkrData->reply nulling, freeing & reallocing cla…

bayaro:
…rification

### DIFF
--- a/plugins/omelasticsearch/omelasticsearch.c
+++ b/plugins/omelasticsearch/omelasticsearch.c
@@ -177,6 +177,8 @@ CODESTARTcreateWrkrInstance
 	pWrkrData->curlCheckConnHandle = NULL;
 	pWrkrData->serverIndex = 0;
 	pWrkrData->restURL = NULL;
+    pWrkrData->reply = NULL; /* init curl response buffer */
+    pWrkrData->replyLen = 0;
 	if(pData->bulkmode) {
 		pWrkrData->batch.currTpl1 = NULL;
 		pWrkrData->batch.currTpl2 = NULL;
@@ -368,6 +370,9 @@ checkConn(wrkrInstanceData_t *pWrkrData)
 	int i;
 	int r;
 	DEFiRet;
+
+    pWrkrData->reply = realloc(pWrkrData->reply, 0); /* clear the curl response buffer */
+    pWrkrData->replyLen = 0;
 
 	curl = pWrkrData->curlCheckConnHandle;
 	urlBuf = es_newStr(256);
@@ -777,7 +782,7 @@ parseRequestAndResponseForContext(wrkrInstanceData_t *pWrkrData,fjson_object **p
 
 		fjson_object_object_get_ex(result, "status", &ok);
 		itemStatus = checkReplyStatus(ok);
-		
+
 		char *request =0;
 		char *response =0;
 		if(ctx->statusCheckOnly)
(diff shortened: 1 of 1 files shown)


bayaro:
@rgerhards Hi, I have added a bit lines to avoid memory leaks and a possible segfault.

bayaro:
@rgerhards Thank you for the great rsyslog.

plugins/omelasticsearch/omelasticsearch.c
```diff
@@ -177,6 +177,8 @@ CODESTARTcreateWrkrInstance
 	pWrkrData->curlCheckConnHandle = NULL;
 	pWrkrData->serverIndex = 0;
 	pWrkrData->restURL = NULL;
+    pWrkrData->reply = NULL; /* initial init of pointer because of checkConn can be called before curlPost */
```
rgerhards: The comment is wrong - sorry. This is not possible. However, in the longer term this is the right spot to do it. 

Will you remove the misleading comment or shall I do after merge? If it is left in, you really confuse people working on the action engine.

bayaro: Sorry, I do not know your coding/release process. (And my English is not very good(( )
Please add any changes that required.

bayaro: yeap. I understand your idea.
Yes you can remove it. I have added it just to detalise the changes.

bayaro: removed

rgerhards:
MERGE ONLY after #1888 if then still needed.

rgerhards:
I think the now correctly merged commit c6a70e011b28d098210d9bc5c3662af499689793 had addressed all issues raised in this PR - except for the non-necessary nulling in createWrkrInstance. I think this should go in only after buffer handling is being refactored (#1964). Do you agree? If so, we could close this PR.

bayaro:
Because of in some cases we do not come to `finalize` section, we can call `checkConn` besides `curlPost`, etc.  It is less code and easy understanding if we will use the rule.

`pWrkrData->reply` should always point to real memory (or NULL)
1|) we init it at an application start
2) we clear it before every curl request (in two places: `checkConn` & `curlPost`)
3) we just reallocate it in all other places

With this rule we of course have an uncleared buffer between curl requests. But it is not very big, fortunately. But we will not have memory leaks and potential segfaults

Short example about current finalize in `checkConn` in  `c6a70e01`
Connection was broken after partially received response.
As result we have: `res != CURLE_OK`, not empty buffer ==> `ABORT_FINALIZE`.
===> next `checkConn` call raise memory leak.

rgerhards:
> Connection was broken after partially received response.
> As result we have: res != CURLE_OK, not empty buffer ==> ABORT_FINALIZE.

This will bail out via finalize_it: -- actually this is the whole point. finalize_it is the only exit path.

bayaro:
yeah,
`ABORT_FINALIZE` means "abort and go to finalize" :)
I understand.
so only 3 small comments left. I'll will add then directly to c6a70e01
yes, this PR can be closed.

rgerhards:
exactly. FYI: The ABORT_FINALIZE(errorcode), FINALIZE (return with current errorcode) and finalize_it: parts, together with CHKiRet() and CHKmalloc() mimic something along the lines of structured exception handling. The idea is that there always is an exit cleanup point (finalize_it:), so that we can always clean up properly. Thus we also do not do return other than in finalize_it.

Thanks again for all your good help, much appreciated!

rgerhards:
> ABORT_FINALIZE means "abort and go to finalize" :)

actually more like this: not "go to finalize" but "and finalize processing, do all cleanup" (which technically is goto finalize). The spirit is differently.

PR #1975, opened 2017-11-06
kafka: Fixed multiple issues found using clang 5.0 analyzer.

alorbach:
Added some assert() calls to avoid false positives.

### DIFF
--- a/plugins/imkafka/imkafka.c
+++ b/plugins/imkafka/imkafka.c
@@ -325,6 +325,8 @@ checkInstance(instanceConf_t *inst)
 
 	/* Set custom configuration parameters */
 	for(int i = 0 ; i < inst->nConfParams ; ++i) {
+		/* Avoid false positives in CLANG */
+		assert( inst->confParams == NULL);
 		DBGPRINTF("imkafka: setting custom configuration parameter: %s:%s\n",
 			inst->confParams[i].name,
 			inst->confParams[i].val);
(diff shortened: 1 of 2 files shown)


rgerhards:
will most probably merge via https://github.com/rsyslog/rsyslog/pull/1921

rgerhards:
looks like at least one the "false positives" was not a false positive. In my test run, the assert was triggered here: https://travis-ci.org/rsyslog/rsyslog/jobs/298075805

plugins/imkafka/imkafka.c
```diff
@@ -325,6 +325,8 @@ checkInstance(instanceConf_t *inst)
 
 	/* Set custom configuration parameters */
 	for(int i = 0 ; i < inst->nConfParams ; ++i) {
+		/* Avoid false positives in CLANG */
+		assert( inst->confParams == NULL);
```
rgerhards, line 329: do you really want to assert that it **is** NULL? Or do you wanted ```assert( inst->confParams != NULL);```?

plugins/omkafka/omkafka.c
```diff
@@ -1054,6 +1055,9 @@ loadFailedMsgs(instanceData *const __restrict__ pData)
 	uchar *puStr;
 	char *pStrTabPos;
 
+	/* Avoid false positives in CLANG */
```
rgerhards, line 1058: I think this is NOT a false positive. Without the change in line 1476 it could be NULL. So I would remove the comment. The assert itself probably makes sense, albeit I think it really wants to check against ```!= NULL```.

rgerhards:
I applied my proposed corrections to my merge branch and it worked. So I merged https://github.com/rsyslog/rsyslog/pull/1921 which means this PR has also been merged.

PR #1984, opened 2017-11-07
remove incorrect action option from testsuite conf

jvymazal:


jvymazal:
this links to https://github.com/rsyslog/librelp/pull/58

Issue #1983, opened 2017-11-07
Wrong mmjsonparse cookie causing segfault


jvymazal:
using custom cookie (instead of default `"@cee"`) and enclosing it in single quotes instead of double ones causes whole rsyslog to segfault (possibly in loop, depending on further configuration).

rgerhards:
Can you please post a concrete sample of the failing config? Seeing the PR I am unsure if I understand correctly.

jvymazal:
if you use something like `action(type="mmjsonparse" cookie='')` which is invalid and should be `action(type="mmjsonparse" cookie="")` then mmjsonparse accepts this as cookie and dumps rsyslog upon first processed JSON message. Combined with constant source of JSON messages abd e.g. systemd this leads to restart loop which is not very nice and IMHO config error should be more clearly handled. (took quite some time to figure out reason for the segfaults with current state)

rgerhards:
ah, ok, this is what I initially understood. That means the proposed fix in https://github.com/rsyslog/rsyslog/pull/1985 is not correct as the problem is not in mmjsonparse but instead in the core engine's parameter processing. We need to create a testbench test first to ensure that the problem will actually be fixed.

jvymazal:
Well, IIRC rsyslog itself has no problem with either single or double quotes so how (generically) would you like to fix this? I proposed the fix in the mmjsonparse because this is like the only place where type of quotes matter. The core engine just hands the cookie value to the module, or am I wrong?

rgerhards:
These quotes, if they would work, would not make it into the place where you checked it. The core has to strip them. I also do not think that you can use single quotes for parameters (but I need to check).

jvymazal:
Ok, done more thorough debugging and it turns out there was yet another cause for the original problem, closing this. Thanks for quick reply by the way.

PR #1985, opened 2017-11-07
aborting in case of invalid JSON cookie (leads to segfault if used)

jvymazal:


jvymazal:
attempt to address #1983

rgerhards:
see https://github.com/rsyslog/rsyslog/issues/1983#issuecomment-342742902

PR #1753, opened 2017-09-04
omprog: added error handling and transaction support for external plugins

jsiwrk:
(Submitting this PR again; I have cleaned up my fork.)

I've implemented this planned feature myself as a personal challenge, to have some fun and learn. What's your opinion? Did you intend to implement it this way?

I have manually tested the patch and it seems to work very well. Also, the CI tests show no regression (although I'm not sure if the omprog module was sufficiently covered).

For my part the patch would be ready to merge, EXCEPT that I have not implemented any automated test. Sorry for that. I would need some guidance on how to do it. On the other hand I'm not sure if I will be able to dedicate enough time to this personal project in the short term, so it would be great if someone else could implement at least a first basic test.

Other pending things apart from the tests:
* Update the rest of the documentation. At least, add reference doc about the new omprog settings. I the patch is finally accepted I could contribute to this, I think it'll be easy.
* The clang analyzer gives a (presumably false) warning I don't know how to eliminate.
* I'm pretty confident the Python skeleton is sufficiently pythonic. However I'm a Python newbie, so it would be great if someone with more knowledge could review it.
* It seems rsyslog does not require the standard getline function to be available, and I didn't find any quality implementation of it within the rsyslog code. So I implemented my own getline within omprog.c, but probably should be placed somewhere else, since it is reusable.

Note: I had to reorder functions in omprog.c due to new dependencies between them. I realise this will make comparing changes a bit difficult, sorry for that. I've not attempted to minimize this with forward declarations since I haven't seen them used elsewhere.

PR #1992, opened 2017-11-09
omprog: added tests for transactions with feedback

jsiwrk:
I have added two automated tests for the feedback feature introduced in #1753. They are still incomplete; consider this as work in progress. I have submitted the PR now because I'm stuck at two issues related to some behavior of the rsyslog core that I don't fully understand.

My doubts are related to how the rsyslog core handles transactions. I have marked them with two "TODO" marks in the code.

1) It seems that deferred messages (confirmed with RS_RET_DEFER_COMMIT) in a transaction are not retried if a later message in the transaction fails (with RS_RET_SUSPENDED). Please see the TODO in "omprog-transactions-feedback.sh". Is this the expected behavior?

2) If a commit fails (endTransaction returns a RS_RET_SUSPENDED), the next transaction hangs at the second message. The debug logs show that the action is correctly resumed, and sent the first message (which the action confirms with DEFER_COMMIT), but then action is no longer called, despite the queue is not empty yet (?). To reproduce this problem, see the TODO in "omprog-feedback-bin.sh"

@rgerhards, it would be great if you could take a careful look at the PR and help me with these two doubts. I have saved the rsyslog debug files for the two cases:
* [rsyslog-debug (deferred messages in failed transaction not retried).log](https://github.com/rsyslog/rsyslog/files/1455955/rsyslog-debug.deferred.messages.in.failed.transaction.not.retried.log)
* [rsyslog-debug (next transaction hangs after failed commit).log](https://github.com/rsyslog/rsyslog/files/1455957/rsyslog-debug.next.transaction.hangs.after.failed.commit.log)

I have carefully reviewed the handling of return codes in omprog.c, and I'm quite sure there isn't any bug there. I would say these behaviors are specific to the rsyslog core, and hence common to all plugins. Regarding the second case, I have tried to find out what's causing the hang by inspecting the code in action.c, but I don't fully understand how the retries are managed when a commit fails (it seems there is special retry code for this case).

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -453,7 +453,10 @@ TESTS +=  \
 	omprog-cleanup-with-outfile.sh \
 	omprog-noterm-cleanup.sh \
 	omprog-noterm-default.sh \
-	omprog-feedback.sh
+	omprog-feedback.sh \
+	omprog-transactions.sh \
+	omprog-transactions-failed-messages.sh \
+	omprog-transactions-failed-commits.sh
 if OS_LINUX
 TESTS += \
 	omprog-cleanup-when-unresponsive.sh \
@@ -1380,6 +1383,9 @@ EXTRA_DIST= \
 	omprog-noterm-default.sh \
 	omprog-noterm-unresponsive.sh \
 	omprog-feedback.sh \
+	omprog-transactions.sh \
+	omprog-transactions-failed-messages.sh \
+	omprog-transactions-failed-commits.sh \
 	omprog-multiparam-vg.sh \
 	testsuites/omprog-cleanup.conf \
 	testsuites/omprog-noterm.conf \
@@ -1392,6 +1398,10 @@ EXTRA_DIST= \
 	testsuites/term-ignoring-script.sh \
 	testsuites/omprog-feedback.conf \
 	testsuites/omprog-feedback-bin.sh \
+	testsuites/omprog-transactions-bin.sh \
+	testsuites/omprog-transactions.conf \
+	testsuites/omprog-transactions-failed-messages.conf \
+	testsuites/omprog-transactions-failed-commits.conf \
 	pipe_noreader.sh \
 	testsuites/pipe_noreader.conf \
 	uxsock_simple.sh \
--- a/tests/omprog-transactions-failed-commits.sh
+++ b/tests/omprog-transactions-failed-commits.sh
@@ -1,0 +1,130 @@
+#!/bin/bash
+# This file is part of the rsyslog project, released under ASL 2.0
+
+echo ===============================================================================
+echo '[omprog-transactions-failed-commits.sh]: test omprog with confirmMessages and useTransactions flags enabled, with commit failures'
+
+. $srcdir/diag.sh init
+. $srcdir/diag.sh startup omprog-transactions-failed-commits.conf
+. $srcdir/diag.sh wait-startup
+
+. $srcdir/diag.sh injectmsg 0 10
+
+. $srcdir/diag.sh wait-queueempty
+. $srcdir/diag.sh shutdown-when-empty
+. $srcdir/diag.sh wait-shutdown
+
+# Since the transaction boundaries are not deterministic, we cannot check for
+# an exact expected output. We must check the output programmatically.
+
+transaction_state="NONE"
+status_expected=true
+messages_to_commit=()
+messages_processed=()
+line_num=1
+error=
+
+while IFS= read -r line; do
+    if [[ $status_expected == true ]]; then
+        case "$transaction_state" in
+        "NONE")
+            if [[ "$line" != "<= OK" ]]; then
+                error="expecting an OK status from script"
+                break
+            fi
+            ;;
+        "STARTED")
+            if [[ "$line" != "<= OK" ]]; then
+                error="expecting an OK status from script"
+                break
+            fi
+            transaction_state="ACTIVE"
+            ;;
+        "ACTIVE")
+            if [[ "$line" != "<= DEFER_COMMIT" ]]; then
+                error="expecting a DEFER_COMMIT status from script"
+                break
+            fi
+            ;;
+        "COMMITTED")
+            if [[ "$line" == "<= Error: could not commit transaction" ]]; then
+                messages_to_commit=()
+                transaction_state="NONE"
+            else
+                if [[ "$line" != "<= OK" ]]; then
+                    error="expecting an OK status from script"
+                    break
+                fi
+                messages_processed+=("${messages_to_commit[@]}")
+                messages_to_commit=()
+                transaction_state="NONE"
+            fi
+            ;;
+        esac
+        status_expected=false;
+    else
+        if [[ "$line" == "=> BEGIN TRANSACTION" ]]; then
+            if [[ "$transaction_state" != "NONE" ]]; then
+                error="unexpected transaction start"
+                break
+            fi
+            transaction_state="STARTED"
+        elif [[ "$line" == "=> COMMIT TRANSACTION" ]]; then
+            if [[ "$transaction_state" != "ACTIVE" ]]; then
+                error="unexpected transaction commit"
+                break
+            fi
+            transaction_state="COMMITTED"
+        else
+            if [[ "$transaction_state" != "ACTIVE" ]]; then
+                error="unexpected message outside a transaction"
+                break
+            fi
+            if [[ "$line" != "=> msgnum:"* ]]; then
+                error="unexpected message contents"
+                break
+            fi
+            prefix_to_remove="=> "
+            messages_to_commit+=("${line#$prefix_to_remove}")
+        fi
+        status_expected=true;
+    fi
+    let "line_num++"
+done < rsyslog.out.log
+
+if [[ -z "$error" && "$transaction_state" != "NONE" ]]; then
+    error="unexpected end of file (transaction state: $transaction_state)"
+fi
+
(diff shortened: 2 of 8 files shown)


rgerhards:
are you testing with master branch? There were a couple of issues in earlier versions.

I'll provide more insight later, but am working on a couple of complex things and an explanation is not written within 10 minutes - pls bear a bit with me.

jsiwrk:
> are you testing with master branch? There were a couple of issues in earlier versions.
I've re-run the tests on yesterday's master and the behavior is the same.

jsiwrk:
Good news! After rebasing to master (still locally, will update the PR soon) I cannot reproduce the hang in the 2nd case.

In the changelog I've found this:

> - core/action bugfix: 100% CPU utilization on suspension of output module
    Triggering condition:
    * output module using the legacy transaction interface
      (e.g. omelasticsearch, omlibdbi)
    * output module needs to suspend itself
  In these cases, rsyslog enters a busy loop trying to resolve the
  suspend condition. The bug is rooted in rsyslog core action code.
  (...)
  This patch now also ensures that failed messages are properly
  handled and do not cause eternal hang.
  closes https://github.com/rsyslog/rsyslog/issues/2113

So it seems that bug #2113 was affecting my test (note: the omprog feedback feature suspends the module when the external plugin returns an error). @rgerhards, could you please confirm it was the same bug?

One doubt: from the changelog comment, I assume omprog is using the "legacy" transaction interface (?). I think I added the transaction interface to omprog when I added the feedback feature (in PR #1753). Does this mean that I added it incorrectly? Is there a "modern" transaction interface?

jsiwrk:
The 2nd issue has been solved with #2356, as explained above.

The 1st issue still persists. I have opened #2420 to handle it as such.

I have updated the PR assuming the current behavior of the rsyslog core is correct. I have included a TODO comment in the 'omprog-transactions-failed-messages.sh' test explaining what will need to be modified in the test (simply remove two lines of code) in case issue #2420 is finally considered a bug and resolved.

The PR is now ready to merge from my part. Even if a resolution or clarification to issue #2420 is not expected in the short term, I would suggest to still merge this PR as is, as it adds test coverage for the omprog transactions feature.

jsiwrk:
Rebased to master. @rgerhards: the "DO-NOT-MERGE_work-in-progress" label can be removed. As I suggested in my previous comment, I think it would be a good idea to merge this PR as is, since these tests are important IMO.

rgerhards:
solaris CI failure is unrelated - I'll babysit it (I finally need to find a cure for the racy test...).

rgerhards:
@jsiwrk I just noticed that both the racy test as well as this PR is for omprog -- any appetite to look into the test issue? I was about to disable all omprog tests until I find out what is wrong. Already did some work on that, but obviously not very successful...

In any case, I suggest to have a look at the CI failures as these may probably be in new tests. So for now I **do not** babysit CI.

rgerhards:
oops... I already restarted solaris, it's in the queue. here is the failure log: http://build.rsyslog.com:8010/builders/rsyslog%20solaris11x64%20rsyslog/builds/2083/steps/shell_8/logs/test-suite.log

jsiwrk:
@rgerhards: The new tests being proposed in this PR do not have synchronization issues, AFAIK. At least I have not seen any CI failure caused by these tests so far.

> I just noticed that both the racy test as well as this PR is for omprog -- any appetite to look into the test issue?

A quick glance at the failure you just posted and the code of the failed test suggests the test needs a rethink... Apparently the cause of the failure has been: omprog-noterm.sh is sent messages 000 to 005, then terminates its loop after receiving the first SIGUSR1 (the script explicitly sets a trap to do this), but the test code is expecting it to have processed message 009... That's not even a sync issue, it simply does not make sense (?).

Apart from this case, it seems there are potential sync issues with these tests, indeed. These issues could be related to the handling of signals, and the fact that waiting for the rsyslog queue to become empty does not mean that the script has processed the messages. Maybe they could be corrected by not making the test script terminate when receiving the signals, but making it write something to the output file (and then parsing the output file from the test code). And maybe using the feedback feature could help also. I'm not sure though, it needs careful thinking. I don't have time for this now, maybe in the future...

In any case, it doesn't seem a good idea to try to resolve these sync issues as part of this PR, since the new tests are unaffected, and moreover they cover an unrelated feature.

> I was about to disable all omprog tests until I find out what is wrong.

I suggest disabling only the tests that have sync issues (omprog-cleanup, omprog-noterm, etc.).

rgerhards:
@janmejay if you have a moment, could you look at this?

> A quick glance at the failure you just posted and the code of the failed test suggests the test needs a rethink... Apparently the cause of the failure has been: omprog-noterm.sh is sent messages 000 to 005, then terminates its loop after receiving the first SIGUSR1 (the script explicitly sets a trap to do this), but the test code is expecting it to have processed message 009... That's not even a sync issue, it simply does not make sense (?).
> 
> Apart from this case, it seems there are potential sync issues with these tests, indeed. These issues could be related to the handling of signals, and the fact that waiting for the rsyslog queue to become empty does not mean that the script has processed the messages. Maybe they could be corrected by not making the test script terminate when receiving the signals, but making it write something to the output file (and then parsing the output file from the test code). And maybe using the feedback feature could help also. I'm not sure though, it needs careful thinking. I don't have time for this now, maybe in the future...

rgerhards:

janmejay:
@rgerhards haven't looked at the whole conversation yet, but I did look at omprog-noterm-cleanup.sh. 

The test seems correct. It tests child-process cleanup and re-spawn. We deliver USR1 to get the script ready to kill itself after processing the next message. In the pkill and injectmsg loop, we make child-process exit each time it processes a message.

This test will succeed only if the child process is re-spawned.

In the failure that you linked, it appears the child wasn't re-spawned after message 005, which means either messages were lost during re-spawn or death wasn't detected and re-spawn was never attempted. 

It appears in the end (for the last message) re-spawn indeed happened. So whatever caused the messages to get lost, may be temporary problem or a corner-case.

The other scenario when this can fail is if sleeps are not enough for the process to die. After injectmsg, we wait for 0.5 seconds. The expectation is that the message will be handled by the child process within this time and it'll detect the absence of loop-monitor and terminate. If it detects the condition, but doesn't terminate in 0.5 seconds, that can lead to messages being lost, because rsyslog will continue to deliver it without pipe getting full and script will not process it because it has already dropped out of the loop, but hasn't exit yet.

Has this test been unstable over time?

jsiwrk:
I correct myself, because I realise I had not understood the test well enough:

> Apparently the cause of the failure has been: omprog-noterm.sh is sent messages 000 to 005, then terminates its loop after receiving the first SIGUSR1 (the script explicitly sets a trap to do this), but the test code is expecting it to have processed message 009... That's not even a sync issue, it simply does not make sense (?).

As @janmejay has explained this test is checking that rsyslog correctly restarts the process if it terminates prematurely. So it seems a sync issue indeed.

@janmejay: Maybe these sync problems could be corrected by using the feedback feature (confirmMessages="on"). In this mode, rsyslog will wait for a confirmation from the process when the process is started (or restarted), and after sending each message. This will synchronize rsyslog and the process, and maybe all the sleeps in the test code could be eliminated in this way (?).

rgerhards:
> Has this test been unstable over time?

yes, it was always problematic, but I think it became more problematic after the "big omprog enhancement". I opened this issue tracker, which contains useful extra information: https://github.com/rsyslog/rsyslog/issues/2403

As it looks, the test is indeed inherently racy. Maybe "confirmMessages" will actually work well enough. It's not exactly the same, but IMHO it solves the same test objective (USR1 handling check).

jsiwrk:
@janmejay: I'm currently trying to resolve #2403 using the feedback feature. I will continue commenting there. Just FYI.

rgerhards:
@jsiwrk it looks like the transaction tests also sometimes have problems, here is an example: http://build.rsyslog.com:8010/builders/rsyslog%20solaris11x64%20rsyslog/builds/2151

jsiwrk:
It seems that the test omprog-transactions-failed-commits.sh has caused rsyslog to hang (this test comes just after omprog-transactions-failed-messages.sh): 
```
PASS: rscript_http_request.sh
PASS: omprog-feedback.sh
PASS: omprog-transactions.sh
PASS: omprog-transactions-failed-messages.sh

command timed out: 300 seconds without output running ['gmake', 'check', 'V=0'], attempting to kill
process killed by signal 9
program finished with exit code -1
elapsedTime=1316.230378`
```

Note this test was initially affected by issue #2356, as I explained above. This bug caused rsyslog to hang when RS_RET_SUSPENDED was returned by a plugin when committing a transaction.

I would bet the reason of the failure has not been the test code, but the issue #2356, which maybe was not fully corrected. I see the bug was corrected by essentially [adding a sleep in the rsyslog core](https://github.com/rsyslog/rsyslog/pull/2356/commits/ee104ac13ba499e090d7a0aea6ccdbae742005d9#diff-c245bc2e6f98365bca144d6c877de320R1200), so maybe this wasn't enough...

Have you seen other failures in the new tests? Having more examples could help confirm or discard this theory. On the other hand, it is also possible that these test have some sync issues like the other omprog tests (which I'm currently reviewing), however I don't think the above error has been due to that.

rgerhards:
Just as this came up again, I currently have no idea of what exactly is causing this issue. I am working, as part of the testbench improvements, on better tooling to see what causes the hang. Until we have such, it's really impossible to do even an educated guess about what is going on. My gut feeling is that this kind of problem can be caused by very different reasons. Probably some of them are already solved. Sorry I have no better answer yet.

PR #2008, opened 2017-11-10
fix segfault in json variable subsystem

rgerhards:
experienced via mmexternal #1822, but potentially points to a large issues. Looks like interface spec is inconsistently implemented.


rgerhards:
correct fix #2014 merged, closing this w/o merge

PR #2010, opened 2017-11-10
added setting priority string to Relp TLS testsuite

jvymazal:


tests/testsuites/sndrcv_relp_tls_rcvr.conf
```diff
@@ -2,7 +2,7 @@ $IncludeConfig diag-common.conf
 
 module(load="../plugins/imrelp/.libs/imrelp")
 # then SENDER sends to this port (not tcpflood!)
-input(type="imrelp" port="13515" tls="on")
+input(type="imrelp" port="13515" tls="on" tls.prioritystring="SYSTEM")
```
rgerhards: This should be done in a new test (copy of this one) because otherwise we do not test the default case, which seems equally important.

rgerhards:
@jvymazal any interest in pursuing this further?

rgerhards:
I am assigning the 8.32 release goal. If I do not hear any further when we reach the milestone, I'll close the PR unmerged.

jvymazal:
I have no idea what is wrong with the kafka test on ubuntu16, and even less how could adding onther test affect it, I will try one more re-spin closer to the end of milestone and close this if it will not pass.

rgerhards:
@jvymazal hold on. I had tried to restart it in the hope it would self-heal. I guess it is caused by some failed download (and the test not handling that properly). But it 99.9999% is not affected by what you did. I'll sort it out, but it may take a day. I keep you posted.

rgerhards:
sry -- I didn't make it in 8.32 :-(

rgerhards:
update: we are working on a testbench for librelp natively. Hope to have it within the next days. Once done, we can reconsider how to tackle this beast here.

rgerhards:
It looks like some files are missing in EXTRA_DIST. I'll see that I add them via a separate commit if you do not mind (so that we finally can merge ;-)).

jvymazal:
Yes, thanks, that would be awesome.

rgerhards:
I have amended the commit, otherwise we may get into trouble with git bisect... CI may fail as I am introducing some stricter checks. If it fails for that reason, I'll handle that.

rgerhards:
@jvymazal finally merged :-) -- thanks for the PR and especially bearing with us!

jvymazal:
No problem, thanks for your help and all the CI work.

PR #2018, opened 2017-11-11
core/template bugfis: potential NULL pointer access at config load

rgerhards:
could happen if the config was loaded and the parameters could not
properly be processed. If so, this occured only during the startup
phase.

Detected by Coverity scan, CID 185318

### DIFF
--- a/template.c
+++ b/template.c
@@ -1874,7 +1874,7 @@ finalize_it:
 }
 
 /* Add a new template via the v6 config system.  */
-rsRetVal
+rsRetVal ATTR_NONNULL()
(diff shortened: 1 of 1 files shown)


davidelang:
I wonder if this will solve the occasional 'system crashes at boot' reports 
that we've been getting, turning them into 'invalid config' issues

David Lang

rgerhards:
Yeah, they could have been caused by this, or by a couple of other things that I have worked on. Actually, I came down that path trying to solve some of the very hard to fix and never reproduce things. I hope that 8.31 will be superior here. But I can't proof nor know for sure.

PR #574, opened 2015-11-02
add maxoutqsize counter to omkafka

chenryn:
librdkafka has its own outqueue. We need a maxoutqsize to help us optimize the confparam `queue.buffering.max.messages`.

PR #2041, opened 2017-11-15
omkafka: make maxoutqsize resettable

flisky:
This fixes up #574.

### DIFF
--- a/plugins/omkafka/omkafka.c
+++ b/plugins/omkafka/omkafka.c
@@ -60,7 +60,7 @@ DEFobjCurrIf(strm)
 DEFobjCurrIf(statsobj)
 
 statsobj_t *kafkaStats;
-int ctrQueueSize;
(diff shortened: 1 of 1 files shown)


rgerhards:
@alorbach any concerns re this PR? Looks more like a bugfix for me, but I wonder if there is a hard reason for the current implementation.

rgerhards:
@flisky can you please rebase to current master. That will make CI failure go away, most probably.

flisky:
No lucky, still the same failure - `line too long (121)`.

rgerhards:
ah, I see what goes on. Wonder why it wasn't triggerred before on master. Anyhow. Fix is underway...

rgerhards:
@flisky I would appreciate if you could rebase once again. This should finally solve the issue. There actually was a CI bug in master -- sorry for that.

Issue #2040, opened 2017-11-14
rsyslog doesn't collect log messages from a UDP socket that is bound to a nonlocal address


amg1127:
I am trying to setup an UDP log input whose socket is bound to the nonlocal address **192.168.99.99**:

```
    # ip -4 addr show
    1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1
        inet 127.0.0.1/8 scope host lo
        valid_lft forever preferred_lft forever
    2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP qlen 1000
        inet 10.4.28.91/26 brd 10.4.28.127 scope global eth0
        valid_lft forever preferred_lft forever
```

In order to achieve this result, I created the file "_/etc/rsyslog.d/udp-nonlocaladdr-bug.conf_" with the following content:

```
    # cat /etc/rsyslog.d/udp-nonlocaladdr-bug.conf
    module(load="imudp")

    input(type="imudp" address="127.0.0.1" port="514" RateLimit.Interval="0" ruleset="logtest")

    input(type="imudp" address="192.168.99.99" port="514" RateLimit.Interval="0" ruleset="logtest")

    ruleset(name="logtest") {
        action(type="omfile" file="/tmp/logtest")
    }
```

Then, after restarting "_rsyslog_" service, the intended socket is shown by "_netstat_" output:

```
    # systemctl restart rsyslog

    # netstat -pano | grep :514
    udp        0      0 192.168.99.99:514       0.0.0.0:*                           30134/rsyslogd       off (0.00/0/0)
    udp        0      0 127.0.0.1:514           0.0.0.0:*                           30134/rsyslogd       off (0.00/0/0)
```

However, "_rsyslog_" does not accept datagrams from that socket even though I bind the address to the "_eth0_" interface while the program is running:

```
    # ip -4 addr add dev eth0 192.168.99.99/24

    # logger --udp --server 192.168.99.99 'Nonlocal hello'

    # logger --udp --server 127.0.0.1 'Local hello'

    # cat /tmp/logtest
    Nov 14 18:23:17 localhost root: Local hello
```

The configuration works as intended only after the service is restarted at this point:

```
    # systemctl restart rsyslog

    # ip -4 addr show
[...]

PR #2049, opened 2017-11-17
imudp: proposed fix for issue #2040 - avoid socket descriptor leak while binding to a nonlocal address

amg1127:
https://github.com/rsyslog/rsyslog/issues/2040

### DIFF
--- a/runtime/net.c
+++ b/runtime/net.c
@@ -1490,6 +1490,7 @@ create_udp_socket(uchar *hostname,
 							break;
 						}
 					}
+					/* FALLTHRU */
 				default:
 					close(*s);
 					*s = -1;
--- a/runtime/net.c
+++ b/runtime/net.c
@@ -1490,7 +1490,7 @@ create_udp_socket(uchar *hostname,
 							break;
 						}
 					}
-					/* FALLTHRU */
+					/* FALLTHROUGH */
 				default:
 					close(*s);
 					*s = -1;
--- a/runtime/net.c
+++ b/runtime/net.c
@@ -1485,13 +1485,15 @@ create_udp_socket(uchar *hostname,
 						if (ipfreebind >= IPFREEBIND_ENABLED_WITH_LOG)
 							errmsg.LogMsg(0, RS_RET_OK_WARN, LOG_WARNING,
 								"bound address %s IP free", hostname);
-						continue;
+						goto bindsuccess;
 					}
 				}
 				close(*s);
 				*s = -1;
 				continue;
 			}
+bindsuccess:
+			;
 		}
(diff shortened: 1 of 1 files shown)


runtime/net.c
```diff
@@ -1485,13 +1485,15 @@ create_udp_socket(uchar *hostname,
 						if (ipfreebind >= IPFREEBIND_ENABLED_WITH_LOG)
 							errmsg.LogMsg(0, RS_RET_OK_WARN, LOG_WARNING,
 								"bound address %s IP free", hostname);
-						continue;
+						goto bindsuccess;
```
rgerhards: We should not use ```goto``` other than to exit a function (via it's finalizer).

amg1127:
@rgerhards I was able to remove the `goto` statement while maintaining the original handling of _EADDRNOTAVAIL_ error.

rgerhards:
I admit I am not happy with the solution- looks a bit hackish to me, introduces a kind of goto instead of looking at the actually desired code structure. On the other hand, the original code looks hackish at all, it's probably one of the darkest and oldest pieces inside rsyslog. What makes matters worse is that IPFREEBIND feature cannot be configured, but is hard coded. So if we patch this now, we introduce a change of behaviour. Well, well, this looks like something I really need to check on how to best clean up :-(

Sorry for the hassle, I should have noticed that when I merged the original change for IPFREEDBIND...

rgerhards:
oh oh, this **really** needs a refactoring...

rgerhards:
> What makes matters worse is that IPFREEBIND feature cannot be configured, but is hard coded.

I must correct myself: it's just that there are some definitions, but integer values are used ;-) So the good news is that it is configurable.

rgerhards:
I now have used your original patch (idea) to fix this in the refactored code, see #2080. Thanks a ton for all your help! You could not do this cleaner than you did without refactoring...

rgerhards:
closed via https://github.com/rsyslog/rsyslog/pull/2080

amg1127:
The hassle is not a problem at all. I enjoy your desire for keeping good coding standards into the software and am aware that my fix proposal was not as good as required.

I am used to coding small softwares for hobby, however I have been working as an IT system administrator for a long time, so I would not be able to study all "_rsyslog_" code and write a good patch in a timely manner. I opted to change the code as less as possible instead.

Thank you for solving my issue and restoring the IPFREEBIND feature. Enabling the "_ip_nonlocal_bind_" kernel option on the OS would let me a little annoyed, as it would allow all processes to bind nonlocal addresses. I am looking forward for the next stable RPM package.

PR #2071, opened 2017-11-21
testbench: improvements in elasticsearch testing

### LINKED ISSUE #2072: testbench: es test harness
The soon-to-be-merged code resets database on start and stop of ES. This prevents some types of tests, those that need to check across different start and stop of ES (retry-handling). Thus, the database purge must be done in an explicit step.

I do not want to hold merging the enhanced functionality, as it already provides extra value. But we should upgrade it after merge.

rgerhards:
now also closes #2072

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -28,6 +28,8 @@ if test "$unamestr" = "AIX"; then
 fi
 # AIXPORT END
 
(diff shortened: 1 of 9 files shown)


rgerhards:
closes #2072

Issue #2107, opened 2017-11-25
error during config processing: omfile: creating parent directories for file  'File exists' failed


nipakoo:
We are bumping into the race in makeFileParentDirs quite heavily. Got two rsyslogd threads repeatedly competing for creating parent directories on an NFS mount (service failing and restarting). Journal flooding with following messages until the rsyslog service is manually restarted:

Nov 24 23:16:05 xx-0 rsyslogd[1518]: error during config processing: omfile: creating parent directories for file  'File exists' failed: /var/log/node/xxxxx [v8.30.0 try http://www.rsyslog.com/e/2207 ]
Nov 24 23:16:05 xx-0 rsyslogd[1518]: error during config processing: omfile: creating parent directories for file  'File exists' failed: /var/log/node/yyyyy/auth.log [v8.30.0 try http://www.rsyslog.com/e/2207 ]

On our deployment, this is currently reproduced ~50% of the time.

rgerhards:
Thanks for the report. I am currently preparing next Tuesdays release, so I can unfortunately not look directly into it. Will try during next week.

My first impression is that actually a mutex lock is missing during file path creation.

nipakoo:
Yeah, there is a comment dating back to 2010 that acknowledges this. Seems the lock was not implemented because it will not fix race with other processes. Anyway, should be fixed as the most likely race should be within our process space.

rgerhards:
fixed via #2109

PR #2109, opened 2017-11-25
error during config processing: omfile: creating parent directories for file  'File exists'

nipakoo:
Add mutex for attempting to create parent directories of a log file. https://github.com/rsyslog/rsyslog/issues/2107

### DIFF
--- a/runtime/srutils.c
+++ b/runtime/srutils.c
@@ -194,6 +194,10 @@ uchar *srUtilStrDup(uchar *pOld, size_t len)
  * the creation fails in the similar way, we return an error on that second
  * try because otherwise we would potentially run into an endless loop.
  * loop. -- rgerhards, 2010-03-25
+ * The likeliest scenario for a prolonged contest of creating the parent directiories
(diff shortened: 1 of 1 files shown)


rgerhards:
Thx a ton for the PR!

It looks great on first view, but I need to review it in depth to check for a potential deadlock situation which I think could arise when multiple threads are executed by a dynafile action. Just wanted to let you know that this potentially requires a bit of extra time for the merge.

rgerhards:
I thought a bit about my concern: it is mostly invalid, the proposed cure is the right thing to do. There is just a border case during shutdown which could lead to a hang condition. That happens if the omfile thread needs to be cancelled by rsyslog core because it does not terminate within the timeout period. This is extremely unlikely to happen, though (a bit more likely with NFS I admit). So we need to have a cancel cleanup handler which unlocks the mutex under these circumstances.

Having said that, the newly introduced potential hang is less likely than what this PR fixes. So I think we are good to merge and can clean up the other one afterwards.

PR #2117, opened 2017-11-27
BUGFIX: fixing bugs pointed out by coverity

allanpark:


rgerhards:
It should be noted that this is a regression fix for a PR merged last Friday. So released code is not affected.

Issue #1728, opened 2017-08-17
crash with 1024+ of open TCP sessions due to select limit


devicenull:
8.29.0 seems to crash right around the ~1024 open TCP (+SSL) sessions:
```
(gdb) bt
#0  0x00007ffff69e81d7 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007ffff69e98c8 in __GI_abort () at abort.c:90
#2  0x00007ffff6a27f07 in __libc_message (do_abort=do_abort@entry=2, fmt=fmt@entry=0x7ffff6b31864 "*** %s ***: %s terminated\n") at ../sysdeps/unix/sysv/linux/libc_fatal.c:196
#3  0x00007ffff6ac2077 in __GI___fortify_fail (msg=msg@entry=0x7ffff6b3180a "buffer overflow detected") at fortify_fail.c:31
#4  0x00007ffff6ac0230 in __GI___chk_fail () at chk_fail.c:28
#5  0x00007ffff6ac1fe7 in __fdelt_chk (d=<optimized out>) at fdelt_chk.c:25
#6  0x00007ffff2a25dd4 in Add (pNsdsel=0x7fffdc002b50, pNsd=0x7fffded37860, waitOp=<optimized out>) at nsdsel_ptcp.c:91
#7  0x00007ffff5b8825c in RunSelect (sizeWorkset=128, workset=0x7ffff167c180, pThis=0x55555580c250) at tcpsrv.c:794
#8  Run (pThis=0x55555580c250) at tcpsrv.c:908
#9  0x00005555555b6184 in thrdStarter (arg=0x55555584c400) at ../threads.c:226
#10 0x00007ffff79b1dc5 in start_thread (arg=0x7ffff167d700) at pthread_create.c:308
#11 0x00007ffff6aaa76d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:113
```

I'm guessing this is because it's exceeded FD_SETSIZE.  I'd think this should be detected (although having more then 1024 hosts connected at once would be nice too)

rgerhards:
I have pushed a patch (see above) which includes a new error message handler. Please run this version.

The root problem seems to be that poll() cannot be used and we fall back to select(). The question now is why this happens. And, of course, we also need to find a way to ignore the execess FDs -- or disable the select() based code at all. So I would prefer to see why at all the code goes down that path.

[...]

PR #2123, opened 2017-11-27
network driver subsystem: improve error reporting

rgerhards:
see also https://github.com/rsyslog/rsyslog/issues/1728


rgerhards:
We get failures on platforms that actually need to fallback. We need to adapt the tests for these cases!

rgerhards:
looking at the test results, it must be re-thought if this feature really makes sense...

rgerhards:
This does not really make sense any more - we have improved the state in any case, so falling back to non epoll is no longer an issue.

PR #2141, opened 2017-11-30
Basic Imfile FEN API support for Solaris 

alorbach:
Support is currently limited to single files. Wildcards are not supported yet.


plugins/imfile/imfile.c
```diff
@@ -2522,6 +2562,204 @@ do_inotify(void)
...
+					"for fileinfo failed ");
+				ABORT_FINALIZE(RS_RET_INOTIFY_INIT_FAILED);
+			}
+
+			if ((pLstn->pfinf->fobj.fo_name = strdup((char*)pLstn->pszFileName)) == NULL) {
+				errmsg.LogError(1, RS_RET_INOTIFY_INIT_FAILED, "imfile: do_fen: alloc memory "
+					"for strdup failed ");
+				free(pLstn->pfinf);
+				pLstn->pfinf = NULL;
+				ABORT_FINALIZE(RS_RET_INOTIFY_INIT_FAILED);
```
rgerhards: shouldn't we better use a FEN-specific error code? I mean this might be easier in troubleshooting...

plugins/imfile/imfile.c
```diff
@@ -2522,6 +2562,204 @@ do_inotify(void)
...
+			* port is used to collect events from multiple sources.
+			*/
+			DBGPRINTF("imfile: do_fen: event received, processing ... \n");
+			switch (portEvent.portev_source) {
+				case PORT_SOURCE_FILE:
+					/* Call file events event handler */
+					fen_processEvent((lstn_t*)portEvent.portev_user, portEvent.portev_events);
+					break;
+				default:
+					errmsg.LogError(1, RS_RET_SYS_ERR, "imfile: do_fen: Event from unexpected source "
```
rgerhards: ```LogError()``` is now available natively (without ```errmsg.``` in front of it). I suggest to switch to that variant (it's even slightly faster, what really doesn't matter here ;-))

plugins/imfile/imfile.c
alorbach: Yes thats better, I added specific error code RS_RET_FEN_INIT_FAILED. Tests are running

Issue #1977, opened 2017-11-06
RainerScript function: http_request(url)


rgerhards:
Add a function to do a http request (via libcurl) and provide the response back as a string. This is useful for many applications, among other for querying Amazon AWS config data.

A mock-up of AWS state document has been placed here: http://www.rsyslog.com/files/testbench/aws-sim.php This can be used for testbench tests.

davidelang:
doing this for every log message that's processed would be insanely slow. If 
this is included, it needs to have a LOT of warnings about being protected by 
something that prevents it from being executed every message.

rgerhards:
Ack, but there are many legit uses for it.

davidelang:
On Mon, 6 Nov 2017, Rainer Gerhards wrote:

> Ack, but there are many legit uses for it.

I agree, but I'm not sure if it's valid to have it as a function.

We've talked a few times about external things affecting the config on the fly 
(adding or removing inputs for example to work with containers)

This sounds like a case where what you really want is the ability for an 
external 'thing' to set a global variable and update it periodically. Having 
something in the ruleset do this seems to be in the "I have a hammer, this looks 
like a nail" category :-)

David Lang

rgerhards:
I don't think so. Adding a module is not as flexible. We have had other
requests for similar functionality these days. Of course it's powerfull and
can be abused but there are other things inside rsyslog that can so also.

If you insist in it, I can disable it by default and require a global
directive to opt-in.

Rainer

Sent from phone, thus brief.

Am 06.11.2017 22:58 schrieb "David Lang" <notifications@github.com>:

PR #2146, opened 2017-12-01
rainerscript: new function "http_request"

rgerhards:
closes https://github.com/rsyslog/rsyslog/issues/1977


rgerhards:
@Whissi @mbiebl This creates a new build requirement for libcurl for rsyslog core. I think this is available on almost all modern systems by default. Anyhow, if you see a problem, let me know and I'll add an ability to disable the need for libcurl.

Whissi:
Don't forget embedded devices. Please make this optional, i.e. allow to disable cURL integration via configure.

mbiebl:
Fwiw, I share David's more general concerns in #1977 to build this directly into the core

rgerhards:
OK, I'll first provide a switch to disable it.

Building it outside the core requires a whole new callable interface (for functions). Unfortunately I have no time to design and implement this ATM. So I hope it is OK to provide the switch initially.

In general, everything can be abused. It won't offer better performance if we do it via a message modification module. A prime legit case is to obtain AWS metadata and the like. I don't want to add a specialized function for each and every of such calls. With the ability to do a http request very easily and a bit of script logic, this is nicely and easily done.

rgerhards:
FYI: PR for disabling: https://github.com/rsyslog/rsyslog/pull/2258

PR #2154, opened 2017-12-02
imuxsock: report system socket name received from systemd

rgerhards:
... of course only if this capability is used on systemd systems.


deoren:
Is this info emitted as normal log output or only via debug output?

deoren:
Great idea btw.

Issue #1824, opened 2017-10-09
update postgres to new config syntax.


GrayTShirt:
I'm interesting in updating the configuration syntax of the postgres output module, (refactoring the entire module for the new API?)

In terms of backwards compatibility what's expected?

GrayTShirt:
ah, I see there is a similar issue #818

rgerhards:
Hi Dan,

That would be really great!

Backwards compatibility is easy: old good should continue to work without
need to change anything.

Usually this means the legacy statements need to be preserved and be
functionality be added only via new style parameters.

If you have questions, let me know.

Rainer

Sent from phone, thus brief.

Am 09.10.2017 16:29 schrieb "Dan Molik" <notifications@github.com>:

GrayTShirt:
Is there an example of a module that preserves the legacy syntax parameters while supporting the new style parameters?

rgerhards:
All of them do, except some very new ones which do not have legacy statement. Ommysql might be a natural choice. I've also just done a change to the new transactional interface, so this might also be good to look at.

GrayTShirt:
hmmm, I got a chance to look at the code, how do you feel about a new module instead, I get a hunch that it will be less maintenance moving forward, and considering that new configuration will need to be written user-side, a few letters difference isn't a big deal.

davidelang:
There are a lot of modules with features that only exist in the new syntax. If 
you look through the various documentation pages, most features added during the 
8.x timeframe are going to be new syntax only.

you just don't change the legacy config section and only add things to the v6+ 
config.

David Lang

davidelang:
what's the difference? if the new module is enough different, it could be 
accepted. But if it's just a parameter or two, they should be combined.

rgerhards:
> I got a chance to look at the code, how do you feel about a new module instead

I am not sure I really understand: do you mean just because of the config? If so, that would not be a good reason to create a new module.  There really is no maintenance problem. You need to write (copy&paste) the old-style support code once and it will never be touched again. At least in practice with all other modules no noteworthy changes ever happened.

On contrast, if you duplicate the actual code (everything that does not deal with the config), you constantly would need to update two versions of almost the same code, what is a maintenance nightmare.

GrayTShirt:
That's a fair point. That's the answer I was looking for, in terms of maintenance...On Oct 10, 2017 3:19 AM, Rainer Gerhards <notifications@github.com> wrote:
[...]

Issue #895, opened 2016-03-20
ompgsql: add support for transactions


rgerhards:
Currently, each message is processed a single transaction. This should be changed to support the transactional interfaces in the rsyslog output module interface.

rgerhards:
closed via https://github.com/rsyslog/rsyslog/pull/2172

Issue #818, opened 2016-02-23
ompgsql: add support for v6+ config format


enp:
It is impossible to use module 'ompgsql' in action:

```
action(type="ompgsql" server="127.0.0.1" serverport="5432" db="eye" uid="eye" pwd="HIDDEN")
```

Error is:

```
rsyslogd: module name 'ompgsql' is unknown [v8.16.0 try http://www.rsyslog.com/e/2209 ]
rsyslogd: error during parsing file /etc/rsyslog.conf, on or before line 73: errors occured in file '/etc/rsyslog.conf' around line 73 [v8.16.0 try http://www.rsyslog.com/e/2207 ]
rsyslogd: CONFIG ERROR: there are no active actions configured. Inputs will run, but no output whatsoever is created. [v8.16.0 try http://www.rsyslog.com/e/2103 ]
```

But this works fine:

```
*.* :ompgsql:127.0.0.1,eye,eye,HIDDEN
```

Is it possible to adopt 'ompgsql' module for new configuration style?

rgerhards:
The module simply does not yet support the new-style format. Someone needs to convert it.

rgerhards:
closed via https://github.com/rsyslog/rsyslog/pull/2172

PR #2172, opened 2017-12-06
update postgres output module

GrayTShirt:
Fixes issue #1824, invalidates #1800, implements #895, #648, #818

### DIFF
--- a/tests/pgsql-basic-cnf6-vg.sh
+++ b/tests/pgsql-basic-cnf6-vg.sh
@@ -5,10 +5,18 @@
 
 psql -h localhost -U postgres -f testsuites/pgsql-basic.sql
 
-. $srcdir/diag.sh startup pgsql-basic.conf
+. $srcdir/diag.sh generate-conf
+. $srcdir/diag.sh add-conf '
+module(load="../plugins/ompgsql/.libs/ompgsql")
+if $msg contains "msgnum" then {
+	action(type="ompgsql" server="127.0.0.1"
+		db="syslogtest" user="postgres" pass="testbench")
+}'
+. $srcdir/diag.sh startup-vg
 . $srcdir/diag.sh injectmsg  0 5000
 . $srcdir/diag.sh shutdown-when-empty
-. $srcdir/diag.sh wait-shutdown
+. $srcdir/diag.sh wait-shutdown-vg
+. $srcdir/diag.sh check-exit-vg
 
 psql -h localhost -U postgres -d syslogtest -f testsuites/pgsql-select-msg.sql -t -A > rsyslog.out.log
 . $srcdir/diag.sh seq-check  0 4999
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,11 @@ compiler:
   - gcc
   - clang
 
-services: mysql
+services:
+  - elasticsearch
+  - mysql
+  - postgresql
+
 
 env:
   global:
@@ -112,9 +116,6 @@ matrix:
        compiler: "gcc"
        dist: trusty
        env: DO_COVERITY="YES",DO_CRON="YES", KAFKA="YES", GROK="YES"
-
-services:
-  - elasticsearch
 
 before_install:
   - if [ "$TRAVIS_OS_NAME" == "linux" ] && [ "${AD_PPA}x" == "x" ] ; then AD_PPA="v8-stable"; fi
--- a/plugins/ompgsql/ompgsql.c
+++ b/plugins/ompgsql/ompgsql.c
@@ -50,6 +50,9 @@
 
 MODULE_TYPE_OUTPUT
 MODULE_TYPE_NOKEEP
+MODULE_CNFNAME("ompgsql")
+
+static rsRetVal resetConfigVariables(uchar __attribute__((unused)) *pp, void __attribute__((unused)) *pVal);
 
 /* internal structures
  */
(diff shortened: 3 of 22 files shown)


rgerhards:
Thx for the patch set. The Solaris buildbot slave seems to have an unrelated issue. I will investigate and either re-start or report back if something is really with the patch.

GrayTShirt:
this PR also resolves #818

GrayTShirt:
Quick question; what's the difference between CODEqueryEtryPt_STD_OMODTX_QUERIES and CODEqueryEtryPt_TXIF_OMOD_QUERIES and generally which one is preferred?

GrayTShirt:
For prepared statements, I'm stuck trying to figure out the number of params a SQL statement should have. For instance, given a particular template, how many variables will it have? Additionally is that something I can get before `doAction`?

rgerhards:
> For prepared statements, I'm stuck trying to figure out the number of params a SQL statement should have. For instance, given a particular template, how many variables will it have? Additionally is that something I can get before doAction?

The whole interface is modeled after string passing. There *was* an array-interface that provided what you look for. However, the only user was omoracle, which was contributed and abandoned. So I have removed that interface a couple of years ago. I could look into reviving it, but it's unfortunately not done in an instant :-(

I would suggest that we go ahead with this PR as-is and then discuss how we can work towards an even better one. What do you think? Along the same lines, would you be willing to lend me a helping hand actually activating the tests during CI runs? I probably could use some advise on postgres packages and commands -- it's very long since I last used them...

rgerhards:
Note: the travis failure is caused by this: https://bugs.llvm.org/show_bug.cgi?id=35572

I have disabled that static analyzer for now in current master (buildbot still has it!). If you rebase, that error will go away.

GrayTShirt:
Sure, I can ensure we run tests on travis, I suggest we use the postgres provided apt repository. How do you feel about adding another build slot?

Too bad about the array interface. I'm still interested in an array/structured interface even if it's lots of work.

rgerhards:
> Sure, I can ensure we run tests on travis, I suggest we use the postgres provided apt repository. How do you feel about adding another build slot?

I see you already added commits (Y) I wouldn't mind another build slot, but I think it is not vitally necessary. I need to refactor Travis scripts in any case soon:
1. you really notice they have grown and been rewoven over time, it's no longer straightforward
1. we will soon hit the 50 minute single-build timeout, so I need to distribute over more VMs (I am very grateful that Travis permits us to run such an extended test suite on their systems!)

> Too bad about the array interface. I'm still interested in an array/structured interface even if it's lots of work.

I  need to check what needs to be touched, but I fear it is a lot. IIRC the interface was functional the last time in the v6 or v7 timeframe, and many things have changed since then. Most importantly, I fear that we need to do quite a bit of core changes. Given the current todo, this sounds a bit challenging :-(

If you would like to have a shot at it, we can see how we could work together on it. But I think I would at least need to get you decently started, else it probably is too frustrating (due to the many fine things to be taken into account).

rgerhards:
oh... in any case, I think we should merge **this** PR once the testbench runs. It really, really is very valuable! Thx for the great work.

GrayTShirt:
My pleasure, I'm happy to help

GrayTShirt:
There appears to be some backwards compatibility issues and a small memory leak. I'll continue debugging.

rgerhards:
> not really, other than grouping it near other diag.sh commands, and the only command before it is SQL prep

@GrayTShirt The reason it should be the first is that it does post-run cleanup plus it provides the test heading. In theory, future versions of the test harness may file if init is not the first statement. Highly suggest moving it up.

GrayTShirt:
Yea I realized that putting the test heading first might be a good idea :)

GrayTShirt:
Ok, the backwards compatibility has been fixed, but there still is a memory leak in the new configuration syntax without a defined template.

GrayTShirt:
I've added some more memory checks to figure out what the problem is, but I'm have trouble.

GrayTShirt:
When I run the valgrind tests locally I fail all of them with 36000 bytes still reachable. But I'm not getting the 150000 bytes that travis is getting.

rgerhards:
Which test do you refer to? Just do that I can find it quicker. Can you
also post a link to a relevant Travis log/run?

Sent from phone, thus brief.

GrayTShirt:
https://travis-ci.org/rsyslog/rsyslog/jobs/313672627
Test: pgsql-basic-cnf6-vg.sh

Also on phone.

rgerhards:
@GrayTShirt did you notice that the test was valgrind-clean but had a problem with the output data? From the bottom of its test log:
```
rsyslogd run exited with 0     --> valgrind is ok
scanf error in index i=0          --> expected output data seems not to be present
sequence error detected
```
The difference in remaining assigned memory can be due to --enable-valgrind option, if set differently (if set, it doesn't unload modules). It can also be due different libraries etc. Valgrind is used in the testbench so that any remaining allocs that are not reported as errors are 99.99% sure irrelevant.

GrayTShirt:
Aye yi yi!

GrayTShirt:
Shall I rebase into a single commit?

rgerhards:
I think that's a great idea!

tests/pgsql-basic-cnf6-vg.sh
```diff
@@ -0,0 +1,30 @@
...
+	action(type="ompgsql" server="127.0.0.1"
+		db="syslogtest" user="postgres" pass="testbench")
+}'
+. $srcdir/diag.sh startup-vg
+. $srcdir/diag.sh injectmsg  0 5000
+. $srcdir/diag.sh shutdown-when-empty
+. $srcdir/diag.sh wait-shutdown-vg
+. $srcdir/diag.sh check-exit-vg
+
+# we actually put the message in the SysLogTag field, so we know it doesn't use the default
```
rgerhards: I guess this comment is not fully correct ;-)

rgerhards: If I am right, a ```git commit --amend``` should do for the fix ;-)

GrayTShirt:
I deleted that confusing comment :)

GrayTShirt:
this PR will also resolve #895 and #648

rgerhards:
I updated the initial description. Let me know (or update) if I missed some issues that it also closes ;-)

GrayTShirt:
Those are all the open issues and pull-requests I found that relate to this change, good to go!

rgerhards:
@GrayTShirt great work! Thanks a ton, this will be very useful for rsyslog users :-) I'll merge and close this PR now, feel free to open a new issue tracker for the transactional interface. We can then check what needs to be done to make it happen!

rgerhards:
oh... I forgot: it would be great if you could update the doc as well: https://github.com/rsyslog/rsyslog-doc/blob/master/source/configuration/modules/ompgsql.rst

GrayTShirt:
Hahaha! I'm already working on it!

Issue #2164, opened 2017-12-04
[mmexternal] invalid pointer issue in v8.31.0


simnyc:
Hi rsyslog team,
We're seeing an issue with rsyslog v8.31.0 that we're not seeing with other hosts running v8.30.0 with identical config:

```
*** Error in `rsyslogd': free(): invalid pointer: 0x00007fa39c00254f ***
```

Our system:

```
$ uname -a
Linux i-0fa276cee897620ac 3.13.0-123-generic #172-Ubuntu SMP Mon Jun 26 18:04:35 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
$ dpkg -l | grep rsyslog
ii  rsyslog                          8.31.0-0adiscon3trusty1                    amd64        a rocket-fast system for log processing
ii  rsyslog-gnutls                   8.31.0-0adiscon3trusty1                    amd64        TLS protocol support for rsyslog
ii  rsyslog-mmjsonparse              8.31.0-0adiscon3trusty1                    amd64        Parsing/handling of CEE/Lumberjack JSON messages in rsyslog
ii  rsyslog-relp                     8.31.0-0adiscon3trusty1                    amd64        RELP protocol support for rsyslog
$ rsyslogd -N 1
rsyslogd: version 8.31.0, config validation run (level 1), master config /etc/rsyslog.conf
rsyslogd: End of config validation run. Bye.
```

When trying to start the service here is what we see in `dmesg`:

```
dmesg -T
[Mon Dec  4 20:34:02 2017] init: rsyslog main process (31994) killed by ABRT signal
[Mon Dec  4 20:34:02 2017] init: rsyslog main process ended, respawning
[Mon Dec  4 20:34:02 2017] init: rsyslog main process (32024) killed by ABRT signal
[Mon Dec  4 20:34:02 2017] init: rsyslog main process ended, respawning
[Mon Dec  4 20:34:03 2017] init: rsyslog main process (32047) killed by ABRT signal
[Mon Dec  4 20:34:03 2017] init: rsyslog main process ended, respawning
[Mon Dec  4 20:34:03 2017] init: rsyslog main process (32070) killed by ABRT signal
[Mon Dec  4 20:34:03 2017] init: rsyslog main process ended, respawning
[Mon Dec  4 20:34:03 2017] init: rsyslog main process (32093) killed by ABRT signal
[Mon Dec  4 20:34:03 2017] init: rsyslog main process ended, respawning
[...]

PR #2174, opened 2017-12-07
mmextrenal: fix double-free

rgerhards:
regression from 8.31.0

fixes https://github.com/rsyslog/rsyslog/issues/2164


rgerhards:
fail is missing clang-5.0 packages -- happens every now and then at LLVM repo. Not a showstopper.

PR #2181, opened 2017-12-09
case preservation for RFC5424 SD-PARAMs

alanrobson:
please also see Adiscon [36205--786c48a]

### DIFF
--- a/plugins/mmpstrucdata/mmpstrucdata.c
+++ b/plugins/mmpstrucdata/mmpstrucdata.c
@@ -216,7 +216,7 @@ parseSD_NAME(uchar *sdbuf, int lenbuf, int *curridx, uchar *namebuf)
 		if(   sdbuf[i] == '=' || sdbuf[i] == '"'
 		   || sdbuf[i] == ']' || sdbuf[i] == ' ')
 			break;
-		namebuf[j] = tolower(sdbuf[i]);
+		namebuf[j] = sdbuf[i];
 		++i;
 	}
 	namebuf[j] = '\0';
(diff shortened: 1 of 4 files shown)


plugins/mmpstrucdata/mmpstrucdata.c
```diff
@@ -216,7 +216,7 @@ parseSD_NAME(uchar *sdbuf, int lenbuf, int *curridx, uchar *namebuf)
 		if(   sdbuf[i] == '=' || sdbuf[i] == '"'
 		   || sdbuf[i] == ']' || sdbuf[i] == ' ')
 			break;
-		namebuf[j] = tolower(sdbuf[i]);
+		namebuf[j] = sdbuf[i];
```
rgerhards, line 219: We cannot do this unconditionally as others may depend on existing behavior. So we need to introduce an option to keep the lowercasing or not.

alanrobson: I think that's beyond my coding abilities - if you can give me some pointers I may be able to add the option.

rgerhards: If you don't mind, I would add it -- it's just a couple of minutes for me as I know exactly where to place it. Actually quicker than to describe it with words. I could post the resulting PR for your review, so you know the next time (if any). What do you think?

rgerhards: I take the non-response as a go-ahead ;-)

rgerhards:
**Please do NOT close** -- this will automatically be merged via https://github.com/rsyslog/rsyslog/pull/2364

PR #2193, opened 2017-12-10
ompgsql - fix unsafe string copy in defaults

GrayTShirt:


GrayTShirt:
looks like raspbian failure is unrelated.

rgerhards:
yeah, keep cool ;-) I'll follow these things, especially as I know some parts of the CI are still somewhat fragile (especially the newly added raspberry). Actually, I've already resubmitted the build, but it's right now in the queue. If an issue comes up that looks unrelated, just wait a litle but, I'll babysit them (except if I forget ;-)).

plugins/ompgsql/ompgsql.c
```diff
@@ -363,8 +363,8 @@ setInstParamDefaults(instanceData *pData)
 	pData->trans_commit  = 100;
 	pData->trans_age     = 60;
 	pData->port          = 5432;
-	strncpy(pData->user, "postgres", strlen("postgres"));
-	strncpy(pData->pass, "postgres", strlen("postgres"));
+	strncpy(pData->user, "postgres", strlen("postgres") + 1);
```
rgerhards: thinking twice... shouldn't that be ```..., sizeof(pData->user));``` as ```strncpy()``` intends to guard that the buffer is not overrun?

dmolik: That makes sense.

rgerhards:
raspbian is having some strange issue... This PR is fine for sure, I merge it now despite raspbian.

Issue #1979, opened 2017-11-06
RainerScript: add capability to parse JSON string


rgerhards:
It is useful to be able to parse arbitrary JSON strings without the need of an external message modification module. For example, together with #1977 this can be used to parse out specific fields from Amazon AWS configuration info. There are also other ample uses for such functionality.

rgerhards:
proposed interface:
```
int json_parse(json_string, var_root);
```
returns 0 if success, something else on failure.
Supports the ```script_error()``` interface.

davidelang:
On Sun, 10 Dec 2017, Rainer Gerhards wrote:

> proposed interface:
> ```
> int json_parse(json_string, var_root);
> ```
> returns 0 if success, something else on failure.
> Supports the ```script_error()``` interface.

Why this instead of a mmnormalize call with a inline format?

Yes, it's shorter, but is it enough faster to matter? (to make up for the 
long-term maintinance? or is it so trivial that long-term maintinance is 
non-existant?)

rgerhards:
It's a bit faster, but it is more about the usability. It's far more intuitive for a very common use case. Maintainance-wise it's not a real problem, it's around 10 lines of code.

davidelang:
Ok, I try to double check such things to keep too much from building up.

and while I don't subscribe to the Python "there should only be one way to do 
something" I don't like adding too many different ways. :-)

rgerhards:
I am with you in the same camp and appreciate you keeping an eye on those things. It's better if you have to reason, so nothing accidently slips in.

PR #2198, opened 2017-12-11
config: implement function parse_json()

rgerhards:
It is useful to be able to parse arbitrary JSON strings without the need
of an external message modification module. For example, together with
    https://github.com/rsyslog/rsyslog/issues/1977
this can be used to parse out specific fields from Amazon AWS
configuration info. There are also other ample uses for such functionality.

closes https://github.com/rsyslog/rsyslog/issues/1979


davidelang:
hmm, does it make sense to redefine mmjsonparse to be a wrapper (either of this 
function or of a trivial mmnormalize call), I don't think there is anything that 
mmjsonparse does better than these two.

rgerhards:
> hmm, does it make sense to redefine mmjsonparse to be a wrapper (either of this
> function or of a trivial mmnormalize call), I don't think there is anything that
> mmjsonparse does better than these two.

I am not sure... Maybe we can do this via the optimizer, but other than that it looks pretty complex to do. If we should pursue this further, I would appreciate a dedicated issue tracker, as this here will be closed upon merge.

Issue #2048, opened 2017-11-16
Building with "--enable-debug --enable-rtinst" is broken: ‘pdbgFuncDB’ undeclared


Whissi:
```
In file included from rsyslog.h:654:0,
                 from dnscache.c:30:
debug.h:134:44: error: ‘pdbgFuncDB’ undeclared here (not in a function)
  int dbgCALLStaCK_POP_POINT = dbgEntrFunc(&pdbgFuncDB, __FILE__, __func__, __LINE__);
                                            ^~~~~~~~~~
debug.h:134:66: error: ‘__func__’ is not defined outside of function scope [-Werror]
  int dbgCALLStaCK_POP_POINT = dbgEntrFunc(&pdbgFuncDB, __FILE__, __func__, __LINE__);
                                                                  ^~~~~~~~
debug.h:146:66: error: expected declaration specifiers or ‘...’ before string constant
  dbgprintf("%s:%d: %s: log point\n", __FILE__, __LINE__, __func__);
                                                                  ^
debug.h:146:12: error: expected declaration specifiers or ‘...’ before string constant
  dbgprintf("%s:%d: %s: log point\n", __FILE__, __LINE__, __func__);
            ^
debug.h:129:46: note: in definition of macro ‘dbgprintf’
 #define dbgprintf(...) r_dbgprintf(__FILE__, __VA_ARGS__)
                                              ^~~~~~~~~~~
debug.h:146:66: error: expected declaration specifiers or ‘...’ before string constant
  dbgprintf("%s:%d: %s: log point\n", __FILE__, __LINE__, __func__);
                                                                  ^
debug.h:129:46: note: in definition of macro ‘dbgprintf’
 #define dbgprintf(...) r_dbgprintf(__FILE__, __VA_ARGS__)
                                              ^~~~~~~~~~~
debug.h:146:66: error: expected declaration specifiers or ‘...’ before numeric constant
  dbgprintf("%s:%d: %s: log point\n", __FILE__, __LINE__, __func__);
                                                                  ^
debug.h:129:46: note: in definition of macro ‘dbgprintf’
 #define dbgprintf(...) r_dbgprintf(__FILE__, __VA_ARGS__)
                                              ^~~~~~~~~~~
debug.h:146:58: error: expected declaration specifiers or ‘...’ before ‘__func__’
  dbgprintf("%s:%d: %s: log point\n", __FILE__, __LINE__, __func__);
                                                          ^
[...]

PR #2205, opened 2017-12-12
cleanup: remove unused runtime instrumentation code

rgerhards:
The --enable-rtinst configure switch and associated instrumentation
code dates back to the time pre-(valgrind, address sanitizer, ...).
It has not been actively used by rsyslog developers for years and is
inferior to the new debugging tools.

With this commit, we remove the --enable-rtinst option and do some
mild refactoring of the code itself. As it turned out, fully removing
the debug code requires a bit more of time, and we push this as a
follow-on activity. It's not as urgent as getting us to build
properly under all configure switches.

closes https://github.com/rsyslog/rsyslog/issues/2048

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -616,21 +616,6 @@ AC_ARG_ENABLE(debug_symbols,
 )
 
 
-# runtime instrumentation
-AC_ARG_ENABLE(rtinst,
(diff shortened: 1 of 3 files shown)


rgerhards:
new issue for final cleanup: https://github.com/rsyslog/rsyslog/issues/2211

Issue #1459, opened 2017-03-09
Missing epoll impementation(nsdpoll_gtls) for gnutls


nirmoy:
We are facing issue where  rsyslog crashes after 1024 connections reached with gnutls. It seems it's famous issue of select() and there is no implementation for epoll for gnutls. I wonder why its missing and how much efforts will be required to implement it.

npajkovsky:
I have simple fix for this, but I agree, that rsyslog badly needs *nsdpoll_gtls*

``` C
--- rsyslog-8.4.0.orig/runtime/unlimited_select.h                                                                                                             
+++ rsyslog-8.4.0/runtime/unlimited_select.h                                                                                                                  
@@ -32,6 +32,10 @@
 #ifdef USE_UNLIMITED_SELECT                                                                                                                                  
 # undef FD_ZERO                                                                                                                                              
 # define FD_ZERO(set) memset((set), 0, glbl.GetFdSetSize());                                                                                                 
+                                                                                                                                                             
+/* remove the fortify size checking ... assume we know what we do */                                                                                         
+#undef __FD_ELT                                                                                                                                              
+#define __FD_ELT(d) (d / __NFDBITS)                                                                                                                          
 #endif 
```

rgerhards:
Sorry for the late reply. I have been hyper-busy this year and am only recovering the past couple of weeks. This issue still exits, and I hope to be finally able to work on it. As can be seen, I have tanatively assigned the 8.32 release goal -- if we don't make it by then, 8.33 will be it.

tech note: at least an interim cure is to support unlimited select in nsdsel_ptcp, which would work-around the issue here.

rgerhards:
mhhh... actutally nsdsel_ptcp already does this. However, it requires the ```--enable-unlimited-select``` configure switch during build time. I probably need to experiment if we can turn it on by default without causing issues...

rgerhards:
The segfault will be solved by #2619

PR #2209, opened 2017-12-13
build system: enable --unlimited-select by default

rgerhards:
This is essential and seems to be often forgotten. We enable it
in the hope that it does not break anything. If so, we need to
re-evaluate the default and situation.

see also https://github.com/rsyslog/rsyslog/issues/1459

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -554,13 +554,13 @@ fi
 
 # support for unlimited select() syscall
 AC_ARG_ENABLE(unlimited_select,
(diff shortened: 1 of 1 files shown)


rgerhards:
OK, this fails miserably enough so that we need to consider an alternate path ;-)

rgerhards:
abandoning this in favor of https://github.com/rsyslog/rsyslog/issues/2615, which is the proper cure.

Issue #2168, opened 2017-12-05
build fails with older versions of kafka versions


mbiebl:
Today I tried to build rsyslog 8.31.0 on Debian stretch, which has librdkafka 0.9.3-1.

This failed with: 
```
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I../.. -I../../runtime -I../.. -I../../grammar -I/usr/include/libfastjson -W -Wall -Wformat-security -Wshadow -Wcast-align -Wpointer-arith -Wmissing-format-attribute -Werror=implicit-function-declaration -g -fno-strict-aliasing -Wall -Wextra -Wundef -Wnested-externs -Wwrite-strings -Wpointer-arith -Wmissing-declarations -Wmissing-prototypes -Wstrict-prototypes -Wredundant-decls -Wno-unused-parameter -Wno-missing-field-initializers -Wdeclaration-after-statement -Wformat=2 -Wold-style-definition -Wcast-align -Wformat-nonliteral -Wformat-security -Wsign-compare -Wstrict-aliasing -Wshadow -Winline -Wpacked -Wmissing-format-attribute -Wmissing-noreturn -Winit-self -Wmissing-include-dirs -Wunused-but-set-variable -Warray-bounds -Wimplicit-function-declaration -Wreturn-type -Wswitch-enum -Wswitch-default -Wno-error=unused-parameter -Wno-error=missing-field-initializers -pthread -Wdate-time -D_FORTIFY_SOURCE=2 -DPATH_PIDFILE=\"/run/rsyslogd.pid\" -g -O2 -fdebug-prefix-map=/build/rsyslog-8.31.0=. -fstack-protector-strong -Wformat -Werror=format-security -c omkafka.c  -fPIC -DPIC -o .libs/omkafka_la-omkafka.o
omkafka.c: In function 'writeKafka':
omkafka.c:1597:0: error: unterminated argument list invoking macro "DBGPRINTF"
 ENDmodInit
 
omkafka.c:658:4: error: 'DBGPRINTF' undeclared (first use in this function)
    DBGPRINTF("omkafka: Failed to produce to topic '%s' (rd_kafka_produce)"
    ^~~~~~~~~
omkafka.c:658:4: note: each undeclared identifier is reported only once for each function it appears in
omkafka.c:658:4: error: expected ';' at end of input
omkafka.c:658:4: error: expected declaration or statement at end of input
omkafka.c:658:4: error: expected declaration or statement at end of input
[...]

PR #2210, opened 2017-12-13
omkafka bugfix: build fails with older versions of librdkafka

rgerhards:
closes https://github.com/rsyslog/rsyslog/issues/2168


rgerhards:
New Debian 9 buildslave has been installed with distro-standard librdkafka, so we can reliably test.

Issue #2218, opened 2017-12-14
lmsig_ksi_ls12: fix coverity issue 185331


allanpark:
Coverity reported an an issue 185331 in lib_ksils12.c: FORWARD_ NULL.

PR #2219, opened 2017-12-14
Issue $2218: lmsig_ksils12: fixing coverity issue 185331

allanpark:
fixes https://github.com/rsyslog/rsyslog/issues/2218


rgerhards:
@allanpark note: I am currently working on the CI system. This is most probably the cause of the CI failures. I'll automatically restart when done. If a real problem comes up, I'll notify you.

rgerhards:

PR #2221, opened 2017-12-14
pmrfc3164: fix invalid whitespace-skip

rgerhards:
regression from yesterday's commit, deteced by Coverity Scan,
CID 186292


rgerhards:

Issue #1684, opened 2017-07-20
refactor code to use new error/internal message interface


rgerhards:
The errmsg object previously used, as many others do, indirect calls via function pointers exchanged at module load. The core intent was to support platforms that were unable to dynamically link loaded modules to symbols already in core. Nowadays, this problem is (long) gone away, and so we can use the regular direct calls. This may even speed up some process, but reduces complexity in any case.

Note that there is no hard need for a quick migration. I keep this issue primarily as a reference whenever we change some code in this regard. It is anticipated that it will take quite a while until it is closed (so do not assign a milestone unless you are sure you are serious in doing this non-essential work).

PR #2222, opened 2017-12-14
omzmq3: use new errmsg interface

jgerhards:
see also https://github.com/rsyslog/rsyslog/issues/1684


rgerhards:

PR #2227, opened 2017-12-15
imfile: implemented wildcard support into FEN mode (solaris)

alorbach:
Partitionally moved inotify code to global imfile code as needed
for FEN Mode. Enabled all wildcard tests that are supported by FEN API.


plugins/imfile/imfile.c
```diff
@@ -251,19 +250,32 @@ typedef struct fileTable_s fileTable_t;
...
 static int allocMaxDirs;
 static int currMaxDirs;
+#endif /* #if defined(HAVE_INOTIFY_INIT) || (defined(OS_SOLARIS) && defined (HAVE_PORT_SOURCE_FILE)) --- */
 
 /* the following two macros are used to select the correct file table */
 #define ACTIVE_FILE 1
 #define CONFIGURED_FILE 0
 
+#ifdef HAVE_INOTIFY_INIT
+/* support for inotify / fen mode */
```
rgerhards: Is HAVE_INOTIFY_INIT overloaded for FEN as well? This might be confusing...

plugins/imfile/imfile.c
```diff
@@ -1666,22 +1691,88 @@ dirsAdd(uchar *dirName, int* piIndex)
...
+		psztmp = strchr((char*)dirs[newindex].dirNameBfWildCard, '*');
+		if (psztmp != NULL) {
+			*psztmp = '\0';
+			/*	Now set NULL Byte on last directory delimiter occurrence,
+			*	This makes sure that we have the current base path to create a watch for! */
+			psztmp = strrchr((char*)dirs[newindex].dirNameBfWildCard, '/');
+			if (psztmp != NULL) {
+				*psztmp = '\0';
+			} else {
+				DBGPRINTF("dirsAdd unexpected error #1 for dir '%s'\n",
```
rgerhards: would it be better to expose this as a user error message? It looks like this may cause trouble later on...

plugins/imfile/imfile.c
```diff
@@ -1759,6 +1850,81 @@ dirsAddFile(lstn_t *__restrict__ pLstn, const int bActive)
 	RETiRet;
```
rgerhards, line 1844: I think cleanup handlers for memory leaks (one malloc succeeded, next one not) may be missing. But I may be wrong.

plugins/imfile/imfile.c
```diff
@@ -1759,6 +1850,81 @@ dirsAddFile(lstn_t *__restrict__ pLstn, const int bActive)
...
+		pThis->pszDirName = existing->pszDirName; /* read-only */
+	} else {
+		pThis->pszDirName = (uchar*)strdup((char*)newdirname);
+	}
+	pThis->pszBaseName = (uchar*)strdup((char*)newname);
+	if(asprintf((char**)&pThis->pszFileName, "%s/%s", (char*)pThis->pszDirName, (char*)newname) == -1) {
+		DBGPRINTF("lstnDup: asprintf failed, malfunction can happen\n");
+		ABORT_FINALIZE(RS_RET_OUT_OF_MEMORY);
+	}
+	pThis->pszTag = (uchar*) strdup((char*) existing->pszTag);
```
rgerhards: is failure of strdup() ok? Also needs to be checked.

plugins/imfile/imfile.c
```diff
@@ -1759,6 +1850,81 @@ dirsAddFile(lstn_t *__restrict__ pLstn, const int bActive)
...
+	lstn_t *pThis;
+	CHKiRet(lstnAdd(&pThis));
+
+	/* Use dynamic dirname if newdirname is set! */
+	if (newdirname == NULL) {
+		pThis->pszDirName = existing->pszDirName; /* read-only */
+	} else {
+		pThis->pszDirName = (uchar*)strdup((char*)newdirname);
+	}
+	pThis->pszBaseName = (uchar*)strdup((char*)newname);
```
rgerhards: check strdup fail

plugins/imfile/imfile.c
```diff
@@ -2589,76 +2686,108 @@ fen_printevent(int event, char *pname)
...
-			, fobjp->fo_name, errno);
+	DBGPRINTF("fen_removeFile removing listener for '%s', dirIdx %d\n",
+		pLstn->pszFileName, dirIdx);
+	if(	pLstn->bRMStateOnDel) {
+		statefn = getStateFileName(pLstn, statefile, sizeof(statefile), NULL);
+		/* Get full path and file name */
+		getFullStateFileName(statefn, toDel, sizeof(toDel));
+		bDoRMState = 1;
 	} else {
-		DBGPRINTF("imfile: fen_monitorDirectory: Failed to stat directory: %s - errno %d\n"
```
rgerhards, line 2603: should this be exposed to the user?

plugins/imfile/imfile.c
```diff
@@ -2589,76 +2686,108 @@ fen_printevent(int event, char *pname)
...
 	}
 
-	/* Register file event */
-	fobjp->fo_atime = statFile.st_atim;
-	fobjp->fo_mtime = statFile.st_mtim;
-	fobjp->fo_ctime = statFile.st_ctim;
-	if (port_associate(port, PORT_SOURCE_FILE, (uintptr_t)fobjp,
-				pLstn->pfinf->events, (void *)pLstn) == -1) {
-		/* Add error processing as required, file may have been deleted/moved. */
-		errmsg.LogError(1, RS_RET_SYS_ERR, "imfile: fen_monitorDirectory: Failed to register directory: "
```
rgerhards, line 2615: This is called incorrectly. I do not know what the ```1``` as first parameter means. Errno should go here, with no explicit reference to errno later on.

plugins/imfile/imfile.c
```diff
@@ -2589,76 +2686,108 @@ fen_printevent(int event, char *pname)
...
-		pLstn->bPortAssociated = 1;
+	pollFile(pLstn, NULL);	/* one final try to gather data */
+	lstnDel(pLstn);		/* Delete Listener now */
+	fileTableDelFile(&dirs[dirIdx].active, pLstn); /* Remove from filetable */
+	if(bDoRMState) {
+		DBGPRINTF("fen_removeFile unlinking '%s'\n", toDel);
+		if(unlink((char*)toDel) != 0) {
+			char errStr[1024];
+			rs_strerror_r(errno, errStr, sizeof(errStr));
+			errmsg.LogError(0, RS_RET_ERR, "fen_removeFile: could not remove state "
```
rgerhards: errno is automatically expanded to the error string if given as first parameter. Please change.

plugins/imfile/imfile.c
```diff
@@ -2589,76 +2686,108 @@ fen_printevent(int event, char *pname)
...
 	/* Compare filename first */
 	if (strcmp(fobjp->fo_name, (const char*)pLstn->pszFileName) == 0){
-		DBGPRINTF("imfile: fen_processEvent: matching file found: %s %s\n", fobjp->fo_name);
+		DBGPRINTF("fen_processEventFile: matching file found: '%s'\n", fobjp->fo_name);
 
 		/* Get File Stats */
 		if (!(revents & FILE_EXCEPTION) && stat(fobjp->fo_name, &statFile) == -1) {
-			DBGPRINTF("imfile: fen_processEvent: Failed to stat file "
-				": %s - errno %d\n", fobjp->fo_name, errno);
+			DBGPRINTF("fen_processEventFile: Failed to stat file: %s - errno %d\n",
```
rgerhards: expose error to user?

plugins/imfile/imfile.c
```diff
@@ -2667,46 +2796,268 @@ fen_processEvent(port_event_t* pportEvent, lstn_t *pLstn, int revents)
...
 				pollFile(pLstn, NULL);
+			} else if (revents & FILE_RENAME_FROM) {
+				/* File has been renamed which means it was deleted. remove the file*/
+				fen_removeFile(pLstn);
+				FINALIZE;
 			}
 
 			/* If exception, no need to re-register. */
 			if (revents & FILE_EXCEPTION) {
+				fen_removeFile(pLstn);
```
rgerhards: is this still an error case (ABORT in next stmt)?

plugins/imfile/imfile.c
```diff
@@ -2667,46 +2796,268 @@ fen_processEvent(port_event_t* pportEvent, lstn_t *pLstn, int revents)
...
+				DBGPRINTF("fen_DirSearchFiles file '%s' idx %d already being monitored ... \n", file, ftIdx);
+			} else {
+				DBGPRINTF("fen_DirSearchFiles setup new monitor for dynamic file '%s' \n", file);
+
+				/* duplicate listener firs */
+				pLstnNew = pLstn;
+				if(lstnDup(&pLstnNew, basefilename, dirs[dirIdx].dirName) != RS_RET_OK) {
+					DBGPRINTF("fen_DirSearchFiles failed to duplicate listener for '%s'\n",
+						pLstn->pszFileName);
+					ABORT_FINALIZE(RS_RET_IO_ERROR);
```
rgerhards, line 2896: original error code from lstnDup should be preserved and use here!

plugins/imfile/imfile.c
```diff
@@ -2667,46 +2796,268 @@ fen_processEvent(port_event_t* pportEvent, lstn_t *pLstn, int revents)
...
+					free(pLstnNew->pfinf);
+					pLstnNew->pfinf = NULL;
+					ABORT_FINALIZE(RS_RET_IO_ERROR);
+				}
+				/* Event types to watch. */
+				pLstnNew->pfinf->events = FILE_MODIFIED; /* not needed/working |FILE_DELETE|FILE_RENAME_TO|FILE_RENAME_FROM;*/
+				pLstnNew->pfinf->port = glport;
+
+				/* Add Listener to configured dirs tab */
+				dirsAddFile(pLstnNew, ACTIVE_FILE);
```
rgerhards, line 2918: does it for sure always succeed?

plugins/imfile/imfile.c
```diff
@@ -2667,46 +2796,268 @@ fen_processEvent(port_event_t* pportEvent, lstn_t *pLstn, int revents)
...
+							dirs[dirIdx].configured.listeners[iListIdx].pLstn,
+							FILE_MODIFIED /*dirs[dirIdx].configured.listeners[iListIdx].pLstn->pfinf->events*/);
+					} else {
+						DBGPRINTF("fen_processEventDir Listener for %s already associated\n",
+							dirs[dirIdx].configured.listeners[iListIdx].pLstn->pszFileName);
+					}
+				}
+			}
+		}
+	} else
```
rgerhards: please use { .. }

plugins/imfile/imfile.c
alorbach: No, that's a wrong comment. Going to fix

plugins/imfile/imfile.c
alorbach: Yes we can change these debug prints to errors

plugins/imfile/imfile.c
alorbach: I don't see mallocs in this function?

plugins/imfile/imfile.c
alorbach: there are currently many other positions where there is no CHKmalloc() applied at the moment. 
Maybe that should be fixed in another commit / branch?

plugins/imfile/imfile.c
alorbach: same as above

plugins/imfile/imfile.c
alorbach: that function doesn't exist anymore.

plugins/imfile/imfile.c
alorbach: The errmsg.LogError was copied from do_inotify.

rgerhards: so we have two places to fix? ;-)

plugins/imfile/imfile.c
rgerhards: Well, we need to have this fixed before we can merge, so it doesn't matter so much to me. If we don't fix, Coverity will report a software defect and I think its not smart to willingly introduce that.

plugins/imfile/imfile.c
alorbach: Changed. There are also some other positions in existing code where errno is preprocessed like this.

rgerhards: yup, it's at a couple of places still in the source code. If you have seen some, can you fix them as well.

plugins/imfile/imfile.c
alorbach: That's a good question. I would say not in this case. Because if user configured a logfile which has been created on rsyslog startup, rsyslog would output this error in each loop.

plugins/imfile/imfile.c
rgerhards: I should add that these are due to historical development. The LogError originally did not support errno-expansion. When we added it, we just semi-automatically changed the existing calls. However, that old stuff should finally be phased out.

plugins/imfile/imfile.c
rgerhards: ok - let's create an issue for this specifically. Probably requires wider audience feedback.

plugins/imfile/imfile.c
alorbach: FILE_EXCEPTION is generated by the system when something really bad happened.  Unfortunetally there is no documentation on when FILE_EXCEPTION is generated. Just that it needs to be checked in the few code samples that can be found.

rgerhards: I would add this as a comment. Saves questions ;-)

plugins/imfile/imfile.c
alorbach: lstnDup can only fail with RS_RET_OUT_OF_MEMORY and RS_RET_ERR at the moment. 
Also other calls of lstnDup look the same. I would consider changing this in another issue (refactoring the code?).

rgerhards: ok, so let's do that - but isn't it a 5 minute job?

plugins/imfile/imfile.c
alorbach: Fixed

plugins/imfile/imfile.c
alorbach: Yes unless in out of memory conditions.

plugins/imfile/imfile.c
alorbach: added CHKmalloc

plugins/imfile/imfile.c
alorbach: added CHKmalloc

plugins/imfile/imfile.c
alorbach: Fixed it in fen code. 
Other places are in inotify code. 
Should be done in another commit.

plugins/imfile/imfile.c
alorbach: Ok I am going to fix them with this commit as well.

plugins/imfile/imfile.c
alorbach: added comment

rgerhards:
@alorbach there seems to be a real issue in CI testing

Issue #2224, opened 2017-12-14
script: add new function to_hex()


rgerhards:
This should convert the provided integer value to a hexadecimal string representation.

jay7x:
Can we have some hashing functions as well? It's frequently used to distribute flow evenly between multiple rulesets/receivers. For now it's hard to do using RainerScript only w/o external modules.

rgerhards:
Please open a new issue for it  :-)

davidelang:
another useful function to distribute traffic would be a random() function. In 
addition, it could be used to be able to do things like "log 1% of traffic" on a 
high volume server

rgerhards:
@davidelang actually, ```random()``` exists, I think for quite a while. Maybe doc is missing (could not check right now).

PR #2242, opened 2017-12-17
rscript: add int2Hex

jgerhards:
closes https://github.com/rsyslog/rsyslog/issues/2224


rgerhards:
There is a CI problem on CentOS 7, but it is unrelated. Nevertheless, I re-initiated the run to see if we can find out more about this unrelated problem so that we can nail it.

rgerhards:
@jgerhards could you please update the doc?

rgerhards:
a rebase to current master will most likely make the CI failures go away -- it looks like we use some new settings that does not match the base source version of this branch.

PR #2246, opened 2017-12-17
Add is_time() RainerScript function.

mrworkman:
Related to #1624.

@jgerhards, what do you think?

@rgerhards,
Some Travis tests are yielding the following wrt inlining `wtiSetScriptErrno()`. I'm not entirely sure why it can't do the inlining. Any insights? I haven't even figured out how to reproduce the warning/error on my system yet..
```sh
In file included from rainerscript.c:54:0:
In function ‘doFuncCall’:
../runtime/wti.h:134:1: error: inlining failed in call to ‘wtiSetScriptErrno.part.14’: call is unlikely and code size would grow [-Werror=inline]
 wtiSetScriptErrno(wti_t * const pWti, uint8_t newval)
 ^
cc1: error: called from here [-Werror=inline]
../runtime/wti.h:134:1: error: inlining failed in call to ‘wtiSetScriptErrno.part.14’: call is unlikely and code size would grow [-Werror=inline]
 wtiSetScriptErrno(wti_t * const pWti, uint8_t newval)
.
.
.
```


grammar/rainerscript.c
```diff
@@ -1291,7 +1291,7 @@ str2num(es_str_t *s, int *bSuccess)
 	}
 	if(c[0] == '-') {
 		neg = -1;
-		i = -1;
```
mrworkman: This seems to be a bug! 

`i` gets set to `-1`, and then is used in the while loop below - reading outside the bounds of the array.  The result is that it cannot parse negative numbers. 

I'm pretty sure this should be `1`, not `-1`.

rgerhards:
> Some Travis tests are yielding the following wrt inlining wtiSetScriptErrno()

argh, that's a compiler decision. IMHO inline is pretty broken in C99. It's probably not 100% correct, let me check. I wonder this came not up earlier.

rgerhards:
If you rebase to master, the inline function error should go away.

grammar/rainerscript.c
```diff
@@ -2368,6 +2368,79 @@ doFuncCall(struct cnffunc *__restrict__ const func, struct svar *__restrict__ co
 		varFreeMembers(&r[0]);
 		break;
 	}
+	case CNFFUNC_IS_TIME: {
+		char *fmt = NULL;
+
+		cnfexprEval(func->expr[0], &r[0], usrptr, pWti);
+		str = (char*) var2CString(&r[0], &bMustFree);
+
+		// Check if the optional 2nd parameter was provided
```
rgerhards: It would be good to do the "real" work not inside the switch, but rather in a dedicated function. We try to keep the code managable size. I am sorry I seem to have overlooked this in previous PRs.

grammar/rainerscript.c
rgerhards: created https://github.com/rsyslog/rsyslog/issues/2255

grammar/rainerscript.c
mrworkman: No problem. I'll do that with this one, then do a separate PR to get the other ones set up like that too.

rgerhards: great, thx a lot! I admit I may have missed that myself as well... ;-)

mrworkman: Done.

grammar/rainerscript.c
```diff
@@ -2374,66 +2447,16 @@ doFuncCall(struct cnffunc *__restrict__ const func, struct svar *__restrict__ co
 		cnfexprEval(func->expr[0], &r[0], usrptr, pWti);
 		str = (char*) var2CString(&r[0], &bMustFree);
 
+		bMustFree2 = 0;
```
rgerhards: This smells strongly like a constant expression... (see next comment)

grammar/rainerscript.c
```diff
@@ -2374,66 +2447,16 @@ doFuncCall(struct cnffunc *__restrict__ const func, struct svar *__restrict__ co
...
-				}
-			}
-		}
-
-		// If not a valid date/time string, set 'errno'
-		if (ret->d.n == 0) wtiSetScriptErrno(pWti, RS_SCRIPT_EINVAL);
+		ret->d.n = doFunc_is_time(str, fmt, &r[0], pWti);
 
 		if(bMustFree) free(str);
 		if(bMustFree2) free(fmt);
```
rgerhards: ... which means both of the lines can be removed -- or am I overlooking something?

grammar/rainerscript.c
mrworkman: Just initializing it, because it is conditionally set in the following `if ()`. I think one of the Solaris build systems was giving a warning because `bMustFree2` "contains garbage".

rgerhards: yeah ... but have you read the other comment below? I think the problem is not that it is not initialized, but that it is queried albeit it is not used.

Let me re-phrase the question: what is ```fmt``` and when does it actually need to be free'ed? Maybe it's even a logic bug.

mrworkman: Yes, I read it ;)

`fmt` is the optional second parameter to `is_time()`. If `is_time()` is called without the format specifier, then `fmt` won't have to be freed. If it is specified, it has to be converted to a cstr, and freed after. No?

Maybe I should have mentioned the usage in the pull request details. I can do that later when I have access to an actual keyboard (in a few hours).

rgerhards: let me check... I did not see whre bMustFree2 was written except in the initializer. Maybe gotten lost in diff... (and, yeah, I know that with the real keyboards -- you are doing extremely well ;-)).

rgerhards: damn, my bad -- I really got lost in the diff. Overlooked the place where bMustFree2 is written...

rgerhards:
@mrworkman all looks good to me - are you also OK with merge? Could you do a quick doc update?

mrworkman:
Yes, I'm OK with merging. I will see if I can update the docs later this afternoon.

mrworkman:
Opened PR rsyslog/rsyslog-doc#450 for the doc update.

Issue #2128, opened 2017-11-28
remove support for outdated GuardTime signature interface


rgerhards:
The code still contains code for the original, outdated and no longer functioning pre-KSI signature interface. This should be removed.

PR #2261, opened 2017-12-19
remove obsolete pre-KSI GuardTime signature interface

rgerhards:
this is no longer functional starting Jan 2018 as it
reached end of life.

closes https://github.com/rsyslog/rsyslog/issues/2128


rgerhards:
debian9 buildbot slave is not yet ready for use -- ignore

Issue #2228, opened 2017-12-15
omfile with DynaFile parameter truncates file path to 200 bytes


Dieken:
tools/omfile.c uses macro MAXFNAME which is 200 defined in runtime/syslogd-types.h,  I'm not sure which line truncates my /very/long/file/path.

Probably these lines in tools/omfile.c, prepareFile():
```
 609     /* the copies below are clumpsy, but there is no way around given the
 610      * anomalies in dirname() and basename() [they MODIFY the provided buffer...]
 611      */
 612     uchar szNameBuf[MAXFNAME+1];
 613     uchar szDirName[MAXFNAME+1];
 614     uchar szBaseName[MAXFNAME+1];
 615     ustrncpy(szNameBuf, newFileName, MAXFNAME);
 616     szNameBuf[MAXFNAME] = '\0';
 617     ustrncpy(szDirName, (uchar*)dirname((char*)szNameBuf), MAXFNAME);
 618     szDirName[MAXFNAME] = '\0';
 619     ustrncpy(szNameBuf, newFileName, MAXFNAME);
 620     szNameBuf[MAXFNAME] = '\0';
 621     ustrncpy(szBaseName, (uchar*)basename((char*)szNameBuf), MAXFNAME);
 622     szBaseName[MAXFNAME] = '\0';
```

Dieken:
I have a DynaFile template "/very/long/path-YYYY-mm-dd-HH", this bug breaks my hourly log rotation,  the file size of truncated file path "/very/long/path-YYYY" keeps increasing :-(

rgerhards:
Thanks for the report. This seems not to be intentional. Nevertheless the question: what is your platform?

Dieken:
@rgerhards  I'm using Ubuntu 14.04 and Ubuntu 16.06,  thanks for your quick response!

davidelang:
I've seen OS limits in that range (250 or so characters IIRc), so this may 
actually be an OS limitation you are running into instead of a rsyslog one

doing a little googling, the max file name length in linux is 255 characters, 
and max path size is 4096 characters, but there is some ability to change these.

On other operating systems, the limits are going to be different.

Dieken:
@davidelang Yes, different OSes have different limits, and people can change the limit(but I doubt anybody will change).

On Linux there is function to check the limits: http://man7.org/linux/man-pages/man3/pathconf.3.html, but I think it’s unnecessary to limit, if it’s too long the OS will fail the file operation.

[...]

PR #2262, opened 2017-12-19
core bugfix: MAXFNAME was set too low

rgerhards:
it just permitted 200 chars, with almost all systems permitting for
more. I tried to find a more portable way to find out the actual max,
but this turned out horrible. The next solution would have been to use
dynamic alloc, but that often is overkill. So I now settle to just
increasing the value to 4KiB. This is the Linux limit, and it is by
far the highest I could find. This should be good to go for quite
some while but should not put too much stressure on the stack alloc.

closes https://github.com/rsyslog/rsyslog/issues/2228

### DIFF
--- a/runtime/syslogd-types.h
+++ b/runtime/syslogd-types.h
@@ -4,7 +4,7 @@
  *
  * File begun on 2007-07-13 by RGerhards (extracted from syslogd.c)
  *
- * Copyright 2007-2014 Adiscon GmbH.
+ * Copyright 2007-2017 Adiscon GmbH.
  *
  * This file is part of the rsyslog runtime library.
  *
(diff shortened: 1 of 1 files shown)


Dieken:
The limit is used as both max file name length and max file path length, anyway, this is fine, the OSes will do the final check.

But Rsyslog allocates file name or path on stack, for example, in omfile.c,  this might be a risk.

rgerhards:
> But Rsyslog allocates file name or path on stack, for example, in omfile.c, this might be a risk.

why do you think so? Do you see a missing length check?

Dieken:
I mean it may use too much stack, for example 3*4KB in omfile.

Long time not program in C, my memory still stays in the old times when stack space is small, this is probably not true now;-)

rgerhards:
That's no issue, we request 10MiB stack space. Not doing on stack would mean a full malloc/free cycle, which is very low. So we use stack space for such things.

Issue #2249, opened 2017-12-18
imfile: FEN Mode, expose error on stat failure of configured files? 


alorbach:
When static configured files do not exist on startup, they are checked in each processing loop in do_fen if they exist. On failure an error is printed into the debug log. 

```
		/* Get File Stats */
		if (!(revents & FILE_EXCEPTION) && stat(fobjp->fo_name, &statFile) == -1) {
			DBGPRINTF("fen_processEventFile: Failed to stat file: %s - errno %d\n",
				fobjp->fo_name, errno);
			ABORT_FINALIZE(RS_RET_FILE_NO_STAT);
		}

```

Question is should we LogError instead? it would generated an error each time the configured file is checked but does not exist.

rgerhards:
@alorbach  can you please add a bit of context so that others can make a decision.  Thx

alorbach:
Updated description

rgerhards:
Question: did this also apply to the inotify case?

alorbach:
I need to review the code in order to tell more.

alorbach:
So one problem is that FEN API doesn't provide us a file name once a DIR update is processed. Or I am too stupid to find a property for the filename when something changes in a directory? I have searched the documentation up and down ;). 

So we check for all configured files when a change happens to the parent directory. 
If we would output a LogError here, it would be generated each time if a file is modified in the parent directory. This could become very busy. 

It would be easier to decide if there was a global LogError limit by module. For example max 10 errors in 60 minutes per module for example. In this case we could use LogError much more often without having to care about spamming ;).

rgerhards:
@kortemik do you have an opinion on this issue?

rgerhards:
> It would be easier to decide if there was a global LogError limit by module. For example max 10 errors in 60 minutes per module for example. In this case we could use LogError much more often without having to care about spamming ;).

well, there kind of is ;-) The internal message rate limiter trims messages. But it does not work on a per-module basis.

kortemik:
I see the point, notify event's can come with quite high frequency and therefore log flood of misconfigured entires is very much a likely problem. But how often the DIR events get fired? Only when file is moved,removed,created and such?

http://www.sptci.com/docs/cpp/fen.html

rgerhards:
First thing I see is this:
```
==28533== Process terminating with default action of signal 6 (SIGABRT)
==28533==    at 0x5CC6428: raise (raise.c:54)
[...]

PR #2267, opened 2017-12-19
imfile: let user know if file was not found during FEN event notifica…

rgerhards:
…tion

closes https://github.com/rsyslog/rsyslog/issues/2249

** PROPOSAL** for #2249


rgerhards:
no objections, so I merge...

PR #2269, opened 2017-12-19
omhttpfs: use new errmsg interface

jgerhards:
see also https://github.com/rsyslog/rsyslog/pull/2259


rgerhards:
centos7 failure is known unrelated

PR #2285, opened 2017-12-21
codestyle: partially reduce line length to 120

PascalWithopf:
lines are still checked for a length of 125 because
these are just some of the lines


rgerhards:
fixed compiler errors (via maintainer edit)

Issue #1625, opened 2017-06-20
Redis module should RPUSH instead of LPUSH


pemontto:
Most all logging systems that push to Redis use RPUSH as it pushes to the tail of a list rather than the head. I think this behaviour is desirable, if consumers can't keep up then newer logs could be left unprocessed.

Perhaps it could be a configurable option.

See [omhiredis.c#L193](https://github.com/rsyslog/rsyslog/blob/2576a4cdf49b33a8a0ac39e5ee5faa1698a6d2ee/contrib/omhiredis/omhiredis.c#L193)

Redis documentation for [RPUSH](https://redis.io/commands/rpush) and [LPUSH](https://redis.io/commands/lpush)

rgerhards:
Thanks for opening this issue. Unfortunately, I was hyper-busy with many other things for quite some month and not all issues were properly followed up. I am currently triaging everything. 

That said, I wonder if what you request is really the change of the constant string you point to. If so, would you please suggest a parameter name for this and a bit of doc that would describe it? Then I can proably make the change myself albeit I have no idea of what I am doing ;-)

@taotetek Or would you like to comment on this?

taotetek:
@rgerhards aha - looks like logstash does indeed default to an RPUSH. I don't think I'd want to switch the default, as it could result in a nasty surprise for someone, but configurable would be easy enough.

rgerhards:
@taotetek so it would just be the change of the constant?

rgerhards:
for the record: I have asked @jgerhards to add a config parameter for this, as it looks like this was all that was asked for. But we do not know what we are doing, albeit it looks right... ;-)

PR #2305, opened 2017-12-27
omhiredis: add option to use RPUSH instead of LPUSH

jgerhards:
if userpush is on, RPUSH will be use instead of LPUSH

closes https://github.com/rsyslog/rsyslog/issues/1625


rgerhards:
@jgerhards can you please update the doc?

rgerhards:
testbench failures are unrelated, so merging (I am still working on more intense testbench, thus the problems).

PR #2306, opened 2017-12-27
omelasticsearch: add eol to every elastic response for error file

bayaro:
At this time all jsons are writing to the error file without an eol. So it is not very continent to monitor, catch changes, observe and process this file.


rgerhards:
The CI failures look unrelated -- we recently made the CI system more picky. I'll have a deeper look at the PR soon.

bayaro:
Thank you. As I have seen some issue with connectivity appears.

Issue #2143, opened 2017-11-30
Don't write a pid file when in socket activation mode


mbiebl:
When rsyslog is used under systemd, it uses socket activation and the sd_notify API to signal service readiness. In that case, writing (and cleaning up) a pid file is not necessary. It's actually harmful if you want to lock down rsyslog.service with the systemd builtin mechanisms, as you need one more writable directory.

Please consider not writing a pid file when using socket activation mode or provide a command line switch which allows to turn off the pid file creation and can be used in rsyslog.service.

Thanks for considering

rgerhards:
very good suggestion, thx!

Would it be appropriate to do not write (nor check) pid files if ```-n``` option is given. This means no backgrounding, and so the need for a pid file is more than questionable. That's also used in rsyslog.service file. On the other hand, it's a (slight) change in behaviour. But I would say that's fine.

Comments? @davidelang ?

mbiebl:
> Would it be appropriate to do not write (nor check) pid files if -n option is given.

Sounds ok to me. The altenative would be to check for getenv("LISTEN_PID") resp. if nFds > 0

rgerhards:
@mbiebl What would you find to be more appropriate. I think that "fixing" ```-n``` might even be appropriate in cases without systemd, in the (unusual) case that you run two instances in non-background in parallel... It kind of feels more like the right thing to me, but I am not sure if going overboard...

mbiebl:
I don't have a strong preference. Both would be ok to me.

rgerhards:
thx -- I'll keep it open a bit for further comments. Will change it in 8.32 timeframe for sure.

jay7x:
I'd say it's better to keep current behaviour for backward compatibility but provide specific cmdline switch to disable pidfile creation (`--no-pidfile` or `-I` e.g.). So systemd unit's options may be adjusted later.

Just remember there are old logrotate scripts which are not aware of systemd still (check /etc/logrotate.d/syslog in CentOS RPM for example).

mbiebl:
> Just remember there are old logrotate scripts which are not aware of systemd still (check /etc/logrotate.d/syslog in CentOS RPM for example).

I assume those old logrotate scripts check for the existence of a pid file and would be broken as well once the .service file start using ```--no-pidfile``` ?

rgerhards:
I thought that is a good point. Just curios: how will logrotate work when no pidfile is given?

mbiebl:
[...]

PR #2309, opened 2017-12-28
rsyslogd: add capability to specify that no pid file shall be written

rgerhards:
Command line option -iNONE provides this capability. This utilizes the
pre-existing -i option, but uses the special name "NONE" to turn of the
pid file check feature. Turning off is useful for systems where this no
longer is needed (e.g. systemd based).

closes https://github.com/rsyslog/rsyslog/issues/2143


rgerhards:
unrelated testbench failures (things I currently work on...)

PR #2318, opened 2017-12-29
tcpsrv: use new errmsg interface

jgerhards:
see also https://github.com/rsyslog/rsyslog/issues/1684


rgerhards:
still issues with the CI system, working on it. PR is OK, merging

PR #2319, opened 2017-12-29
testbench: increase default main queue timeouts

rgerhards:
we increase the default timeouts - CI machines tend to be slow and
timing out causes cancelation, which results in irrelevant small
memory leaks (immediately before termination!). So we do not want
these. They make the testbench look instable.


rgerhards:
This obviously doesn't work -- bad idea. We should increase the timeouts individually.

Issue #2271, opened 2017-12-20
imfile: misses files very quickly created after directory creation


alorbach:
Review test failures when timing changes in imfile-wildcards-dirs2.sh. 
In some cases inotify or fen API misses a file during the test when sleetime  is changed or removed.

rgerhards:
test ```imfile-wildcards-dirs.sh``` seems to have similar issues

rgerhards:
actually, ```imfile-wildcard-dirs.sh```seems to fail more frequently. I did some test runs, and it looks like a real bug, not a testbench issue. I currently think that files already existing in a directory when the directory notification is processed are not properly detected and handled.

rgerhards:
actually, this commit describes the problem, we just need a different soltution that actualyl fixes the problem inside code: https://github.com/rsyslog/rsyslog/commit/faca9c8ced118c2aa47f2c38c3f8c680be01ddca

rgerhards:
I disable the test that fails due to this bug so that the CI system can work normally again. This needs to be re-enabled once this bug here is fixed.

PR #2324, opened 2017-12-30
testbench: TEMPORARILY disable tests to permit "normal" CI runs

rgerhards:
... because we know they sometimes fail due to a confirmed bug
inside imfile. These need to be re-enabled once the bug is fixed.

see also https://github.com/rsyslog/rsyslog/issues/2271


rgerhards:
OK, CI failures are still related to ongoing work -- merge OK

Issue #2046, opened 2017-11-16
action.c: bHadAutoCommit handled in thread-unsafe way


rgerhards:
As it looks, this data item actually belongs into wrkrInstance and was not properly moved over when worker instances were introduced. Brings up errors under thread debugger. Needs to be better investigated and most probably needs a fix.

rgerhards:
This looks like we need to move this to the ```pWti->actWrkrInfo[pThis->iActionNbr]``` structure.

PR #2326, opened 2017-12-30
action bugfix: bHadAutoCommit handled in thread-unsafe way

rgerhards:
This internal state was improperly handled and most probably caused
(hard to see) issues when action instances were run on multiple worker
threads. It looks like the state variable was forgotten to move over
to worker state when action workers were introduced. This patch
fixes that.

closes https://github.com/rsyslog/rsyslog/issues/2046


rgerhards:
CI failure was unrelated -- well known kafka issue

PR #2341, opened 2018-01-02
 doFuncReplace(): add tail of src if tail is partially similar to 'replaceWith'

bayaro:
there is a bug. To catch it just try to use config:
```
template(name="test" type="list") {
    property(name="$!all-json")
}
ruleset(name="test") {
    reset $!xxx=replace($msg, "==1234567890==", "=54321=");
    action(type="omfile" file="tmp/aaa" template="test")
}
*.* call test
```

do two calls 
` logger "A==1234567890=="`
`logger "A==1234567890="`

and see results into `/tmp/aaa`

So I submit a PR to resolve this bug. I have auto-replacement of trailing spaces. Also I have decreased the amount of calculations a bit. And have renamed/added a couple variables to align their names with other.

### DIFF
--- a/grammar/rainerscript.c
+++ b/grammar/rainerscript.c
@@ -1636,35 +1636,34 @@ doFuncReplace(struct svar *__restrict__ const operandVal, struct svar *__restric
     uchar *replaceWith = es_getBufAddr(replaceWithStr);
     uint lfind = es_strlen(findStr);
     uint lReplaceWith = es_strlen(replaceWithStr);
-    uint size = 0;
+    uint lSrc = es_strlen(str);
+    uint lDst = 0;
     uchar* src_buff = es_getBufAddr(str);
     uint i, j;
-    for(i = j = 0; i <= es_strlen(str); i++, size++) {
+    for(i = j = 0; i <= lSrc; i++, lDst++) {
         if (j == lfind) {
-            size = size - lfind + lReplaceWith;
+            lDst = lDst - lfind + lReplaceWith;
             j = 0;
         }
-		if (i == es_strlen(str)) break;
+		if (i == lSrc) break;
 		if (src_buff[i] == find[j]) {
 			j++;
 		} else if (j > 0) {
 			i -= (j - 1);
-			size -= (j - 1);
+			lDst -= (j - 1);
 			j = 0;
 		}
     }
-    es_str_t *res = es_newStr(size);
+    es_str_t *res = es_newStr(lDst);
     unsigned char* dest = es_getBufAddr(res);
     uint k, s;
-    for(i = j = k = s = 0; i <= es_strlen(str); i++, s++) {
+    for(i = j = s = 0; i <= lSrc; i++, s++) {
         if (j == lfind) {
-            for (k = 0; k < lReplaceWith; k++) {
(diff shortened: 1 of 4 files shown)


rgerhards:
Thanks for the PR, I'll review ASAP, but I may take until next week as I am putting finishing touches on 8.32.

rgerhards:
8.32 milestone tentatively assigned - will probably become 8.33.

bayaro:
thank you,
it will be very appreciate for me if it resolves with 8.32.
travis-ci fails two times:
1) full job timeout
2) again strange test that shows a losing 1 record of 10.

bayaro:
the most important commit is https://github.com/rsyslog/rsyslog/pull/2341/commits/7d988d36d971f0e85bc82738dc4c2bae6ed4e51b

one of other - nothing except trailing spaces
third - the new test

rgerhards:
Travis: the failing test is due to a bug in imfile, I just got aware yesterday evening that this test is also affected. We will decide today or tomorrow if we can fix imfile in the 8.32 timeframe of need to postpone this to 8.33.

I'll try to merge your PR really ASAP, but I need to finish off things that I already have begun -- keep with me :-)

bayaro:
thank you,
I'll be in deep hope :)

rgerhards:
@bayaro Is this the correct description of the bug that the PR fixes?

The ```replace()``` script function potentially truncates strings during replacement. The larger the source string is, the higher the probability is that it is incorrectly truncated.

Please let me know.

bayaro:
no.
if the trail of message similar to replacement string. Example:
msg: 1234567890
replacement: 8906
the last N (amount of similar symbols in tail of message and the begin of replacement, in current example = 3) will never be assigned in dst message. And will contain a garbage. As longer replacement string as more garbage will be in dst message.

The length of dst message is right. The tail of dst message filled with garbage. Because of memory for dst message is allocated with right size, but didn't filled if the tail of message *partially* equal to a beginning of a replacement string.

bayaro:
see 3 lines from https://github.com/rsyslog/rsyslog/pull/2341/files#diff-7c86a4d92cde5e553d7e6f5b0b0c3d2bR1678

rgerhards:
@bayaro thx! It  would be great if you could include that information into the commit message. Think about it: if in a couple of month from now someone tries to understand what the patch did, it is not really convenient to be forced to setup a lab to understand the need for the patch.

bayaro:
@rgerhards 
>  It would be great if you could include that information into the commit message. Think about it: if in a couple of month from now someone tries to understand what the patch did, it is not really convenient to be forced to setup a lab to understand the need for the patch.

I have described it in the PR name. And this information is present into the name of most important commit.
Could you please add description as you think to be best?
Also maybe you add the label 'bug' to this PR?

rgerhards:
If that's ok with you, I can ammend the commit myself. I wanted to have the text from https://github.com/rsyslog/rsyslog/pull/2341#issuecomment-355630797 inside the commit message.

bayaro:
I understand. Thank you. Added.

rgerhards:
thx!

bayaro:
thank you.

rgerhards:
thank **you** for the patch! Much appreciated

PR #2342, opened 2018-01-03
Update imfile.c

portante:
A few insignificant changes resulting from a review of the code recently.  Please close if not interested.


rgerhards:
thx, much appreciated!

Side-note: if you look at the issues, there are quite a bit in regard to imfile. I have confirmed quite a bit of them by now. I think it probably is too late to get a full fix into 8.32, but my personal goal is to have fixes and/or a larger refactor for 8.33.

Issue #1489, opened 2017-03-31
imfile with wildcard processes files in wrong order


peick:
When using `imfile` with wildcards the log files are processed in the wrong order.

### How to reproduce

```
echo first >> /var/log/sample.log
logrotate /etc/logrotate.d/sample
echo second >> /var/log/sample.log
logrotate /etc/logrotate.d/sample
echo third >> /var/log/sample.log
```

#### Content of omfile output in /var/log/sample-rsyslog-transported.log

```
2017-03-31T12:45:46.019822+02:00 vagrant sample: third
2017-03-31T12:45:46.020095+02:00 vagrant sample: second
2017-03-31T12:45:46.020152+02:00 vagrant sample: first
```

#### Expected output of  /var/log/sample-rsyslog-transported.log

```
2017-03-31T12:45:46.019822+02:00 vagrant sample: first
2017-03-31T12:45:46.020095+02:00 vagrant sample: second
2017-03-31T12:45:46.020152+02:00 vagrant sample: third
```

#### (partial) rsyslog configuration

```
module(load="imfile" mode="inotify")
module(load="builtin:omfile")

input(
  type="imfile"
  stateFile="sample-state"
  File="/var/log/sample.log*"
  Tag="sample:"
  ruleset="sample-out"
  Facility="local0"
)

ruleset(name="sample-out") {
  action(
    type="omfile"
    File="/var/log/sample-rsyslog-transported.log"
  )
  stop
}
```

#### logrotate configuration

```
/var/log/sample.log {
    rotate 5
    create 640 daemon daemon
    size 1
    missingok
}
```

#### version used
```
$ rsyslogd -v
rsyslogd 8.25.0, compiled with:
        PLATFORM:                               x86_64-unknown-linux-gnu
        PLATFORM (lsb_release -d):
        FEATURE_REGEXP:                         Yes
        GSSAPI Kerberos 5 support:              No
        FEATURE_DEBUG (debug build, slow code): Yes
        32bit Atomic operations supported:      Yes
        64bit Atomic operations supported:      Yes
        memory allocator:                       jemalloc
        Runtime Instrumentation (slow code):    No
        uuid support:                           Yes
        Number of Bits in RainerScript integers: 64
```

alorbach:
What happens if you add more than 3 loglines to the test? Do they all appear it wrong order?

rgerhards:
[...]

PR #2344, opened 2018-01-03
imfile: add sortfiles parameter

jgerhards:
if set, files will be input in a sorted order

closes https://github.com/rsyslog/rsyslog/issues/1489


rgerhards:
@jgerhards can you please add doc for this feature?

Issue #2187, opened 2017-12-09
RainerScript: unknown function names do not trigger error msg?


rgerhards:
briefly seen during other testing. We should verify this.

rgerhards:
closed by https://github.com/rsyslog/rsyslog/pull/2351, which we also treat as bug confirmation ;-)

PR #2351, opened 2018-01-03
Report unknown function calls as parser errors

mattaezell:
I was using 8.24.0 that came with RHEL 7.4 and tried to use the `ltrim()` function.  It simply returned the value `0`.  Turns out that `ltrim()` didn't exist in 8.24.0, and the warning about that is only printed with debug enabled.  This will now print messages like:
```
rsyslogd: error during parsing file /path/to/rsyslog.conf, on or before line 34: Invalid function ztrim [v8.32.0.master try http://www.rsyslog.com/e/2207 ]
```

*maintainer edit: auto-close existing issue on merge*
closes https://github.com/rsyslog/rsyslog/issues/2187


rgerhards:
thx, much appreciated!

Issue #2230, opened 2017-12-15
imfile wildcard directory can't discover new files


Dieken:
I'm running Rsyslogd as Kubernetes DaemonSet to collect console log and file log,  I construct a directory layout with a cron job:

```
## console log
# can't directly symlink to log file due to #2229
/archive/NAMESPACE/DEPLOYMENT/POD/CONTAINER -> /var/lib/docker/containers/CONTAINER/

## file log
/archive/NAMESPACE/DEPLOYMENT/POD/logs -> /var/lib/kubelet/pods/POD_ID/volumes/kubernetes.io~empty-dir/
```

I tried wildcard directory `/archive/*/*/*/*/*/*.log` to collect file log, for example:
```
/archive/test/svcA/svcA-6d648bff95-22hgd/logs/log-vol/error.log
```

But when I create "new.log" in an existed matched directory "/archive/test/svcA/svcA-6d648bff95-22hgd/logs/log-vol/",  imfile in inotify mode can't automatically pick it.  Restarting Rsyslog works.

And newly created directory matching the wildcard directory also can't be automatically picked by Rsyslog.

I bypassed this issue to have a cron job to generate Rsyslog configuration files with wildcard file name "/archive/xxx/xxx/xxx/xxx/xxx/*.log", this works fine.

platform: Ubuntu 14.04 & 16.04 with ppa:adicson/v8-stable.

rgerhards:
@alorbach it looks like we have some issues with files not being picked up in inotify mode. IIRC there have been other issues raised in the past weeks. Could you have a look? Maybe we can craft a dedicated testbench test? I *think* the issue came up not only with containers.

davidelang:
I don't think we support lots of wildcards in the path (or did that get changed)

for a long time we didn't support any wildcards in the path, only in the 
filename, but since you are running a current rsyslog, you are running a new 
enough version to have that support. But I don't remember if it allowed multiple 
layers of wildcards.

alorbach:
I can create a testcase for this scenario. perhaps there is a limit of wildcards that the GLOB api can handle here. The debug logfile will tell.

alorbach:
Could not reproduce this issue yet. 
Maybe related to this bug as here https://github.com/rsyslog/rsyslog/issues/1716 ? 

When /archive/ does not exist on startup of rsyslog, it willl fail to detect /archive/test/ and all other subdirectories.

khogeland:
Experiencing what appears to be the same issue.

[...]

Issue #1716, opened 2017-08-11
imfile with multiple, consecutive, wildcards does not pick up new files in folders that don't already exist


dansteen:
Hi!

First of all, I wanted to thank you guys for the great work you do on rsyslog.  It's an amazing work, and only gets better with each release!

I did run into a quick problem.  I noticed that for imfile, if a path has multiple wildcards consecutively, new files will not be picked up in folders that don't already exist.   A concrete example will probably make this more clear:

given the following imfile stanza:
```
input( type="imfile" File="/var/log/*/*/application.log" Severity="info" Tag="nomad" addMetadata="on" ReadMode="0" PersistStateInterval="10000" ruleset="application.log")
```
if `/var/log/bob` does not exist when rsyslog is started, then `/var/log/bob/dave/application.log` will not be picked up when it is eventually created.

However, in the following case:
```
input( type="imfile" File="/var/log/*/sam/*/application.log" Severity="info" Tag="nomad" addMetadata="on" ReadMode="0" PersistStateInterval="10000" ruleset="application.log")
```
a file named `/var/log/bob/sam/dave/application.log` will be picked up regardless of whether `/var/log/bob` exists when rsyslog starts up.

A quick workaround I have found that works is to just put two consecutive slashes next to each other.   so instead of this `/var/log/*/*/application.log`,  do this: `/var/log/*//*/application.log`, and things will work.

Thanks!

mostolog:
Have you tried /var/log/**/application.log ?

josiahp:
I'm encountering this same issue (currently on 8.25.0 after trying 8.30.0). My early tests seemed to indicate that 8.25.0 did not have this issue, but it indeed seems to. 

@mostolog 's suggestion does not seem to alleviate the issue, and @dansteen 's workaround is not working in my case either. A restart of rsyslog causes all new files to be picked up immediately.

A new file in a directory that was already being watched seems to be picked up whereas a new file in a directory that previously was not being watched seems not to work. By "previously not being watched" I mean that no files were previously being updated, but that it was included in the wildcard specified in the imfile directive.

mostolog:
After asking on lists, seems ** is unsupported.
Perhaps a sigHUP would do the trick.

Anyhow, due to our needs, we aren't longer using glob wilcards for multiple directories.

rgerhards:
@alorbach we seem to have overlooked this one. I guess it might be relevant to your current imfile-related work.

alorbach:
[...]

PR #2353, opened 2018-01-04
imfile: Fixed issues not detecting files in directory when wildcards …

alorbach:
…are used.

When directories and files were created at the same time,
imfile missed the file in some cases (Timing issue).

The handling has been fixed to scan newly created directories ALWAYS for
matching files.

This should fix following issues:
closes https://github.com/rsyslog/rsyslog/issues/2271

However we still have a problem when a directory exist that matches a multilevel wildcard pattern before rsyslog is started. This is discussed here https://github.com/rsyslog/rsyslog/issues/2354

### DIFF
--- a/tests/imfile-wildcards-dirs-multi3.sh
+++ b/tests/imfile-wildcards-dirs-multi3.sh
@@ -16,7 +16,9 @@ do
 done
 
 # Start rsyslog now before adding more files
-. $srcdir/diag.sh startup imfile-wildcards-dirs-multi2.conf
+. $srcdir/diag.sh startup imfile-wildcards-dirs-multi3.conf
+# sleep a little to give rsyslog a chance to begin processing
+sleep 2
 
 for j in `seq 1 $IMFILEINPUTFILESSTEPS`;
 do
@@ -26,7 +28,8 @@ do
 		echo "Make rsyslog.input.dir$i/dir$j/testdir"
 		mkdir rsyslog.input.dir$i/dir$j
 		mkdir rsyslog.input.dir$i/dir$j/testdir
-		./inputfilegen -m 1 > rsyslog.input.dir$i/dir$j/testdir/file.logfile
+		mkdir rsyslog.input.dir$i/dir$j/testdir/subdir$j
+		./inputfilegen -m 1 > rsyslog.input.dir$i/dir$j/testdir/subdir$j/file.logfile
 	done
 	ls -d rsyslog.input.*
 
@@ -37,8 +40,8 @@ do
 	# Delete all but first!
 	for i in `seq 1 $IMFILEINPUTFILES`;
 	do
-		rm -rf rsyslog.input.dir$i/dir$j/testdir/file.logfile
-		rm -rr rsyslog.input.dir$i/dir$j
+		rm -rf rsyslog.input.dir$i/dir$j/testdir/subdir$j/file.logfile
+		rm -rf rsyslog.input.dir$i/dir$j
 	done
 done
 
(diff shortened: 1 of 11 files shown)


rgerhards:
> This should fix following issues:

**should fix** or **fixes**? We need to know if this is actually solved.

> May fix issues (needs to be verified
> #1716
> #2230

Can you please verify and change the commit text accordingly.

tests/imfile-wildcards-dirs-multi2.sh
```diff
@@ -28,19 +28,17 @@ do
...
 		mkdir rsyslog.input.dir$i/dir$j
-		./msleep 25
 		mkdir rsyslog.input.dir$i/dir$j/testdir
-		./msleep 25
 		./inputfilegen -m 1 > rsyslog.input.dir$i/dir$j/testdir/file.logfile
-		./msleep 50
 	done
 	ls -d rsyslog.input.*
 
+	./msleep 250
```
rgerhards: why do we need this?

tests/imfile-wildcards-dirs-multi3.sh
```diff
@@ -0,0 +1,52 @@
...
+	# Delete all but first!
+	for i in `seq 1 $IMFILEINPUTFILES`;
+	do
+		rm -rf rsyslog.input.dir$i/dir$j/testdir/subdir$j/file.logfile
+		rm -rf rsyslog.input.dir$i/dir$j
+	done
+done
+
+# sleep a little to give rsyslog a chance for processing
+sleep 1
```
rgerhards: any hard reason for this sleep?

tests/imfile-wildcards-dirs-multi4.sh
```diff
@@ -0,0 +1,55 @@
...
+# Start rsyslog now before adding more files
+. $srcdir/diag.sh startup imfile-wildcards-dirs-multi4.conf
+# sleep a little to give rsyslog a chance to begin processing
+sleep 1
+
+for i in `seq 1 $IMFILEINPUTFILES`;
+do
+	echo "Make rsyslog.input.dir$i"
+	mkdir rsyslog.input.dir$i
+	./msleep 100
```
rgerhards: this needs to go away -- it will make the test not match reality

tests/testsuites/imfile-wildcards-dirs-multi4.conf
```diff
@@ -0,0 +1,35 @@
+$IncludeConfig diag-common.conf
+$WorkDirectory test-spool
+
+/* Filter out busy debug output, comment out if needed */
+global(
```
rgerhards, line 5: Is it really a good idea to disable debug log parts by default?

tests/imfile-wildcards-dirs-multi2.sh
alorbach: Yes imfile needs a little time to read and process the file at least. Otherwise we would generate same files again without any pause.

tests/imfile-wildcards-dirs-multi3.sh
alorbach: maybe deleted later. Test is not activated yet.

tests/imfile-wildcards-dirs-multi4.sh
alorbach: can be deleted, also test not active yet

tests/testsuites/imfile-wildcards-dirs-multi4.conf
alorbach: Yes for these imfile tests. Debug log gets very huge with 99% unuseful informationen.

rgerhards:
While not a complete fix for all cases, this fixes issues for very common cases. Most importantly, we can get this fix into 8.32, but not a more complete one. So I opt to merge this partial fix and leave the final solution for refactoring (and most probably 8.33).

Issue #2113, opened 2017-11-26
rsyslog eats 100% cpu if no omelasticsearch server is listening


julthomas:
Issue occurs on rsyslog master@289d4ce (friday nov 24), 8.30.0
Issue does not occur on rsyslog 8.29.0

How to reproduce:
* either start rsyslog with no elastic listening, start sending messages before or after rsyslog start
* or stop the elastic server while rsyslog is running and processing messages

Sample configuration with single elastic server and wrong elastic port (nothing listening here):
<pre>
global(net.enableDNS="off")
module(load="imudp")
module(load="omelasticsearch")
input(type="imudp" port="5514" name="net-bsd")

template (name="json" type="list") {
    constant(value="{")
    property(name="timegenerated" dateformat="rfc3339" outname="timegenerated" format="jsonf")
    constant(value=",")
    property(name="hostname" outname="host" format="jsonf")
    constant(value=",")
    property(name="msg" outname="msg" format="jsonf")
    constant(value="}\n")
}

action(
    type="omelasticsearch"
    server="127.0.0.1"
    serverport="9202"
    searchindex="idx"
    template="json"
    bulkmode="on"
    maxbytes="100m"
)

if (1 == 1) then {
    stop
}
</pre>

It is easy to reproduce, I can help testing if needed.

Regards,
Julien

rgerhards:
Thanks for the bug report! I'll see that we craft a testbench test for that (if you are interested in doing that, let me know ;-). Right now, I work on the finishing touches for next Tuesday release, and I am also triaging the large backlog of issues we have. So please bear with me if I am silent the next couple of days.

AndyDorman:
I am not sure this is related and I haven't figured out how to reproduce it reliably, but we may have a similar problem.
We have a cluster of 18 client servers running some heavy duty email security filtering processes that use rsyslog-imrelp to log to a couple of rsyslog hosts.  During periods of peak load (usually caused by disk IO which is our processing choke point) some (not all) of the client servers experience what is described in this ticket...ie, the CPU % rises to or above 100% and the remote logging stops indefinitely. A simple restart of rsyslog on the affected client fixes the problem, sometimes for days, and 4 clients that have much lower loads than the other 14 have never experienced this problem.

[...]

Issue #1836, opened 2017-10-10
core/action: add ability to write an error file for failed messages


rgerhards:
The rsyslog core should provide an ability, on the action level, to write an error file for messages which have finally failed. This is similar to the omelasticsearch error file (which is more application specific and as such should stay).

Such a capability provides users with a tool to
* reliably detect failed messages
* properly act on them

I am untangling this from the very broad issue #974. This is actually a new feature.

PR #2356, opened 2018-01-05
core/action bugfix: 100% CPU utilization on suspension of output module

rgerhards:
Triggering condition:

* output module using the legacy transaction interface
  (e.g. omelasticsearch, omlibdbi)
* output module needs to suspend itself

In these cases, rsyslog enters a busy loop trying to resolve the
suspend condition. The bug is rooted in rsyslog core action code.
This patch fixes it by inserting a 1-second sleep during calls
to the resume handler.

Note: we cannot sleep exactly as long as tryResume needs. This
would require larger refactoring, which probably is not worth for
the legacy interface. The current solution is almost as good, as
the one second sleep has very little overhead on a real system.
Thus we have choosen that approach.

This patch now also ensures that failed messages are properly
handled and do not cause eternal hang.

closes https://github.com/rsyslog/rsyslog/issues/2113

Note: the legacy interface must also be considered when implementing
error file functionality,
see also https://github.com/rsyslog/rsyslog/issues/1836

### DIFF
--- a/action.c
+++ b/action.c
@@ -1189,11 +1189,19 @@ doTransaction(action_t *__restrict__ const pThis, wti_t *__restrict__ const pWti
 			 */
 			iRet = actionProcessMessage(pThis,
 				&actParam(iparams, pThis->iNumTpls, i, 0), pWti);
-			if(iRet != RS_RET_DEFER_COMMIT && iRet != RS_RET_PREVIOUS_COMMITTED &&
-			   iRet != RS_RET_OK)
-				--i; /* we need to re-submit */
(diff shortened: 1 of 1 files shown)


jsiwrk:
One of the tests being proposed in PR #1992 covers this bug (omprog-transactions-failed-commits.sh).

jsiwrk:
Not sure this bug has been completely corrected... See [this comment](https://github.com/rsyslog/rsyslog/pull/1992#issuecomment-377002520).

bosiam:
High Memory usage occurred on suspension of output module ?
image
image

PR #2367, opened 2018-01-08
Fix tcp keepalive settings assignment

loiklo:
Set the right tcp keepalive settings to match the documentation and the linux sysctl original meaning.
TCPKeepAliveIntvl and TPCKeepAliveTime was switched

### DIFF
--- a/plugins/imtcp/imtcp.c
+++ b/plugins/imtcp/imtcp.c
@@ -781,9 +781,9 @@ CODEmodInit_QueryRegCFSLineHdlr
 	CHKiRet(regCfSysLineHdlr2(UCHAR_CONSTANT("inputtcpserverkeepalive_probes"), 0, eCmdHdlrInt,
 			   NULL, &cs.iKeepAliveProbes, STD_LOADABLE_MODULE_ID, &bLegacyCnfModGlobalsPermitted));
 	CHKiRet(regCfSysLineHdlr2(UCHAR_CONSTANT("inputtcpserverkeepalive_intvl"), 0, eCmdHdlrInt,
-			   NULL, &cs.iKeepAliveTime, STD_LOADABLE_MODULE_ID, &bLegacyCnfModGlobalsPermitted));
-	CHKiRet(regCfSysLineHdlr2(UCHAR_CONSTANT("inputtcpserverkeepalive_time"), 0, eCmdHdlrInt,
 			   NULL, &cs.iKeepAliveIntvl, STD_LOADABLE_MODULE_ID, &bLegacyCnfModGlobalsPermitted));
+	CHKiRet(regCfSysLineHdlr2(UCHAR_CONSTANT("inputtcpserverkeepalive_time"), 0, eCmdHdlrInt,
+			   NULL, &cs.iKeepAliveTime, STD_LOADABLE_MODULE_ID, &bLegacyCnfModGlobalsPermitted));
 	CHKiRet(regCfSysLineHdlr2(UCHAR_CONSTANT("inputtcpflowcontrol"), 0, eCmdHdlrBinary,
 			   NULL, &cs.bUseFlowControl, STD_LOADABLE_MODULE_ID, &bLegacyCnfModGlobalsPermitted));
 	CHKiRet(regCfSysLineHdlr2(UCHAR_CONSTANT("inputtcpserverdisablelfdelimiter"), 0, eCmdHdlrBinary,


rgerhards:
Thanks for the PR. I admit I think it is not an easy decision if we should apply it. Of course, you are fixing a bug. But that bugfix break compatibility with existing configuration. And the bug fix touches config code that is obsolete and shall not be used for new configurations. In sum, it looks like we break things but do not bring any good.

@davidelang @deoren @Whissi @portante I would appreciate your feedback.

portante:
I agree this is a tough call.  Since the new parameters are not swapped, could we consider emitting a warning when these swapped parameters are used, giving us a change to direct the users to use the new configuration in order to get the properties correct?

Whissi:
Well, a new version allows you to break existing configurations. And such a bugfix is a valid reason for me. If you are following a traditional version schema like

1.0.0
1.0.1
1.0.2
1.1.0
1.1.1
1.2.0
1.2.1
1.2.1
2.0.0

You would see software doing this already in 1.1.0 or 1.2.0 releases. If you really believe this change is affecting most installations and would affect adoption rate you maybe want to postpone for 2.0.0 release (I cannot find this option set in Gentoo, Debian or RH default configuration so I would think this change is safe for 1.1.0 or 1.2.0 releases).

Back to rsyslog:  I am not aware of any roadmap. I.e. when do you plan to release 9.0? Do you plan to remove legacy configuration in 9.0 at all?

If no 9.0 is in sight or we cannot get rid of legacy configuration at all, maybe introduce an additional option like "workaround_switched_tcp_parameters=true" which is enabled per default and add a logic to switch parameter when the new parameter is disabled.
The new parameter will trigger a deprecation warning.
After 2-3 releases you maybe want to disable the parameter per default, i.e. assume the new fixed logic per default...

deoren:
@rgerhards I agree with @portante that emitting a warning is a good approach, provided that they're read. What is the worst that can happen if the values are switched internally and users do not update their configuration to account for that?

On a related note, I think it would be useful to clarify how much you are willing to invest in issues regarding the obsolete legacy format. I think from a user perspective that if the format is supported, it is natural to expect bug reports related to the format to be accepted and fixed.

I'm biased in that I would like to see support for the old format completely dropped, but there should be a well advertised upcoming milestone indicating that change. If you opt to no longer accept changes regarding the old format, we need to make sure that the docs heavily emphasize that stance.

portante:
+1 to not spending efforts on the legacy format.

+1 to announcing the removal of the legacy format in a future version, and be aggressive about making that happen.

GrayTShirt:
I vote for v9 in November 2020

rgerhards:
on the version number: see http://blog.gerhards.net/2017/10/time-for-better-version-numbering-scheme.html and http://blog.gerhards.net/2014/12/rsyslogs-new-release-cycle-and.html (a bit outdated).

In short, as of current policy v9 will never happen. It'll always be 8.x.0. If you need to start a discussion on this, please do NOT pollute this tracker but create a new one. We might be bold enough one time to switch to a real one-number version number, e.g. v833 instead of v8.33.0, just so that we have a new starting point. BTW: Nov 2020 would roughly be 8.57.0.

Looks like we really need to think about retiring obsolete legacy format. That would break with another long-standing policy: so far, rsyslog users could always be sure that there was no risk associated with an update: old configs continued to work as before. But maybe that really is old-fashioned data center thinking...

rgerhards:
I am all about fixing obsolete legacy format, but this issue here IMHO is very much on the border. The fix involves changes to existing configs, and the prime goal for supporting obsolete legacy format is to avoid **exactly this**.

deoren:
> @rgerhards: I am all about fixing obsolete legacy format bugs, but this issue here IMHO is very much on the border. The fix requires changes to existing configs, and the prime goal for supporting obsolete legacy format is to avoid exactly this.

Understood. What is the worst that can happen if due to these changes the values in the config file are switched when parsed?

loiklo:
@deoren If it can help to take a decision. From my side, I work with thousands of servers centralizing logs to rsyslog servers in tcp. All of them are running RHEL 6 and 7.

The default rsyslog settings in RHEL is TCPKeepAlive off. Since there are thousands of connections on the rsyslog servers, I had to turn the TCPKeepAlive on to setup an agressive policy regarding ghost connections (following firewall tcp-timeout issue).

Basically, I set up: intvl=10 - probe=5 - time=2
If intvl and time are switched, it won't break anything. It will just send 5 more empty tcp packets for 10 seconds (5 probe x 2 seconds), which is painless (any professional grade 100€ router can send more than 1.000.000 PPS). For 3000 servers, it means 3000 pps for 5 seconds (3000 servers x 2 back-and-forth / 2 seconds).

Let's take another worst example: intvl=3600 - probe=5 - time=1. It means one keepalive every hour and a 5 retry every 1s after a network issue. If the time and intvl values are switched, it will generate 1 keepalive every seconds, It's a LOT more, but after 5 probe or packet, it will stop.

To summarize, I think it won't break anything:
- Keepalive is off by default is many linux distribution
- When we enable it, it's to set up an agressive policy. And setting up a 10-5-1, 60-5-2, 2-5-60 or 1-20-1 policy is almost the same.

Bonus: For people setting up their rsyslog from the documentation, it will now work as expected.

rgerhards:
@loiklo That's a good explanation! I think you have a point here...

deoren:
@loiklo,

As @rgerhards said, thank you for the detailed feedback.

davidelang:
breaking existing configs should always be a last resort, only when there is no other sane way forward

bumping the version number as an excuse to break existing configs is something bad projects do.

If documentation and reality disagree about legacy configs (especially advanced things), it's probably better to update the config to match what is really happening rather than changing behavior to match the documentation.

I think this is one of those cases. Update the documentation to say something like "In the legacy config, there is a bug that these parameters don't work as defined, instead they work this way 'X' actually means 'Y' and 'Y' actually means 'X'. This is included only to keep existing configs working and explain the behavior of the old configs. Anyone creating a new config should use the new action() style statements instead"

deoren:
@davidelang Unless someone else beats me to it, I will open a new ticket re dropping the legacy format and loop you in. Even if a major version isn't going to be released to kill it, I think there are still changes that can be made to greatly discourage using the older format.

deoren:
@davidelang @portante @Whissi @GrayTShirt 

I opened #2383 for further discussion. 

@davidelang I meant to include these remarks my last post: If the team's official policy is to put a code freeze on `obsolete legacy` configuration related code, your workaround is a reasonable one: document well the inconsistency, keep the code the way it is and (optionally) have rsyslog emit a warning or info statement whenever one of the parameters are used.

If there is not a code freeze, then making small changes to fix bugs that are not likely to cause problems could also be acceptable, especially if they're documented in the changelog, a remark is made in the docs (though that is problematic if we're retiring all `obsolete legacy` format coverage from current docs and not maintaining the old docs branches) and the change is communicated in the same way the recent CentOS 7 RPM package changes were covered (website, mailing list, changelog).

davidelang:
On Fri, 12 Jan 2018, Deoren Moor wrote:

> I opened #2383 for further discussion.
>
> @davidelang I meant to include these remarks my last post: If the team's 
> official policy is to put a code freeze on `obsolete legacy` configuration 
> related code, your workaround is a reasonable one: document well the 
> inconsistency, keep the code the way it is and (optionally) have rsyslog emit 
> a warning or info statement whenever one of the parameters are used.
>
> If there is not a code freeze, then making small changes to fix bugs that are 
> not likely to cause problems could also be acceptable, especially if they're 
> documented in the changelog, a remark is made in the docs (though that is 
> problematic if we're retiring all `obsolete legacy` format coverage from 
> current docs and not maintaining the old docs branches) and the change is 
> communicated in the same way the recent CentOS 7 RPM package changes were 
> covered (website, mailing list, changelog).

It's not that there is a code freee on the 'legacy' or 'obsolete legacy' config 
code, but that the entire reason for that code existing (especially the 
'obsolete legacy' code) is so that an upgrade of rsyslog doesn't change behavior 
with the existing config code.

So changing how rsyslog behaves given the exact same config in contrary to the 
purpose of that code existing.

If there is a case where 'obsolete legacy' config code is so broken that it 
needs to be changed, the 'right' change would be to have it loudly complain that 
the config exists (ideally writing out what the new-style line would be to 
implement what is in the old config) for a few years, and then error out if the 
broken config is used.

I don't see any scenario where changing what the 'obsolete legacy' config does, 
without disabling it, would be the right thing to do.


The code to implement these config options can change to make the implementation 
simpler (converting them to new style variables internally) or updating the 
implementation of those config options as the core changes to make them keep 
working. So it's not a code freeze at all, it's a functionality freeze.

David Lang

rgerhards:
@davidelang I thought long and hard, multiple times, and I think this really is one of the rare cases where we should merge albeit it may unexpectedly change existing behavior. Reasons:

- probably most users are not aware the system is operating other than the have configured it
- the fix aligns with existing doc
- even if current users actually configured the parameters correctly and counter to the doc, the change will not cause really bad harm

All in all, I tend far more to merge than not. Still keep this open for violent objections ;-)

deoren:
> @davidelang: if someone has an existing config and this change goes in, what's the worst that
happens?

Checkout https://github.com/rsyslog/rsyslog/pull/2367#issuecomment-356015304

rgerhards:
In addition to what @deoren quoted, I have come to the conclusion that most people using these parameters have probably never validated that they actually work as intended. As such, I would expect that the far majority of configurations are actually currently broken, but do not know so. That's my main point in merging this.

rgerhards:
moving to 8.34 as I am would like to be available when this is rolled out, which I am not at the 8.33 release.

rgerhards:
@loiklo thanks again for the patch, the explanations and your persistence. I have finally merged it and it will be part of 8.34.0. README will contain elaborate description.

PR #2370, opened 2018-01-09
action: fix memory leak when errorfile capability is used

rgerhards:
regression from recent commit c1a601e72f779328994c6c344ef4e7576a57e5c3
Not present in any released code.

Detected by Coverity Scan, CID 186455


rgerhards:
unrelated CI failures

Issue #2374, opened 2018-01-11
Disabling cURL via "--disable-libcurl" doesn't work / build error without curl


Whissi:
### Expected behavior
I expected to end with *rsyslogd* **not** linked against *libcurl*.

### Actual behavior
```
/var/tmp/portage/app-admin/rsyslog-8.32.0/image/usr/sbin/rsyslogd (interpreter => /lib64/ld-linux-x86-64.so.2)
    libcurl.so.4 => /usr/lib64/libcurl.so.4
    [...]
```

### Steps to reproduce the behavior
Build rsyslog with `--disable-libcurl` and one of `--enable-elasticsearch` or --enable-omhttpfs` (at least I suspect that this is the root cause).

Summary will list
```
libcurl enabled:                          no
```
but *rscript_http_request.sh* is scheduled to run and *rsyslogd* will be linked against *libcurl.so*.

Yes, when you build with `--disable-libcurl --disable-elasticsearch --disable-omhttpfs`, *rsyslog* will actually fail to build:
```
libtool: link: x86_64-pc-linux-gnu-gcc -O2 -pipe -march=ivybridge -mtune=ivybridge -mno-xsaveopt -Wno-error=missing-prototypes -Wno-error=enum-compare -Wno-error=unused-function -Wno-error=deprecated-declarations -frecord-gcc-switches -ggdb -Wl,-O1 -o rsyslogd rsyslogd-syslogd.o rsyslogd-rsyslogd.o rsyslogd-omshell.o rsyslogd-omusrmsg.o rsyslogd-omfwd.o rsyslogd-omfile.o rsyslogd-ompipe.o rsyslogd-omdiscard.o rsyslogd-pmrfc5424.o rsyslogd-pmrfc3164.o rsyslogd-smtradfile.o rsyslogd-smfile.o rsyslogd-smfwd.o rsyslogd-smtradfwd.o rsyslogd-iminternal.o -Wl,--export-dynamic  -Wl,--as-needed ../grammar/.libs/libgrammar.a ../compat/.libs/compat.a -lz -lpthread ../runtime/.libs/librsyslog.a -ldl -lrt -ljemalloc -lestr -lfastjson -luuid -llogging-stdlog
../grammar/.libs/libgrammar.a(libgrammar_la-rainerscript.o): In function `doFunc_http_request':
[...]

PR #2378, opened 2018-01-11
Fix build without libcurl

Whissi:
Please check "\n" in *grammar/rainerscript.c* before merge.
```C
		LogError(0, RS_RET_INTERNAL_ERROR,
			"rainerscript: internal error: HTTP_Fetch not supported, not built with libcurl support\n");
```

This PR will fix https://github.com/rsyslog/rsyslog/issues/2374

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -454,6 +454,8 @@ if test "$enable_libcurl" = "yes"; then
 	)
 fi
 
+AM_CONDITIONAL(ENABLE_LIBCURL, test "x${enable_libcurl}" = "xyes")
+
 # klog
 AC_ARG_ENABLE(klog,
         [AS_HELP_STRING([--enable-klog],[Integrated klog functionality @<:@default=yes@:>@])],
--- a/grammar/rainerscript.c
+++ b/grammar/rainerscript.c
@@ -1935,6 +1935,7 @@ done:
 	return(estr);
 }
 
+#ifdef HAVE_LIBCURL
 /* curl callback for doFunc_http_request */
 static size_t
 curlResult(void *ptr, size_t size, size_t nmemb, void *userdata)
@@ -2010,6 +2011,7 @@ finalize_it:
 	}
 	RETiRet;
 }
+#endif
 
 static int ATTR_NONNULL(1,3,4)
 doFunc_is_time(const char *__restrict__ const str,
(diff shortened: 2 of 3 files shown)


rgerhards:
Thx - you are right , the ```\n``` needs to go away. I'll fix this after merge as I do not want to break the commit -- or will you update?

rgerhards:
@Whissi BTW: I must have been totally confused ... did I really forgot the real code? Shame on me...

Whissi:
I updated the PR and removed the "\n".

rgerhards:
excellent, thx!

mbiebl:
Hi, I don't think this patch is sufficient.
I just tried building rsyslog with --disable-libcurl. 
The resulting rsyslogd binary had a dependency on libcurl.
The issue is, that libcurl is actually installed (I need it to build the elasticsearch plugin).

```
$ ./configure --disable-libcurl --enable-elasticsearch
...
    libcurl enabled:                          no
    omelasticsearch module will be compiled:  yes

$ make 
...

$ objdump -x tools/rsyslogd | grep NEEDED
  NEEDED               libcurl-gnutls.so.4
  NEEDED               libz.so.1
  NEEDED               libpthread.so.0
  NEEDED               libdl.so.2
  NEEDED               librt.so.1
  NEEDED               libestr.so.0
  NEEDED               libfastjson.so.4
  NEEDED               libsystemd.so.0
  NEEDED               libuuid.so.1
  NEEDED               liblogging-stdlog.so.0
  NEEDED               libc.so.6
```
I don't think we actually use ENABLE_LIBCURL when deciding what to link against rsyslogd?

mbiebl:
The culprit is grammar/Makefile.am afaics. I has
```
libgrammar_la_CPPFLAGS =  $(RSRT_CFLAGS) $(LIBLOGGING_STDLOG_CFLAGS)
#libgrammar_la_LIBADD = $(CURL_LIBS) $(RSRT_LIBS) $(SOL_LIBS)
libgrammar_la_LIBADD = $(CURL_LIBS)
```

No matter if ```ENABLE_LIBCURL``` is set to yes or no, CURL_LIBS is added unconditionally.

mbiebl:
btw, it would make sense to use ```CURL_CFLAGS``` here. As otherwise the include paths will not be properly set if libcurl has been installed in a non-standard location.

mbiebl:
I posted a PR at #2387, please have a look

PR #2385, opened 2018-01-13
omprog: added test for feedback feature (without using transactions)

jsiwrk:
A simple test for the feedback feature implemented in #1753, which didn't have any test yet. This test only checks the 'confirmMessages' flag without the use of transactions.

Note: This PR is based on #1992, but in contrast to that PR, this one is ready to merge. I will rewrite PR #1992 to restrict it to the problematic transaction-related cases.


rgerhards:
sorry for the delay, been working heavily on the CI system and containers. This PR really is good news!

PR #2387, opened 2018-01-14
build system: Don't link core against libcurl if explicitly requested

mbiebl:
When using --disable-libcurl, libgrammar should not be linked against
libcurl unconditionally as CURL_LIBS might be set if one of the modules
using libcurl, like elasticsearch, is enabled.

See #2378


Whissi:
Great catch. We will need the same for *omhttpfs* Makefile.

mbiebl:
Can omhttpfs be built without CURL support?

Whissi:
Scratch what I said. Was about
>it would make sense to use CURL_CFLAGS here

but we are already using this in the Makefile. And no, *omhttpsfs* can't be build without cURL.

rgerhards:
thx @mbiebl and @Whissi for following up on this. Much appreciated!

PR #2389, opened 2018-01-14
travis: do not unnecessarily run checks under cron

rgerhards:


rgerhards:
unrelated CI failures

PR #2398, opened 2018-01-16
fix memory leak in parse_json() 

rgerhards:


rgerhards:
Travis has a major outage, as all else is well, I'll merge.

Issue #2408, opened 2018-01-17
new (as of 8.32) segfault when parsing simple configuration


derekjsonar:
Works in 8.31, segfaults in 8.32. The example conf file includes syntax that does successfully parse in 8.32. If you have a temporary work around that lets me keep the full syntax, please let me know. I prefer to not pin our rsyslog to 8.31.

### Environment
- rsyslog version: 8.32.0
- platform: RHEL 7.4 (via AWS instance)
```
[root@ip-172-30-2-51 ec2-user]# cat /etc/yum.repos.d/rsyslog.repo 
[rsyslog_v8]
name=Adiscon CentOS-$releasever - local packages for $basearch
baseurl=http://rpms.adiscon.com/v8-stable/epel-7/$basearch
enabled=1
gpgcheck=0
gpgkey=http://rpms.adiscon.com/RPM-GPG-KEY-Adiscon
protect=1
```
- `yum install rsyslog`

### Steps to reproduce the behavior

1. install rsyslog 8.32
2. replace `/etc/rsyslog.conf` with:
```
module(load="imptcp" processOnPoller="off")
module(load="omprog")
input(type="imptcp" port="10564" ruleset="gateway")

template(name="delim_rawmsg" type="list") {
    property(name="rawmsg")
    constant(value="D3L1M1T3R")
}

## parses successfully with 8.31 but NOT 8.32
ruleset(name="gateway") {
    action(type="omprog" 
           binary="/usr/lib/sonar/gateway/sonargateway --input_delimiter D3L1M1T3R --prenorm_delimiter PR3N0RM"
           template="delim_rawmsg")
    stop
}

## parses successfully with 8.32
#ruleset(name="gateway") {
#    action(type="omprog" 
#           binary="/usr/lib/sonar/gateway/sonargateway --input_delimiter D3L1M1T3R --prenorm_delimiter PR3N0RM")
#    stop
#}

## parses successfully with 8.32
#ruleset(name="gateway") {
#    action(type="omprog" 
#           binary="/usr/lib/sonar/gateway/sonargateway"
#           template="delim_rawmsg")
#    stop
#}

```
3. `sudo bash`
4. `systemctl disable rsyslog`
5. `rsyslogd -dn`:
```
1072.094018668:main thread    : action 0 queue: queue.c: queue.dequeuetimeend: 25
1072.094021566:main thread    : ../action.c: Action 0x5636eda01200: queue 0x5636eda01550 created
[...]

PR #2410, opened 2018-01-18
fix issues in external command line parser

rgerhards:
see #2408


rgerhards:
Travis Mac builds do not work right now.

Issue #2416, opened 2018-01-19
provide a way to use env variable for config parameter


rgerhards:
We should be able to use an environment variable in place of any string parameter. The variable will be evaluated **at time of config processing** so it is essentially still a static string. This is especially useful to configure rsyslog via external configuration, and so especially useful for containers.

PR #2419, opened 2018-01-19
config: add ability to use environment variables for config params

rgerhards:
We introduce strings enclosed in backticks, which is along the lines
of what the shell does. However, we currently only support
"echo $VARNAME" -- but this may be extended later on.

closes https://github.com/rsyslog/rsyslog/issues/2416

### DIFF
--- a/grammar/rainerscript.c
+++ b/grammar/rainerscript.c
@@ -555,8 +555,8 @@ objlstPrint(struct objlst *lst)
 	}
 }
 
-struct nvlst*
-nvlstNewStr(es_str_t *value)
+struct nvlst* ATTR_NONNULL(1)
+nvlstNewStr(es_str_t *const value)
 {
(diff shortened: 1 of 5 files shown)


grammar/rainerscript.c
```diff
@@ -570,6 +570,44 @@ nvlstNewStr(es_str_t *value)
...
+	const char *realval;
+
+	char *const param = es_str2cstr(value, NULL);
+	if(param == NULL)
+		goto done;
+
+	if(strncmp(param, "echo $", sizeof("echo $")-1) != 0) {
+		parser_errmsg("invalid backtick parameter `%s` currently "
+			"only `echo $<var>` is supported - replaced by "
+			"empty strong (\"\")", param);
```
urysegal, line 586: Spelling mistake I suspect -- should be "string" not "strong"

Issue #2151, opened 2017-12-01
RFE: Add ability to optionally include specific files via direct reference (e.g., not via wildcard)


deoren:
## Current functionality

1. `$IncludeConfig /etc/rsyslog.d/*.conf`
1. `$IncludeConfig /path/to/file/that/must/exist/or/rsyslog/will/complain/rsyslog-forward-to-test-receiver.conf`

The first allows for conditional loading if files are found by wildcard match, or not if they're not matched (e.g., file.conf.disabled).

## New functionality (concept)

- `$IncludeOptionalConfig /etc/rsyslog-forward-to-test-receiver.conf`

It could be combined with the existing functionality like this:

```
$IncludeOptionalConfig /etc/rsyslog-forward-to-test-receiver.conf
$IncludeConfig /etc/rsyslog-forward-to-prod-receiver.conf
```

If the test conf is found, then forwarding to the test node would be enabled, otherwise only forwarding to the prod node. 

I know similar functionality could be enabled via environment variables or some other conditional check, but the ability to use a directive like `$IncludeOptionalConfig` would open up the potential for conditional loading of modules, inputs and other items not (AFAIK) currently possible with the existing configuration support.

rgerhards:
When we implement this, we should probably provide a new-style counterpart to $includeConfig and control the optionality via a parameter.

PR #2426, opened 2018-01-22
config: add include() script object

rgerhards:
This permits to include files (like legacy $IncludeConfig) via a
script object. Needless to say, the script object offers more
features:

- include files can now be
  - required, with rsyslog aborting when not present
  - required, with rsyslog emitting an error message but otherwise
    continuing when not present
  - optional, which means non-present include files will be
    skipped without notice
  This is controlled by the "mode" parameter.
- text can be included form e.g. an environment variable
  --> ex: ```include(text=`echo $ENVVAR`)```

This finally really obsoletes $IncludeConfig.

closes https://github.com/rsyslog/rsyslog/issues/2151

### DIFF
--- a/grammar/grammar.y
+++ b/grammar/grammar.y
@@ -63,6 +63,7 @@ extern int yyerror(const char*);
 %token <estr> FUNC
 %token <objType> BEGINOBJ
 %token ENDOBJ
+%token BEGIN_INCLUDE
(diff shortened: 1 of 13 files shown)


rgerhards:
mixing closely related new feature into this pr ;-)

Issue #2436, opened 2018-01-24
imjournal does not appear to follow the journal correctly when the journal rotates


portante:
### Expected behavior

When the journal rotates only new logs from the previous location spot in the journal should be collected when the journal is re-opened.

### Actual behavior

It appears that a large number of logs are re-read.

### Steps to reproduce the behavior

Create a logging rate that will cause the journald configuration to rotate journal files often enough to see that logs from the journal are duplicated periodically. 

### Environment
- rsyslog version:
```
rsyslogd 8.24.0, compiled with:
	PLATFORM:				x86_64-redhat-linux-gnu
	PLATFORM (lsb_release -d):
	FEATURE_REGEXP:				Yes
	GSSAPI Kerberos 5 support:		Yes
	FEATURE_DEBUG (debug build, slow code):	No
	32bit Atomic operations supported:	Yes
	64bit Atomic operations supported:	Yes
	memory allocator:			system default
	Runtime Instrumentation (slow code):	No
	uuid support:				Yes
	Number of Bits in RainerScript integers: 64

See http://www.rsyslog.com for more information.
```
- platform: RHEL 7.4
- for configuration questions/issues, include rsyslog.conf:
```
module(load="imjournal" StateFile="imjournal.state" UsePidFromSystem="on" RateLimit.Burst="500000" RateLimit.Interval="1" IgnorePreviousMessages="on" PersistStateInterval="1000")
```

*Note: rsyslog documentation is available here*:
  [pre-release](http://www.rsyslog.com/doc/master/),
  [current stable release](http://www.rsyslog.com/doc/v8-stable/)

portante:
This logic looks wrong: https://github.com/rsyslog/rsyslog/blob/master/plugins/imjournal/imjournal.c#L506-L516

Basically, before closing the journal shouldn't we capture the current cursor, then close, then re-open the journal, and then seek to the saved cursor and continue?  Persisting that cursor to disk seems like a reasonable approach that would be natural here, no?

portante:
Also tracking in RHEL 7.4 as: https://bugzilla.redhat.com/show_bug.cgi?id=1538372

portante:
See also: systemd/systemd#7998, which highlights that a journalctl -f appears to use an internal mechanism to close open file descriptors on rotation instead of the proper sd_journal_close()/sd_journal_open() dance.

portante:
[...]

PR #2437, opened 2018-01-25
Fix to reloading journal

portante:
WIP: psuedo code for fix to reloading journal after journald rotates its internal journal files.

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -114,6 +114,9 @@ static const char *pidFieldName;	/* read-only after startup */
 static int bPidFallBack;
 static ratelimit_t *ratelimiter = NULL;
 static sd_journal *j;
+static int j_inotify_fd;
+
+#define J_PROCESS_PERIOD 1024  /* Call sd_journal_process() every 1,024 records */
 
 static rsRetVal persistJournalState(void);
 static rsRetVal loadJournalState(void);
@@ -126,6 +129,12 @@ static rsRetVal openJournal(void) {
 		LogError(-r, RS_RET_IO_ERROR, "imjournal: sd_journal_open() failed");
 		iRet = RS_RET_IO_ERROR;
 	}
+	if ((r = sd_journal_get_fd(j)) < 0) {
+		LogError(-r, RS_RET_IO_ERROR, "imjournal: sd_journal_get_fd() failed");
+		iRet = RS_RET_IO_ERROR;
+	} else {
+		j_inotify_fd = r;
+	}	
 	RETiRet;
 }
 
@@ -134,6 +143,7 @@ static void closeJournal(void) {
 		persistJournalState();
 	}
 	sd_journal_close(j);
+	j_inotify_fd = 0;
 }
 
 
@@ -424,7 +434,7 @@ persistJournalState(void)
 	char *cursor;
 	int ret = 0;
 
-	/* On success, sd_journal_get_cursor()  returns 1 in systemd
+	/* On success, sd_journal_get_cursor() returns 1 in systemd
 	   197 or older and 0 in systemd 198 or newer */
 	if ((ret = sd_journal_get_cursor(j, &cursor)) >= 0) {
                /* we create a temporary name by adding a ".tmp"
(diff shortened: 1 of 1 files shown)


portante:
This is a proposed fix for https://github.com/rsyslog/rsyslog/issues/2436.

rgerhards:
@portante Thx for your work. Do you consider it complete or still WiP?

portante:
Yes, still trying to figure out why the cursor sometimes fails to be fetched.  It looks like we have to fetch the cursor before we call get_fd, but not sure yet.

rgerhards:
> Yes, still trying to figure out why the cursor sometimes fails to be fetched. It looks like we have to fetch the cursor before we call get_fd, but not sure yet.

thx! please let me know when you consider it ready for merge. Will flag it as "no not merge" in the meantime.

portante:
@rgerhards, this is ready for review, but realize I have not built and tested this myself yet.

See the conversation at https://github.com/systemd/systemd/issues/7998 for more background.

portante:
FWIW, here is initial work on fixes to the docker journald log driver client: https://github.com/projectatomic/docker/pull/296

portante:
Rather than polling every 1,024 messages calling `sd_journal_process()`, we could create a thread to "listen" to the inotify fd and do that, but it does not seem worth the effort given the intricate details of journal we'd need to know.

rgerhards:
@portante There is a brace ```(``` missing ;-)

rgerhards:
> Rather than polling every 1,024 messages calling sd_journal_process(), we could create a thread to "listen" to the inotify fd and do that, but it does not seem worth the effort given the intricate details of journal we'd need to know.

IMHO we should not go too deep into (non-documented?) journal details if we can avoid it. So I would stick with the current approach and see if any new problems come up with this. @portante do you agree? Do I overlook something?

portante:
I agree @rgerhards.

rgerhards:
@portante ok, so then we need to dig into undocumented journal land... Are you up to this?

portante:
@rgerhards, I'm up for it.  I am hoping the discussion at https://github.com/systemd/systemd/pull/8032 will give us the necessary insight we need to do it correctly.

portante:
Do you consider it worthwhile to take this patch as is, and then we work a follow-on patch for the out-of-band handling of journal file rotation?

rgerhards:
@portante much appreciated, thank you!

rgerhards:
> Do you consider it worthwhile to take this patch as is, and then we work a follow-on patch for the out-of-band handling of journal file rotation?

If I don't get it wrong, I think it **improves** the situation without any new negative effect. As such, I would consider it useful to merge. If you agree, I'll do later today (working on s/t else at the moment which I would like to finish first).

portante:
Great, I'll see if I can't build a package for this with help from co-workers at Red Hat to confirm the behavior before we merge.  Then I'll rebase on top of latest and we can merge.

rgerhards:
excellent - just let me know explicitly (not just by git push) when you think it is ready to merge. Will do than ASAP.

portante:
@rgerhards, almost ready, I just found a small error handling bug and updated the commit.  We'll be testing soon.

rgerhards:
@portante side-note: I identified some instability in CI testing, that's causing the failures you see. If you rebase to current master that should go away. Else I'll just re-submit. I am working on the root cause, it's probably a race in the test setup itself, but may be an actual failure in an unrelated component. So it takes some time to investigate. Current master has the prime culprit temporarily disabled.

portante:
@rgerhards, this is ready to merge, thanks!

Later this evening I’ll rebase it just to be safe if you want.

portante:
@rgerhards, rebased the patch and pushed it again.  It is ready to be merged.

We did some local testing with this patch applied to RHEL 7.4 v8.24.0 and did not see any file handle leaks, nor did we see any duplicate logs captured due to starting from the previously saved cursor.

I'll be posting the general methodology used for verifying all is good in another dedicated git repo.

portante:
@rgerhards, any further word on merging this patch?

rgerhards:
@portante sorry for the delay, will happen today. I got side-stepped by intense discussion on rsyslog-doc...

rgerhards:
@portante sorry, took me a day more. The past ~ two weeks things tend to be a bit unexpected ;-) But now finally merged, and thanks once again for the contribution!

portante:
Thanks @rgerhards!  Glad to help.  Now to consider imfile and symlinks ...

rgerhards:
@portante careful with imfile! This is up for a big rewrite -- @alorbach and I originally planned this for 8.33, but due to all the doc work etc we can't make that goal. There are some design deficits that really need to be fixed. I guess the symlink functionality could be brought to work in current code, but that's probably a very unfriendly experience...

rgerhards:
@portante see https://github.com/rsyslog/rsyslog/projects/4

PR #2440, opened 2018-01-25
testbench: do not override default Alpine CFLAGS

rgerhards:
... as this brought us away from the abuild environment. Most importantly,
-Os was overriden by -O2, which did make a lot of warning go away.


rgerhards:
Note: this should fail CI on first incarnation. The necessary fixes to make it work are already present in another branch, but I would like to see it actually fails.

PR #2450, opened 2018-01-30
BUGFIX, Issue 2446. Fix directory creation code in lmsig_ksils12 plugin

### LINKED ISSUE #2446: lmsig_ksi_ls12: KSI LS12 plugin doesn't set the log signature folder owner properly
### Expected behavior
The dirOwner and dirGourp conf. values should be applied to log signature folder.

### Actual behavior
The log signature folder owner stays root/root.

### Environment
- rsyslog version: 8.32
- platform: *

allanpark:
There was a bug in lmsig_ksils12 code which rendered the dirOwner/dirGroup configuration vlaues useless. This patch solves the problem and the respective values will be used as intended. 

fixes #2446

### DIFF
--- a/runtime/lib_ksils12.c
+++ b/runtime/lib_ksils12.c
@@ -489,14 +489,17 @@ static int mkpath(char* path, mode_t mode, uid_t uid, gid_t gid) {
 
 	for (char *p = strchr(path + 1, '/'); p; p = strchr(p + 1, '/')) {
 		*p = '\0';
-		if (mkdir(path, mode) == -1) {
-			if (errno != EEXIST) {
-				*p = '/';
-				return -1;
-			}
-			if (chown(p, uid, gid)) {
+		if (mkdir(path, mode) == 0) {
+			if (uid != (uid_t) -1 || gid != (uid_t) -1) {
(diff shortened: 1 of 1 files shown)


runtime/lib_ksils12.c
```diff
@@ -489,14 +489,17 @@ static int mkpath(char* path, mode_t mode, uid_t uid, gid_t gid) {
...
 		*p = '\0';
-		if (mkdir(path, mode) == -1) {
-			if (errno != EEXIST) {
-				*p = '/';
-				return -1;
-			}
-			if (chown(p, uid, gid)) {
+		if (mkdir(path, mode) == 0) {
+			if (uid != (uid_t) -1 || gid != (uid_t) -1) {
+				if (chown(path, uid, gid)) {
```
rgerhards, line 494: I would emit a warning message here. Otherwise this might be hard to diagnose. Note that current similar code may not already do it ... that's one of the long-running projects we have. If you prefer, I can also happily apply the change.

allanpark: Thanks a lot, that would be completely fine with me!

Issue #2376, opened 2018-01-11
omstdout: refactor to use new config interface


rgerhards:
This was originally only a testbench tool, but it has been turned out that it is useful for container deployments. The code needs some brush-up to abide the latest rsyslog features, e.g. advanced configuration format.

PR #2455, opened 2018-02-02
omstdout: add new configuration format

PascalWithopf:
Omstdout can now be configured in the new style
and a parameter for the template was added.

closes https://github.com/rsyslog/rsyslog/issues/2376


rgerhards:
@PascalWithopf it would be great if you could also update the doc to match the new format

deoren:
@PascalWithopf @rgerhards Created rsyslog/rsyslog-doc#547 to track the doc work.

PR #2460, opened 2018-02-06
Fix typo rsyslogd.8

bbigras:
unless it's a mnemonic to remember "-N"


rgerhards:
actually, it was meant as an explanation ... but probably that idea did not work out ;-) Do you think it would make sense to explain the relationship?

bbigras:
If it's an explanation it's fine by me.

rgerhards:
I think if we needed to discuss it, it's not obvious ;-) "Correcting" case is not bad, so let's do it...

PR #2465, opened 2018-02-08
adding hash64 and hash64mod functions support in rainerscript

CodeHarsh:
Rsyslog supports collecting data/logs from local boxes and sending it across to aggregating service for processing and storage. A typical setup of a client to any aggregating service is an array of systems arranged as a pipeline (A->B->C->D->E). 
A message/information enters to pipeline from machine/service A, is processed and sent to machine/service B, which does similar action and at last machine/service E does this final steps and completes the flow.

When collecting data/logs from these machines/services Rsyslog support various filtering and sampling features. One of the ways to achieve is to use random (if we have no prior idea of data/log);
set $.rand = random(100);
if ($.rand < 20) then {
  //send out
}

or parsing using mmnormalize (if we know our data/log)
action(type="mmnormalize" ruleBase="/etc/rsyslog.d/parse.rulebase" path="$!msg")
if ($!msg!method == "POST") then {
//send out
}

When we have no prior idea of data and we use random for sampling/filtering we end up in situations where we have data/logs from all the machines/services (A->B->C->D->E) but we can't tie an event end to end using logs. With this pull request, I introduce two new functions (hash64 and hash64mod) in rainerscript to support this use-case. Using hash64 or hash64mod, at runtime we can sample/filter data/logs while guaranteeing that data/logs are available to tie an event end to end (For the minimum sample percent of pipeline). Eg.
action(type="mmnormalize" ruleBase="/etc/rsyslog.d/parse.rulebase" path="$!msg")
if (hash64mod($!msg!someUID,100) < 30) then {
//send out
}

While implementing the support I added wrapper hash-impl (along with default implementation) so Rsyslog can easily integrate with any hash implementation supported outside and use it (for large strings etc).


rgerhards:
thx for the PR, this sounds really useful. It would be great if you could add some doc, especially as this probably requires some more elaborate info to use in the actual case.

Side-not: recent PRs show that we really need to provide a loadable interface for functions -- else we drag in too many dependencies, aka too many of the great extensions will be disabled other than in custom builds.

CodeHarsh:
Thanks, I will add/update docs with some detailed examples.

CodeHarsh:
Created a pull request on [rsyslog-doc](https://github.com/rsyslog/rsyslog-doc/pull/558)

CodeHarsh:
*bump*

rgerhards:
Please fix CI failures.

Sent from phone, thus brief.

Am 20.02.2018 08:12 schrieb "Harshvardhan Shrivastava" <
notifications@github.com>:

CodeHarsh:
Fixed all the CI issues. The ubuntu failure is due to virtual memory on build box.

CodeHarsh:
bump

rgerhards:
We are now working on a way to enable rsyslog to actually load function module dynamically, see also http://jan.gerhards.net/2018/03/making-rainerscript-functions-loadable.html

IMHO the best plan is to get this done, and then finally merge in the new code. So we avoid that it is disabled by all distros by default, aka "nobody can use it". I would guess that we could still make the next release, but that needs to be seen.

CodeHarsh:
Sure @rgerhards  
Loadable function module sounds interesting. On a second note, this feature is available on all distros. Only the hash implementation will change from default to specialised based on config flags.

rgerhards:
> On a second note, this feature is available on all distros. Only the hash implementation will change from default to specialised based on config flags.

We are misunderstanding. I am not talking that the base plumbing is missing. I am talking about the fact that distros tend to disable non-loadable features that they do not consider essential and that may require more dependencies than they like. Just like http_request() didn't make it into distro packages. I guess it can not be our goal to have rsyslog features only available to those that build from source or use project packages.

davidelang:
having different distros pick different hashes is one reason I am saying that 
the hash implementation should not be visible in the function name.

There is value in having multiple systems get the same hash result from the same 
data.

David Lang

jgerhards:
So the refactor to use modules is waiting for tests to run and be merged at the moment (see PR: https://github.com/rsyslog/rsyslog/pull/2562). Could you modify your PR and implement the functions as a separate module? If you do not want to do that, I could do as well. Just let me know (also if you want to do it yourself and have any questions).

CodeHarsh:
@jgerhards I will modify this PR. Thanks for being considerate.

jgerhards:
FYI: The PR is now merged

plugins/fmhash/fmhash.c
```diff
@@ -0,0 +1,414 @@
...
+	*functArray = functions;
+ENDgetFunctArray
+
+
+BEGINmodExit
+CODESTARTmodExit
+ENDmodExit
+
+#if !defined(__clang__)
+#pragma GCC diagnostic ignored "-Wstrict-prototypes"
```
rgerhards: This looks strange to me, what is the error that comes up without this pragma? Maybe there is a bug in module-interface macros...

CodeHarsh: http://build.rsyslog.com:8010/builders/rsyslog%20centos7%20rsyslog/builds/2936/steps/shell_2/logs/stdio

make[2]: Leaving directory `/var/lib/buildbot/slave/rsyslog_centos7_rsyslog/build/plugins/fmhttp'
Making all in plugins/fmhash
make[2]: Entering directory `/var/lib/buildbot/slave/rsyslog_centos7_rsyslog/build/plugins/fmhash'
  CC       fmhash_la-fmhash.lo
fmhash.c:396:1: error: function declaration isn't a prototype [-Werror=strict-prototypes]
 BEGINqueryEtryPt
 ^
fmhash.c:402:1: error: function declaration isn't a prototype [-Werror=strict-prototypes]
 BEGINmodInit()
 ^
fmhash.c:402:1: error: function declaration isn't a prototype [-Werror=strict-prototypes]
fmhash.c:402:1: error: function declaration isn't a prototype [-Werror=strict-prototypes]
fmhash.c:402:1: error: function declaration isn't a prototype [-Werror=strict-prototypes]
make[2]: *** [fmhash_la-fmhash.lo] Error 1

rgerhards: thx - that is strange. I will look into it. Please do not wonder if I "clone" your PR and submit it again - it then is just to have an example where this problem occurs. I'll delete mine later on.

rgerhards:
The failure on Solaris also looks strange. Will check that one as well.

plugins/fmhash/fmhash.c
CodeHarsh: Sure, Thanks for your help.

rgerhards:
sry, looks like I was a bit too agressive cancelling out "unnecessary" builds on buildbot. I think I have restarted all that I cancelled. However, this is a real problem: http://build.rsyslog.com:8010/builders/rsyslog%20ubuntu16%20rsyslog/builds/1151/steps/shell_10/logs/test-suite.log

rgerhards:
@CodeHarsh I tried to find the need for the pragmas in my test PR, but there are too many other failures for this to make sense. I would suggest that you first go ahead and see that you can clean CI as much as possible, then I'll take over with my tests again. Does that sound OK?

plugins/fmhash/fmhash.c
```diff
@@ -0,0 +1,416 @@
...
+    if(seedVal) {
+        seed = var2Number(seedVal, &success);
+        if (!success) {
+            DBGPRINTF("fmhash: hashXX(string, seed) didn't get a valid 'seed' limit, defaulting hash value to 0");
+            iRet = RS_RET_ERR;
+            FINALIZE;
+        }
+    }
+
+    char *hashStr = NULL;
```
rgerhards: This is the source of the "maybe-unitialized". Please move it to after line 164 and remove the disabling of the compiler diagnostic. We always try to compile clean with max warnings and disable them only if really, really there is no other way to solve the issue.

CodeHarsh: My bad. Never realised this line was the issue. I have fixed it.

rgerhards: I have become pretty used to this kind of issues - do them often enough myself ;-)

rgerhards: Just for others coming across this thread: the root cause is that the compiler actually "allocates" the variable on the stack as soon as the function is entered. But is is uninitialized until that line is reached. In the concrete code here that was no issue due to the free-guard variable, but the compiler complains that this is dangerous practice as code may change in the future. It's often overlooked that C99 definition and assignment still means mem alloc on function entry.

rgerhards:
I have also restarted my test PR #2592 as things look pretty good now - will post results as soon as I have them.

plugins/fmhash/fmhash.c
rgerhards: Please remove these two sets of pragmas. My test PR shows that the problem does no longer occur (checked centos 7 as in the sample). This is as expected. I guess the previous error was due to some other code that has been corrected in the mean time.

CodeHarsh:
@rgerhards  I moved fmhash to contrib. 
Travis ci failed due to connection timed out while getting release key.

rgerhards:
Thx. I have restarted Travis.

CodeHarsh:
@rgerhards Sorry for changing the code after your review. I thought it will be good to have Valgrind tests added. Also, removed malloc call that can be easily be bypassed by using typedefs.

Travis build failed because it was unable to verify package hash.

rgerhards:
> @rgerhards Sorry for changing the code after your review. I thought it will be good to have Valgrind tests added. Also, removed malloc call that can be easily be bypassed by using typedefs.

no problem at all - just keep improving! I'll happily review again, it usually gets quicker each time.

> Travis build failed because it was unable to verify package hash.

I have restarted Travis. Unfortunately we see these type of problems when third-party packages are being rebuild. I've solved part of that via containers, but a couple still remain as we right now cannot run the full testbench inside container.

contrib/fmhash/Makefile.am
```diff
@@ -0,0 +1,8 @@
+#
+# fmhash support
+# 
+pkglib_LTLIBRARIES = fmhash.la
+fmhash_la_SOURCES = fmhash.c
+fmhash_la_CPPFLAGS = $(PTHREADS_CFLAGS) $(RSRT_CFLAGS) $(LIBLOGGING_STDLOG_CFLAGS)
```
rgerhards: I think we can remove the LIBLOGGING macros here. Can you pls check.

contrib/fmhash/fmhash.c
```diff
@@ -0,0 +1,385 @@
...
+ * http://www.isthe.com/chongo/src/fnv/hash_32.c
+ */
+#if defined(__clang__)
+#pragma GCC diagnostic ignored "-Wunknown-attributes"
+#endif
+static hash_t
+#if defined(__clang__)
+__attribute__((no_sanitize("unsigned-integer-overflow")))
+#endif
+fnv_32(const void* input, size_t len, seed_t seed) {
```
rgerhards, line 84: rsyslog code style is to use tabs for indention (with standard width 8). I would highly appreciate if this could be obeyed to.

contrib/fmhash/fmhash.c
```diff
@@ -0,0 +1,385 @@
...
+		free(hashStr);
+	}
+	RETiRet;
+}
+
+static rsRetVal
+hash_wrapper3(struct svar *__restrict__ const sourceVal, struct svar *__restrict__ const modVal
+		, struct svar *__restrict__ const seedVal, hash_context_t* hcontext) {
+
+	DEFiRet;
```
rgerhards, line 183: this looks like mix of tab and sp indention, which renders things badly readble

contrib/fmhash/fmhash.c
```diff
@@ -0,0 +1,385 @@
...
+    int success = 0;
+    hash_t xhash = 0;
+	hash_t mod = var2Number(modVal, &success);
+	if (! success) {
+		DBGPRINTF("fmhash: hashXXmod(string, mod)/hash64mod(string, mod, seed) didn't"
```
rgerhards: just a remark: is this something that should really go to the debug log, or is it something the user should be warned about?

contrib/fmhash/fmhash.c
```diff
@@ -0,0 +1,385 @@
...
+}
+
+static rsRetVal ATTR_NONNULL(1)
+init_fmHash32mod(struct cnffunc *const func)
+{
+	DEFiRet;
+	hash_context_t *hash_context = NULL;
+	if(func->nParams < 2) {
+		parser_errmsg("fmhash: hash32mod(string, mod)/hash32mod(string, mod, seed)"
+				" insufficient params. %d of file %s\n",
```
rgerhards: source code locations should only be given for internal errors -- else it tends to confuse users, which then often think it **is** an internal error (I made this too often myself...).

CodeHarsh:
@rgerhards I have fixed indentation and change parse debug msgs to parser_warnmsg

rgerhards:
Thanks a lot for all your work. Looks good to me. We should squash the commits. Do you prefer to do it yourself or should I during merge? Just let me know.

We probably need to put a bit of attention to the doc pr at https://github.com/rsyslog/rsyslog-doc/pull/558 -- I guess at least some wording needs to be changed.

So it looks like we are extremely close to merge :-)

rgerhards:
labelled "doc missing" to keep me aware https://github.com/rsyslog/rsyslog-doc/pull/558 needs some work

CodeHarsh:
@rgerhards I have done doc changes. If possible can you squash commits at the time of merge?
Also, If you can check doc PR.

rgerhards:
@CodeHarsh thanks again for all your work!

contrib/fmhash/fmhash.c
```diff
@@ -0,0 +1,382 @@
+/*
+ * Copyright (c) 2018-2020, Harshvardhan Shrivastava
```
mbiebl, line 2: Copyright 2020?

CodeHarsh: Wrong copy paste

CodeHarsh: I fixed it in https://github.com/rsyslog/rsyslog/pull/2722

PR #2466, opened 2018-02-08
Add the mmkubernetes plugin

portante:


portante:
It would be really powerful to have this plugin enabled for Kubernetes environments where we can fetch additional metadata about the log source to enrich when sending to Elasticsearch for indexing.

See https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter

rgerhards:
Thanks, this looks very useful! I appreciate your help in improving the support for Kubernetes. In case you have not yet seen, we have also setup a project for better container support:

https://github.com/rsyslog/rsyslog/projects/5

Any ideas are welcome, as are pointers to things we should look at (of course, time is unfortunately limited, so we may need a bit -- just setting expectations straight ;-)).

davidelang:
sorry for the slight hijack, I think there's a issue asking for this, but it's 
many years old

It would be really useful to have the ability to create and destroy new unix 
socket inputs at runtime. That way as a container is created, you can create a 
/dev/log in the container that goes directly to the main rsyslog on the system.

davidelang:
ping, what are the next steps with this? is there anything people can do to help?

portante:
This plugin fetches additional metadata from Kubernetes about the container's pod and namespace.

In order to do that, the plugin needs to know a pod name and a namespace name.

Typically, when kubernetes is using docker with journald log driver, the pod name and namespace name are embedded in the container name metadata property docker constructs for journal metadata.  So something in the pipeline needs to put those values into variables this plugin can read.

In the case of k8s using docker's json-file log driver, the `/var/log/containers/*.log` file name is actually a symlink to the real log file, where the symlink name contains the pod and namespace name.  Similarly, something needs to parse the symlink name to get those values into variables for this plugin.

With the pod name and the namespace name, this plugin can fetch the pod ID and namespace ID from k8s, and also the labels and annotations associated with the pod and namespace.

Next steps are to (off the top of my head):

 * ensure it builds correctly with latest rsyslog
 * verify it works to connect to the k8s API with proper certs as parameters
 * scales properly (memory footprint is known from cache size parameters)
 * and that we have a good test subsystem for it

davidelang:
On Tue, 20 Feb 2018, Peter Portante wrote:

> This plugin fetches additional metadata from Kubernetes about the container's pod and namespace.
>
> In order to do that, the plugin needs to know a pod name and a namespace name.
>
> Typically, when kubernetes is using docker with journald log driver, the pod 
> name and namespace name are embedded in the container name metadata property 
> docker constructs for journal metadata.  So something in the pipeline needs to 
> put those values into variables this plugin can read.
>
> In the case of k8s using docker's json-file log driver, the 
> `/var/log/containers/*.log` file name is actually a symlink to the real log 
> file, where the symlink name contains the pod and namespace name.  Similarly, 
> something needs to parse the symlink name to get those values into variables 
> for this plugin.

if this info is in the filename (symlink), why not just read everything in the 
directory via imfile with wildcards and then parse the filename to extract the 
pod and namespace name?

why is there a need for a separate plugin?

or is it a matter of opening the files via imfile, parsing the pod and namespace 
names from the filename, and passing them to the mmkubernetes code to reach out 
to k8s to fetch additional data?

Is this data cached, or does it need to be fetched each time? 
Is there a race condition where the pod may go away before the data can be 
fetched?

Is there a hook in k8s that would let this data be fed to rsyslog directly (and 
ideally feed the logstream as well), rather than going through the indirect 
method of having things written to disk and then reading them off disk?

at the risk of trading one indirection for another, I'll point out that with 
table_lookup we have a way of reading data in on the fly and looking it up in 
ram

> With the pod name and the namespace name, this plugin can fetch the pod ID and 
> namespace ID from k8s, and also the labels and annotations associated with the 
> pod and namespace.

what other data can be fetched? I'm looking at an environment where the labels 
and annotations are mostly created from outside k8s, so if I need these, I can 
probably get them more efficiently from the upstream sources rather than 
querying k8s.


> Next steps are to (off the top of my head):
>
> * ensure it builds correctly with latest rsyslog
> * verify it works to connect to the k8s API with proper certs as parameters
> * scales properly (memory footprint is known from cache size parameters)
> * and that we have a good test subsystem for it
>
>

portante:
> if this info is in the filename (symlink), why not just read everything in the directory via imfile with wildcards and then parse the filename to extract the pod and namespace name?
> why is there a need for a separate plugin?

The symlink only has pod name and namespace name, and we *have* to read it from the symlink.  We can't make an API call without those names (one API call for pod data, one for namespace data), and we *only* get the UUID for the pod and namespace from the API call today (see https://github.com/kubernetes/kubernetes/issues/58638).  For multi-tenancy applications, fetching the UUID is required to ensure logs are attributed to the proper namespace.

> or is it a matter of opening the files via imfile, parsing the pod and namespace names from the filename, and passing them to the mmkubernetes code to reach out to k8s to fetch additional data?

Yes, exactly.

> Is this data cached, or does it need to be fetched each time?

It should be cached.

> Is there a race condition where the pod may go away before the data can be fetched?

Yes, which should result in "orphaned" log data, so that it is at least marked that way.

> Is there a hook in k8s that would let this data be fed to rsyslog directly (and ideally feed the logstream as well), rather than going through the indirect method of having things written to disk and then reading them off disk?

Yes, docker supports a syslog logging driver (see https://docs.docker.com/config/containers/logging/syslog/).

However, the docker daemon support in kubernetes is only for the journald and json-file logging drivers of docker.  So under kube, using the `kubectl logs` command one can see the logs from those two logging drivers, but when using any other logging driver, k8s cannot provide a view of the logs.

> at the risk of trading one indirection for another, I'll point out that with table_lookup we have a way of reading data in on the fly and looking it up in ram

Basically, mmkubernetes is the dynamic feed to such a table.  One possible avenue would be to have table misses filled by mmkubernetes.

> what other data can be fetched? I'm looking at an environment where the labels and annotations are mostly created from outside k8s, so if I need these, I can probably get them more efficiently from the upstream sources rather than querying k8s.

If a use case does not need labels and annotations on logs, or the pod UUID and namespace UUID, then there is no need for the mmkubernetes plugin to be placed in that configuration.  But if the requirement is to have all logs annotated with those UUIDs, and associated labels and annotations we need this plugin.

If you annotate logs in a database with a join, for using the UUID as the join key for a table with logs and separate tables with pod labels, pod annotations, namespace labels, and namespace annotations, you would still need the API call to get the UUID.  If you annotate using pod names and namespace names, you run the risk of collisions due to name reuse, but then you could do a post join.

Basically the mmkubernetes plugin is a pre-join of k8s metadata with the logs.

davidelang:
@rgerhards is there a way to trigger a fetch of data when a file is opened (or a new connection is made to a unix socket) so that every time a file is opened, the data can be fetched and cached?

davidelang:
If the logs are delivered to rsyslog via the syslog logging driver, how do we get the pod and environment names?

portante:
@davidelang, I believe the [labels](https://docs.docker.com/config/containers/logging/syslog/#options) feature of the driver, but not entirely sure as I have not experimented with that logging driver.

davidelang:
The travis build fails due to a line being too long, that line is:

#define DFLT_FILENAME_REGEX "var\\.log\\.containers\\.([a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*)_([^_]+)_(.+)-([a-z0-9]{64})\\.log$"

I don't know if there is any way to reasonably break this up. Is there a way to  annotate this line so that the line length checker will leave it alone? @rgerhards

richm:
@portante please close this PR in favor of https://github.com/rsyslog/rsyslog/pull/2507

portante:
Rebased and added Rich's work.

davidelang:
the build failures are the result of a lot of 'line to long' errors.

The style guide is enforced by Travis. There is the one line with the regex that is hard to split, but the rest of them are just long comments, please fix those.

portante:
Sure, I'll fix them up.

richm:
Anyone know how to tell Eclipse to enforce line lengths of 120 or less?

rgerhards:
On a conference, so just brief (@pascalwithopf) pinged me:

Am 01.03.2018 08:25 schrieb "David Lang" <notifications@github.com>:

The travis build fails due to a line being too long, that line is:

#define DFLT_FILENAME_REGEX "var\.log\.containers\.(a-z0-9?(\.a-z0-9?)*)
*([^*]+)_(.+)-([a-z0-9]{64})\.log$"

I don't know if there is any way to reasonably break this up.



Very easy. For example, break the line after the macro name and put the
string on the next line, indented with a tab.

Rainer

portante:
The long log lines have been fixed, the failure case now is:

```
RUN="run.sh",BUILD_FROM_TARBALL="YES", CHECK="YES", CFLAGS="-g -O2", RS_TESTBENCH_VALGRIND_EXTRA_OPTS="--suppressions=travis/trusty.supp --gen-suppressions=all", EXTRA_CONFIGURE="--disable-testbench2"

sudo apt-get install -qq --force-yes libestr-dev librelp-dev libfastjson-dev liblogging-stdlog-dev \
	liblognorm-dev \
	libcurl4-gnutls-dev
E: Unable to locate package libfastjson-dev
E: Unable to locate package liblogging-stdlog-dev
```

davidelang:
Also, is there a corresponding PR for rsyslog-docs to document this?

davidelang:
Is there any way to do test coverage with this other than having a full k8s install available? (say a dummy https server that responds in place of k8s?)

davidelang:
looking at the TODOs in the code
￼  * - cache cleanup
￼  * - SIGHUP handling
I think having SIGHUP clear the cache is a quick-and-dirty, but useuable solution to both of these. Even if you are someone like me who may HUP really frequently to roll logs, it shouldn't be a problem (unless k8s is slow in responding to the queries)

￼  * - authentication
Is there more to authentication than the SSL handshake?
￼  * - statistics generation
other than stats on the cache, I think anything else should be done by user stats
￼  * - other missing configuration options
what other config options do you think are needed
￼  * - documentation
yes! (and this answers my question a couple posts ago)
￼  * - more robust error-handling + debugging information
what happens now if k8s doesn't respond or returns garbage?
￼  * - batching, parallel queries, failover, ...
Parallel queries only make sense if you are doing batching (otherwise you only have one log message to process at a time), just make sure that the code is thread-safe so that if there are multiple threads processing logs, this code can be running in them simultaniously

Even with batches, I'm not sure trying to do parallel queries makes a lot of sense. It adds a LOT of complexity and I question how much value it provides. How well does k8s handle multiple simultaneous queries?

For failover, if the connection to the k8s server doesn't work, shouldn't normal retry logic going to a name provide a restored connection to k8s on a different server?

Is this something that should be run on the same host system as the pod generating the logs? or is this something that should be run on the next hop system aggregating info from many hosts?
￼

richm:
> Is there any way to do test coverage with this other than having a full k8s install available? (say a dummy https server that responds in place of k8s?)

Probably - are there other such test fixtures or http/rest mocking used by other rsyslog plugins?

>  Is there more to authentication than the SSL handshake?

This plus the token auth - so we can remove this todo.

>  other than stats on the cache, I think anything else should be done by user stats

What needs to be done by this plugin?

> what other config options do you think are needed

There are a few more listed in the corresponding fluentd plugin that should eventually be implemented.

>  what happens now if k8s doesn't respond or returns garbage?

There is some error handling, but there are probably other failure modes I didn't think of.

> re: batching, parallel queries

I'm not really sure what this means - I'm not familiar with the rsyslog internals - I did try to make the code thread safe, but it will need some more eyes

> How well does k8s handle multiple simultaneous queries?

Very well, from my experience.

> For failover, if the connection to the k8s server doesn't work, shouldn't normal retry logic going to a name provide a restored connection to k8s on a different server?

Yes, although I'm not sure if the fluentd plugin handles this.  It should be pretty straightforward to add support for multiple URLs here - there are a few examples in some of the other plugins e.g. omelasticsearch.

> Is this something that should be run on the same host system as the pod generating the logs? or is this something that should be run on the next hop system aggregating info from many hosts?

Both.  In some cases, you may want the local nodes to perform all of the k8s metadata processing.  In other cases e.g. with hundreds of nodes, you may want to reduce resource consumption on each node as much as possible, and instead have a farm of rsyslog aggregators doing the processing.

davidelang:
I know there are some cases of text fixtures, but I don't know how extensive they are

As for stats, my thinking is that if the stats are something that be figured out by things external to this plugin (by looking at the resulting json), don't bother trying to do stats for them (especially for version 1)

Things that can only be known inside the module (cache size/memory cache hit count and cache miss count usage come to mind) need to be added at some point.

If you have accounted for the error conditions that you can think of, let's remove that as a todo item (it's something that can always be improved, but listing it as a todo item makes people think the module isn't safe to use until error handling is improved)

for the additional config options, can you list what the fluentd plugin has under this (we can then discuss specifics)

davidelang:
batches and parallel processing.

Rsyslog is based around queues and threads, even though it looks to the user like it's calling modules, that's not really what is happening.

inbound traffic hits an input thread that adds the log to a queue.

there are one or more worker threads that read from the queue (by default, one message at a time)

These workers execute the actions for that queue (which can include making a copy of the log into a different queue), these actions are code scattered across any number of different modules. This code is executed directly by the worker thread (i.e. it's not a matter of passing the message to the mmkubernetes thread, the worker thread calls the functions that are part of the module in the context of the worker thread)

this is why the code needs to be thread safe, and why parallel requests to the outside don't always make sense.

Because the queue(s) are so central to rsyslog, we found years ago that contention on the queue was a very significant factor, and allowing a thread to add/remove multiple messages from the queue and then process them before going back to the queue significantly speeds things up (it adds one more way that logs can be re-ordered, but there were already lots of ways, so we deemed that unimportant)

There are a few output modules that combine multiple log messages together before sending them to the destination (omelasticsearch and database outputs) so that the multiple messages are processed as one transaction on the destination

but logically, just about all log processing in rsyslog is done one message at a time.

So I don't think that parallel processing or parallel fetching of things from k8s is likely to fit well, so I think it makes sense to drop that from the todo list

davidelang:
re: failover, in general, rsyslog tends to try and leave failover/load balancing mechanisms to external tools. There are a few cases where we support it directly (omelasticsearch being one), but only in fairly primitive ways (simple round-robin across possible destinations). One thing that it's common for us to support is a 'reconnect every X messages' functionality so that we give an external load balancer a chance to direct us to a different destination.

also note, if there are multiple worker threads, each one will open a connection to k8s if it's called for.

I think it's valuable to split the todo list into a "needs to be done before the module is ready for production use" and "ideas for future enhancements" sections.

davidelang:
this is noted to only work with logs from /var/log/containers or with logs that go through the journal. Is there no way to have the logs delivered directly to rsyslog with the appropriate data in the message?

richm:
You can construct a `CONTAINER_NAME` message property containing the appropriate Kubernetes metadata in the format matching the `containerregex`, and specify a `CONTAINER_ID_FULL` message property containing the full docker id.

The defaults and sample configuration are for the typical small to medium sized deployment which will want to annotate logs at the originating node.

davidelang:
possible premature optimization :-)

there seems to be a lot of regex matching/parsing in here, would it make sense to leverage mmnormalize/liblognorm for this instead?

richm:
> possible premature optimization :-)

yes - my goal is to be able to replace the existing functionality of the fluentd plugin as quickly as possible, so I kept it the same.

> use mmnormalize

yes, good idea - will likely be much faster - not sure if I can do that with the annotations_match though, where users will expect a good old-fashioned regex

rgerhards:
sorry for the silence, been very busy (and away) the past ~2 weeks. I haven't grasped everything of the long tracker, but I have a general recommendation: I would suggest the we get to a state that is useful, but not perfect ASAP -- actually I have the impression we are mostly there already. If so, let's get this into the code and than work on improvements.

But I may have missed to big picture -- so just take this as a rough suggestion.

davidelang:
I agree that we seem to be at a usable state (I haven't been able to test, and won't have a chance to this week), i'm just nervous about defining a new module where the default and documented thing to do is to use multiple (fairly nasty) regexes. It's really hard to change the default, and almost impossible to disable that mode entirely after it hits the real world.

richm:
I am working on changing the use of regex for the filename match and containername match to use mmnormalize, since that is in the critical path (every msg).  I don't think it will be changed much - any change in ordering or fields will require a corresponding code change.  For minor changes (whatever that may be) the user will need to change the rulebase.

Ideally, we would have a default mmnormalize pattern, hard coded in the mmkubernetes.c, for the filename match and the container name match.  Then, we would allow the user to override them by explicitly providing their own mmnormalize action + rulebase file.

richm:
I want to do something very similar to mmnormalize - have the user specify an array of rules or pass in a rulebase file.  Here is the mmnormalize rule processing: https://github.com/rsyslog/rsyslog/blob/master/plugins/mmnormalize/mmnormalize.c#L354
```c
			for(int j=0; j < pvals[i].val.d.ar->nmemb; ++j) {
				tStr = (char*)es_str2cstr(pvals[i].val.d.ar->arr[j], NULL);
				size += strlen(tStr);
				free(tStr);
                         }
```
Why does the code use `es_str2cstr` to allocate a string in order to use `strlen`, rather than just use ` es_strlen(pvals[i].val.d.ar->arr[j])`?

davidelang:
On Mon, 5 Mar 2018, Richard Megginson wrote:

> I am working on changing the use of regex for the filename match and 
> containername match to use mmnormalize, since that is in the critical path 
> (every msg).  I don't think it will be changed much - any change in ordering 
> or fields will require a corresponding code change.  For minor changes 
> (whatever that may be) the user will need to change the rulebase.

mmnormalize isn't going to be thrown off by the order (unlike the regex) as 
mmnormalize will name each field.

> Ideally, we would have a default mmnormalize pattern, hard coded in the 
> mmkubernetes.c, for the filename match and the container name match.  Then, we 
> would allow the user to override them by explicitly providing their own 
> mmnormalize action + rulebase file.

exactly, and in the last year or so, mmnormalize gained the ability to pass the 
rulebase as a string, not as a filename, so we should set this up to use that 
mode and just let the admin override the default rulebase strings.

David Lang

davidelang:
I would guess programming habit :-)

double check that the two methods produce identical (i.e. not off-by-one) 
results.

richm:
great - `ln_loadSamplesFromString` not supported in 2.0.2, so won't work on rhel7 :-(
should I create some sort of hack to write out the in memory rules to a tmpfile and pass the tmpfile to `ln_loadSamples`?

rgerhards:
> great - ln_loadSamplesFromString not supported in 2.0.2, so won't work on rhel7 :-(
> should I create some sort of hack to write out the in memory rules to a tmpfile and pass the tmpfile to ln_loadSamples?

2.0.3 is the minimum requirement for current rsyslog (exactly for this reason btw). So whatever you do, it would not build with 2.0.2 in any case ;-)

richm:
Well, I need something that works on rhel7 asap, and at this point in the cycle they will be reluctant to upgrade/rebase liblognorm simply in order to add features, so I'll have to figure some way to make this work with both the current code and rhel7

richm:
I have a PR for a test for this in openshift logging - https://github.com/openshift/origin-aggregated-logging/pull/1032

rgerhards:
@portante from you PoV, is this still WiP or ready to merge? If the latter, I'd do a final round of review and see that I merge ASAP

portante:
@rgerhards, yes, I believe this is ready to merge.  Thanks!

rgerhards:
@portante 
> @rgerhards, yes, I believe this is ready to merge. Thanks!

great. I just saw we have some CI failures from those unreliable tests I work on. Would it be possible to rebase to current master - that should improve the situation. Due to the failing tests, some others are skipped, so I would perfer to try to get CI clean. But it's not totally vital.

rgerhards:
Note: I need to enable mmkubernetes in CI testing (as it is not default). will do so soon.

portante:
Yes, I'll rebase.

portante:
@rgerhards, rebased and squashed the commits down to two.

rgerhards:
ok, I see at least one real compile failure on ubuntu16 after enableing mmkubernates. I guess ```curlCB``` should be static? I also notice that I seem to have missed one racy omprog test, which brings up false CI failures (this time on Travis, which I restarted).

rgerhards:
I have (hopefully correctly) reconfigured the rest of the buildbot farm to now also build mmkubernetes.

contrib/mmkubernetes/mmkubernetes.c
```diff
@@ -0,0 +1,1409 @@
...
+ *       see COPYING.ASL20 in the source distribution
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* todo: needed for MVP
```
rgerhards: Is this still a TODO? If so,  I would suggest to also open a github tracker. Else remove ;-)

contrib/mmkubernetes/mmkubernetes.c
```diff
@@ -0,0 +1,1409 @@
...
+ENDfreeInstance
+
+size_t curlCB(char *data, size_t size, size_t nmemb, void *usrptr)
+{
+	wrkrInstanceData_t *pWrkrData = (wrkrInstanceData_t *) usrptr;
+	char * buf;
+	size_t newlen;
+
+	newlen = pWrkrData->curlRplyLen + size * nmemb;
+	buf = realloc(pWrkrData->curlRply, newlen);
```
rgerhards: this could lead to a NULL pointer exception when realloc() fails

rgerhards:
It would be great if you could also add doc for the new module at https://github.com/rsyslog/rsyslog-doc/tree/master/source/configuration/modules (@deoren FYI)

rgerhards:
my bad - doc is already there, but the PR has some issues: https://github.com/rsyslog/rsyslog-doc/pull/605

rgerhards:
@portante sorry for being such a pest, it would be great if you could rebase again to master -- I overlooked some racy tests, which are now disabled in master branch (they are inside the Makefile, so I cannot just do that via CI definitions...).

The CI failures you currently see seem all valid to me - it's actually so far a single issue and wasn't caught earlier because CI did not yet build that module.

richm:
@portante I did the rebase on my richm/mmk8s branch but now that conflicts with yours.  My branch has the compiler error fixes and other issues mentioned by @rgerhards. If you rebase your mmk8s branch on master then I can rebase my mmk8s branch on yours and submit a PR to you.  https://github.com/richm/rsyslog/tree/mmk8s

richm:
I also filed all of the old major todo items as github issues - look for mmkubernetes - I suppose these should be added to https://github.com/rsyslog/rsyslog/projects/5

portante:
@richm, done!

davidelang:
could you make references in the todo items that link them to this one (referring to this thread via #tags in comments should be enough)

davidelang:
a reminder that we are a week away from the next release, has @rgerhards requested change been addressed?

and the testbench builds are failing due to curlCB() calls, is there a missing dependency on curl?

richm:
@davidelang I have one more fix for the long line lengths.  Once ci is clear, I'll ask @portante to merge my PR to his repo, then he can push to your repo.  Then, if everything checks out ok, I'll do one more commit to squash all of these separate commits into one, then @portante can merge and re-push.

richm:
@davidelang hmm - I thought the curlCB problem was because it should have been declared static, since it is defined in mmkubernetes.c?  Can you show me the error?

davidelang:
https://travis-ci.org/rsyslog/rsyslog/jobs/356659684

mmkubernetes.c:716:8: error: no previous prototype for 'curlCB' 
[-Werror=missing-prototypes]
  size_t curlCB(char *data, size_t size, size_t nmemb, void *usrptr)
         ^



On Tue, 27 Mar 2018, Richard 
Megginson wrote:

> Date: Tue, 27 Mar 2018 23:12:42 +0000 (UTC)
> From: Richard Megginson <notifications@github.com>
> Reply-To: rsyslog/rsyslog
>     <reply+0021b24f99104f1fcb980f7ef2ce879f598822ce83e592ee92cf0000000116d291e
>     a92a169ce119e64e9@reply.github.com>
> To: rsyslog/rsyslog <rsyslog@noreply.github.com>
> Cc: David Lang <david@lang.hm>, Mention <mention@noreply.github.com>
> Subject: Re: [rsyslog/rsyslog] Add the mmkubernetes plugin (#2466)
> 
> @davidelang hmm - I thought the curlCB problem was because it should have been declared static, since it is defined in mmkubernetes.c?  Can you show me the error?
>
>

richm:
>  size_t curlCB(char *data, size_t size, size_t nmemb, void *usrptr)

Yes, I fixed that by changing this to `static size_t curlCB..`

davidelang:
Ok, that's part of the fixes that are waiting to be pushed up, correct?

richm:
> Ok, that's part of the fixes that are waiting to be pushed up, correct?

yes

rgerhards:
Given the long easter weekend here in Germany (Friday and Monday are public holidays) the timing is actually a bit more critical. Offical close of the merge window is, as usual, this Friday. I usually do late merges if possible and low risk, but I can't guarantee it this time. Anyhow, I am very interested in getting this PR in ;-)

@richm I would suggest doing a second PR from your branch so that you get full CI coverage. The buildbot slaves actually add some checks, especially in this case. I have no problem with two "identical" PRs, just mention it in the second one. I'll then flag it as "do not merge" and we can delete the second PR once this here is merged. I think this gives you the most productive environment.

richm:
@rgerhards I don't understand - you want me to drop this PR #2466 and submit a completely different PR?

rgerhards:
> @rgerhards I don't understand - you want me to drop this PR #2466 and submit a completely different PR?

If I understood you correctly, you have some private branch that you work on. As there is no PR for it, you do not get the full CI exposure. So once it finally gets merged, full CI will run and maybe (maybe!) that will find another issue.

What I propose is
* create a PR P for your private branch, get it CI clean
* do NOT drop or otherwise modify this PR
* once your private branch is merged to this here, we drop PR P and continue here

Does that clarify? If not, let me know.

richm:
I'm going through @portante as an intermediary - because he opened this PR initially, and there was quite a bit of discussion here that we didn't want to lose, we kept this PR, and I opened a PR against his fork of rsyslog - so I submit a PR against his fork https://github.com/portante/rsyslog/pull/1 and he merges/rebases and updates this PR.

richm:
btw, my latest PR finally passed all of the CI - https://github.com/portante/rsyslog/pull/3 - see https://travis-ci.org/portante/rsyslog/builds/359160444?utm_source=github_status&utm_medium=notification

rgerhards:
To be precise, it passed Travis, so there are 17 more checks you only get one this is updated. My intent was not to replace your workflow but get you early access to full CI as this is time critical. But I am happy with all that good work and don't want to talk you into something you don't like. All well, let's wait on @portante and then see how the rest of CI works.

richm:
@rgerhards I think the latest commits addressed your review comments - let me know if I need to fix anything else to lift the "requested changes" flag

davidelang:
a quick question on this plugin. When you use the syslog log-driver, it includes a tag value, that according to the same system logging to journald is something passed on the command line (i.e. tag=\"docker-lcbb-d09\" ), is this enough to be able to fetch all the other metadata to add it to the log?

sample message via journald

journald:
{ "__CURSOR" : "s=c53d70f5331145fc80552964d2945009;i=23b016;b=438f7e642c0d4c7b853ea6a5ed84f407;m=17c564f995b;t=5686db030fe51;x=98a7e6bbc8a12774", "__REALTIME_TIMESTAMP" : "1522195200278097", "__MONOTONIC_TIMESTAMP" : "1633535629659", "_BOOT_ID" : "438f7e642c0d4c7b853ea6a5ed84f407", "_TRANSPORT" : "journal", "_SYSTEMD_SLICE" : "system.slice", "_MACHINE_ID" : "6da0573b97b741d2b5f9059f40533c2c", "_HOSTNAME" : "lcbb-d09.lc.dc.openx.org", "PRIORITY" : "6", "CONTAINER_TAG" : "\"docker-lcbb-d09\"", "SYSLOG_IDENTIFIER" : "\"docker-lcbb-d09\"", "CONTAINER_ID" : "cb21c5b05bf3", "CONTAINER_ID_FULL" : "cb21c5b05bf3ba3e3ddd2312d2cc0f4da165bc1a48a847d85f9450a8c1a608ff", "CONTAINER_NAME" : "k8s_calico-node_calico-node-vpx5n_kube-system_f4716447-2e5c-11e8-8fdb-90e2babf6c3c_1", "_PID" : "13023", "_UID" : "0", "_GID" : "0", "_COMM" : "dockerd", "_EXE" : "/usr/bin/dockerd", "_CMDLINE" : "docker --selinux-enabled --log-driver=journald --log-opt tag=\"docker-lcbb-d09\" -s overlay2 --insecure-registry docker.openx.org --bip=192.168.7.1/24 --ip-masq=true --mtu=1450 --bip=192.168.7.1/24 --ip-masq=true --mtu=1450", "_CAP_EFFECTIVE" : "3fffffffff", "_SYSTEMD_CGROUP" : "/system.slice/docker.service", "_SYSTEMD_UNIT" : "docker.service", "MESSAGE" : "2018-03-28 00:00:00.277 [INFO][129] int_dataplane.go 690: Applying dataplane updates", "_SOURCE_REALTIME_TIMESTAMP" : "1522195200277830" }

similar message from the syslog log-driver

<30>Mar 28 23:35:51 "docker-lcbb-d09"[32219]: 2018-03-28 23:35:51.655 [INFO][129] int_dataplane.go 690: Applying dataplane updates

richm:
I don't know - not sure if you can lookup the other docker metadata from the docker api given the CONTAINER_TAG value.
As you suggested on the mailing list, I think the best approach would be to enhance the docker syslog log driver to add CONTAINER_ID or CONTAINER_ID_FULL or other metadata.

rgerhards:
Thanks everyone for the great collaboration! As a side-note, I am all ears if you have suggestions what else we can change to make rsyslog work even better in containers (in general).

PR #2487, opened 2018-02-17
[ommongodb] Adding possibility to ignore some insertion error code

HugoSoszynski:
Added new configuration (not mandatory) option `allowed_error_codes`.
The directive argument MUST be in the following format (handling a maximum of 256 codes): "11000,...,17,64"

The behavior of ommongodb was for a long time to `ABORT_FINALIZE(RS_RET_SUSPENDED)` on any MongoDB insertion error; keeping this behavior, I wanted to add the possibility to ignore some error code that may be allowed in some (really rare) configuration (*ex. `11000: DuplicateKey` in case of collection containing one or more unique field(s)*).

Hugo.

### DIFF
--- a/plugins/ommongodb/README
+++ b/plugins/ommongodb/README
@@ -22,7 +22,7 @@ A very basic example is:
 *.*     action(type="ommongodb" db="logs" collection="syslog")
 
 A more complex example:
-*.*     action(type="ommongodb" uristr="mongodb://vulture:9091,vulture2:9091/?replicaset=Vulture&ssl=true" ssl_cert="/var/db/mongodb/mongod.pem" ssl_ca="/var/db/mongodb/ca.pem" db="logs" collection="syslog")
+*.*     action(type="ommongodb" uristr="mongodb://vulture:9091,vulture2:9091/?replicaset=Vulture&ssl=true" ssl_cert="/var/db/mongodb/mongod.pem" ssl_ca="/var/db/mongodb/ca.pem" db="logs" collection="syslog" allowed_error_codes="11000")
 
 Please see the script clean-mongo-syslog for an example of how to
 purge old records from MongoDB using PyMongo.  It can be run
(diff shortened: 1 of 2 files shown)


rgerhards:
Bonjour @HugoSoszynski , sorry for the long silence. I had a week off and immediately thereafter went to a conference. Would you be willing to update the doc at https://github.com/rsyslog/rsyslog-doc/blob/master/source/configuration/modules/ommongodb.rst ? That would be fantastic.

Side-Note: CI fails due to a recent fix I needed to apply. If you want to have the error go away, you can rebase to current master. But this is NOT strictly necessary, I know what the problem is and am happy to merge as is as soon as we have the doc covered.

HugoSoszynski:
Hi,

Sorry for the long silence, I had a incredibly busy month.
I created a PR for the documentation (https://github.com/rsyslog/rsyslog-doc/pull/636).

All checks seems to be ok now.

Regards.

rgerhards:
@HugoSoszynski we all have busy times, thanks again for the contribution. In the doc PR I noticed that we could further simplify this PR, if you like. The config system has "array" parameters, which permit a standard way to provide multiple values. As a plus, it also gives you the number so that the array can dynamically be sized. For a sample, look at the "broker" parameter of omkafka: https://github.com/rsyslog/rsyslog/blob/master/plugins/omkafka/omkafka.c#L1458

What do you think? If you like, I can also ammend your commit, which might be useful to get used to this mode.

rgerhards:
@HugoSoszynski I have amended the PR so that the array interface is used. This will unstall the PR.

rgerhards:
@HugoSoszynski I needed to rebase to current master as otherwise CI does not work properly.

rgerhards:
@HugoSoszynski thanks again for the PR! much appreciated :-)

HugoSoszynski:
Thank you very much @rgerhards.

I am incredibly sorry for not ending what I have begun.
I had a lot going on the last few month.

I will make sure this won't happen again.

Thanks again.

Yours faithfully.

PR #2490, opened 2018-02-19
omfile: action now suspends, also with writes. suspend conditions: no…

kortemik:
…t enough inodes, not enough blocks, file-system read-only or output path is not found

this allows omfile to act correctly in case message write retries are intended.

old behaviour should no change as actions default to no retry http://www.rsyslog.com/doc/v8-stable/configuration/actions.html:
```
action.resumeRetryCount integer

[default 0, -1 means eternal]

Sets how often an action is retried before it is considered to have failed. Failed actions discard messages.
```

### DIFF
--- a/runtime/rsyslog.h
+++ b/runtime/rsyslog.h
@@ -510,6 +510,9 @@ operation not carried out */
 	/* RainerScript error messages (range 1000.. 1999) */
 	RS_RET_SYSVAR_NOT_FOUND = 1001, /**< system variable could not be found (maybe misspelled) */
 	RS_RET_FIELD_NOT_FOUND = 1002, /**< field() function did not find requested field */
+	// <kortemik>
+	RS_RET_FS_ERR = -2443, /**< file-system error */
+	// </kortemik>
 
 	/* some generic error/status codes */
 	RS_RET_OK = 0,			/**< operation successful */
--- a/runtime/rsyslog.h
+++ b/runtime/rsyslog.h
@@ -506,13 +506,11 @@ operation not carried out */
 	RS_RET_UDP_MSGSIZE_TOO_LARGE = -2440, /**< a message is too large to be sent via UDP */
 	RS_RET_NON_JSON_PROP = -2441, /**< a non-json property id is provided where a json one is requried */
 	RS_RET_NO_TZ_SET = -2442, /**< system env var TZ is not set (status msg) */
+	RS_RET_FS_ERR = -2443, /**< file-system error */
 
 	/* RainerScript error messages (range 1000.. 1999) */
 	RS_RET_SYSVAR_NOT_FOUND = 1001, /**< system variable could not be found (maybe misspelled) */
 	RS_RET_FIELD_NOT_FOUND = 1002, /**< field() function did not find requested field */
-	// <kortemik>
-	RS_RET_FS_ERR = -2443, /**< file-system error */
-	// </kortemik>
 
 	/* some generic error/status codes */
 	RS_RET_OK = 0,			/**< operation successful */
--- a/tools/omfile.c
+++ b/tools/omfile.c
@@ -693,8 +693,9 @@ fsCheck(instanceData *__restrict__ const pData, const uchar *__restrict__ const 
 			LogError(0, iRet, "too few available blocks in %s", path);
 			FINALIZE;
 		}
-	/* there must be enough inodes left  */
-	if (stat.f_favail < 1)
+	/* there must be enough inodes left, one is left for administrative purposes
+	 * check is not done if file system reports 0 total inodes, such as btrfs */
+	if (stat.f_favail < 2 && stat.f_files > 0)
 		{
 			iRet = RS_RET_FS_ERR;
 			LogError(0, iRet, "too few available inodes in %s", path);
--- a/tools/omfile.c
+++ b/tools/omfile.c
@@ -1138,9 +1138,8 @@ CODESTARTcommitTransaction
 	}
 
 finalize_it:
-
 	if (iRet != RS_RET_OK) {
-		if (runModConf->bDynafileDoNotSuspend == 0) {
+		if (runModConf->bDynafileDoNotSuspend == 0 || !(pData->bDynamicName)) {
 			LogError(0, iRet, "suspending action");
 			iRet = RS_RET_SUSPENDED;
 		}
--- a/tools/omfile.c
+++ b/tools/omfile.c
@@ -1140,12 +1140,14 @@ CODESTARTcommitTransaction
 finalize_it:
 
 	if (iRet != RS_RET_OK) {
-		LogError(0, iRet, "suspending action");
-		iRet = RS_RET_SUSPENDED;
+		if (runModConf->bDynafileDoNotSuspend == 0) {
+			LogError(0, iRet, "suspending action");
+			iRet = RS_RET_SUSPENDED;
+		}
+		else {
+			LogError(0, iRet, "discarding message");
+		}
 	}
-/* FIXME: what to do with runModConf->bDynafileDoNotSuspend because
- * now it suspends even with non-dynafiles
- */
 	pthread_mutex_unlock(&pData->mutWrite);
 ENDcommitTransaction
 
--- a/tools/omfile.c
+++ b/tools/omfile.c
@@ -37,6 +37,9 @@
  */
 #include "config.h"
 #include "rsyslog.h"
+// <kortemik>
+#include "glbl.h"
+// </kortemik>
 #include <stdio.h>
 #include <stdarg.h>
 #include <stdlib.h>
@@ -49,6 +52,9 @@
 #include <unistd.h>
 #include <sys/file.h>
 #include <fcntl.h>
+// <kortemik>
+#include <sys/statvfs.h>
+// </kortemik>
 #ifdef HAVE_ATOMIC_BUILTINS
 #	include <pthread.h>
 #endif
@@ -83,6 +89,7 @@ static rsRetVal resetConfigVariables(uchar __attribute__((unused)) *pp, void __a
  */
 DEF_OMOD_STATIC_DATA
 DEFobjCurrIf(errmsg)
+DEFobjCurrIf(glbl)
 DEFobjCurrIf(strm)
 DEFobjCurrIf(statsobj)
 
@@ -132,7 +139,6 @@ typedef struct s_dynaFileCacheEntry dynaFileCacheEntry;
 #define USE_ASYNCWRITER_DFLT 0 	/* default buffer use async writer */
 #define FLUSHONTX_DFLT 1 	/* default for flush on TX end */
 
-
 typedef struct _instanceData {
 	pthread_mutex_t mutWrite; /* guard against multiple instances writing to single file */
 	uchar	*fname;	/* file or template name (display only) */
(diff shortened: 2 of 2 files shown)


rgerhards:
I could re-init some CI-System induced CI failures, but the ubuntu16 slave will not work without a rebase. I would appreciate if you could rebase to current master.

runtime/rsyslog.h
```diff
@@ -510,6 +510,9 @@ operation not carried out */
 	/* RainerScript error messages (range 1000.. 1999) */
 	RS_RET_SYSVAR_NOT_FOUND = 1001, /**< system variable could not be found (maybe misspelled) */
 	RS_RET_FIELD_NOT_FOUND = 1002, /**< field() function did not find requested field */
+	// <kortemik>
```
rgerhards: I would appreciate if you could remove these author lines. git log shows that and so it only clutters the source.

tools/omfile.c
```diff
@@ -660,6 +679,58 @@ prepareFile(instanceData *__restrict__ const pData, const uchar *__restrict__ co
 	RETiRet;
 }
 
+// <kortemik>
```
rgerhards: this author note IMHO is fine, I would maybe add a date to it

rgerhards:
I did not have time to go through the code in depth, but a general question. Assuming action retries was configured, I think then the behavior actually changes. If so, I tend to think that we need a config switch to enable the new mode. @davidelang what do you think?

rgerhards:
@kortemik did you see my comments? I just ask as I would like to make sure we can merge for 8.34.

kortemik:
i suspect cwd is in /proc for the failed tests, and the functionality brings it up now

rgerhards:
... that was probably intensional, at least looking at one test. So it looks like default behavior changed, just like I suspected. So I think we need to have a config switch that permits to opt in to the new functionality but otherwise keep the previous default active.

kortemik:
i may be wrong but it is a possibility still that the functionality did not change that much.
test didn't fail previously and i did only change the comments, so i suspect there is a race condition in the failing tests:

the test waits until "main queue" is empty, but because the omfile action is now suspending the "action queue" may actually have messages still in it. This would cause the observed behaviour as the "action queue" does not have time to drain before the instance goes down.

if this is true, then increasing time before bringing instance down should make the tests pass as the "action queue" would then have time to work the messages through after the message intended to fail gets discarded.

rgerhards:
Well, you are saying yourself the behavior changes. We do not permit this (not me but the project). So this needs to be optional. Either you add the option or I do or we cannot merge. That's simply how it is. We do changes only in extreme emergencies.

rgerhards:
As a side note: the testbench ensures that a wait is done on shutdown.

rgerhards:
To be precise, at least one of the tests explicitly checks that a wrong file name does not block the action. The comment says it's modelled after actual use case.

kortemik:
latest commit now re-applies the bDynafileDoNotSuspend functionality to retain the old behaviour.

the documentation regarding this variable needs to be updated as well as http://www.rsyslog.com/doc/v8-stable/rainerscript/global.html states the variable as a global one 
although it's module specific and should be on omfile page.

kortemik:
@rgerhards now it should match the expectations. i must admit that the behaviour has still changed as now it does quite a lot more error checking on the writes, file descriptor creation, stream flushes and so on. This means that non-dynafile omfile output is more reliable than previously and that may cause environment where problems with those topics existed to suspend the omfile action. II see this as acceptable change but in case not, then I propose to guard all the changes with if (reliable) { changes } else { old } .

rgerhards:
thx, I'll have an in-depth review in spit of the functionality change. Note that CI has some issues, working on that (Solaris is probably caused by some system update, at least that's what I currently think).

@davidelang I would appreciate your thoughts on the potential change of behavior.

rgerhards:
No objection so far, so I think we give the new code a go. If problems pop up, we can change the behavior.

doing a merge PR to "rebase" and see if we get it CI clean.

rgerhards:
@kortemik the Suse CI failure also comes up on my merge PR. It looks like a valid issue. Unfortunately, this looks like I cannot merge it as-is.

kortemik:
@rgerhards are you able to run df -i on the Suse CI?

rgerhards:
Interesting:
```
Filesystem     Inodes IUsed  IFree IUse% Mounted on
devtmpfs       185590   405 185185    1% /dev
tmpfs          187511     6 187505    1% /dev/shm
tmpfs          187511   530 186981    1% /run
tmpfs          187511    15 187496    1% /sys/fs/cgroup
/dev/sda2           0     0      0     - /
/dev/sda2           0     0      0     - /var/lib/mariadb
/dev/sda2           0     0      0     - /tmp
/dev/sda2           0     0      0     - /var/lib/libvirt/images
/dev/sda2           0     0      0     - /var/tmp
/dev/sda2           0     0      0     - /var/opt
/dev/sda2           0     0      0     - /var/lib/pgsql
/dev/sda2           0     0      0     - /var/lib/named
/dev/sda2           0     0      0     - /var/crash
/dev/sda2           0     0      0     - /usr/local
/dev/sda2           0     0      0     - /.snapshots
/dev/sda2           0     0      0     - /var/spool
/dev/sda2           0     0      0     - /home
/dev/sda2           0     0      0     - /srv
/dev/sda2           0     0      0     - /opt
/dev/sda2           0     0      0     - /var/lib/mysql
/dev/sda2           0     0      0     - /var/lib/mailman
/dev/sda2           0     0      0     - /var/log
/dev/sda2           0     0      0     - /boot/grub2/x86_64-efi
/dev/sda2           0     0      0     - /boot/grub2/i386-pc
```
Obviously there **are** free inodes, as all other tests work well (and the system operates flawlessly).

kortemik:
of course we can leave the inode check out and rely on fopen errno but my point was to leave one unused as it would allow an administrator to roll up a tar-file on the same filesystem to free up inodes. 

other filesystems may not have enough space or can not be contaminated with log information.

which filesystem is on /dev/sda2 ?

rgerhards:
@kortemik btrfs - by virtue of SuSe default

All in all, I think we should consider refactoring the PR - it seems to have a couple of things that can really break existing installations, and so I am hesitant to merge it for next Tuesday's release. @whissi @mbiebl @deoren I would appreciate your feedback.

rgerhards:
@kortemik I forgot the say that I think it is useful functionality, especially the part of leaving some space. But it sounds like we need a couple of opt-in switches.

deoren:
> @rgerhards: All in all, I think we should consider refactoring the PR - it seems to have a couple of things that can really break existing installations, and so I am hesitant to merge it for next Tuesday's release. @Whissi @mbiebl @deoren I would appreciate your feedback.

It's probably worth considering. The last release was unintentionally bumpy, so putting out another with code that you don't have a lot of confidence in yourself is probably cause for further review.

davidelang:
sorry for being late to the conversation.

first off, I think this needs to be refactored a bit. there is a fair bit of cleanup work (int -> uint, etc), separate from the new functionality. I think it would be great to have the cleanup stuff completely separate from the new functionality.

As for the new functionality where for every message you do a bunch of checks to check and see if there is enough disk space, inodes to flush the entire queue m(and check for readonly filesystems)

This is a performance nightmare, think of doing this tens or hundreds of thousands of times a second.

I think it's worth doing more testing to figure out what is wrong after you get a write error, but trying to 'prove' that a write to the filesystem is not going to fail before doing the write is an impossible task. Even if you do every check that you propose, you could still fail to write because some other program wrote to some other file on the filesystem (or one of a hundred other things happened)

When you have delayed writes to the filesystem, the application can check that there was no error when doing the write() call to the filesystem, but it cannot guarantee that the filesystem will actually get the data to disk (without doing fsync() calls), there is just too much else that can happen.

Going a step further and trying to be sure that data that you haven't tried to write yet will be written without an error is significantly more ambitious.

It may be worth doing something like this every few seconds, but doing it on every write to disk not only won't work, it will be a drastic performance hit.

You need to have monitoring tools watching your filesystem and alerting you when you get low on space (where low is defined by the sysadmin as something they consider suitable)

kortemik:
I say we need to profile this. However if that's too much effort then we must rely on theory statvfs() is quite cheap syscall compared to filesystem related syscalls in place i.e. write. In our performance test and neither in production environments we experienced no difference in performance.

The omfile does not make any claims that something in queue is writeable/storeable, as that's on user configuration, they need to make sure that their queues are reliable and can store events when something unexpected happens with the actions configured. Only claim omfile should and can make is that message inserted via action() is written on the disk. 

Using the synchronous feature, where sync syscall is made after each write one can achieve it. However synchronous writing is terribly slow and unusable for most usecases. The functionality in this PR adds features that usual mission critical systems have for asynchronous writing so that only a few messages are lost in a process crash. Maximum lost being the number that will fit the write buffer of rsyslog.

OS crash or something else filling up the filesystem which rsyslog writes to is something we can not make any claims on and therefore should not care about. It's up to the system operator to configure the system in reliable way and properly for their use case.

@rgerhards would you agree on "if total inodes eq 0 then skip the verify free inode clause"?

rgerhards:
With what we have seen so far, I think this PR tries to optimize omfile to one specific use case. It is obviously an important use case, but needs to be used in the minority of cases. In contrast, it looks like omfile semantics are pretty much changed.

All in all, I would propose this path forward: let's create a new module "omfile-hardned" (or so) in the ./contrib folder. Then, let's experiment in practice with this version of imfile. This also gives @kortemik the ability to do some further changes for the use case without the need to think about the rest of the rsyslog project. Once a couple of month has passed, and users have actually opted to use the new code and did not have any issues, we may consider moving it as an optional capability into omfile itself.

IMHO there is much benefit in such a stages approach. If nobody objects, I can take the PR and create a new PR out of it that creates the new module. Any objections, please let me know.

kortemik:
+1 for the staged approach!

davidelang:
On Fri, 30 Mar 2018, Mikko Kortelainen wrote:

> I say we need to profile this. However if that's too much effort then we must 
> rely on theory statvfs() is quite cheap syscall compared to filesystem related 
> syscalls in place i.e. write. In our performance test and neither in 
> production environments we experienced no difference in performance.

when I started looking at rsyslog, we had a number of 'cheap' gettimeofday() 
calls in the processing path, and we discovered that removing them signficantly 
increased the max throughput. That it didn't hurt you in production could just 
mean you are not near enough to the edge :-)

> The omfile does not make any claims that something in queue is 
> writeable/storeable, as that's on user configuration, they need to make sure 
> that their queues are reliable and can store events when something unexpected 
> happens with the actions configured. Only claim omfile should and can make is 
> that message inserted via action() is written on the disk.
>
> Using the synchronous feature, where sync syscall is made after each write one 
> can achieve it. However synchronous writing is terribly slow and unusable for 
> most usecases. The functionality in this PR adds features that usual mission 
> critical systems have for asynchronous writing so that only a few messages are 
> lost in a process crash. Maximum lost being the number that will fit the write 
> buffer of rsyslog.

applications can't guarantee that when they issue a write() call, that the 
result will get to disk. They can issue a sync call to force the write and 
detect a failure, but a success on write() just means the filesystem accepted 
the call, not that it will make it to disk.

The max that can be lost in a system crash is the max that the filesystem 
buffers, not just rsyslogs buffers (including the queues)

as you have said, reliable writes are incredibly slow (I did testing years ago 
and found the slowdown was on the order of 1000x, and that was when using a very 
expensive PCI based SSD, writing to a more normal disk was another 10x slower)

I believe that your checks are too expensive for the common case, and not enough 
when you have problems.

> OS crash or something else filling up the filesystem which rsyslog writes to 
> is something we can not make any claims on and therefore should not care 
> about. It's up to the system operator to configure the system in reliable way 
> and properly for their use case.

I would say that it's up to the system operator to monitor the filesystem to 
alert them when it gets close to full or otherwise has problems, not the job of 
rsyslog (and every other application) to try and detect these problems on every 
write.

David Lang

davidelang:
On Fri, 30 Mar 2018, Rainer Gerhards wrote:

> IMHO there is much benefit in such a stages approach. If nobody objects, I can take the PR and create a new PR out of it that creates the new module. Any objections, please let me know.

sounds good to me.

rgerhards:
Will be merged via 2606. I'll close this one, for github that than looks unmerged - just FYI.

PR #2494, opened 2018-02-20
tests/Makefile: Fix EXTRA_DIST and add testsuites/include-std-omfile-action.conf

Whissi:
This commit fixes EXTRA_DIST list which was broken via commit
87f296fd2e667e95f38991a8a7caf84c3458e7c9.

In addition, this commit adds the missing testsuites/include-std-omfile-action.conf file
to EXTRA_DIST.

Fixes: https://github.com/rsyslog/rsyslog/issues/2493

Issue #2493, opened 2018-02-20
include-obj-text-from-file.sh is invalid


Whissi:
With rsyslog-8.33.0 release I am seeing:
```
------------------------------------------------------------
Test: ./include-obj-text-from-file.sh
------------------------------------------------------------
testbench: TZ env var not set, setting it to UTC
testconf.conf is:
     1  $IncludeConfig diag-common.conf
     2
     3  template(name="outfmt" type="string" string="%msg:F,58:2%\n")
     4
     5  if $msg contains "msgnum:" then {
     6          include(text=`cat testsuites/include-std-omfile-action.conf`)
     7  }
rsyslogd: error during parsing file testconf.conf, on or before line 6: invalid backtick parameter `cat testsuites/include-std-omfile-action.conf` - replaced by empty string ("") [v8.33.0 try http://www.rsyslog.com/e/2207 ]
rsyslogd started, start msg not yet seen, pid  56687
rsyslogd startup msg seen, pid  56687
injecting 10 messages
imdiag[13500]: 10 messages injected
imdiag[13500]: mainqueue empty
cp: cannot stat 'rsyslog.out.log': No such file or directory
./diag.sh: line 539: rsyslog.out.log: No such file or directory
work: No such file or directory
error opening file 'work'
sequence error detected
FAIL include-obj-text-from-file.sh (exit status: 1)
```

Whissi:
Not sure why, but it looks like we are only missing 4 files:
```
--- /tmp/rsyslog-8.34.0.txt     2018-02-20 20:49:36.480837997 +0100
+++ /tmp/rsyslog-8.33.0.txt     2018-02-20 20:49:19.114604903 +0100
@@ -385,7 +385,6 @@
 ├── rsyslog/tests/omrelp_dflt_port.c
 ├── rsyslog/tests/omruleset-queue.sh
 ├── rsyslog/tests/omruleset.sh
-├── rsyslog/tests/omstdout-basic.sh
 ├── rsyslog/tests/omtcl.sh
 ├── rsyslog/tests/omtcl.tcl
 ├── rsyslog/tests/ourtail.c
@@ -536,14 +535,12 @@
 ├── rsyslog/tests/tcp_forwarding_retries.sh
 ├── rsyslog/tests/tcp_forwarding_tpl.sh
 ├── rsyslog/tests/tcp-msgreduc-vg.sh
-├── rsyslog/tests/template-const-jsonf.sh
[...]

PR #2498, opened 2018-02-26
Test for CI system itself -- should break

rgerhards:
This checks why the CI system did not detected the problem described in #2493 and fixed in #2494.


rgerhards:
test was successful, PR no longer needed.

PR #2507, opened 2018-03-01
Kubernetes Metadata plugin - mmkubernetes

richm:
This plugin is used to annotate records logged by Kubernetes containers.
It will add the namespace uuid, pod uuid, pod and namespace labels and
annotations, and other metadata associated with the pod and namespace.
It will work with either log files in `/var/log/containers/*.log` or
with journald entries with `CONTAINER_NAME` and `CONTAINER_ID_FULL`.

For file logs, the filename must match this regex:

    /var/log/containers/([a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*)_([^_]+)_(.+)-([a-z0-9]{64})\\.log$

The first match is the pod name, the second is the container hash (not
currently used), the third is the namespace name, the fourth is the
container name, and the fifth is the container id.  You can specify a
different regex by using the module or action configuration parameter
`filenameregex` but keep in mind that the field positions are hardcoded
so the regex must have the same fields in the same order.

For journald logs, there must be a field `CONTAINER_NAME` which matches this:

    ^[^_]+_([^\\._]+)(\\.([^_]+))?_([^_]+)_([^_]+)_[^_]+_[^_]+$

The first match is the container name, the second is the container hash
(not currently used), the third is the pod name, and the fourth is the
namespace name.  The record must also have the field `CONTAINER_ID_FULL`.
You can specify a different regex by using the module or action
configuration parameter `containerregex` but keep in mind that the field
positions are hardcoded so the regex must have the same fields in the same
order.

The Kubernetes metadata is added to the record in the top-level
fields `kubernetes` and `docker`.
See https://github.com/ViaQ/elasticsearch-templates/blob/master/namespaces/kubernetes.yml
and
https://github.com/ViaQ/elasticsearch-templates/blob/master/namespaces/docker.yml
for more details.

*Configuration*
`kubernetesurl` - Required - URL of the Kubernetes API server e.g.
                  `https://localhost:8443`
`tls.cacert` - Required - full path and file name of file containing
               the CA cert of the Kubernetes API server cert issuer
`tokenfile` - Required (or `token`) - the file containing the token
              to use to authenticate to the Kubernetes API server
`token` - Required (or `tokenfile`) - the token to use to
          authenticate to the Kubernetes API server
`annotation_match` - Optional - by default no pod or namespace annotations
                     will be added to the records - this parameter is an
                     array of patterns to match the keys of the `annotations`
                     field to include in the `annotations` field or the
                     `namespace_annotations` field.

*Example*

    module(load="imfile" mode="inotify")
    module(load="mmkubernetes" kubernetesurl="https://localhost:8443"
           tls.cacert="/etc/rsyslog.d/mmk8s.ca.crt"
           tokenfile="/etc/rsyslog.d/mmk8s.token" annotation_match=["."])

    template(name="tpl" type="list") {
        property(name="jsonmesg")
        constant(value="\n")
    }

    ruleset(name="k8s") {
        action(type="mmkubernetes")
        action(type="omfile" file="/var/log/k8s.log" template="tpl")
    }

    input(type="imfile" file="/var/log/containers/*.log" tag="kubernetes" addmetadata="on" ruleset="k8s")
    if ($!_SYSTEMD_UNIT == "docker.service") and (strlen($!CONTAINER_NAME) > 0) then {
        call k8s
    }

*Credits*
This work is based on https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter
and has many of the same features.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -279,6 +279,11 @@ if ENABLE_OMTCL
 SUBDIRS += contrib/omtcl
 endif
 
+# mmkubernetes
+if ENABLE_MMKUBERNETES
+SUBDIRS += contrib/mmkubernetes
+endif
+
 # tests are added as last element, because tests may need different
 # modules that need to be generated first
 SUBDIRS += tests
--- a/configure.ac
+++ b/configure.ac
@@ -2085,6 +2085,25 @@ AM_CONDITIONAL(ENABLE_OMTCL, test x$enable_omtcl = xyes)
 
 # END TCL SUPPORT
 
+# mmkubernetes - Kubernetes metadata support
+
+AC_ARG_ENABLE(mmkubernetes,
+        [AS_HELP_STRING([--enable-mmkubernetes],
+            [Enable compilation of the mmkubernetes module @<:@default=no@:>@])],
+        [case "${enableval}" in
+         yes) enable_mmkubernetes="yes" ;;
+          no) enable_mmkubernetes="no" ;;
+           *) AC_MSG_ERROR(bad value ${enableval} for --enable-mmkubernetes) ;;
+         esac],
+        [enable_mmkubernetes=no]
+)
+if test "x$enable_mmkubernetes" = "xyes"; then
+        PKG_CHECK_MODULES([CURL], [libcurl])
+fi
+AM_CONDITIONAL(ENABLE_MMKUBERNETES, test x$enable_mmkubernetes = xyes)
+
+# END Kubernetes metadata support
+
 # man pages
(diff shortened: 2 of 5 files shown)


richm:
@portante

portante:
Since we have a discussion history in the first, why not just add this as a patch to that one?

richm:
> Since we have a discussion history in the first, why not just add this as a patch to that one?

How?  submit a PR against your repo and PR?

PR #2508, opened 2018-03-02
Fix data alignment problems in plugins/imuxsock/imuxsock.c

jduerstock:
gcc would (rightly) complain that the cred and ts pointers
would cause alignment problems, so they were converted to structs
and the necessary data was memcpy()'ed to them.

the aux[] buffer was also potentially misaligned, so making a union
out of it and struct cmsghdr insured it was properly aligned.

The problems were especially visible on ia64 machines:
```
[6143331.481422] ia64_handle_unaligned: 2 callbacks suppressed
[6143331.481463] in:imuxsock(31387): unaligned access to 0x2000000000e2e554, ip=0x20000000008594c0
[6143331.481519] in:imuxsock(31387): unaligned access to 0x2000000000e2e55c, ip=0x20000000008594e0
```

### DIFF
--- a/plugins/imuxsock/imuxsock.c
+++ b/plugins/imuxsock/imuxsock.c
@@ -1043,11 +1043,6 @@ finalize_it:
  * of the socket which is to be processed. This eases access to the
  * growing number of properties. -- rgerhards, 2008-08-01
  */
-#if !defined(_AIX)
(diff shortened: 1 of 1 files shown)


jduerstock:
Oops, forgot something.  Will resubmit.

PR #2516, opened 2018-03-05
coverity: work around expired/invalid scan.coverity.com cert

rgerhards:
DO NOT MERGE THIS AT THE MOMENT

I just found out that the problem is not only the certificate, but Coverity Scan does not work at all since February, 20th. See Coverity's Twitter channel: https://twitter.com/CoverityScan

I hope we get this up again -- Coverity was a very good addition to your toolbox.


rgerhards:
This doesn't help, because not only the cert is dead -- the whole site and service does no longer work.

PR #2527, opened 2018-03-09
tests/glbl-umask.sh: Improve test stability

hanazuki:
This patch improves the stability of `tests/glbl-umask.sh`, which now fails when running after `tests/rawmsg-after-pri.sh`, since the glob pattern `rsyslog.o*` matches `rsyslog.out.compare`, created by `rawmsg-after-pri.sh`.

### DIFF
--- a/tests/glbl-umask.sh
+++ b/tests/glbl-umask.sh
@@ -21,7 +21,7 @@ $srcdir/diag.sh injectmsg 0 1
 . $srcdir/diag.sh shutdown-when-empty
 . $srcdir/diag.sh wait-shutdown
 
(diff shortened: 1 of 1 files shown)


rgerhards:
Thx, good catch! It also points to the fact that the cleanup routine is not working thouroughly.

Note that I will probably need a couple of days to merge as it also serves as a reminder item for the latter to me and I am right now deep it some very complex refactoring work.

rgerhards:
I have changed the testbench framework to do a more proper cleanup. Nevertheless, I think you patch still makes a lot of sense, as I do not see any real reason for using the wildcard.

PR #2543, opened 2018-03-14
Fetching cursor on readJournal() and simplified pollJournal()

jvymazal:
Fetching journal cursor in persistJournal could cause us to save
invalid cursor leading to duplicating messages further on, now we are
saving it on each readJournal() where we now that the state is good.
This result in simplyfing persisJournalState() a bit as well.

pollJournal() is now cleaner and faster, correctly handles INVALIDATE
status from journald and is able to continue polling after journal
flush.

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -80,6 +80,7 @@ static struct configSettings_s {
 	int iDfltFacility;
 	int bUseJnlPID;
 	char *usePid;
+	int bWorkAroundJournalBug;
 } cs;
 
 static rsRetVal facilityHdlr(uchar **pp, void *pVal);
@@ -95,7 +96,8 @@ static struct cnfparamdescr modpdescr[] = {
 	{ "defaultseverity", eCmdHdlrSeverity, 0 },
 	{ "defaultfacility", eCmdHdlrString, 0 },
 	{ "usepidfromsystem", eCmdHdlrBinary, 0 },
-	{ "usepid", eCmdHdlrString, 0 }
+	{ "usepid", eCmdHdlrString, 0 },
+	{ "workaroundjournalbug", eCmdHdlrBinary, 0 }
 };
 static struct cnfparamblk modpblk =
 	{ CNFPARAMBLK_VERSION,
@@ -129,6 +131,7 @@ static struct {
 	uint64 ratelimitDiscardedInInterval;
 	uint64 diskUsageBytes;
 } statsCounter;
+static char *last_cursor = NULL;
 
 #define J_PROCESS_PERIOD 1024  /* Call sd_journal_process() every 1,024 records */
 
@@ -288,6 +291,7 @@ readjournal(void)
 	char *message = NULL;
 	char *sys_iden;
 	char *sys_iden_help = NULL;
+	char *c = NULL;
 
 	const void *get;
 	const void *pidget;
@@ -433,6 +437,15 @@ readjournal(void)
 		tv.tv_usec = timestamp % 1000000;
 	}
 
+	if (cs.bWorkAroundJournalBug) {
+		/* save journal cursor (at this point we can be sure it is valid) */
+		sd_journal_get_cursor(j, &c);
+		if (c) {
+			free(last_cursor);
+			last_cursor = c;
+		}
+	}
+
 	/* submit message */
 	enqMsg((uchar *)message, (uchar *) sys_iden_help, facility, severity, &tv, json, 0);
 
@@ -451,87 +464,82 @@ persistJournalState(void)
 	DEFiRet;
 	FILE *sf; /* state file */
 	char tmp_sf[MAXFNAME];
-	char *cursor;
-	int ret = 0;
+	size_t n;
 
-	/* On success, sd_journal_get_cursor() returns 1 in systemd
-	   197 or older and 0 in systemd 198 or newer */
-	if ((ret = sd_journal_get_cursor(j, &cursor)) >= 0) {
-		/* we create a temporary name by adding a ".tmp"
-		* suffix to the end of our state file's name
-		*/
-		snprintf(tmp_sf, sizeof(tmp_sf), "%s.tmp", cs.stateFile);
-		if ((sf = fopen(tmp_sf, "wb")) != NULL) {
-			if (fprintf(sf, "%s", cursor) < 0) {
-				iRet = RS_RET_IO_ERROR;
-			}
-			fclose(sf);
-			free(cursor);
-			/* change the name of the file to the configured one */
-			if (iRet == RS_RET_OK && rename(tmp_sf, cs.stateFile) == -1) {
-				LogError(errno, iRet, "imjournal: rename() failed: "
-					"for new path: '%s'", cs.stateFile);
-				iRet = RS_RET_IO_ERROR;
-			}
-
-		} else {
-			LogError(errno, RS_RET_FOPEN_FAILURE, "imjournal: fopen() failed "
-				"for path: '%s'", tmp_sf);
-			iRet = RS_RET_FOPEN_FAILURE;
+	if (cs.bWorkAroundJournalBug) {
+		/* first check that we have valid cursor */
+		if (!last_cursor) {
+			ABORT_FINALIZE(RS_RET_OK);
 		}
 	} else {
-		LogError(-ret, RS_RET_ERR, "imjournal: sd_journal_get_cursor() failed");
-		iRet = RS_RET_ERR;
+		int ret;
+		if ((ret = sd_journal_get_cursor(j, &last_cursor))) {
+			LogError(-ret, RS_RET_ERR, "imjournal: sd_journal_get_cursor() failed");
+			ABORT_FINALIZE(RS_RET_ERR);
+		}
 	}
+
+	/* we create a temporary name by adding a ".tmp"
+	 * suffix to the end of our state file's name
+	 */
+	n = snprintf(tmp_sf, sizeof(tmp_sf), "%s.tmp", cs.stateFile);
+	if (n >= sizeof(tmp_sf)) { /* backup in case of too long path name */
+		strncpy(tmp_sf, cs.stateFile, sizeof(tmp_sf) - 5);
+		strcat(tmp_sf, ".tmp");
+	}
+
+	sf = fopen(tmp_sf, "wb");
+	if (sf == NULL) {
+		LogError(errno, RS_RET_FOPEN_FAILURE, "imjournal: fopen() failed for path: '%s'", tmp_sf);
+		ABORT_FINALIZE(RS_RET_FOPEN_FAILURE);
+	}
+
+	if(fputs(last_cursor, sf) == EOF) {
+		LogError(errno, RS_RET_IO_ERROR, "imjournal: failed to save cursor to: '%s'", tmp_sf);
+		ABORT_FINALIZE(RS_RET_IO_ERROR);
+	}
+
+	if (fclose(sf) == EOF) {
+		LogError(errno, RS_RET_IO_ERROR, "imjournal: fclose() failed for path: '%s'", tmp_sf);
+		ABORT_FINALIZE(RS_RET_IO_ERROR);
+	}
+
+	/* change the name of the file to the configured one */
+	if (rename(tmp_sf, cs.stateFile) < 0) {
+		LogError(errno, iRet, "imjournal: rename() failed for new path: '%s'", cs.stateFile);
+		ABORT_FINALIZE(RS_RET_IO_ERROR);
+	}
+
+finalize_it:
 	RETiRet;
 }
 
 
 static rsRetVal skipOldMessages(void);
-/* Polls the journal for new messages. Similar to sd_journal_wait()
- * except for the special handling of EINTR.
- */
 
-#define POLL_TIMEOUT 1000 /* timeout for poll is 1s */
+#define POLL_TIMEOUT 900000 /* timeout for poll is 900ms */
 
 static rsRetVal
 pollJournal(void)
 {
 	DEFiRet;
-	struct pollfd pollfd;
-	int err; // journal error code to process
-	int pr = 0;
+	int err;
 
-	pollfd.fd = j_inotify_fd;
-	pollfd.events = sd_journal_get_events(j);
-#ifdef NEW_JOURNAL
-	pr = poll(&pollfd, 1, POLL_TIMEOUT);
-#else
-	pr = poll(&pollfd, 1, -1);
-#endif
-	if (pr == -1) {
-		if (errno == EINTR) {
-			/* EINTR is also received during termination
-			 * so return now to check the term state.
-			 */
-			ABORT_FINALIZE(RS_RET_OK);
-		} else {
-			LogError(errno, RS_RET_ERR, "imjournal: poll() failed");
-			STATSCOUNTER_INC(statsCounter.ctrPollFailed, statsCounter.mutCtrPollFailed)
+	err = sd_journal_wait(j, POLL_TIMEOUT);
+	if (err == SD_JOURNAL_INVALIDATE) {
+		STATSCOUNTER_INC(statsCounter.ctrRotations, statsCounter.mutCtrRotations);
+		closeJournal();
+
+		iRet = openJournal();
+		if (iRet != RS_RET_OK) {
 			ABORT_FINALIZE(RS_RET_ERR);
 		}
-	}
-#ifndef NEW_JOURNAL
-	assert(pr == 1);
-#endif
 
-	err = sd_journal_process(j);
-	if (err < 0) {
-		LogError(-err, RS_RET_ERR, "imjournal: sd_journal_process() failed");
-		ABORT_FINALIZE(RS_RET_ERR);
+		if (cs.stateFile) {
+			iRet = loadJournalState();
+		}
+		LogMsg(0, RS_RET_OK, LOG_NOTICE, "imjournal: journal reloaded...");
 	}
-	if (err == SD_JOURNAL_INVALIDATE)
-		STATSCOUNTER_INC(statsCounter.ctrRotations, statsCounter.mutCtrRotations);
 
 finalize_it:
 	RETiRet;
@@ -776,6 +784,7 @@ CODESTARTbeginCnfLoad
 	cs.iDfltFacility = DFLT_FACILITY;
 	cs.bUseJnlPID = -1;
 	cs.usePid = NULL;
+	cs.bWorkAroundJournalBug = 0;
 ENDbeginCnfLoad
 
 
@@ -912,6 +921,8 @@ CODESTARTsetModCnf
 			cs.bUseJnlPID = (int) pvals[i].val.d.n;
 		} else if (!strcmp(modpblk.descr[i].name, "usepid")) {
 			cs.usePid = (char *)es_str2cstr(pvals[i].val.d.estr, NULL);
+		} else if (!strcmp(modpblk.descr[i].name, "workaroundjournalbug")) {
+			cs.bWorkAroundJournalBug = (int) pvals[i].val.d.n;
 		} else {
 			dbgprintf("imjournal: program error, non-handled "
 				"param '%s' in beginCnfLoad\n", modpblk.descr[i].name);
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -28,6 +28,9 @@ check_PROGRAMS = $(TESTRUNS) ourtail tcpflood chkseq msleep randomgen \
 	omrelp_dflt_port \
 	mangle_qi \
 	have_relpSrvSetOversizeMode
+if ENABLE_IMJOURNAL
+check_PROGRAMS += journal_print
+endif
 TESTS = $(TESTRUNS)
 
 if ENABLE_DEFAULT_TESTS
@@ -445,7 +448,7 @@ endif
 
 if ENABLE_IMJOURNAL
 TESTS +=  \
-	imjournal-basic.sh
+	imjournal-basic.sh \
 	imjournal-statefile.sh
 if HAVE_VALGRIND
 TESTS +=  \
@@ -1977,6 +1980,10 @@ minitcpsrv_LDADD = $(SOL_LIBS)
 syslog_caller_SOURCES = syslog_caller.c
 syslog_caller_CPPFLAGS = $(LIBLOGGING_STDLOG_CFLAGS)
 syslog_caller_LDADD = $(SOL_LIBS) $(LIBLOGGING_STDLOG_LIBS)
+
+journal_print_SOURCES = journal_print.c
+journal_print_CPPFLAGS = $(LIBSYSTEMD_JOURNAL_CFLAGS)
+journal_print_LDADD = $(LIBSYSTEMD_JOURNAL_LIBS)
 
 diagtalker_SOURCES = diagtalker.c
 diagtalker_LDADD = $(SOL_LIBS)
--- a/tests/imjournal-basic-vg.sh
+++ b/tests/imjournal-basic-vg.sh
@@ -11,17 +11,22 @@
 . $srcdir/diag.sh require-journalctl
 generate_conf
 add_conf '
-module(load="../plugins/imjournal/.libs/imjournal" IgnorePreviousMessages="on")
+module(load="../plugins/imjournal/.libs/imjournal" IgnorePreviousMessages="on"
+	RateLimit.Burst="1000000")
 
 template(name="outfmt" type="string" string="%msg%\n")
 action(type="omfile" template="outfmt" file="rsyslog.out.log")
 '
 startup_vg
 TESTMSG="TestBenCH-RSYSLog imjournal This is a test message - $(date +%s)"
-systemd-cat echo "$TESTMSG" 
+./journal_print "$TESTMSG"
+if [ $? -ne 0 ]; then
+        echo "SKIP: failed to put test into journal."
+        exit 77
+fi
 journalctl -an 200 | fgrep -qF "$TESTMSG"
 if [ $? -ne 0 ]; then
-        echo "SKIP: some problem with journald service."
+        echo "SKIP: cannot read journal."
         exit 77
 fi
 ./msleep 500
--- a/tests/imjournal-basic.sh
+++ b/tests/imjournal-basic.sh
@@ -11,17 +11,22 @@
 . $srcdir/diag.sh require-journalctl
 generate_conf
 add_conf '
-module(load="../plugins/imjournal/.libs/imjournal" IgnorePreviousMessages="on")
+module(load="../plugins/imjournal/.libs/imjournal" IgnorePreviousMessages="on"
+	RateLimit.Burst="1000000")
 
 template(name="outfmt" type="string" string="%msg%\n")
 action(type="omfile" template="outfmt" file="rsyslog.out.log")
 '
 startup
 TESTMSG="TestBenCH-RSYSLog imjournal This is a test message - $(date +%s)"
-systemd-cat echo "$TESTMSG"
+./journal_print "$TESTMSG"
+if [ $? -ne 0 ]; then
+	echo "SKIP: failed to put test into journal."
+	exit 77
+fi
 journalctl -an 200 | fgrep -qF "$TESTMSG"
 if [ $? -ne 0 ]; then
-	echo "SKIP: some problem with journald service."
+	echo "SKIP: cannot read journal."
 	exit 77
 fi
 ./msleep 500
--- a/tests/imjournal-statefile.sh
+++ b/tests/imjournal-statefile.sh
@@ -11,16 +11,21 @@
 . $srcdir/diag.sh require-journalctl
 generate_conf
 add_conf '
-module(load="../plugins/imjournal/.libs/imjournal" StateFile="imjournal.state")
+module(load="../plugins/imjournal/.libs/imjournal" StateFile="imjournal.state"
+	RateLimit.Burst="1000000")
 
 template(name="outfmt" type="string" string="%msg%\n")
 action(type="omfile" template="outfmt" file="rsyslog.out.log")
 '
 TESTMSG="TestBenCH-RSYSLog imjournal This is a test message - $(date +%s)"
-systemd-cat echo "$TESTMSG"
+./journal_print "$TESTMSG"
+if [ $? -ne 0 ]; then
+        echo "SKIP: failed to put test into journal."
+        exit 77
+fi
 journalctl -an 200 | fgrep -qF "$TESTMSG"
 if [ $? -ne 0 ]; then
-        echo "SKIP: some problem with journald service."
+        echo "SKIP: cannot read journal."
         exit 77
 fi
 # do first run to process all the stuff already in journal db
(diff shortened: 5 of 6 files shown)


jvymazal:
The resuming polling after journald flush (after for example configuration change) is currently not working, as result of #2436

jvymazal:
@rgerhards the CI on fedora26x64 is working correctly? The error there does not make much sense to me...

plugins/imjournal/imjournal.c
```diff
@@ -431,83 +440,70 @@ persistJournalState(void)
...
+		ABORT_FINALIZE(RS_RET_OK);
 	}
+
+	/* we create a temporary name by adding a ".tmp"
+	 * suffix to the end of our state file's name
+	 */
+	snprintf(tmp_sf, sizeof(tmp_sf), "%s.tmp", cs.stateFile);
+
+	sf = fopen(tmp_sf, "wb");
+	if (sf != NULL) {
```
rgerhards: I guess you meant ==, not !=

rgerhards:
Well, the compiler is right in 9999999% of the cases. I flagged the bug in a comment ;-)

jvymazal:
Should I force push empty commit since there is an errored job? Thanks for spotting the typo BTW.

rgerhards:
I currently run some tests on the testbench itself, this can result in temporary errors. I'll baby-sit these. I let you know if something needs your attention.

rgerhards:
@portante I'd appreciate if you could also have a look after your recent journal wrangling ;-)

plugins/imjournal/imjournal.c
```diff
@@ -431,83 +440,70 @@ persistJournalState(void)
...
+	 */
+	snprintf(tmp_sf, sizeof(tmp_sf), "%s.tmp", cs.stateFile);
+
+	sf = fopen(tmp_sf, "wb");
+	if (sf == NULL) {
+		LogError(errno, RS_RET_FOPEN_FAILURE, "imjournal: fopen() failed for path: '%s'", tmp_sf);
+		ABORT_FINALIZE(RS_RET_FOPEN_FAILURE);
+	}
+
+	iRet = fputs(last_cursor, sf);
```
portante: We need an `fsync()` here as well.

plugins/imjournal/imjournal.c
```diff
@@ -431,83 +440,70 @@ persistJournalState(void)
...
+	}
+
+	iRet = fclose(sf);
+	if (iRet < 0) {
+		LogError(errno, RS_RET_IO_ERROR, "imjournal: fclose() failed for path: '%s'", tmp_sf);
+		ABORT_FINALIZE(RS_RET_IO_ERROR);
+	}
+
+	/* change the name of the file to the configured one */
+	iRet = rename(tmp_sf, cs.stateFile);
```
portante: On success, we need to `fsync()` the directory so that the `rename` is persisted.

plugins/imjournal/imjournal.c
```diff
@@ -413,6 +415,13 @@ readjournal(void)
 		tv.tv_usec = timestamp % 1000000;
 	}
 
+	/* save journal cursor (at this point we can be sure it is valid) */
+	sd_journal_get_cursor(j, &c);
```
portante: This is a potential performance issue.  I have not had a chance to verify the overhead of this.  It is on my plate to do.

plugins/imjournal/imjournal.c
jvymazal: Are you sure about performance impact of such a change? AFAIK the fsync() was never in place here and if somebody have has small persistStateInterval set combined with busy disk it could cause quite a problem. And if you have to ensure flush, I would go with fdatasync() at least to speed it up a bit.

plugins/imjournal/imjournal.c
jvymazal: What about combined slowing of two consecutive blocking calls? Does it really help reliability? The window between these two is so small that I am not  sure it makes any difference (aside of having two slow calls execute instead of one).

plugins/imjournal/imjournal.c
msekletar: Obtaining cursor involves memory allocation, but otherwise should be cheap. All metadata that cursor is composed of are very likely in memory already. 

IOW, pointer dereferences inside ```sd_journal_get_cursor()``` shouldn't cause major page fault because ```EntryObject```  was loaded into memory previously since the plugin calls ```sd_journal_get_realtime_usec()``` anyway.

There is one more dereference outside of ```EntryObject```, i.e. fetching journal file ```seqnum_id```, that could potentially cause disk round-trip. However, ```seqnum_id``` will be accessed for every message in current file so it is unlikely that kernel would decide to swap out the page back to disk.

portante: All great in theory, best to test and measure overhead to be sure.

plugins/imjournal/imjournal.c
richm: Won't the `fclose()` do a sync?

plugins/imjournal/imjournal.c
richm: I don't think fsync is needed here.  Once the data is sync'd previously, the data is on the disk, and rename is atomic.  See "man 2 rename".

plugins/imjournal/imjournal.c
richm: the answer is no, it won't:
```
NOTES
       A  successful  close does not guarantee that the data has been success‐
       fully saved to disk, as the kernel  uses  the  buffer  cache  to  defer
       writes.   Typically,  filesystems  do  not flush buffers when a file is
       closed.  If you need to be sure that the data is physically  stored  on
       the  underlying  disk, use fsync(2).  (It will depend on the disk hard‐
       ware at this point.)
```
so yes, this is a performance hit.

plugins/imjournal/imjournal.c
portante: Rich, see https://lwn.net/Articles/457667/

plugins/imjournal/imjournal.c
portante: See https://lwn.net/Articles/457667/

jvymazal: Yes, particularity [this](https://lwn.net/Articles/458176/ ) comment, I do not really like the idea of such a load here.

plugins/imjournal/imjournal.c
```diff
@@ -431,83 +440,72 @@ persistJournalState(void)
...
+		LogError(errno, RS_RET_IO_ERROR, "imjournal: fclose() failed for path: '%s'", tmp_sf);
+		ABORT_FINALIZE(RS_RET_IO_ERROR);
+	}
+
+	/* change the name of the file to the configured one */
+	if (rename(tmp_sf, cs.stateFile) < 0) {
+		LogError(errno, iRet, "imjournal: rename() failed for new path: '%s'", cs.stateFile);
+		ABORT_FINALIZE(RS_RET_IO_ERROR);
+	}
+
```
richm, line 512: Perhaps the choice of whether or not to do a fsync after the fputs, and whether to do the fsync of the directory, should be left up to the admin?  There are some cases where a deployment cannot tolerate, under any circumstances, the duplication (or is it record loss in this case?) which will ensue if there is a crash and rsyslog does not read the correct journal cursor.  There are other cases where
- performance is the most critical factor - seeing records as soon as possible
- hard crashes are extremely rare
- slight record duplication (or loss?) is tolerable

in which case the user would like to relax the sync constraint.  Or - too many knobs?  We know what's best for all users?

davidelang:
On Fri, 16 Mar 2018, Richard Megginson wrote:

>> +	}
> +
> +	iRet = fputs(last_cursor, sf);
> +	if (iRet < 0) {
> +		LogError(errno, RS_RET_IO_ERROR, "imjournal: failed to save cursor to: '%s'", tmp_sf);
> +		ABORT_FINALIZE(RS_RET_IO_ERROR);
> +	}
> +
> +	iRet = fclose(sf);
> +	if (iRet < 0) {
> +		LogError(errno, RS_RET_IO_ERROR, "imjournal: fclose() failed for path: '%s'", tmp_sf);
> +		ABORT_FINALIZE(RS_RET_IO_ERROR);
> +	}
> +
> +	/* change the name of the file to the configured one */
> +	iRet = rename(tmp_sf, cs.stateFile);
>
> I don't think fsync is needed here either.  After calling fclose, the data is on the disk, and rename is atomic.  See "man 2 rename".

rename is atomic in memory, not on disk. fsync (or similar) is needed to force 
the memory state of the filesystem to be written to disk and can be EXTREMELY 
expensive on some filesystems (ext3 is pathalogically bad at doing fsync)

if you do a rename, anything accessing the live filesystem (with the memory 
portion intact) will see either the old or the new name, but if you loose power 
(and therefor the memory portion of the filesystem) you may see either name, 
neither name, or some other problem.

David Lang

plugins/imjournal/imjournal.c
jvymazal: I would probably agree with that, I definitely do not like the idea of doing unconditional fsync() and would leave the default on current behavior with possibility to force fsync() via config switch

richm:
re: atomicity of the state file write - don't we have the same issue with imfile?  How does imfile write its state files?  Are they atomic?  Can we use the same method in imjournal that imfile uses?

jvymazal:
imfile does not use any fsync (AFAIK), but that plugin is in for bigger rewrite as well and maybe we should consider doing the same there

rgerhards:
sorry for being late, the imfile refactoring (base part) took all of my time.

Atomic state file updates and sync behavior is IMHO something we should have a special discussion on. I know that @portante has brought up this issue several times, and I admit that I have only partly understood the use case. Albeit it helped to work more with docker, which at least gives me a glimpse of what the needs are. If we can clearly document the use case, we can get to a clean solution and we can probably create some code that is to be shared among different rsyslog components.

So if nobody objects I would recommend to take this route:
* let's create a new issue tracker for this discussion and spec what we need
* merge this PR as soon as it is ready without regard to these issues - I guess we are already there
* once the "updates/syncing procedure" has been implemented, update all modules that need this functionality.

Maybe we can create a github project to track the bits.

What do you think?

richm:
Works for me.

The issue as I understand it is this:
- rsyslog updates state file but it is not flushed to disk
- machine crashes hard - no buffered writes are flushed to disk - so now the state file is behind the actual cursor position
- machine starts up - rsyslog starts up
- rsyslog reads state file to determine last cursor position, seeks to that cursor position, reads and sends logs from that position

now the log destination has duplicate records

one way to mitigate this is to assign a uuid to each log record to track it in the log destination e.g. what we do with Elasticsearch is to assign a uuid to each record which corresponds to the Elasticsearch `_id` field - Elasticsearch will reject records (documents) if the `_id` already exists, and the Elasticsearch client is smart enough to ignore expected "already exists" bulk index responses.

rgerhards:
> The issue as I understand it is this:

let's move that to a dedicated tracker... will post it soon...

richm:
Another way to solve this is have some sort of transactional semantics with the statefile e.g. in many databases, there is a transaction log which is an append-only journal - the data is written and fsync'd to the transaction log, and periodically flushed to the actual data file - we could do the same with the state files - every log processing operation in rsyslog triggers a write to the statefile - this is actually written to some sort of statefile "transaction log" - this is periodically flushed to the actual statefile - so that we can delay the rename and fsync of the directory, and not have to pay that penalty for every log record. 
When rsyslog starts up, it "recovers" the statefile by ensuring that the statefile "transaction log" is flushed to the actual statefile datafile.

rgerhards:
ALL: please let us move the discussion on "state file hardening" and use cases to https://github.com/rsyslog/rsyslog/issues/2566

jvymazal:
ok, with the fsync matter settled (at least concerning this PR) @rgerhards should I try enough re-pushes until we get a green CI or are you willing to merge this as-is (provided nobody has additional objections)? I have tried re-spinning it three times a row now and got 3 different failures... 
Thanks

rgerhards:
@portante do you think this is ready for merge (not considering the discussion on atomic state file writes, which we now track more generally via #2566).

portante:
Is there an easy way to build a version of rsyslog with this patch for RHEL 7?

jvymazal:
@portante I have sent you instruction via PM

plugins/imjournal/imjournal.c
richm: @portante what do you require in order to be satisfied that this proposed PR does not cause a noticeable performance regression?  Do you have a test rig ready to go, and just need a rhel7.4 build with this patch in it?  Do you need other resources (manpower, computing)?

jvymazal:
@rgerhards @portante  can we decide on this in 8.36 time-frame please?

rgerhards:
I am fine with merging, but it looks to me like @portante has some concerns. As he has recently worked most on this module, I would like to have his explicit "ok" for merging.

rgerhards:
Let's do the following: if @portante does not object, I'll merge into v8.37. This should un-stall the PR.

portante:
@rgerhards, this approach seems like the wrong one to take to solve this issue.

The core problem appears to be that the journal APIs don't automatically detect and handle the case where journald restarts using a different configuration.  This patch proposes to address that deficiency by re-introducing the close/re-open of journal directory behavior, while trying to maintain the proper behavior of not generating duplicate log entries.  This approach relies on the INVALIDATE status being returned from the journal APIs as the trigger for when to close/re-open the journal.  That status is returned any time journal files rotate, *as well as* when journald no longer uses a particular journal directory.  As a result of this reliance on the INVALIDATE status, clients have to save the journal cursor before closing the journal, so that after re-opening the journal they can continue processing logs without re-processing logs already handled.  At the time an INVALIDATE is detected, saving the cursor does not always work, so this patch fetches the cursor after *every* record before processing the next one.  The alleviates the problem of the cursor being bad in the case where a normal rotation occurs by placing the burden of the cursor fetches on the code path processing every record.

The ideal solution would be for the journal APIs to detect that a particular journal directory is no longer being used, reporting that when all records from that journal directory have been processed.  Then clients can know exactly when they can safely close the current journal and re-open the new journal directory, without requiring a fetch of the cursor on every record.

Is it wise to take the change in before seeking a fix to systemd first?  Is it wise to take this change in before properly vetting the change to see if adding a fetch to the cursor on every record is not a problem?

I have gathered baselines for determining the answer to the performance problem, but have not completed the work yet.

Hopefully somebody has already posted an issue to systemd upstream on this topic.

jvymazal:
@portante there is https://github.com/systemd/systemd/issues/8262 opened, but no activity for long time there

rgerhards:
@portante thx for the info - I now know your are still actively looking at this PR. So I won't merge without your explicit "go ahead".

davidelang:
On Mon, 25 Jun 2018, Peter Portante wrote:

> Is it wise to take the change in before seeking a fix to systemd first?

Based on past interactions with the systemd team, yes, it is very wise to take 
the changes rather than trying to get systemd to fix anything.

David Lang

portante:
@davidelang, is it helpful to make technical decisions based on inter-community interactions?

Why would rsyslog want to carry technical debt like this when there are alternatives?

Another solution to this problem is to leave the rsyslog code as is, and watch the journal configuration for changes, restarting rsyslog from the watcher.  That keeps the burden off the fast-path and reduces complexities of the interactions.

davidelang:
On Mon, 25 Jun 2018, Peter Portante wrote:

> @davidelang, is it helpful to make technical decisions based on inter-community interactions?
>
> Why would rsyslog want to carry technical debt like this when there are alternatives?

It's not that we want to, it's that repeated experiences of trying to work with 
the systemd people have been strongly rebuffed, so unless the people in charge 
have changed, trying to work with them fits the definition of insanity (doing 
the same thing and expecting different results)

David Lang

> Another solution to this problem is to leave the rsyslog code as is, and watch the journal configuration for changes, restarting rsyslog from the watcher.  That keeps the burden off the fast-path and reduces complexities of the interactions.
>
>

msekletar:
@davidelang I think you are a bit too harsh in this case. I think that in this specific instance systemd community (Lennart to be precise) acknowledged that API which @portante talks about is missing and could be introduced. However, there are implementation considerations since it would require adding even more "ugly" ```inotify()``` code to ```sd_journal```.

https://github.com/systemd/systemd/issues/8262

portante:
@rgerhards, adding this to the docs is definitely worth while.

Regarding keeping a work-around in place until systemd fixes the problem ... wouldn't that depend on the work-around?  It seems we could all get behind a work-around that does not affect the processing of every log record.  Until we know the performance implications, wouldn't it be questionable to take this change in now?

rgerhards:
> @rgerhards, adding this to the docs is definitely worth while.

I would appreciate if someone who knows the details could do that.
 
> Regarding keeping a work-around in place until systemd fixes the problem ... wouldn't that depend on the work-around? It seems we could all get behind a work-around that does not affect the processing of every log record. Until we know the performance implications, wouldn't it be questionable to take this change in now?

Fully agree, except if this is a really a big problem. I don't have enough insight into what really breaks and for how many users. I mean *in reality*, not in theory. At least I have not yet seen any bug report from users. So it may not be needed to work around ... or maybe it is? Don't know... If we have compelling reason to work-around, a compromise would IMHO be to add a config parameter "WorkAroundJournalBug", which defaults to "off" and do the slow code if explicitly set to "on".

jvymazal:
> add a config parameter "WorkAroundJournalBug", which defaults to "off" and do the slow code if explicitly set to "on".

Ok, added, also added doc change for it, probably best what we can do until journald agrees to make the necessary changes on their side.

rgerhards:
@portante with https://github.com/rsyslog/rsyslog/pull/2543#issuecomment-408013471, I think we can merge (if no issues come up in CI). Do you agree?

jvymazal:
@rgerhards do you think that the failing job is CI flake or valid? If it is valid it looks like some problem in skipOldMessages(), respectively journal call used there.

rgerhards:
> @rgerhards do you think that the failing job is CI flake or valid? 

I would think it is a CI problem - actually I am short of permanently turning off that slave - it's Fedora 23, really! I kept it in because of 32 bit OS, but I have almost replaced it with an armhf box, which I am stabilizing. I have re-initiated the build. Let's see what happens. 

I will double-check CI that journal tests run where possible - doesn't look so atm. Give me a bit, I'll probably reconfigure and re-initiate some tests. I let you know when I have a result. Do not wonder if in the meantime some builders start and stop ;-)

rgerhards:
@jvymazal FYI: we need to be very careful, imjournal was actually NOT tested in the initial run except for the fedora23 slave. Let's see what happens when I enable more places...

rgerhards:
@jvymazal this test was also not properly executed: https://build.rsyslog.com/#/builders/87/builds/566/steps/9/logs/test-suite_log

But the message was from the same user. Is there some issue with the test itself?

rgerhards:
journal now fully enabled on

* centos7
* debian9
* ubuntu16

I should really create a dedicated builder for this ;-)

jvymazal:
> But the message was from the same user. Is there some issue with the test itself?

If you mean the 
```
Hint: You are currently not seeing messages from other users and the system.
      Users in the 'systemd-journal' group can see all messages. Pass -q to
      turn off this notice.
```
that is normal unless you are root (or systemd-journal group), however in the cases when tests currently skip, there appears that the test user does not even have access permissions on the journal files so it cannot insert the message to journal.

rgerhards:
> in the cases when tests currently skip, there appears that the test user does not even have access permissions on the journal files so it cannot insert the message to journal.

mhhh... that sound pretty unusual for a standard user. the buildbot user under which this runs has no special restrictions. IF the issue persists, can you add some debug instrumentation to the test (so that we can see what fails and which what reason)?

rgerhards:
debian9 now also skips with the same message, same goes for ubuntu16... (centos not yet finished)

rgerhards:
I did a quick test on a local ubuntu 18.04 machine, there the test works. But...

jvymazal:
> IF the issue persists, can you add some debug instrumentation to the test (so that we can see what fails and which what reason)?

I will watch it for sure and maybe resurrect the journal_print.c code (which originally did the basically same as systemd-cat ) and extend it a bit so we can see what the problem is in more detail.

rgerhards:
oh wait - maybe we should remove the ```-n 200``` journalctl switch? The machines might be busy. In any case it may be a good idea to show us the output of journalctl - that will also tell us if this is indeed the problem. I would suggest we do this first simply via ```tee``` - so we can see fault and hopefully cause.

rgerhards:
sry, accidental close, where re-open seems to have retriggered CI ;-)

rgerhards:
on a local ubuntu16.04 machine I ran ```find / |systemd-cat ``` in parallel to the test - it sometimes worked, it sometimes failed with:
```
in startup
testconf.conf is:
     1	$ModLoad ../plugins/imdiag/.libs/imdiag
     2	$IMDiagServerRun 13500
     3	
     4	:syslogtag, contains, "rsyslogd"  ./rsyslogd.started
     5	###### end of testbench instrumentation part, test conf follows:
     6	
     7	module(load="../plugins/imjournal/.libs/imjournal" IgnorePreviousMessages="on")
     8	
     9	template(name="outfmt" type="string" string="%msg%\n")
    10	action(type="omfile" template="outfmt" file="rsyslog.out.log")
rsyslogd: NOTE: RSYSLOG_DEBUG_TIMEOUTS_TO_STDERR activated
rsyslogd started, start msg not yet seen, pid  107540
rsyslogd startup msg seen, pid  107540
SKIP: some problem with journald service.
```
which seems to be journal file rotation induced?

I guess journal rate limiting kicked in the cases where the test worked (and suppressed the message generator).

rgerhards:
> which seems to be journal file rotation induced?

no, I guess you have changed the error message. I ran against the version from master. Anyhow, that strongly backs the -n 200 is the problem. Still I think we should do the tee-trick to make sure ;-)

jvymazal:
> maybe we should remove the -n 200 journalctl switch? The machines might be busy.

Does not look like it...

> In any case it may be a good idea to show us the output of journalctl - that will also tell us if this is indeed the problem. 

I am displaying it, although only in case of failure, and in that case 200 records back usually is several **days**, which should be enough for debugging purposes

rgerhards:
ohhh... I logged in into the  testbench ubuntu16 vm using my user account:

```
$ journalctl
Hint: You are currently not seeing messages from other users and the system.
      Users in the 'systemd-journal' group can see all messages. Pass -q to
      turn off this notice.
No journal files were opened due to insufficient permissions.
```

Looks like it is exceptionally hard to do testing for the journal ;-)

rgerhards:
I can confirm that the journal includes the test messages. As a side-note, might this mean rsyslog itself can also not read the messages? Maybe that was actually the root cause why these tests were disabled...

richm:
Correct.  Doing any sort of testing with the system journal requires the ability to use `sudo journalctl ...`. If that is not possible, then it will be very difficult to debug issues.  OTOH, this is what rsyslog admins are expected to do.

rgerhards:
looks like a debian-distro issue - at least on centos7 slave I can run journalctl with my user - ok, let's see what the CentOS 7 test shows.

jvymazal:
> I guess journal rate limiting kicked in the cases where the test worked (and suppressed the message generator).

Well, that was a case before the tests edit, now with `IgnorePreviousMessages="on"` it should correctly get only new things (basically out test message)

> Maybe that was actually the root cause why these tests were disabled...

seems correct, if rsyslog is running under same user as test itself (I suppose it is), than this is expected.

rgerhards:
> Correct. Doing any sort of testing with the system journal requires the ability to use sudo journalctl .... If that is not possible, then it will be very difficult to debug issues. OTOH, this is what rsyslog admins are expected to do.

yeah, but I hope you don't suggest that I permit PRs to run with root privileges inside the CI? Think about the consequences... That won't happen for sure.

rgerhards:
seems like it doesn't work on Centos 7 either (different issue here): https://build.rsyslog.com/#/builders/69/builds/534/steps/6/logs/test-suite_log 

So maybe we need to keep the Fedora 23 worker to be able to test the journal ;-) When I have time I will see that if add a Fedora28 slave.

@jvymazal  Back to to the original discussion: as the only machine capable to check the journal failed again, I would think this is a valid failure.

rgerhards:
I think I managed to reconfigure the testbench VMs (Ubuntu16 and Centos7) so that the journal permits regular users to read data. ```/etc/systemd/journal.conf``` now has:
```
Storage=persistent
SplitMode=uid
```
The ubuntu16 build seems to have succeeded, centos I just changed and re-started. Let's check details tomorrow. I think it would be good to check the fedora builder why it aborts - I'll re-start that one now as well.

rgerhards:
@jvymazal ah, have a look at the current (before restart) fedora log:
```
imjournal from <fedora23x32:rsyslogd>: begin to drop messages due to rate-limiting
imjournal: 1879 messages lost due to rate-limiting
 [origin software="rsyslogd" swVersion="8.37.0.master" x-pid="26685" x-info="http://www.rsyslog.com"] exiting on signal 15.
```
Here, it seems to read really all the log, thus resulting in rate-limiting. We should probably turn this off to make the test robust. I can check if I can create a different environment that explicitely generates that behavior. Does anybody know if this can be configured on the fly **without** root permissions?

richm:
not sure what you mean by "on the fly" but this is controlled by the imjournal settings in rsyslog e.g.
```
module(load="imjournal" StateFile="imjournal.state" UsePidFromSystem="on" RateLimit.Burst="1000000" RateLimit.Interval="1" PersistStateInterval="1000")
```
That is, set the rate limits extremely high so the test has no chance of running into them

rgerhards:
Sorry, i meant the capability to read all journal content, so the journal
part, not rsyslog. I guess it requires changing group membership and so I
can not change dynamically. Now that I wrote it, I think that's a strange
idea and I actually need to clear up my mind what I really think... ;-) thx!

Sent from phone, thus brief.

Richard Megginson <notifications@github.com> schrieb am Do., 26. Juli 2018,
21:35:

rgerhards:
No idea why in CentOS 7 changing the journal settings does not work. I have now added the CI user to the ```systemd-journal``` group. Let's hope this solves the issue.

jvymazal:
> Let's hope this solves the issue.

@rgerhards seems like one of them was skipped again, we will see for sure when it finishes however given how complicated this grows I think I will try to write some simple program that will tell us more about what is actually problem with journal in the CI environments

rgerhards:
> I think I will try to write some simple program that will tell us more about what is actually problem with journal in the CI environments

That's a great idea, much appreciated.

jvymazal:
and going trough the results.. strange is that I cannot see the new test I wrote https://github.com/rsyslog/rsyslog/blob/863bee8a726bc922a2e3fdf996e030eae151827d/tests/Makefile.am#L431 anywhere.. @rgerhards can you please confirm?

rgerhards:
> and going trough the results.. strange is that I cannot see the new test I wrote
> 
> rsyslog/tests/Makefile.am
> 
> Line 431 in 863bee8
> 
>  imjournal-statefile.sh 
> anywhere.. @rgerhards can you please confirm?

doh... there is a backslash missing in the line before! Unfortunately autotools does not detect this as an error.

jvymazal:
> there is a backslash missing in the line before

thanks for spotting it, will commit it with the print_test tool

rgerhards:
@jvymazal I have updated the CI: "rsyslog centos7" is the "old" centos 7 machine (with 3.7) and "rsyslog centos7-5" are the new workers with 7.5. The old one will go away once we are done here, just leaving it in for the moment as you wanted to check it out.  Fedora 23 is also gone now, as it was really too old (32 bit testing now done via arm machine).

rgerhards:
I have re-initiated CI. Note that the fedora 23 fail will not go away as it is not updated. Just ignore it for now. Nevertheless, you now get the new centos 7 config (as do I, so I can check ;-)).

Fedora 23 will go away with the next PR update, should that be necessary.

jvymazal:
@rgerhards thanks, as for me, I think it looks good now, so unless there are additional comments it should be ready to merge (also with rsyslog/rsyslog-doc/pull/712 please).
Thank you very much for all the help with CI, I realize it was lot of work to make it functional and stable.

plugins/imjournal/imjournal.c
```diff
@@ -912,6 +920,8 @@ CODESTARTsetModCnf
 			cs.bUseJnlPID = (int) pvals[i].val.d.n;
 		} else if (!strcmp(modpblk.descr[i].name, "usepid")) {
 			cs.usePid = (char *)es_str2cstr(pvals[i].val.d.estr, NULL);
+		} else if (!strcmp(modpblk.descr[i].name, "workaroundjournalbug")) {
+			cs.bWorkAroundJournalBug = (int) pvals[i].val.d.n;
```
richm, line 925: What is the default value for `cs.bWorkAroundJournalBug`?  Where is it set?

jvymazal: My understanding is that default value for binary arguments is `off`, unless set otherwise (see for example other binary arguments).

rgerhards: actually depends on the initialization. The journal plugins use the old-style cs. structure, which is something we need to change at some point in time but so far I didn't care much about it. I'll also do a final review on the code, but check for where the other settings are initialized (well, they should, maybe it's missing altogether).

jvymazal: Well, when I look on the other binary arguments I do not see explicit initialization in any of them (only set on them is from the params parsing), but as implementation wise they are 'static int' they are guaranteed to be zero by standard (unless I understood current code wrong).

rgerhards: yeah... that's actually the case. But I am not sure if it is guaranteed by the standard. At least in the past there were some platforms that did not zero out that memory. They may long be history...

From an "esay to read" PoV it may be better though to initialize them. If only it saves discussions ;-) But I understand that is something you inherited. There are quite some things in that module that are a bit diffrent than in the rest of rsyslog...

jvymazal: I have no problem adding the explicit initialization for the sake or readability (will look to other modules for examples). However `static` zero-initialization *is* part of the C standard, see for example [here](http://www.open-std.org/jtc1/sc22/WG14/www/docs/n1256.pdf#%5B%7B%22num%22%3A272%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C-27%2C816%2Cnull%5D) - point 10. Of course that may be minus some non-standard compliant compilers :-).

plugins/imjournal/imjournal.c
```diff
@@ -776,6 +784,7 @@ CODESTARTbeginCnfLoad
 	cs.iDfltFacility = DFLT_FACILITY;
 	cs.bUseJnlPID = -1;
 	cs.usePid = NULL;
+	cs.bWorkAroundJournalBug = 0;
```
jvymazal, line 787: @richm , @rgerhards added the explicit initialization here. If there is some new config structure I could do that as a separate PR after this is merged...

plugins/imjournal/imjournal.c
rgerhards: @jvymazal thx, I hadn't the reference at hand.

rgerhards:
thx for everyone's help and persistence! If @portante tells that the performance implications are negligible, we can (and should) switch the default. And of course revert should the journal team ever fix that bug.

PR #2545, opened 2018-03-15
omfile: Fix wrong order of arguments for errmsg

hrak:
This small PR fixes two error messages in omfile that had the wrong order of arguments.


rgerhards:
Thx for the PR, much appreciated.

Note: I am currently working on CI, there may be temporary unrelated failures. I'll baby-sit them. Obviously, CI should not fail with this PR ;-)

PR #2551, opened 2018-03-17
QA: make lgtm.com analyze more modules

rgerhards:


rgerhards:
merging without full CI, as only lgtm.yml file is changed -- want to have the changes in as quickly as possible.

PR #2554, opened 2018-03-18
fix build issues on Solaris with gcc

rgerhards:


rgerhards:
looks like we need to work a bit more on this -- I seem to overlook some solaris buildbot slave specialities ;-)

rgerhards:
while this does not totally solve all isuses, it's a step forward - so I merge. Let's do more work with a separate PR.

Issue #1296, opened 2016-12-01
WorkDirectory set to non-existing directory creates imfile-state on /


mostolog:
Using 8.23 with this config seems to be enough to replicate this.

```
global(
	workDirectory="/home/user/this_doesnt_exists"
)
input(type="imfile" file="/logs/apps/app1.log" tag="group1/app1" addMetadata="on" readTimeout="5" PersistStateInterval="1")
```

Log shows:
> rsyslogd: $WorkDirectory: /home/user/this_doesnt_exists can not be accessed, probably does not exist - directive ignored [v8.23.0 try http://www.rsyslog.com/e/2181 ]

So perhaps this is intended behavior and it's only a doc issue. Just let me know.

rgerhards:
We need to check if this problem still persists. IIRC it was solved.

rgerhards:
Testing brings up that if we enforce this, it may affect existing installations. Maybe we should just go with a warning. Need to think a bit about this. See also https://github.com/rsyslog/rsyslog/issues/2558, which I think really makes sense.

davidelang:
loud warning, otherwise we break 'working' configs.

yes it's a surprise to find them in /, but it's better than not working at all.

David Lang

rgerhards:
@davidelang thx for the confirmation, that's exactly what concerned me.

PR #2557, opened 2018-03-18
imfile: emit error on startup if no working directory is set

rgerhards:
When the work directory has not been set or is invalid, state files
are created in the root of the file system. This is neither expected
nor desirable. We now complain loudly about this fact. For backwards
compatibility reasons, we still need to support running imfile in
this case.

closes https://github.com/rsyslog/rsyslog/issues/1296


rgerhards:
Obviously, we need to find a somewhat different approach ;-) [CI fails due to many tests not provinding a workdir...]

rgerhards:
changed the scope of this commit/PR to now only emit an error message. Aborting is not appropriate (see updated initial message).

Issue #2403, opened 2018-01-17
testbench: omprog tests sometimes fails


rgerhards:
failures occur most often on CentOS/Fedora. This could be an indication of a real problem.

rgerhards:
the real culprit seems to be the wait time used when signaling rsyslog. This looks like it is just a testbench issue, which happens to occur on some systems with "matching timing" whatever that means. "Slow" would be the wrong word, as some of the slaves that experience this are quite fast. But I can create the issue immediately if I reduce the sleep time.

rgerhards:
The test needs also to cleanup the omprog children when it errors out - else further test will hang.

rgerhards:
closing this as it seems to be solved. should problems re-occur, will re-open.

rgerhards:
partially reappeared, reopening

rgerhards:
It looks like some of the tests are still racy. Cause continous trouble with CI. Will temporarily disable them. Maybe some need really go away, see https://github.com/rsyslog/rsyslog/pull/1992#issuecomment-374010264

PR #2559, opened 2018-03-19
testbench: disable some racy omprog tests

rgerhards:
these test are still instable and some concerns about their
validity at all have been raised. We now disable them to make
the testbench more reliable. Root cause needs to be investigated
and tests finally fixed or removed.

see also https://github.com/rsyslog/rsyslog/issues/2403


rgerhards:
CI failure due to another test bug ... so I finally merge this to break the potential endless cycle ;-)

PR #2561, opened 2018-03-19
imptcp: increase UNIX socket backlog for performance under heavy load

mgraff:
Increase the UNIX socket backlog to the same value as TCP socket.
I encountered performance issues under heavy load with a backlog set to 5.

### DIFF
--- a/plugins/imptcp/imptcp.c
+++ b/plugins/imptcp/imptcp.c
@@ -441,7 +441,7 @@ static rsRetVal startupUXSrv(ptcpsrv_t *pSrv) {
 		ABORT_FINALIZE(RS_RET_ERR_CRE_AFUX);
 	}
 
-	if (listen(sock, 5) < 0) {
+	if (listen(sock, 511) < 0) {
 		errmsg.LogError(errno, RS_RET_ERR_CRE_AFUX, "imptcp: unix socket listen error");
 		ABORT_FINALIZE(RS_RET_ERR_CRE_AFUX);
 	}
--- a/plugins/imptcp/imptcp.c
+++ b/plugins/imptcp/imptcp.c
@@ -131,6 +131,7 @@ struct instanceConf_s {
 	int bSuppOctetFram;		/* support octet-counted framing? */
 	int bSPFramingFix;
 	int iAddtlFrameDelim;
+	int socketBacklog;
 	sbool multiLine;
 	uint8_t compressionMode;
 	uchar *pszBindPort;		/* port to bind to */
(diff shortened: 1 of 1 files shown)


rgerhards:
@mgraff many thanks for the PR. Much appreciated. 

I just wonder if it would make sense to have a config option for this - or am I overdoing here? @deoren @davidelang what do you think?

deoren:
> @rgerhards: I just wonder if it would make sense to have a config option for this - or am I overdoing here? @deoren @davidelang what do you think?

I'm personally always in favor of good defaults ("principle of least surprise" as I heard it referred to as recently) with the option to override them via a user-configurable parameter. On this topic though, I'm afraid that I don't know enough to really contribute, so I'll have to defer to others.

Even so, I turned to Google and found [this](http://veithen.github.io/2014/01/01/how-tcp-backlog-works-in-linux.html):

> To conclude, let’s try to see why the design choice made by Linux would be superior to the traditional BSD implementation. Stevens makes the following interesting point:
> 
>
> >  The backlog can be reached if the completed connection queue fills (i.e., the server process or the server host is so busy that the process cannot call accept fast enough to take the completed entries off the queue) or if the incomplete connection queue fills. The latter is the problem that HTTP servers face, when the round-trip time between the client and server is long, compared to the arrival rate of new connection requests, because a new SYN occupies an entry on this queue for one round-trip time. […]
> > 
> > The completed connection queue is almost always empty because when an entry is placed on this queue, the server’s call to accept returns, and the server takes the completed connection off the queue.
>
> 
> The solution suggested by Stevens is simply to increase the backlog. The problem with this is that it assumes that an application is expected to tune the backlog not only taking into account how it intents to process newly established incoming connections, but also in function of traffic characteristics such as the round-trip time. The implementation in Linux effectively separates these two concerns: the application is only responsible for tuning the backlog such that it can call accept fast enough to avoid filling the accept queue); a system administrator can then tune /proc/sys/net/ipv4/tcp_max_syn_backlog based on traffic characteristics.

The change in this PR sounds like it falls into *the application is only responsible for tuning the backlog* concern, while the latter concern is handled by the sysadmin tuning `/proc/sys/net/ipv4/tcp_max_syn_backlog` directly.

Would the proposed parameter be specific to this module or the TCP traffic handling behavior for rsyslog as a whole?

davidelang:
I am also in favor of good defaults, in this case you don't want too much space 
in the backlog as that makes it possible to loose logs, so this is something to 
make a tunable.

David Lang

rgerhards:
giving @davidelang's comment, I move the release target to 8.35. @mgraff would you like to add the config parameter? If not, no problem at all. If I don't hear back within 1 week, I'll add it myself.

mgraff:
@rgerhards I implemented a new config parameter to control the socket backlog.
It is used by TCP and UNIX sockets, so both have now the same default value (5).

rgerhards:
@thx a ton! - I know I am a pest, but would you also be willing to update the doc at https://github.com/rsyslog/rsyslog-doc ?

mgraff:
@rgerhards I submitted the requested documentation in this PR : https://github.com/rsyslog/rsyslog-doc/pull/643

rgerhards:
Again, thanks a lot for your hard work. Happy to merge!

mgraff:
You're welcome, it was not that hard...
Happy to have contributed ;)

PR #2562, opened 2018-03-19
rscript: make function modules loadable

jgerhards:
function modules add functions to rainerscript dynamically,
change http_request into such a module (enabled by default).
This module can be used as a sample for future function modules.


rgerhards:
Merging this before doc has fully updated, because this lays ground for further work.

@jgerhards please update rsyslog-doc project accordingly

PR #2569, opened 2018-03-20
omkafka: fix 'instanceData' has no member named 'mutCurrPartition'

bruce87en:


rgerhards:
Thx for the PR. The issue is actually a bit more complicated. The root cause of the problem is that a conditional definition (macro) for the mutex is missing. The init and destruct is actually available. This also points out that the testbench does not build all components on systems without atomic operations. I will see how I can add that.

Given the fact that your PR fixes the core issue, I tend to merge it, but later on change the code so that it is fully inline with the regular rsyslog atomic op methods. Any concerns in that regard?

PR #2570, opened 2018-03-20
Make configuration setting SystemLogRateLimitBurst unsigned int

KaleviKolttonen:
rsyslog uses unsigned short for configuration setting
SystemLogRateLimitBurst. Being just 16 bits, unsigned short cannot
hold values bigger than 65535. On our LDAP server rsyslog misbehaved with
SystemLogRateLimitBurst being bigger than 65535. Make it unsigned int to
accept bigger values.


rgerhards:
thx, good catch!

Issue #2231, opened 2017-12-15
[FEAUTURE REQUEST] use some hash for imstate-file file name


Dieken:
Currently imfile modue uses "imstate-file:-full-path-name" as state file name,  truncation happens for very long path name (seems max length of path name can be 4096 bytes).

It's better to use something like "imstate-file:SHA256" for the state file name,  imfile can double check the hash by calculating it from file path in the state file.

rgerhards:
thanks. That is a good suggestion, but we must think a bit about backwards compatibility. To me personally, it's not an easy decision, but I tend towards your suggestion.

rgerhards:
@kortemik @janmejay @davidelang @alorbach @taotetek @deoren ... and whoever I forgot -- what do you think? It also does not make immediately clear which state file this is for. On the other hand, it removes all potential error cases (path length, char escaping).

I set the milestone to 8.33, so we should decide in that time frame.

deoren:
@rgerhards Is there a limit on the name parameter for imfile that would keep it from approaching the path name length restriction? Perhaps combine that value with the hash to build state files?

davidelang:
since filenames can only be 255 bytes long, and paths can be 4096 bytes long, 
this is a problem.

But I don't like loosing the easy ability to see what file the statefile is for.

To preserve backwards compatibility, I would only use a hash if the full path 
name results in a filename that's 'too long'. That way all existing files 
continue to work and we only use the hash when we can't use the filename (and I 
would append the tail of the path name after the hash to try and have as much 
chance as possible of being able to identify what file it's for)

Dieken:
@rgerhards  the imstate-file is created by imfile module: http://www.rsyslog.com/doc/v8-stable/configuration/modules/imfile.html#state-files

@deoren,  I did find imfile module truncated file name "/var/spool/rsyslog/imstate-file:-a-very-long-path", I don't know whether imfile has a smaller filename length limit than default OS limit 255.

@davidelang  Usually people won't check /var/spool/rsyslog often, Rsyslog should just work, but if necessary, we can easily grep the path name from the imstate-files.

@rgerhards @davidelang About the backward-compatibility, I like more this logic:
```
open-imstate-file(a-very-long-path) {
     a-possibly-truncated-file-name = "imstate-file:a-very-long-path-truncated-to-242"; # length of "imstate-file:" is 13.
     if (existed("/var/spool/rsyslog/a-possibly-truncated-file-name")) {
[...]

PR #2592, opened 2018-03-28
Trying to reproduce errors in PR #2465

rgerhards:
**DO NOT MERGE** - delete after testing

see #2465

PR #2593, opened 2018-03-29
imfile: re-enable support for readind old-style state files

rgerhards:
see also https://github.com/rsyslog/rsyslog/issues/2231#issuecomment-376862280

### DIFF
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -111,6 +111,7 @@ static struct configSettings_s {
 
 struct instanceConf_s {
 	uchar *pszFileName;
+	uchar *pszFileName_forOldStateFile; /* we unfortunately needs this to read old state files */
 	uchar *pszDirName;
 	uchar *pszFileBaseName;
 	uchar *pszTag;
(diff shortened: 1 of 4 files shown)


sjvs:
Hi @rgerhards: can you point me at the lgtm failure that you report in d8504e4? It's not immediately obvious to me why enabling mmkubernetes shouldn't work. As far as I can see on lgtm.com the C/C++ analysis after enabling mmkubernetes worked fine:

Here's the info for the original merge commit: https://lgtm.com/projects/g/rsyslog/rsyslog/rev/1c78e8c59ef98d4934d9cb19959c4fa86dd9eff2

And here's the analysis status for a more recent commit: https://lgtm.com/projects/g/rsyslog/rsyslog/rev/c4577375c21eac555592e07c94a8454ad424965d

rgerhards:
@sjvs thx again for your help! I was a bit stunned as well. The commit was amended, so I do not know how to get the previous URL for that one. If you see it inside the system, it is the previous PR on this branch. If you also do not see it, I can craft a new testing PR based on master. Pls let me know.

sjvs:
I can see that some commits in #2592 were analysed by lgtm.com, but I can't find any analysis or build failures in our logs. Are you sure the problem wasn't related to the branch being unmergeable or something like that?

I'll keep on digging and will let you know if I found anything! In the meantime it may be worth re-enabling mmkubernetes in this PR to see whether the analysis succeeds.

rgerhards:
@sjvs strange - now it works. I know for sure it did not for at least two runs in a row :-S I'll keep it enabled and ping you should the problem re-occur. 

Ahh, I see one issue with mmkubernetes: it does not check for liblognorm but uses it. It looks like some of the lgtm.com build VMs have 2.0.3+ available, some not (transition process?)

richm:
@rgerhards I've been mostly using liblognorm 2.0.2 (RHEL7) - to build with 2.0.2, you have to set `-DNO_LOADSAMPLESFROMSTRING=1` in CFLAGS since `ln_loadSamplesFromString` is only available in 2.0.3 and later.  I could not find another way to automatically detect the presence of this function, or the liblognorm version, at compile time.  I suppose I should have added an AC_CHECK_FUNC or something like that.

as for libfastjson, I've been using libfastjson-0.99.4 (RHEL7)

PR #2594, opened 2018-03-29
Merge-PR for PR 2490

rgerhards:
see #2490


rgerhards:
did not work out, see base PR

PR #2596, opened 2018-03-29
"fix" some Coverity Scan cosmetic issues

rgerhards:


rgerhards:
freebsd CI failure is unreleated -- if it reoccurs in the future, we need to look into this as a potential bug.

PR #2601, opened 2018-03-31
omprog: corrected fd leak with confirmMessages=on

jsiwrk:
There was a fd leak in the feedback feature I added in v8.31.0 (PR #1753). The leak occurred when omprog was used with the `confirmMessages=on` setting and no `output` setting. One fd was leaked every time the external program was started.

I have found this when reviewing the omprog tests. A later PR I'm currently preparing will include a regression test for this case. Meanwhile, the correction should be merged ASAP.


rgerhards:
The freebsd CI failure is unrelated, will look into it.

rgerhards:
I think I fixed the freebsd issue, but I merge without rebase as this really seems unnecessary ;-)

PR #2604, opened 2018-04-01
fix issues with atomic operations

### LINKED ISSUE #2612: wrong configure switch for atomic ops
configure.ac in 8.34.0 contains
```
AC_ARG_ENABLE(regexp,
        [AS_HELP_STRING([--enable-atomic-operations],[Enable atomic operation support @<:@default=yes@:>@])],
```

I'm pretty sure the ```regexp``` is a C&P mistake

rgerhards:
... on platforms that do not support them natively. None of them actually break building except when warning-free compile is enforced.

closes #2612

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -254,7 +254,7 @@ AC_LINK_IFELSE([AC_LANG_PROGRAM(, return __builtin_expect(main != 0, 1))],
 # speeds up processig.
 # note that when automic operations are enabled but not available, they
 # will silently NOT be used!
-AC_ARG_ENABLE(regexp,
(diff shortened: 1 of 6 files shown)


rgerhards:
@janmejay If you have second, I would appreciate if you could check the failed Travis build. An obvious cure is to cast away the sign difference, but I have not yet checked if that is problematic. If you do not have time, I'll do that later on. But if you know, that would be much appreciated.

janmejay:
@rgerhards fixed here: https://github.com/rgerhards/rsyslog/pull/40 ( over #2604 )

PR #2617, opened 2018-04-04
add tests for rainerscript functions

PascalWithopf:
Tests for cnum(), int2hex() and substring() created.

### DIFF
--- a/tests/rs-int2hex.sh
+++ b/tests/rs-int2hex.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+# add 2018-04-04 by Pascal Withopf, released under ASL 2.0
+. $srcdir/diag.sh init
+. $srcdir/diag.sh generate-conf
+. $srcdir/diag.sh add-conf '
+module(load="../plugins/imtcp/.libs/imtcp")
+input(type="imtcp" port="13514")
+
+set $!var = int2hex(61);
+set $!var2 = int2hex(-13);
+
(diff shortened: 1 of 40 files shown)


tests/rs-int2hex.sh
```diff
@@ -0,0 +1,27 @@
...
+
+template(name="outfmt" type="string" string="-%$!var%- -%$!var2%-\n")
+:syslogtag, contains, "tag" action(type="omfile" template="outfmt"
+			         file="rsyslog.out.log")
+'
+. $srcdir/diag.sh startup
+. $srcdir/diag.sh tcpflood -m1 -M "\"<129>Mar 10 01:00:00 172.20.245.8 tag: msgnum:1\""
+. $srcdir/diag.sh shutdown-when-empty
+. $srcdir/diag.sh wait-shutdown
+echo '-3d- -fffffffffffffff3-' | cmp rsyslog.out.log
```
rgerhards: cmp needs a dash as first argument, else the test will fail on someplatforms, e.g. BSD

rgerhards:
canceling out buildbot runs to save ressources - resubmit is needed in any case

anatolyborodin:
Hi @PascalWithopf !

Unfortunately, the test for `substring()` doesn't take into account the fact that `$msg` starts with a whitespace, so `$msg = " msgnum:1"`, and the bug #3344 goes unnoticed.

I would also propose to use a constant string instead of `$msg` to make the test more obvious.

PR #2618, opened 2018-04-04
Fix version

rgerhards:
DO NOT MERGE - accidently contains a different commit


rgerhards:
I let this run as the testbench results are useful in any case

PR #2629, opened 2018-04-07
omprog: refactor tests, fix child closing issues

### LINKED ISSUE #2403: testbench: omprog tests sometimes fails

### LINKED ISSUE #2599: omprog not properly closing child process when signalOnClose=on
There are three issues with omprog when it is used with the `signalOnClose=on` setting:
* an unresponsive child is not killed within the supposed timeout of 5 secs
* a message "child-process wasn't reaped within timeout" is continously shown in the output
* EOF is not sent to the child

I have found these issues while reviewing the existing omprog tests to try to resolve issue #2403. Below I give more details and my analysis for each of these issues.

When `signalOnClose=on`, omprog sends a TERM signal to the child process and waits for the child to terminate within a ([hardcoded](https://github.com/rsyslog/rsyslog/blob/v8.33.0/plugins/omprog/omprog.c#L64)) timeout of 5 seconds. After this timeout, omprog should supposedly kill the child. However, this does not work.

It is easy to reproduce by executing the existing test `omprog-cleanup-when-unresponsive.sh` with an increased main queue shutdown timeout, e.g. 1 minute: `main_queue(queue.timeoutShutdown="60000")`. The test terminates after 1 minute, not after 5 seconds. By forcing the test to fail, the output in `test-suite.log` shows that omprog finally kills the child, but AFTER the worker thread has been canceled because of the main queue timeout: see [test-suite.log](https://github.com/rsyslog/rsyslog/files/1865601/test-suite.omprog-cleanup-unresponsive.with.increased.queue.timeoutShutdown.log).

My analysis: the auxiliary thread controlling the timeout is not able to [interrupt the waiting thread](https://github.com/rsyslog/rsyslog/blob/v8.33.0/plugins/omprog/omprog.c#L387). The code seems correct, but apparently the signal is being handled by another thread. I have not managed to make it work using an interrupted wait.

Besides, during the shutdown timeout, omprog continously outputs the following message. (Note: I had to truncate the file, since its size was around 200 Mb because of this repeated message.)
```
rsyslog internal message (3,-2428): omprog: child-process wasn't reaped within timeout (5000 ms) preparing to force-kill. [v8.34.0.master try http://www.rsyslog.com/e/2428 ]
rsyslog internal message (3,-2428): omprog: child-process wasn't reaped within timeout (5000 ms) preparing to force-kill. [v8.34.0.master try http://www.rsyslog.com/e/2428 ]
rsyslog internal message (3,-2428): omprog: child-process wasn't reaped within timeout (5000 ms) preparing to force-kill. [v8.34.0.master try http://www.rsyslog.com/e/2428 ]
rsyslog internal message (3,-2428): omprog: child-process wasn't reaped within timeout (5000 ms) preparing to force-kill. [v8.34.0.master try http://www.rsyslog.com/e/2428 ]
...
```

The reason for this repeated message is an absolute timeout not being recalculated [in a pthread_cond_timedwait call](https://github.com/rsyslog/rsyslog/blob/v8.33.0/plugins/omprog/omprog.c#L381).

Regarding the third issue: omprog waits for the child to terminate [without previously closing the pipe](https://github.com/rsyslog/rsyslog/blob/v8.33.0/plugins/omprog/omprog.c#L507-L526), so the child does not receive an EOF. The child should be given an opportunity to terminate on EOF, even if it ignores the TERM signal.

### Environment
- rsyslog version: 8.34.0
- platform: Ubuntu 16.04

### LINKED ISSUE #2600: omprog not waiting for the child process to terminate when signalOnClose=off
This issue is related to bug #2599, but it is a different issue (maybe an enhancement).

When omprog is used with the `signalOnClose=off` setting, it does not wait for the child process to terminate after closing the pipe. In principle this is not a problem since [the rsyslogd main loop will reap it](https://github.com/rsyslog/rsyslog/blob/8e71447e04f34b585c96f477d1771b9a0f79f166/tools/rsyslogd.c#L1799). However, if rsyslog shuts down before the child terminates, the child will become a zombie. I can reproduce this easily with omprog tests.

I think omprog should wait for the child during a timeout, as it does with `signalOnClose=on`. This timeout should ideally be configurable. And the same configured timeout would apply to `signalOnClose=on`.

If the timeout expires, then omprog could force-kill the child, as it does with `signalOnClose=on`. Also whether the child is killed or not should ideally be configurable, and apply also to `signalOnClose=on`.

So I propose to correct this issue by **adding two new omprog settings**:
* `closeTimeout`: how long must omprog wait for the child to terminate
* `killUnresponsive`: whether the child should be terminated with SIGKILL after `closeTimeout`

These settings would apply to both `signalOnClose=on` and `signalOnClose=off`. The value of `signalOnClose` indicates whether omprog must send a TERM signal to the child (apart from a pipe EOF) before waiting for it to terminate.

To keep backwards compatibility with `signalOnClose=on`, `closeTimeout` would have a default value of 5000 millis (this value is currently hardcoded in omprog), and `killUnresponsive` a default value of `on`.

Keeping full backwards compatibility with `signalOnClose=off` seems not necessary, since waiting for the child to terminate should be the default behavior, I think. So, I propose that the default value of `closeTimeout` is also 5000 millis in this case. However, if the process does not terminate within the timeout, sending SIGKILL without having previously sent SIGTERM doesn't seem good practice. Therefore, I propose that the default value of `killUnresponsive` is the value of `signalOnClose`.

Therefore, the only change in behavior (with regard to an existing configuration that is not modified to make use of these new parameters) will be that, with `signalOnClose=off`, omprog will now wait for the child process to finish during at most 5000 millis. This will avoid zombie processes in most cases, while maintaining compatibility with existing configurations.

jsiwrk:
Refactor omprog tests. Fix sync issues in these tests by
using the feedback mode (confirmMessages=on) to synchronize
the test with the external program. Closes #2403 (I hope)

Fix omprog not properly closing child process when
signalOnClose=on. Needed for the new tests. Closes #2599

Fix omprog not waiting for the child process to terminate
when signalOnClose=off. Needed for the new tests. Closes #2600

Close all file handles before executing the child even when
valgrind is enabled (--enable-valgrind). Needed for the
new tests.

Fix memory leak when the xxxTransactionMark parameters were
used.


jsiwrk:
Sorry for such a big PR. I have combined the big test refactor with the fixes for various (minor) issues, because to make the tests non-racy I needed to correct the issues, and at the same time I didn't want to correct the issues without adding the tests for them.

The new tests are strongly based on @janmejay's existing tests. However I have renamed them to better reflect the main purpose of each test, while at the same time adding some checks. Since rsyslog only has functional tests, there is a compromise in trying to keep each test functionally focused while at the same time trying to check as many special cases as possible (it would be unmanageable to define a separate functional test for every possible special case to be covered, I think).

jsiwrk:
@rgerhards : I have removed the conditional compilation directive in this code fragment:
```
	/* we close all file handles as we fork soon
	 * Is there a better way to do this? - mail me! rgerhards@adiscon.com
	 */
#	ifndef VALGRIND /* we can not use this with valgrind - too many errors... */
	for(i = 3 ; i <= 65535 ; ++i) {
		close(i);
	}
#	endif
```
Without this code, the child process inherits the open fd's from the parent, including the open ports. This caused the `omprog-close-unresponsive` tests to leave a dangling child process (an sleep) which kept the imdiag port opened, and in turn this caused the next test in the Travis CI to fail because the port was in use (with the message `Could not create tcp listener, ignoring port 13500 bind-address (null)`). This drove me crazy because I could'nt reproduce it locally, until I realized the Travis CI is compiled with `--enable-valgrind` which caused the above code to be omitted.

Because I haven't seen any problem with valgrind related to the above code, I have removed the condition. Do you remember why it was added? May be it gave problems with earlier valgrind versions...?

rgerhards:
> Because I haven't seen any problem with valgrind related to the above code, I have removed the condition. Do you remember why it was added?

The issue is that in many test valgrind issues ~64K lines of warning messages (invalid handle access). That has two problems:
* makes debug output hard to read
* broke Travis output size limit (IIRC)

The real cure would be to close only open files, but there seems to be no good portable way to do so.

rgerhards:
adding "in-depth-review" label as I need to think about the implications of removing the valgrind guard.

@jsiwrk maybe we should consider dropping valgrind for the time being? Well, not the best idea either, I know...

jsiwrk:
@rgerhards : It seems valgrind only complains about closing invalid file descriptors if the program is closing fds above the maximum allowed number of open fds for the process. (Because I run rsyslog within a docker container, in which this limit is unrestricted by default, I wasn't seeing the warnings. By setting `ulimit -n 1024` in `diag.sh` I can reproduce the ~64K lines of warnings you mentioned.)

I have updated the PR with a solution: close all fds up to `sysconf(_SC_OPEN_MAX)`. This avoids the valgrind warnings, while at the same time improving efficiency (with respect to blindly closing 65K fds). I have kept a maximum of 65K fds as a fallback in case sysconf fails, or returns more than this value (to avoid closing a lot of fds when ulimit -n is unrestricted). Also, it seems that [valgrind reserves the last 10 fds for itself](http://valgrind.10908.n7.nabble.com/Invalid-file-descriptor-1019-td43056.html), so we must not close these to completely avoid the warnings.

BTW, this way of closing all fds is very similar to what it is done in [rsyslogd.c](https://github.com/rsyslog/rsyslog/blob/5d2cbe5ad00b991ae1c1c7ab54fb7d6d05429a75/tools/rsyslogd.c#L379), but using `sysconf` instead of `getdtablesize` (which isn't POSIX, it seems, so the former should be more portable). Do you know why `sysconf` isn't used there? Maybe it is not available in all platforms rsyslog is being compiled for? (Let's see if the CI complains.)

rgerhards:
> BTW, this way of closing all fds is very similar to what it is done in rsyslogd.c, but using sysconf instead of getdtablesize (which isn't POSIX, it seems, so the former should be more portable). Do you know why sysconf isn't used there? Maybe it is not available in all platforms rsyslog is being compiled for? (Let's see if the CI complains.)

I think this is not really done with intent - the valgrind part was more or less a hack to get things going (and we only have it on Linux). Let's see. Having said that, it just occurs to me that it would be pretty ok in this space to use Linux specifics, e.g. ```/proc```!

The CI failures are due to a change that we needed to apply. Please do a rebase to current master, and we'll get rid of the many errors that originate back to that.

rgerhards:
> (it would be unmanageable to define a separate functional test for every possible special case to be covered, I think).

late comment: yup ;-) Combinatorical explosion working here. As a side-note, we tried to add unit tests, but we cannot get them to link properly. That's why still none exists. Several folks, including from RH, tried to make the link work, but autotools and library link order in the linker seems to be pandorra's box..

PR #2630, opened 2018-04-09
testbench: switch external testbench URL

rgerhards:
this becomes necessary as the main site switches to https


rgerhards:
I know we should better have a local http server in the testbench, but someone would need to do that... ;-)

PR #2635, opened 2018-04-11
Several fixes for mmkubernetes

richm:
Changes:
- add tests for mmkubernetes
- clean up sample mmkubernetes file
- fix indentation
- check for Kubernetes response code
- only examine filename if non-null and length is not 0
- add default kubernetesurl https://kubernetes.default.svc.cluster.local:443
- fix broken lnrule for CONTAINER_NAME
- add pkg check for lognorm >= 2.0.3 for mmkubernetes for
  HAVE_LOADSAMPLESFROMSTRING
- fix lnrules

### DIFF
--- a/tests/mmkubernetes-basic-vg.sh
+++ b/tests/mmkubernetes-basic-vg.sh
@@ -1,0 +1,89 @@
+#!/bin/bash
+# added 2018-04-06 by richm, released under ASL 2.0
+#export RSYSLOG_DEBUG="debug"
+. $srcdir/diag.sh init
+
+testsrv=mmk8s-test-server
+python ./mmkubernetes_test_server.py 18443 rsyslog${testsrv}.pid rsyslogd${testsrv}.started > mmk8s_srv.log 2>&1 &
+BGPROCESS=$!
+. $srcdir/diag.sh wait-startup $testsrv
+echo background mmkubernetes_test_server.py process id is $BGPROCESS
+
+pwd=$( pwd )
+. $srcdir/diag.sh generate-conf
+. $srcdir/diag.sh add-conf '
+module(load="../plugins/imfile/.libs/imfile")
+module(load="../plugins/mmjsonparse/.libs/mmjsonparse")
+module(load="../contrib/mmkubernetes/.libs/mmkubernetes" token="dummy" kubernetesurl="http://localhost:18443"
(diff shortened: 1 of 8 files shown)


richm:
checks passed - please review - once the review is done I'll squash all of these commits into one (or I can do it now, before review, if preferable)
@portante @nhosoi

richm:
@jvymazal

richm:
corresponding docs update https://github.com/rsyslog/rsyslog-doc/pull/647

tests/mmkubernetes-basic-vg.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
+# added 2018-04-06 by richm, released under ASL 2.0
+#export RSYSLOG_DEBUG="debug"
+. $srcdir/diag.sh init
+
+python ./mmkubernetes_test_server.py 18443 > mmk8s_srv.log 2>&1 &
+BGPROCESS=$!
```
rgerhards: do we need a check here that the startup really was completed? I ask because we have some other spots inside the testbench where this causes startup races, especially on slow machines. side-note: we may want to consider adding some generic plumbing to the testbench to do this - so maybe we should just file an enhancement issue for now (if this is a problem at all).

richm: I was able to use the plumbing in diag.sh - how's this?

rgerhards: smart! Looks very good - thx.

richm:
@rgerhards shall I squash into 1 commit and resubmit?  or is separate commits ok?

davidelang:
There are enough commits that fix prior commits that either squashing, or 
re-ordering and squashing some commits is probably a good thing to do.

richm:
et voila

rgerhards:
@richm do you think I should merge? I saw you asked for additional feedback, thus I ask.

richm:
@rgerhards I think all of my requested reviewers are very busy now - @davidelang any comments?

rgerhards:
How about this: I think it looks good at least to me. It improves things, e.g. by introducing tests and fixing some known issues. There is a risk not everything is addressed and we may have some regression.

All in all, I would say "let's merge" and correct anything we overlooked in a later increment. Would you feel comfortable with that?

richm:
@rgerhards works for me - thanks!

PR #2651, opened 2018-04-17
omfwd: add support for bind-to-address for UDP

mikebcom:
To allow the same source address to be used regardless of the egress
interface taken, an option is added for an address to bind the datagram
socket to. Similarly to imudp, it is necessary to add an ipfreebind
option which is set by default, so as to avoid an excess of errors at
startup before the network interface has come up. This enhancement
allows a usecase on networking devices, by which a source interface
that is typically a loopback is specified, on which an address to bind
to is configured. This is so that the same source address is used for
all packets from rsyslog. Support for the TCP case can be added later.

Signed-off-by: Mike Manning <mmanning@att.com>

### DIFF
--- a/tools/omfwd.c
+++ b/tools/omfwd.c
@@ -84,6 +84,7 @@ typedef struct _instanceData {
 	permittedPeers_t *pPermPeers;
 	int iStrmDrvrMode;
 	char	*target;
+	char	*address;
 	char	*device;
 	int compressionLevel;	/* 0 - no compression, else level for zlib */
 	char *port;
@@ -96,6 +97,7 @@ typedef struct _instanceData {
 	int iKeepAliveProbes;
 	int iKeepAliveTime;
 	uchar *gnutlsPriorityString;
+	int ipfreebind;
 
 #	define	FORW_UDP 0
 #	define	FORW_TCP 1
@@ -164,6 +166,7 @@ static struct cnfparamblk modpblk =
 /* action (instance) parameters */
 static struct cnfparamdescr actpdescr[] = {
 	{ "target", eCmdHdlrGetWord, 0 },
+	{ "address", eCmdHdlrGetWord, 0 },
 	{ "device", eCmdHdlrGetWord, 0 },
 	{ "port", eCmdHdlrGetWord, 0 },
 	{ "protocol", eCmdHdlrGetWord, 0 },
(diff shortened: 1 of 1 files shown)


tools/omfwd.c
```diff
@@ -96,6 +97,10 @@ typedef struct _instanceData {
 	int iKeepAliveProbes;
 	int iKeepAliveTime;
 	uchar *gnutlsPriorityString;
+	/*  0 means:  IP_FREEBIND is disabled
+	1 means:  IP_FREEBIND enabled + warning disabled
```
rgerhards: I would prefer #define's here for readability. It also looks like the comment is in error ;-)

tools/omfwd.c
```diff
@@ -863,9 +871,11 @@ static rsRetVal returnToOriginalNs(instanceData *const pData __attribute__((unus
 static rsRetVal doTryResume(wrkrInstanceData_t *pWrkrData)
 {
 	int iErr;
-	struct addrinfo *res;
+	struct addrinfo *res, *addr;
 	struct addrinfo hints;
 	instanceData *pData;
+	int bBindRequired = 0;
+	char *address;
```
rgerhards: can it be ```const char*``` ?

tools/omfwd.c
```diff
@@ -885,12 +895,26 @@ static rsRetVal doTryResume(wrkrInstanceData_t *pWrkrData)
 				pData->target, pData->port, gai_strerror(iErr));
 			ABORT_FINALIZE(RS_RET_SUSPENDED);
 		}
+		address = pData->target;
+		if(pData->address) {
+			/* The AF of the bind addr must match that of target */
+			hints.ai_family = res->ai_family;
+			hints.ai_flags |= AI_PASSIVE;
+			if((iErr = getaddrinfo(pData->address, pData->port, &hints, &addr)) != 0) {
+				errmsg.LogError(0, RS_RET_SUSPENDED,
```
rgerhards: you can drop the ```errmsg.``` prefix. We are currently in a project to gradually phase the old method out. Not a hard request, though.

tools/omfwd.c
```diff
@@ -1284,6 +1313,11 @@ CODESTARTnewActInst
 					"cannot be used with tcp transport -- ignored");
 		}
 	}
+
+	if(pData->address && (pData->protocol == FORW_TCP)) {
+		errmsg.LogError(0, RS_RET_PARAM_ERROR,
```
rgerhards: same here re ```errmsg.``` (see line 1258 for current use)

tools/omfwd.c
mikebcom: Done, the #defines are already in runtime/net.h and well-commented, so I have removed this comment altogether from the copy & paste I did from imudp.c cf original commit 5f12c9b5 there.

tools/omfwd.c
mikebcom: Done

tools/omfwd.c
mikebcom: Done

tools/omfwd.c
mikebcom: Done

rgerhards: thx!

rgerhards:
rsapbian CI failure is not real, it's caused by us switching to a different version of the build master (but I can't get rid of the seemingly failure)

rgerhards:
CI failures are caused by switch of buildbot version; so not pointing at real issue

rgerhards:
thx again for the PR!

PR #2653, opened 2018-04-18
fmhash: fix invalid table end entry

rgerhards:


rgerhards:
docker-ubuntu is experimental slave, fail expected

Issue #1782, opened 2017-09-20
Messages are stuck in rsyslog queue when using RELP delivery


jay7x:
I have following action configured in ruleset on API server. This ruleset receive messages from PHP process (`fwrite()` to `stream_socket_client()`-connected socket) via `imuxsock` input.

```
  action(type="omrelp" name="XXX_sock_relay"
    target="10.X.X.X"
    port="2510"
    tls="off"
    tls.compression="off"
    template="XXX_relay_msg"
    action.reportSuspensionContinuation="on"
    action.resumeRetryCount="-1"
    action.resumeInterval="10"
    queue.type="fixedArray"
    queue.size="250000"
    queue.dequeueBatchSize="1024"
    queue.workerThreads="4"
    queue.workerThreadMinimumMessages="50000"
    queue.spoolDirectory="/shop/logs/rsyslog"
    queue.fileName="XXX_sock"
    queue.maxFileSize="16m"
    queue.maxDiskSpace="2G"
    queue.highWaterMark="200000"
    queue.lowWaterMark="100000"
    queue.checkpointInterval="30000"
    queue.saveOnShutdown="on"
    queue.timeoutEnqueue="1"
  )
```

Few days ago I've started to investigate why this queue `XXX_sock_relay` always have 1000-4000 messages in it when other API's queues have < 10 messages usually.

I've realised this queue on every API server have few old messages stuck in it. Every few mins those messages are delivered to log collector. But those messages seems are not marked as delivered after! So they are remaining in queue to be delivered again and again. Delivery attempt happens every 10s.

I have saved tcpdump for this but it contains private info I don't like to put in public. So if you're interested in investigation of this case please let's contact directly.

I've switched partially to TCP delivery (omfwd/imptcp). There is no issues at all. So it's exactly about `omrelp`/`imrelp`.

CentOS 7 x86_64
rsyslog 8.27
librelp 1.2.14

mostolog:
I would bet we have seen similar issues in the past. Did you try deleting your queues? A fresh start could help solving the issue, although possible error may not be fixed.

jay7x:
New messages will appears stuck in queue in about 24h after fresh start.

deoren:
[...]

PR #2656, opened 2018-04-19
 imrelp: add parameter oversizeMode 

PascalWithopf:
Parameter defines the way oversized messages are handled.
Default is that messages are truncated.
Test was added for message truncation.


rgerhards:
> two 'nice to have'/enhancement requests (not blocking merging)
> 
> could this be moved from the relp code to the generic action() code so that it can apply to any type of action?

That's unfortunately not that easy. We need to handle it on the protocol layer. The core reason for having a max message size (in regard to relp protocol) is that we want to guard against malicious messages which have giantic size and may be misused to trigger all kinds of problems. Thus the RELP protocol specifies they should not be accepted. Changing that to truncation is one thing, always completely receiving them is quite another. 

I still intend to add such an option to librelp, together with a big warning note. Then imrelp can utilize that mode, if so configured by the user.

> 
> could you add a 'write to file' option?
> ideally using the same errorfile directive that omelasticsearch uses (so that eventually we can eliminate that code and unify it in action())
> if you can make it so that if errorfile !='' then default to 'write to file' mode unless otherwise specified that would be even better.

This needs to be done at the rsyslog/imrelp level. It requires above-mentioned change to librelp so that it accepts overly long messages.

rgerhards:
As an example for the security issue: attacker spams us with a continous stream of 2 GB messages of garbagges. relp accepts, and so we easily eat up all system memory quickly. That is why the RELP spec specifies such messages shall not be accepted.

davidelang:
On Fri, 20 Apr 2018, Rainer Gerhards wrote:

> As an example for the security issue: attacker spams us with a continous stream of 2 GB messages of garbagges. relp accepts, and so we easily eat up all system memory quickly. That is why the RELP spec specifies such messages shall not be accepted.

Ahh, I was missing that this was IMrelp, I was thinking this was talking about 
output, and there something like oversizemode and errorfile would be very good 
to have for all types of output

rgerhards:
> Ahh, I was missing that this was IMrelp, I was thinking this was talking about
> output, and there something like oversizemode and errorfile would be very good
> to have for all types of output

actually we need to limit it on both input and output. But the inputs usually need to be handled at the procotol level, so this is harder to do generically. Neverthelss, I'll work towards this, but at later stages.

davidelang:
On Fri, 20 Apr 2018, Rainer Gerhards wrote:

>> Ahh, I was missing that this was IMrelp, I was thinking this was talking about
>> output, and there something like oversizemode and errorfile would be very good
>> to have for all types of output
>
> actually we need to limit it on both input and output. But the inputs usually 
> need to be handled at the procotol level, so this is harder to do generically. 
> Neverthelss, I'll work towards this, but at later stages.

because I do so much parsing of my logs, but want to keep the original text 
around, I am commonly making my output significantly larger than the input. I've 
done some manual detection/work-arounds for this, but it would be _very_ nice 
for this to be a common feature of all action() calls. When I saw the topic show 
up, I didn't pay enough attention to notice it was inputs causing the problem, 
sorry.

PascalWithopf:
@deoren The issue from your comment:
https://github.com/rsyslog/rsyslog/issues/1782#issuecomment-352027757
with maxDataSize in imrelp not being set by maxMessageSize is solved with this commit: https://github.com/rsyslog/rsyslog/pull/2656/commits/57abebfea212a2f273540c471f252b067a1bb716

The problem was that the maxDataSize default was set before maxMessageSize was set and so the default of maxMessageSize was always used.

deoren:
> @PascalWithopf: @deoren The issue from your comment:
> #1782 (comment)
> with maxDataSize in imrelp not being set by maxMessageSize is solved with this commit: a94a414
> 
> The problem was that the maxDataSize default was set before maxMessageSize was set and so the default of maxMessageSize was always used.

Ah, many thanks for tracking that down and fixing it. I look forward to giving these changes a spin when I get some time.

PR #2657, opened 2018-04-19
Increased limit of concurrent FDs.

jvymazal:
This is for solving situations when dealing with many files and
also to keep aligned with systemd value (because of journal
integration).

### DIFF
--- a/rsyslog.service.in
+++ b/rsyslog.service.in
@@ -13,3 +13,7 @@ Restart=on-failure
 [Install]
 WantedBy=multi-user.target
 Alias=syslog.service
+
+# Increase the default a bit in order to allow many simultaneous
+# files to be monitored, we might need a lot of fds.
+LimitNOFILE=16384


jvymazal:
The systemd default can be found [here](https://github.com/systemd/systemd/blob/master/units/systemd-journald.service.in#L41).

jvymazal:
@rgerhards  do you want me to re-spin CI or are you willing to accept the flake? (travis on my rsyslog fork which is up-to-date with yours passed with this commit - [see here](https://travis-ci.org/jvymazal/rsyslog/builds/368635981))

Thanks

rgerhards:
@jvymazal I'll babysit CI - it looks like we still have some race. I'll let you know should anything require your attention.

and... thx for the PR! :-)

mbiebl:
Hi,

unfortunately this patch is incorrect. ```LimitNOFILE=``` has to be added to the ```[Service]``` section, not ```[Install]```

mbiebl:
Fwiw, I find the comment "also to keep aligned with systemd value (because of journal
integration)." a bit strange. What's that supposed to mean?

Why do we need to raise the nofile limit for the journal integration?

jvymazal:
Good spotting @mbiebl, made amends in regards of setting placement.

As for why we need it, in more complex setups watching many journald files and/or many files trough imfile wildcards  the current limit might not be enough.

rgerhards:
> As for why we need it, in more complex setups watching many journald files and/or many files trough imfile wildcards the current limit might not be enough.

The question is if this is for a very special use case. If so, then this should be done by user configuration. If it is generally required, then we should propose it as default.

I am not sure here - @mbiebl I would appreciate if you carry on that discussion. I personally tend to feel like it is a special use case.

jvymazal:
I followed a logic that if systemd considered many journald files common enough to warrant increase in default then it is generally good idea to follow course (because with imjournal we have the same load FD-wise as journald itself has)

rgerhards:
Ah, I think I finally get you - the journald definition already has that many of files. OK, did not understand it initially (shame on me). So this sounds reasonable to me. Let's wait what @mbiebl says.

davidelang:
I have, but it was because I had a lot of output files open. This was prior to 
systemd so I just set the ulimit higher in the init script and kept going.

people who use dynafiles (and remember to set the cache high enough) can very 
easily use a _lot_ of FDs, so many that there's no way to set the limit to work 
for everyone. It just needs to be easy to change the limit.

jvymazal:
Also, the increased limit in systemd is only partly to (potentially) many services running. The second (and valid for us) reason is because journald can easily have tons of files (if you use splitMode, which I believe is default in at least some cases, you can hit the limit quite quickly). And in this case it can lead to exhausting the limit and rsyslog for example not opening stdin, stdout and stderr because no FDs are available anymore.

rgerhards:
@mbiebl I know quite some users who have several thousand "fds" open for network connections. 

question: does it hurt if we keep the max fds as high as proposed? I am not requiring it, but if we talk a handful of KiB it may not really worth trying too hard (granted on embedded its probably a different issue).

I am really undecided on that one.

Issue #2615, opened 2018-04-04
replace all select() calls with poll()


rgerhards:
This is far easier and portable than epoll, but relieves us of the problematic issues with select.

rgerhards:
couple of select() remain:
* those solely used for waiting (just timeout param non-NULL)
* gss-misc and imgssapi - this is contributed functionality we cannot even test
* one select on initial fork of rsyslog - no problem here, not fd can be >= 1024

So the following remain to do:
- [ ] imudp
- [ ] imuxsock

We may consider to change some other places just for sake of removing the API at all from the code, That might be useful for automated checks, but is not yet a priority.

towolf:
This entirely fixed our problems with rsyslog and we can now hold inbound TCP connections from 2500 hosts. Thank you, @rgerhards!

rgerhards:
Thx for confirming, much appreciated!

PR #2680, opened 2018-05-02
imuxsock: replace select() call by poll()

rgerhards:
re-try of early merge of this PR

While extremely unlikely, imuxsock could abort if a file descriptor
> 1024 was received during the startup phase (never occured in
practice, but theoretically could if imfile monitored a large number
of files and were loaded before imuxsock - and maybe other
strange cases).

see also https://github.com/rsyslog/rsyslog/issues/2615


rgerhards:
LGTM fail is irrelevant (just a JavaScript check that failed and we really don't need it).

PR #2696, opened 2018-05-10
testbench: fix race in threadingmqaq test

rgerhards:
As related testing has show this test sometime fails because
the enq operation on the action queue times out. This is valid,
but causes false positives. The simple (and correct!) solution is
to increase the timeout.


rgerhards:
travis fails due to connectivity issues inside travis

Issue #2462, opened 2018-02-08
mmnormalize always returns parsesuccess="OK"


radu-gheorghe:
### Expected behavior
$parsesuccess shouldn't return "OK" if the mmnormalize parsing failed. This works properly with mmjsonparse.

### Actual behavior
It always returns "OK". I thought #429 is back, but according to @rgerhards it's a different issue.

### Steps to reproduce the behavior
parse something with mmnormalize, so that it fails to parse. In my case, the config snippet is this:
```
    action(
        name="main_cee_parser"
        type="mmnormalize"
        useRawMsg="off"
        ruleBase="/etc/json.rulebase"
    )
```
where json.rulebase looked like:

```
rule=:%data:json%
```

If I send something that's not a JSON, $!data is "{ }", but @parsesucess is still "OK".

### Environment
- rsyslog version: 8.32
- platform: Ubuntu 16.04

rgerhards:
@radu-gheorghe flagged as "potential bug" as I need a (testbench) reproducer. I trust you with your analysis ;-)

radu-gheorghe:
Got it. And thanks for the testbench pointer, I'll do that whenever I can.

rgerhards:
At this point, I think it is our chore to craft the test. I don't object of you look into it, but I did not meant to say I request it. Sorry if it sounded so.

radu-gheorghe:
No worries, I think I got the tone exactly as you meant it :)

rgerhards:
@PascalWithopf could you craft a testbench test for this?

PascalWithopf:
@rgerhards the problem can be reproduced. When parameter rulebase is used parsesuccess is always OK, but when parameter rule is used instead it works just fine.

PascalWithopf:
@radu-gheorghe to make it work simply add ```version=2``` as the first line in your rulebase file.

The parameter rulebase uses version 1 as the default, and the error only occurs in version 1.
Because the parameter rule uses version 2 as the default it works without adding the line.

vrotenberg:
Hi, I tried it in the version 2.0.5-1 and had the same problem. The "version=2" doesn't change anything.
Please help.

rgerhards:
I know it's an old issue, but I finally got to it. Using the current version I cannot reproduce. @radu-gheorghe could you give 8.1903.0 a try?

rgerhards:
[...]

PR #2699, opened 2018-05-11
mmnormalize: add tests for parsesuccess

PascalWithopf:
Parsesuccess always is "OK" when using a rulebase with version 1,
even if it wasn't successfull.
These are tests to reproduce the issue.

See https://github.com/rsyslog/rsyslog/issues/2462


rgerhards:
re-directed to my private repo, as PR just provides the test without a fix.

Issue #2549, opened 2018-03-16
Feature Request - imjournal statistics gathering.


monobaila:
I'm currently a very happy user running 8.24.0 on CentOS7. I've been playing around with the impstats module with a view to setting up some monitoring of my rsyslog pipeline. So far it's offered some great insight to the internals which I'll be able to create alerts from (failed counters and discarded counters in particular).

One thing I'd like to be able to track is the number of messages submitted via imjournal, the majority of my inputs come from here so it's a really good metric to baseline, it could be useful for identifying runaway loggers spamming the pipeline. 

It doesn't look like statistics gathering is setup for imjournal - I'd love to see this in a future release. From my naive reading of the code in tcpsrv.c that does the same thing it doesn't look like a massive amount of changes but my C is no where near good enough to make an attempt!

Once I've made some more progress with my monitoring project I'll happily make a contribution to the documentation with things I've learnt which might be useful for others.

rgerhards:
Thanks for your kind words and the suggestion. This looks like a valuable addition. Tentatively assigning 8.34 release goal, but I am not sure we can make it as it closes soon.

PR #2708, opened 2018-05-15
add statistics counter to imjournal

PascalWithopf:
following statistics counter are now supported by imjournal
- submitted = total number of messages submitted for processing

closes https://github.com/rsyslog/rsyslog/issues/2549

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -43,6 +43,7 @@
 #include "datetime.h"
 #include "net.h"
 #include "glbl.h"
+#include "statsobj.h"
 #include "parser.h"
 #include "prop.h"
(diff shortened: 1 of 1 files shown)


rgerhards:
@PascalWithopf please keep in mind to update the doc as well

@mbiebl @portante I am thinking about a test for this functionality. I think I remember that it should be possible to use non-standard journal files, which would be ideal for such a test. I looked up the journal API, but did not find any reference. If you happen to know out of your head, could you give us a hint of where to look?

rgerhards:
Travis fail is due to missing ability to do "apt-get update" - so all well

rgerhards:
@PascalWithopf I merged a bit too quick - please add doc. But I don't revert now as I guess you'll do.

Issue #2712, opened 2018-05-15
imrelp: librelp too old, oversizemode defaults to "abort" [v8.35.0 try http://www.rsyslog.com/e/2186 ]


atc0005:
### Expected behavior

Perhaps an info level message noting what oversizemode option is set on startup

### Actual behavior

This message is shown:

```
imrelp: librelp too old, oversizemode defaults to "abort" [v8.35.0 try http://www.rsyslog.com/e/2186 ]
```

### Steps to reproduce the behavior

1. `sudo apt-add-repository ppa:adiscon/v8-stable`
1. `sudo apt-get update`
1. `sudo apt-get install rsyslog rsyslog-relp packages`
1. Modify `/etc/rsyslog.conf` to load `imrelp`
1. `sudo service rsyslog restart`

Illustration of the changes made to `/etc/rsyslog.conf`:

```diff
--- /tmp/rsyslog.conf	2018-05-15 18:02:58.680917832 -0500
+++ /etc/rsyslog.conf	2018-05-15 18:03:18.318732296 -0500
@@ -22,6 +22,9 @@
 #$ModLoad imtcp
 #$InputTCPServerRun 514
 
+$ModLoad imrelp
+$ModLoad omrelp
```

### Environment
- rsyslog version: 8.35.0
- platform: Ubuntu 16.04

`dpkg -l | grep relp`

```
ii  librelp0                             1.2.16~20180507151350-0adiscon1xenial      amd64        Reliable Event Logging Protocol (RELP) library
ii  rsyslog-relp                         8.36.0~20180515102407-0adiscon1xenial      amd64        RELP protocol support for rsyslog
```

atc0005:
Confirmed same behavior on Ubuntu 14.04.

`dpkg -l | grep relp`

```
ii  librelp0                                              1.2.16-0adiscon1trusty1                             amd64        Reliable Event Logging Protocol (RELP) library
ii  rsyslog-relp                                          8.35.0-0adiscon2trusty1                             amd64        RELP protocol support for rsyslog
```

davidelang:
there was a new version of relp released a few days ago, are you including that 
in your updates?

rgerhards:
This looks strange. Maybe it's a problem with packaging. We are investigating.

Re the information message on startup: I don't think we should do that. If we did, we would need to do so consistently, and that would end up in a config dump on startup. Of course, I'll let me overrule by good arguments.

rgerhards:
[...]

PR #2714, opened 2018-05-16
imrelp bugfix: error message "librelp too old" is always emitted

rgerhards:
... even if librelp is current. The codition check was actually missing.
This commit adds it.

closes https://github.com/rsyslog/rsyslog/issues/2712


rgerhards:
Travis fail is due to missing ability to do "apt-get update" - so all well

PR #2715, opened 2018-05-16
moved FD limit setting to correct section of unit file

jvymazal:
This is to fix directive misplaced in #2657, thanks @mbiebl for the pointer.


rgerhards:
Let's follow the discussion at https://github.com/rsyslog/rsyslog/pull/2657#issuecomment-389322214 before final merge decision.

Issue #2661, opened 2018-04-25
Writing just a newline after regex timeout dumps memory


StrongestNumber9:
### Expected behavior
I would assume it is just a normal event with exactly no content, just "".

### Actual behavior
~~Memory leaks~~ Dumps garbage from memory, nothing gets processed?
```sh
3887.368000466:imfile.c       : stream.c: stream: generated msg based on timeout: ▒Q
```

### Steps to reproduce the behavior
Briefly

Input
```sh
    startmsg.regex="^\\[[0-9]{4}-[0-9]{2}-[0-9]{2}"
```

Module (small for just testing purposes)
```sh
module(load="imfile" readTimeout="5" timeoutGranularity="4")
```

Reproduce
```sh
echo -ne "[$(date +"%Y-%m-%dT%T+01:00")] foo!\n" >> /input.log
sleep 15;
echo >> /input.log

-- now wait until new timeout occurs

```

### Environment
```sh
rsyslogd 8.32.0, compiled with:
        PLATFORM:                               x86_64-unknown-linux-gnu
        PLATFORM (lsb_release -d):
        FEATURE_REGEXP:                         Yes
        GSSAPI Kerberos 5 support:              No
        FEATURE_DEBUG (debug build, slow code): Yes
        32bit Atomic operations supported:      Yes
        64bit Atomic operations supported:      Yes
        memory allocator:                       system default
        Runtime Instrumentation (slow code):    No
        uuid support:                           Yes
        systemd support:                        No
        Number of Bits in RainerScript integers: 64
```

```sh
CentOS Linux release 7.4.1708 (Core)
```

rgerhards:
> Memory leaks, nothing gets processed?

I don't understand that fully. Do you experience a memory leak?

Re the expectation, empty messages will not be generated. But it looks like instead garbage occurs. Did I understand that correctly?

StrongestNumber9:
My bad with the wording, what I meant to say is that it seems like it accesses invalid location in memory and dumps it

rgerhards:
We could not reproduce exactly, but @PascalWithopf crafted a testbench test that fails under ASAN with what looks like you see. Just there is no garbage in our debug log, so we could not see it. I guess it all depends on the runtime memory layout. That said, I would assume that there are cases where it also could cause a segfault.

rgerhards:
[...]

PR #2718, opened 2018-05-16
imfile/core bugfix: potential misadressing in string copy routine

rgerhards:
This can be exposed via imfile, as follows:

- use a regex to process multiline messages
- configure timeouts
- make sure imfile reads a partial message
- wait so that at least one timeout occurs
- add the message termination sequence

This leads to a misadressing, which may have no obvious effects potentially
up to a segfault.

closes https://github.com/rsyslog/rsyslog/issues/2661

### DIFF
--- a/runtime/stringbuf.c
+++ b/runtime/stringbuf.c
@@ -11,7 +11,7 @@
  * e.g. search. Further refactoring and simplificytin may make
  * sense.
  *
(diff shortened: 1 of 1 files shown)


rgerhards:
bad time for CI: travis fail caused by connectivity issues on travis side, freebsd actually worked but buildbot did not manage to get the notification through... So ready to merge actually.

PR #2726, opened 2018-05-18
mmkubernetes test python must encode response

richm:
https://github.com/rsyslog/rsyslog/issues/2721
Was not working on python3 - must use `encode()` to convert the
string to a `bytes` object.
Also run the server with python -u to make sure we get the log
output from the kubernetes test server.


rgerhards:
thx for the very quick turnaround!

PR #2732, opened 2018-05-22
imjournal: add unit names to output

bluca:
When dealing with a large number of templated instances it is hard to
correlate PID names to arguments, especially if logs are shipped
remotely and archived.
If available, add the systemd unit and user unit names to the output
together with the syslog identifiers.

Example:
```
  2018-05-18T16:46:02.511217+00:00 localhost
   diamond[diamond@blue.service]: Usage: diamond [options]
```

```@blue``` is the templated instance argument.

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -277,6 +277,8 @@ readjournal(void)
 	char *message = NULL;
 	char *sys_iden;
 	char *sys_iden_help = NULL;
+	char *unit = NULL;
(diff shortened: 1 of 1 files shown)


davidelang:
Please write your messages with the template RSYSLOG_DebugFormat

at that point you can see all the variables that you have available. You can 
then create a template with whatever data you want in it.

David Lang

bluca:
Ah I see what you mean, thanks for the pointer!

PR #2733, opened 2018-05-23
omelasticsearch: write op types; bulk rejection retries

richm:
Add support for a 'create' write operation type in addition to
the default 'index'.  Using create allows specifying a unique id
for each record, and allows duplicate document detection.

Add support for checking each record returned in a bulk index
request response.  Allow specifying a ruleset to send each failed
record to.  Add a local variable `omes` which contains the
information in the error response, so that users can control how
to handle responses e.g. retry, or send to an error file.

Add support for response stats - count successes, duplicates, and
different types of failures.

Add testing for bulk index rejections.

### DIFF
--- a/plugins/omelasticsearch/omelasticsearch.c
+++ b/plugins/omelasticsearch/omelasticsearch.c
@@ -52,6 +52,8 @@
 #include "cfsysline.h"
 #include "unicode-helper.h"
 #include "obj-types.h"
+#include "ratelimit.h"
+#include "ruleset.h"
 
 #ifndef O_LARGEFILE
 #  define O_LARGEFILE 0
@@ -65,6 +67,8 @@ MODULE_CNFNAME("omelasticsearch")
 DEF_OMOD_STATIC_DATA
 DEFobjCurrIf(errmsg)
 DEFobjCurrIf(statsobj)
+DEFobjCurrIf(prop)
+DEFobjCurrIf(ruleset)
 
 statsobj_t *indexStats;
 STATSCOUNTER_DEF(indexSubmit, mutIndexSubmit)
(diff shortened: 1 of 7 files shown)


richm:
new docs: https://github.com/rsyslog/rsyslog-doc/pull/675

richm:
not sure why `buildbot/rsyslog solaris10x64 rsyslog` failed

richm:
sndrcv_kafka-vg-rcvr.sh test failed - unable to produce topic `static`

rgerhards:
sorry for the silence, was on vacation and currently totally swamped with Adiscon work. I just want to say that I think this PR goes into the right direction. I just need some time for review, may take a couple of days.  @richm thx for all that work!

rgerhards:
@richm can you eloborate a bit on how the retry message is intended to be handled (maybe as a doc PR)? I see it is re-parsed from the message sent to ES. Obviously, this is no longer the original message object (by design) but whatever string was created by the template. That means that the assigned ruleset needs a different template to achieve same results. This is actually what me concerns most - however, if that works pretty well, it actually is a good solution to provide the desired functionality without the need to apply larger changes to rsyslog core.

richm:
> can you elaborate a bit on how the retry message is intended to be handled (maybe as a doc PR)?

should i add it to https://github.com/rsyslog/rsyslog-doc/pull/675 or create a new doc?

> That means that the assigned ruleset needs a different template to achieve same results.

Always, or in certain cases?  e.g. If you use all-json as the output template, then you can reuse it?

Note that the goal was not to preserve exactly the original _message_ when retrying, but to preserve the necessary _data_ in the original message to be resubmitted for the retry.  This is essentially how it was done in the fluentd elasticsearch plugin - preserve the data, but not the original message data structure with all of the metadata (except that the original `time` metadata added by fluentd is preserved, but we don't care about that in the openshift logging cases).

rgerhards:
@richm thx a lot for the clarification. I think it would be a good idea to amend the existing doc PR with a description of the limits of this approach. IMHO that would be good enough for all users, no matter what the platform is. Given that info, they can evaluate if the approach works for them. I guess that also means we will learn if further work is required, and feedback from the field is always the best judgement if we actually need improvement. Deal?

rgerhards:
I forgot:

> Always, or in certain cases? e.g. If you use all-json as the output template, then you can reuse it?

Well, as a user I would tend to expect that it always "just works as before" - which would not be the case in certain cases. Thus I think it makes sense to elaborate on the limits.

richm:
@rgerhards I added a new section - https://github.com/rsyslog/rsyslog-doc/pull/675/commits/6e96e5b53de30da9843aaeb6f7e7f41189aee1a6#diff-bf94d6492cbe0d838f372b4dac32b067R738

rgerhards:
Perfect! Thx

PR #2739, opened 2018-05-25
rsyslog.service.in: create PID file

selivan:
Logrotate scripts use sysv init script with `rotate` option, which expects PID file to be found in /var/run/rsyslogd.pid

### DIFF
--- a/rsyslog.service.in
+++ b/rsyslog.service.in
@@ -6,7 +6,7 @@ Documentation=http://www.rsyslog.com/doc/
 
 [Service]
 Type=notify
-ExecStart=@sbindir@/rsyslogd -n -iNONE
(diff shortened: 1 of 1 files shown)


rgerhards:
We will take a different approach: the sample service file the rsyslog source project provides will continue to ship the option as-is and only change that if package maintainers actually complain about it. So far, that does not seem the case.

The rsyslog-provided packages, however, will remove the ```-iNONE```. 
@friedl have we already done that? Please let us know the status.

friedl:
I am currently looking into producing a patch for the Ubuntu packages that removes the `-iNone` and uses the older `-i` option for creating a PID file.

This may take a little while.

friedl:
Yes, this was completed.

PR #2746, opened 2018-06-01
add GDPR legal notice to contributor doc

rgerhards:
while this is utter nonsense, discussion on GDPR and git
seems unfortunately be rooted in actual legal needs

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,29 @@
+<!--
+LEGAL GDPR NOTICE:
+According to the European data protection laws (GDPR), we would like to make you
+aware that contributing to rsyslog via git will permanently store the
+name and email address you provide as well as the actual commit and the
+time and date you made it inside git's version history. This is inevitable,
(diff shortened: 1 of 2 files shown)


davidelang:
A question. If someone is not willing to provide their identity and a way to contact them, do we want to accept their code? We have no way of contacting them if there is any question about the legality of them contributing the code or if there is a need to talk about changing the license of the code.

rgerhards:
That's a very tricky issue, indeed. The had thing is that I right now do not even know who is responsible for what happens on github - you don't find information anywhere. I guess we need to wait for some high-level court decisions. But I did not even find anything concrete in that regard.

Re the legality: to be sure, we would need written and signed contributor license agreements. That's something few projects do, and they do not do so because it greatly reduces the contributor base. So far, I have used what I consider good common sense: for very small contributions (no real copyright), I do not really care, for larger ones I try to find an indication of the person behind it.

Maybe it's worth opening a meta-issue for legal questions. Any advise would be appreciated.

Issue #2229, opened 2017-12-15
make imfile support symlinks in monitored directories (requires some polling)


Dieken:
I have this file layout with wildcard file "/path/to/xx/yy/zz/*.log":

```
/path/to/xx/yy/zz/a-short-file-name.log ->  /another/path/a-very-long-file-name.log
```

After I delete `/another/path/a-very-long-file-name.log` and recreate it,  imfile in inotify mode doesn't automatically reopen abc.log.  If I delete `a-short-file-name.log` and sym-link again,  imfile continues to read `a-very-long-file-name.log` but it will miss some lines because I use imfile with parameter "freshStartTail=on".

I work around this issue by link `/path/to/xx/yy/zz` to `/another/path`,  imfile works correctly but I have to accept the very-long-file-name.

This issue also affects cronolog(https://linux.die.net/man/1/cronolog) when cronolog keeps a symlink "error.log" to latest "error.log.YYYYmmddHH",  but this is not severe since cronolog quickly recreates the symlink. 

In my actual scenario,  Docker rotates `/var/lib/docker/containers/A-LONG-CONTAINER-ID/A-LONG-CONTAINER-ID.log` to `/var/lib/docker/containers/A-LONG-CONTAINER-ID/A-LONG-CONTAINER-ID.log.1` and recreates `/var/lib/docker/containers/A-LONG-CONTAINER-ID/A-LONG-CONTAINER-ID.log`.  The symlink `/path/to/xx/yy/zz/a-short-file-name.log` is maintained by my own cron job , the symlink won't be recreated from time to time because the target file almost always exists.

platform: Ubuntu 14.04 & 16.06 with ppa:adicson/v8-stable.

Dieken:
Just noticed I changed the path name in code block but didn't change  path name in description, sorry,  fixed it.

rgerhards:
so I guess this is solved

Dieken:
@rgerhards  No, it's not solved, I mean I updated the issue description and fixed some wrong wording.

rgerhards:
@alorbach can you please have a look?

alorbach:
Can you create a debug log which before and after the logrotate?

Dieken:
@alborbach,  if the symlink isn't recreated,  Rsyslog with inotify can't detect the target file is recreated, so Rsyslog doesn't output anything during the log rotation.

rgerhards:
[...]

PR #2749, opened 2018-06-01
Symlink support for imfile

jvymazal:
this introduces symlink detection and following as well
as monitoring changes on them.

### DIFF
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -152,6 +152,7 @@ struct act_obj_s {
 	fs_edge_t *edge;	/* edge which this object belongs to */
 	char *name;		/* full path name of active object */
 	char *basename;		/* only basename */ //TODO: remove when refactoring rename support
+	char *source_name;  /* if this object is target of a symlink, source_name is its name (else NULL) */
 	//char *statefile;	/* base name of state file (for move operations) */
 	int wd;
 #if defined(OS_SOLARIS) && defined (HAVE_PORT_SOURCE_FILE)
@@ -167,6 +168,7 @@ struct act_obj_s {
 	int nRecords; /**< How many records did we process before persisting the stream? */
 	ratelimit_t *ratelimiter;
 	multi_submit_t multiSub;
+	int is_symlink;
 };
 struct fs_edge_s {
 	fs_node_t *parent;
@@ -189,7 +191,7 @@ static rsRetVal persistStrmState(act_obj_t *);
 static rsRetVal resetConfigVariables(uchar __attribute__((unused)) *pp, void __attribute__((unused)) *pVal);
 static rsRetVal ATTR_NONNULL(1) pollFile(act_obj_t *act);
 static int ATTR_NONNULL() getBasename(uchar *const __restrict__ basen, uchar *const __restrict__ path);
-static void ATTR_NONNULL() act_obj_unlink(act_obj_t *const act);
+static void ATTR_NONNULL() act_obj_unlink(act_obj_t *act);
 static uchar * ATTR_NONNULL(1, 2) getStateFileName(const act_obj_t *, uchar *, const size_t);
 static int ATTR_NONNULL() getFullStateFileName(const uchar *const, uchar *const pszout, const size_t ilenout);
 
@@ -483,14 +485,17 @@ in_setupWatch(act_obj_t *const act, const int is_file)
 		goto done;
 
 	wd = inotify_add_watch(ino_fd, act->name,
-		(is_file) ? IN_MODIFY : IN_CREATE|IN_DELETE|IN_MOVED_FROM|IN_MOVED_TO);
+		(is_file) ? IN_MODIFY|IN_DONT_FOLLOW : IN_CREATE|IN_DELETE|IN_MOVED_FROM|IN_MOVED_TO);
 	if(wd < 0) {
-		LogError(errno, RS_RET_IO_ERROR, "imfile: cannot watch object '%s'",
-			act->name);
+		if (errno == EACCES) { /* There is high probability of selinux denial on top-level paths */
+			DBGPRINTF("imfile: permission denied when adding watch for '%s'\n", act->name);
+		} else {
+			LogError(errno, RS_RET_IO_ERROR, "imfile: cannot watch object '%s'", act->name);
+		}
 		goto done;
 	}
 	wdmapAdd(wd, act);
-	DBGPRINTF("in_setupDirWatch: watch %d added for dir %s(%p)\n", wd, act->name, act);
+	DBGPRINTF("in_setupWatch: watch %d added for %s(object %p)\n", wd, act->name, act);
 done:	return wd;
 }
 
@@ -605,7 +610,7 @@ done:	return;
 static void ATTR_NONNULL()
 fen_setupWatch(act_obj_t *const act __attribute__((unused)))
 {
-	DBGPRINTF("fen_setupWatch: DUMMY CALLED - not on Solaris?");
+	DBGPRINTF("fen_setupWatch: DUMMY CALLED - not on Solaris?\n");
 }
 #endif /* FEN */
 
@@ -633,38 +638,48 @@ fs_node_print(const fs_node_t *const node, const int level)
 	}
 }
 
-
 /* add a new file system object if it not yet exists, ignore call
  * if it already does.
  */
-static rsRetVal ATTR_NONNULL()
+static rsRetVal ATTR_NONNULL(1,2)
 act_obj_add(fs_edge_t *const edge, const char *const name, const int is_file,
-	const ino_t ino)
+	const ino_t ino, const int is_symlink, const char *const source)
 {
 	act_obj_t *act;
 	char basename[MAXFNAME];
 	DEFiRet;
 	
-	DBGPRINTF("act_obj_add: edge %p, name '%s'\n", edge, name);
+	DBGPRINTF("act_obj_add: edge %p, name '%s' (source '%s')\n", edge, name, source? source : "---");
 	for(act = edge->active ; act != NULL ; act = act->next) {
 		if(!strcmp(act->name, name)) {
-			DBGPRINTF("active object '%s' already exists in '%s' - no need to add\n",
-				name, edge->path);
-			FINALIZE;
+			if (!source || !act->source_name || !strcmp(act->source_name, source)) {
+				DBGPRINTF("active object '%s' already exists in '%s' - no need to add\n",
+					name, edge->path);
+				FINALIZE;
+			}
 		}
 	}
 	DBGPRINTF("add new active object '%s' in '%s'\n", name, edge->path);
 	CHKmalloc(act = calloc(sizeof(act_obj_t), 1));
 	CHKmalloc(act->name = strdup(name));
-	getBasename((uchar*)basename, (uchar*)name);
-	CHKmalloc(act->basename = strdup(basename));
+	if (-1 == getBasename((uchar*)basename, (uchar*)name)) {
+		CHKmalloc(act->basename = strdup(name)); /* assume basename is same as name */
+	} else {
+		CHKmalloc(act->basename = strdup(basename));
+	}
 	act->edge = edge;
 	act->ino = ino;
+	act->is_symlink = is_symlink;
+	if (source) { /* we are target of symlink */
+		CHKmalloc(act->source_name = strdup(source));
+	} else {
+		act->source_name = NULL;
+	}
 	#ifdef HAVE_INOTIFY_INIT
 	act->wd = in_setupWatch(act, is_file);
 	#endif
 	fen_setupWatch(act);
-	if(is_file) {
+	if(is_file && !is_symlink) {
 		const instanceConf_t *const inst = edge->instarr[0];// TODO: same file, multiple instances?
 		CHKiRet(ratelimitNew(&act->ratelimiter, "imfile", name));
 		CHKmalloc(act->multiSub.ppMsgs = MALLOC(inst->nMultiSub * sizeof(smsg_t *)));
@@ -702,27 +717,24 @@ detect_updates(fs_edge_t *const edge)
 {
 	act_obj_t *act;
 	struct stat fileInfo;
+	int restart = 0;
 
-	for(act = edge->active ; act != NULL ; ) {
+	for(act = edge->active ; act != NULL ; act = act->next) {
 		DBGPRINTF("detect_updates checking active obj '%s'\n", act->name);
-		const int r = stat(act->name, &fileInfo);
+		const int r = lstat(act->name, &fileInfo);
 		if(r == -1) { /* object gone away? */
 			DBGPRINTF("object gone away, unlinking: '%s'\n", act->name);
-			act_obj_t *toDel = act;
-			act = act->next;
-			DBGPRINTF("new next act %p\n", act);
-			act_obj_unlink(toDel);
-			continue;
+			act_obj_unlink(act);
+			restart = 1;
+			break;
 		}
 		// TODO: add inode check for change notification!
 
-		/* Note: active nodes may get deleted, so we need to do the
-		 * pointer advancement at the end of the for loop!
-		 */
-		act = act->next;
 	}
 
-
+	if (restart) {
+		detect_updates(edge);
+	}
 }
 
 
@@ -746,14 +758,39 @@ poll_active_files(fs_edge_t *const edge)
 	}
 }
 
+static rsRetVal ATTR_NONNULL()
+process_symlink(fs_edge_t *const chld, const char *symlink)
+{
+	DEFiRet;
+	char *target = NULL;
+	CHKmalloc(target = realpath(symlink, target));
+	struct stat fileInfo;
+	if(lstat(target, &fileInfo) != 0) {
+		LogError(errno, RS_RET_ERR,	"imfile: process_symlink cannot stat file '%s' - ignored", target);
+		FINALIZE;
+	}
+	const int is_file = (S_ISREG(fileInfo.st_mode));
+	DBGPRINTF("process_symlink:  found '%s', File: %d (config file: %d), symlink: %d\n",
+		target, is_file, chld->is_file, 0);
+	if(!is_file && S_ISREG(fileInfo.st_mode)) {
+		LogMsg(0, RS_RET_ERR, LOG_WARNING,
+			"imfile: '%s' is neither a regular file, symlink, nor a directory - ignored", target);
+		FINALIZE;
+	}
+	act_obj_add(chld, target, is_file, fileInfo.st_ino, 0, symlink);
 
-static void ATTR_NONNULL() poll_tree(fs_edge_t *const chld);
+finalize_it:
+	free(target);
+	RETiRet;
+}
+
 static void ATTR_NONNULL()
 poll_tree(fs_edge_t *const chld)
 {
 	struct stat fileInfo;
 	glob_t files;
 	int need_globfree = 0;
+	int issymlink;
 	DBGPRINTF("poll_tree: chld %p, name '%s', path: %s\n", chld, chld->name, chld->path);
 	detect_updates(chld);
 	const int ret = glob((char*)chld->path, runModConf->sortFiles|GLOB_BRACE, NULL, &files);
@@ -766,18 +803,27 @@ poll_tree(fs_edge_t *const chld)
 				goto done;
 			}
 			char *const file = files.gl_pathv[i];
-			if(stat(file, &fileInfo) != 0) {
+			if(lstat(file, &fileInfo) != 0) {
 				LogError(errno, RS_RET_ERR,
 					"imfile: poll_tree cannot stat file '%s' - ignored", file);
 				continue;
 			}
 
-			const int is_file = S_ISREG(fileInfo.st_mode);
-			DBGPRINTF("poll_tree:  found '%s', File: %d (config file: %d)\n",
-				file, is_file, chld->is_file);
+			if (S_ISLNK(fileInfo.st_mode)) {
+				rsRetVal slink_ret = process_symlink(chld, file);
+				if (slink_ret != RS_RET_OK) {
+					continue;
+				}
+				issymlink = 1;
+			} else {
+				issymlink = 0;
+			}
+			const int is_file = (S_ISREG(fileInfo.st_mode) || issymlink);
+			DBGPRINTF("poll_tree:  found '%s', File: %d (config file: %d), symlink: %d\n",
+				file, is_file, chld->is_file, issymlink);
 			if(!is_file && S_ISREG(fileInfo.st_mode)) {
 				LogMsg(0, RS_RET_ERR, LOG_WARNING,
-					"imfile: '%s' is neither a regular file nor a "
+					"imfile: '%s' is neither a regular file, symlink, nor a "
 					"directory - ignored", file);
 				continue;
 			}
@@ -788,7 +834,7 @@ poll_tree(fs_edge_t *const chld)
 					(chld->is_file) ? "FILE" : "DIRECTORY");
 				continue;
 			}
-			act_obj_add(chld, file, is_file, fileInfo.st_ino);
+			act_obj_add(chld, file, is_file, fileInfo.st_ino, issymlink, NULL);
 		}
 	}
 
@@ -829,8 +875,20 @@ act_obj_destroy(act_obj_t *const act, const int is_deleted)
 	if(act == NULL)
 		return;
 
-	DBGPRINTF("act_obj_destroy: act %p '%s', wd %d, pStrm %p, is_deleted %d, in_move %d\n",
-		act, act->name, act->wd, act->pStrm, is_deleted, act->in_move);
+	DBGPRINTF("act_obj_destroy: act %p '%s' (source '%s'), wd %d, pStrm %p, is_deleted %d, in_move %d\n",
+		act, act->name, act->source_name? act->source_name : "---", act->wd, act->pStrm, is_deleted,
+		act->in_move);
+	if(act->is_symlink && is_deleted) {
+		act_obj_t *target_act;
+		for(target_act = act->edge->active ; target_act != NULL ; target_act = target_act->next) {
+			if(target_act->source_name && !strcmp(target_act->source_name, act->name)) {
+				DBGPRINTF("act_obj_destroy: unlinking slink target %s of %s "
+						"symlink\n", target_act->name, act->name);
+				act_obj_unlink(target_act);
+				break;
+			}
+		}
+	}
 	if(act->ratelimiter != NULL) {
 		ratelimitDestruct(act->ratelimiter);
 	}
@@ -862,6 +920,7 @@ act_obj_destroy(act_obj_t *const act, const int is_deleted)
 	}
 	#endif
 	free(act->basename);
+	free(act->source_name);
 	//free(act->statefile);
 	free(act->multiSub.ppMsgs);
 	#if defined(OS_SOLARIS) && defined (HAVE_PORT_SOURCE_FILE)
@@ -909,7 +968,7 @@ chk_active(const act_obj_t *act, const act_obj_t *const deleted)
  * destruct it.
  */
 static void //ATTR_NONNULL()
-act_obj_unlink(act_obj_t *const act)
+act_obj_unlink(act_obj_t *act)
 {
 	DBGPRINTF("act_obj_unlink %p: %s\n", act, act->name);
 	if(act->prev == NULL) {
@@ -921,6 +980,7 @@ act_obj_unlink(act_obj_t *const act)
 		act->next->prev = act->prev;
 	}
 	act_obj_destroy(act, 1);
+	act = NULL;
 //dbgprintf("printout of fs tree post unlink\n");
 //fs_node_print(runModConf->conf_tree, 0);
 //dbg_wdmapPrint("wdmap after");
@@ -1162,7 +1222,11 @@ enqLine(act_obj_t *const act,
 	msgSetPRI(pMsg, inst->iFacility | inst->iSeverity);
 	MsgSetRuleset(pMsg, inst->pBindRuleset);
 	if(inst->addMetadata) {
-		metadata_values[0] = (const uchar*)act->name;
+		if (act->source_name) {
+			metadata_values[0] = (const uchar*)act->source_name;
+		} else {
+			metadata_values[0] = (const uchar*)act->name;
+		}
 		snprintf((char *)file_offset, MAX_OFFSET_REPRESENTATION_NUM_BYTES+1, "%lld", strtOffs);
 		metadata_values[1] = file_offset;
 		msgAddMultiMetadata(pMsg, metadata_names, metadata_values, 2);
@@ -1389,13 +1453,16 @@ pollFile(act_obj_t *const act)
 {
 	cstr_t *pCStr = NULL;
 	DEFiRet;
+	if (act->is_symlink) {
+		FINALIZE;    /* no reason to poll symlink file */
+	}
 	/* Note: we must do pthread_cleanup_push() immediately, because the POSIX macros
 	 * otherwise do not work if I include the _cleanup_pop() inside an if... -- rgerhards, 2008-08-14
 	 */
 	pthread_cleanup_push(pollFileCancelCleanup, &pCStr);
 	iRet = pollFileReal(act, &pCStr);
 	pthread_cleanup_pop(0);
-	RETiRet;
+finalize_it: RETiRet;
 }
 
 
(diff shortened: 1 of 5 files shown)


jvymazal:
This is attempt to address #2229

plugins/imfile/imfile.c
```diff
@@ -543,6 +546,81 @@ wdmapDel(const int wd)
...
+{
+	const char fullpath[MAXFNAME];
+	const char *ptr = fullpath;
+	struct stat fileInfo;
+	int is_symlink = 0;
+
+	if (file[0] == '/') {
+		ptr = file;
+	} else {
+		snprintf((char *)fullpath, MAXFNAME, "%s/%s", dir, file);
```
richm:
```
fullpath[MAXFNAME-1] = '\0';
```
to ensure null termination

plugins/imfile/imfile.c
```diff
@@ -543,6 +546,81 @@ wdmapDel(const int wd)
...
+	}
+	if(lstat(ptr, &fileInfo) != 0) {
+		perror("lstat");
```
richm: use rs_strerror_r here, as used below

richm: and - should return is_symlink = -1 here too?

plugins/imfile/imfile.c
rgerhards: @richm I had a discussion today on the need to do this sanity check. I also think it's the right thing to do, but I could not find any evidence we are right. Do you have evidence?

tests/imfile-symlink.sh
```diff
@@ -0,0 +1,338 @@
...
+rsyslogconf_bak=`mktemp`
+cp $rsyslogconf $rsyslogconf_bak
+watchdir0="/var/log/test0"
+targetdir0="/var/log/target0"
+reltarget0="../target0"
+watchdir1="/var/log/test1"
+targetdir1="/var/log/target1"
+reltarget1="../target1"
+
+if [ "$EUID" -ne 0 ]; then
```
rgerhards: do we have a chance to get a test without root permissions?

plugins/imfile/imfile.c
richm: @rgerhards I've always had to ensure proper null termination of strings when using snprintf.  I don't have any evidence, just many years of doing it this way.

rgerhards: @richm I feared that answer... much like I knew and many do. But may look like it is no longer true. Should you run over some case where it does not work, let me know ;-)

richm: > may look like it is no longer true.

How so?  Was snprintf changed to always do null termination?  To my knowledge (man snprintf on Fedora 27) it was no.  Was there some change to the compiler to ensure `const char fullpath[MAXFNAME]` is initialized to all nulls?  again, not to my knowledge.

rgerhards: @richm actually yes, snprintf is specified to do that in C99 (here a link to POSIX): http://pubs.opengroup.org/onlinepubs/007904875/functions/fprintf.html

It is interesting that the man page actually does not make a clear guarantee in that regard. I checked the Fedora 27 man pages, as well as Ubuntu and BSD, and indeed they do not specify it precisely - but they claim C99 compatibility. Web searches often bring up discussions where it is said this could be treated as granted.

Red Hat's Florian Weimer wrote a longer post about the problems with ```snprintf```: https://access.redhat.com/blogs/766093/posts/1976193 In it, he states
```
This code assumes that the snprintf implementation does not write an 
unterminated string to the destination buffer on error, which is quite
reasonable as far as such assumptions go.  
```
Given his obviously in-depth resarch, I would guess that he would have noticed if there is a problem.

In conclusion, to me it looks like this is really something that we can rely on - but I am still puzzled. As I said, I was also thinking that there may be a problem on some platforms, but I cannot find hard evidence of where these problems really exist. Having pointers would be most useful.

rgerhards: I should have also provided this quote from POSIX:

*The snprintf() function shall be equivalent to sprintf(), with the addition of the n argument which states the size of the buffer referred to by s. If n is zero, nothing shall be written and s may be a null pointer. Otherwise, output bytes beyond the n-1st shall be discarded instead of being written to the array, and a null byte is written at the end of the bytes actually written into the array.*

It's pretty much at the top of the document.

richm: Ok.  That seems to be the behavior on fedora 27 (gcc (GCC) 7.3.1 20180303 (Red Hat 7.3.1-5)) and rhel7 (gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-16)).  Trying various ways to break snprintf with a simple test program shows that it always null terminates the string and returns the correct result code.

Consider my objection withdrawn.

tests/imfile-symlink.sh
jvymazal: Well, the test needs permissions to restart rsyslog to properly flush the watches while re-creating test dirs, so not likely. Is it a big problem when we have root test?

richm: Yes.  Why not write a test that uses the test framework, which is able to configure and run rsyslog with out redhat/fedora-isms such as running as root, running using systemd, etc.?
For example: https://github.com/rsyslog/rsyslog/blob/master/tests/imfile-wildcards-dirs-multi.sh

rgerhards: +1 - we should have tests for this functionality. If it is just a start, that's "biz as usual" inside the testbench.

jvymazal: Yes, I am planning to rewrite the test, what is a timeframe for this to make the 8.36 please?

rgerhards:
8.36 officially was closed friday, I am doing final merges at the moment. So we are at 8.37. BTW: the timeframe is always shown in the assigned milestone, release dates on https://www.rsyslog.com.

plugins/imfile/imfile.c
```diff
@@ -770,12 +853,91 @@ poll_tree(fs_edge_t *const chld)
...
+				}
+				if(rv == 0) { /* out of memory */
+					LogError(errno, RS_RET_OUT_OF_MEMORY, "imfile: poll_tree: \
+							hastable_insert failed");
+					if (freetarget) {
+						free(target);
+					}
+					continue;
+				}
+				DBGPRINTF("imfile: %s -> %s added to symlink hash\n", slink, target);
```
nhosoi: Do we need to call realloc to expand files.gl_pathv to allocate the space for target prior to the line 931 (as being done in the downstream version)?

jvymazal: Yes, I will edit the PR once I finish first round of testing.

jvymazal:
@rgerhards I have two questions now. First, how can I determine which part of my test causes the "permission denied" failures (as far as I am aware I am not doing anything special there) and more importantly, current implementation has the side-effect of expanding relative file names (not just symlinks, any relative files) in metadata, should I change it to keep the old behavior or adjust affected tests and leave it that way? Many thanks for pointers and opinion.

rgerhards:
> I have two questions now. First, how can I determine which part of my test causes the "permission denied" failures (as far as I am aware I am not doing anything special there) 

Let me check. I am still working on CI (you probably have seen many new buildbot slaves plus Fedora28). This could have caused it. I'll have a look at the whole thing and let you know if something needs attention. From a first glimpse, it looks like we have some path expansion messages that do not belong there. But I may be wrong here. Will have a deeper look when I am through with the current round of CI fixing, as it may affect what you see (so it does not make sense to look immediately).

> and more importantly, current implementation has the side-effect of expanding relative file names (not just symlinks, any relative files) in metadata, should I change it to keep the old behavior or adjust affected tests and leave it that way? 

Good question. @davidelang what do you think? I would tend to say it should stay as it was, but for symlinks it probably is questionable.

davidelang:
On Wed, 18 Jul 2018, Rainer Gerhards wrote:

> Good question. @davidelang what do you think? I would tend to say it should 
> stay as it was, but for symlinks it probably is questionable.

I think we need to keep things as they are, make the expanded version be 
additional metadata

for kubernetes, we depend on three layers of symlinks, and the name of the first 
symlink contains critical information that doesn't exist in the other two 
layers. Ideally, each layer should be documented in the metadata.

David Lang

richm:
For the kubernetes case, we would need to use a function other than `realpath(3)` to get the target.  We would have to make one or more calls to `readlink(2)` to follow an entire chain of links to get every intermediate link.

richm:
The way it works currently (i.e. without this patch/PR) is that the name of the symlink is used for `$!metadata!filename`.  In the kubernetes case, if I have
```
input(type="imfile" file="/var/log/containers/*.log" tag="kubernetes" addmetadata="on")
```
Then the record contains
```
$!:{ "metadata": { "filename": "\/var\/log\/containers\/logging-kibana-1-ctwqh_openshift-logging_kibana-1ce792f7c50f2b2e38d56050f32e6c527589515e299ee2670a099a2d08361f79.log", "fileoffset": "5955682" } }
```
instead of some target file under `/var/log/pods/` or `/var/lib/docker/containers/*`.  This behavior is necessary for mmkubernetes to work.

davidelang:
On Wed, 18 Jul 2018, Richard Megginson wrote:

> For the kubernetes case, we would need to use a function other than 
> `realpath(3)` to get the target.  We would have to make one or more calls to 
> `readlink(2)` to follow an entire chain of links to get every intermediate 
> link.

From a metadata point of view, we can do three things.

1. show the filename of the first link (what rsyslog config is told to look at)

2. show the path to the actual file

3. show the filename of the first link, and then in separate fields show each 
intemediate link and the final path to the actual file

4. show the filename of the first link, then in a separate field show the path 
to the actual file


kubernetes will break if the filename for the first link isn't available.

I can see people wanting the other information.

so in order of preference, I would say 3, 4, 1, and far behind 2.

to avoid breaking existing configs, we need to leave the filename metadata 
showing what it does today (which would rule out 2).

adding the metadata for the final path in a separate variable (4) would be an 
improvement over what we do today. But adding a third variable that's an array 
of any intermediate links would be a nice-to-have.

David Lang

jvymazal:
> I would like to do this in two steps.
> 
> First step - this current PR should be changed to do 1. show the filename of the first link (what rsyslog config is told to look at) - this will preserve the current behavior and not introduce any regressions

Understood, I will edit the PR to carry over the link name to the target. However this should definetely be documented as my first understanding after reading https://github.com/rsyslog/rsyslog-doc/blob/master/source/configuration/modules/imfile.rst#metadata was that filename should behave like 2. which is why I originally written it that way (should have tested that the code actually does that).

rgerhards:
> Understood, I will edit the PR to carry over the link name to the target. However this should definetely be documented as my first understanding after reading https://github.com/rsyslog/rsyslog-doc/blob/master/source/configuration/modules/imfile.rst#metadata was that filename should behave like 2.

PR would be appreciated ;-)

jvymazal:
However, if preserving 

> 1. show the filename of the first link (what rsyslog config is told to look at)

this raises a question of what to list there in case there are more symlinks pointing to the same file. For now I will go with using the first symlink I encounter and ignoring the rest. @richm @davidelang any comments to this?

rgerhards:
> this raises a question of what to list there in case there are more symlinks pointing to the same file. For now I will go with using the first symlink I encounter and ignoring the rest.

As the gouverning rule I would say "it should behave exactly like it did before the patch". So whatever it currently does is the right thing to do.

rgerhards:
side-note: it looks like the CI failures are valid regressions, so something to check out.

I do not know what the permission issue is. Maybe add ```set -x ; set -v``` on top of the test?

rgerhards:
If you have problems to reproduce the issues locally, you can use the docker container CI uses. It's easy to do via ```devtools/devcontainer``` - see that file for the env vars to set.

richm:
imfile.c has a potential uninitialized variable - this fixes it:
```
diff --git a/plugins/imfile/imfile.c b/plugins/imfile/imfile.c
index 6c5ebb65a..8d4a7a6d9 100644
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -810,7 +810,6 @@ poll_tree(fs_edge_t *const chld)
 {
        struct stat fileInfo;
        glob_t files;
-       int issymlink;
        DBGPRINTF("poll_tree: chld %p, name '%s', path: %s\n", chld, chld->name, chld->path);
        detect_updates(chld);
        const int ret = glob((char*)chld->path, runModConf->sortFiles|GLOB_BRACE, NULL, &files);
@@ -818,6 +817,7 @@ poll_tree(fs_edge_t *const chld)
        if(ret == 0) {
                DBGPRINTF("poll_tree: processing %d files\n", (int) files.gl_pathc);
                for(unsigned i = 0 ; i < files.gl_pathc ; i++) {
+                       int issymlink = 0;
                        if(glbl.GetGlobalInputTermState() != 0) {
                                goto done;
                        }
```

richm:
I think the tests will pass if you add `imfile-symlink.sh` to `EXTRA_DIST`in tests/Makefile.am

davidelang:
On Thu, 19 Jul 2018, jvymazal wrote:

> However, if preserving
>
>> 1. show the filename of the first link (what rsyslog config is told to look at)
>
> this raises a question of what to list there in case there are more symlinks pointing to the same file. For now I will go with using the first symlink I encounter and ignoring the rest. @richm @davidelang any comments to this?

for 1, pretend that you know nothing about symlinks, you have a filename that 
you open (either from the config or from matching a pattern), that is what you 
open, and you don't care if it's the only link to a file, an additional hard 
link, a symlink, a named pipe, or anything else. You get back a filehandle that 
you can issue reads on.

That name is what should be reported in the metadata today. As far as the 
application is concerned, that's the filename.

Now, in the case of symlinks, there is additional knowledge that we can gather, 
what the final name is, and what the hops are to get there.

In no case should we be searching to find if we have multiple links to the same 
file.

unless, are you saying that we follow the links to be able to detect rotation of 
the file pointed at by the symlink, and because we do so, we may discover that 
we have multiple filehandles open to the same file?

In this case, my tendency would be to log a warning at file open time, and then 
happily duplicate the logs by reading them through each original filename. After 
all, that's what the admin told us to do. It may seem strange, but we had 
someone just this week who was doing this deliberately because they wanted to 
send the logs to different destinations with different facilities on the log 
message, so he created symlinks and had multiple imfile input() sections?

David Lang

jvymazal:
> for 1, pretend that you know nothing about symlinks, you have a filename that you open (either from the config or from matching a pattern), that is what you open, and you don't care if it's the only link to a file, an additional hard link, a symlink, a named pipe, or anything else. You get back a filehandle that you can issue reads on. That name is what should be reported in the metadata today. As far as the application is concerned, that's the filename. 
This is somehow what actual state of the PR is with the last edits.

> unless, are you saying that we follow the links to be able to detect rotation of the file pointed at by the symlink, and because we do so, we may discover that we have multiple filehandles open to the same file? 
Unless making some major code changes there will never be more open file handles to one file, this is resolved by inotify code, not us.
What is happening currently is that if you configure more symlinks to same file, when creating data for 2nd+ symlink I find that there is already record in the table. To be consistent with original behavior I am skipping the processing of such files now. Only difference is that before you got wd error and now you get specific log notice stating which superfluous links to which file we found.

rgerhards:
> Unless making some major code changes there will never be more open file handles to one file, this is resolved by inotify code, not us.

we do rely on inotify only for notifications. As they are not reliably reported, we still iterate over our data structures to see if a new file popped up. I guess that means your argument is wrong.

rgerhards:
> What is happening currently is that if you configure more symlinks to same file, when creating data for 2nd+ symlink I find that there is already record in the table. To be consistent with original behavior I am skipping the processing of such files now. Only difference is that before you got wd error and now you get specific log notice stating which superfluous links to which file we found.

sorry, I should have taken a bit more time ;-)

I stil think the existing code processed the file, it just did not manage to setup an inotify watch. Is that right? If so, I think it has implications on what we need to preserve.

rgerhards:
also, we must ensure everything works when no inotify is available, e.g. BSD or Solaris. Do we need additional tests to ensure this is the case?

jvymazal:
> also, we must ensure everything works when no inotify is available, e.g. BSD or Solaris. Do we need additional tests to ensure this is the case?

Apart from the watches the code should work in polling mode as well, so targets will be added to active lists and polled.

rgerhards:
> Apart from the watches the code should work in polling mode as well, so targets will be added to active lists and polled.

```should``` or ```does```? ;-)

**Comment by rgerhards** [comment]
In other words: I think we need a test case where two symlinks point to the same file and we must check that in the end result the file is processed twice. @davidelang did I get this right?

**Comment by jvymazal** [comment]
> we must check that in the end result the file is processed twice.

OK, I think I get it now, @davidelang in example you posted I believe polling mode was used, right?  In that case it needs one more tweak to handle this situation. I was not aware that there was symlink support for the polling mode as well.

**Comment by rgerhards** [comment]
The refactored imfile (8.35+?) has unified inotify/FEN/polling. even in inotify there is some polling mode involved, as the inotify API is not reliable (we had many bug reports that rooted in that). So inotify/FEN are primarily used for hinting us in the new code.

**Comment by rgerhards** [comment]
For clarity, I should add that the exiting code does not care about symlinks at all. So whatever currently happens with them, is not by design but by "accident". Nevertheless, people probably count on the existing behavior, so we cannot break it by default. Even though the current handling was never explicitly designed, but it is in place for many years now.

**Comment by jvymazal** [comment]
Understood, I will add a case for this so multiple links are still processed if configured that way.

**Comment by jvymazal** [comment]
@rgerhards one more question, after adding new test to distcheck section I get an error that there are some files left after dist clean, where should I add those to fix it please?

**Comment by rgerhards** [comment]
@jvymazal  thx, sorry for not being specific about the refactoring in the immediately. If in doubt, I think you can run the test against the codebase without this patch. Whatever it does, is expected behavior (I admit I have never looked at that case, so I am really not sure what happens). If the current code really makes strange things, we possibly need to reconsider if there is a bug in existing code, but from @davidelang I got the impression that it works in a defined way.

I would also like to thank you for being so persistent with what looked like a simple patch but turns into something much more elaborate. I know that feeling...

**Comment by rgerhards** [comment]
> @rgerhards one more question, after adding new test to distcheck section I get an error that there are some files left after dist clean, where should I add those to fix it please?

If they are something created by the test, they should be removed by then init/exit here https://github.com/rsyslog/rsyslog/blob/master/tests/diag.sh#L136

We also remove by init to ensure that the next test gets a clean environment (error-exit intentionally does not clean up).

File: plugins/imfile/imfile.c [hunk 5]
```diff
@@ -746,13 +758,44 @@ poll_active_files(fs_edge_t *const edge)
...
+		FINALIZE;
+	}
+	act_obj_add(chld, target, is_file, fileInfo.st_ino, 0, symlink);
+
+finalize_it:
+	free(target);
+	RETiRet;
+}
 
-static void ATTR_NONNULL() poll_tree(fs_edge_t *const chld);
```
rgerhards, line 750: I think I remember this line was required by some analyzer - but I may be wrong. CI will hopefully show,

plugins/imfile/imfile.c
```diff
@@ -1384,6 +1452,9 @@ pollFile(act_obj_t *const act)
 {
 	cstr_t *pCStr = NULL;
 	DEFiRet;
+	if (act->is_symlink) {
+		RETiRet;    /* no reason to poll symlink file */
```
rgerhards: This should be FINALIZE (RETiRet must only be used at end of function, in some settings it can generate extra code).

plugins/imfile/imfile.c
jvymazal: I think that last run almost passed (minus the extra file left but actual tests were green)

davidelang:
On Fri, 20 Jul 2018, Rainer Gerhards wrote:

> In other words: I think we need a test case where two symlinks point to the 
> same file and we must check that in the end result the file is processed 
> twice. @davidelang did I get this right?

Yes.

David Lang

jvymazal:
OK, @rgerhards, changed the implementation to check for both name and source before skipping act_obj_add() and also added another test case for this.

jvymazal:
Fine, I will push it, I just have to solve last possible bug, as scenario was found where everything looks like `if (S_ISLNK(fileInfo.st_mode))` on symlink file is reporting false, will have to investigate it so we do not have to merge this twice

jvymazal:
Ok, found there there is a difference between `stat()` and `lstat()` results.. now it will hopefully work as intended, also edited the newly added test as per your recent testbench commit, @rgerhards if it goes all green we should be fine to merge.

rgerhards:
thx a lot!

rgerhards:
side-note, just FYI: there is still some instability in CI. I am working on that. If something turns red, I'll definitely have a look. It may be unrelated.

tests/imfile-symlink-multi.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+# This is part of the rsyslog testbench, licensed under GPLv3
```
rgerhards: sry, a final nit :-( could you license under ASL 2.0 if you have no concerns with that (if there are, keep as is!). We try to move toward ASL 2.0, but some old tests still bear GPLv3.

tests/imfile-symlink.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+# This is part of the rsyslog testbench, licensed under GPLv3
```
rgerhards: ASL 2.0 preferred

jvymazal:
> I did not find the test mentioned here #2749 (comment) - I would also assume that this requires a new config parameter.

The imfile-symlink-multi tests just that (it points 10 symlinks to one target file and then checks there are 10 messages in out.log
I do not see why we should need any new parameter for that.

rgerhards:
> The imfile-symlink-multi tests just that (it points 10 symlinks to one target file and then checks there are 10 messages in out.log
> I do not see why we should need any new parameter for that.

maybe I am doing too many things in parallel, but... By default, imfile should behave like before the patch. So it must report the unresolved symlink name. Does the test do that? And if it does, how does imfile know when it has to report the resolved symlink?

rgerhards:
> By default, imfile should behave like before the patch. So it must report the unresolved symlink name. Does the test do that?

I withdraw this comment - the test does indeed do that.

Now let me tripple-check the other question.

jvymazal:
@rgerhards regarding the now-failing static analyzer, would you mind taking a look at the report? I have tried to go trough it several times but each time I came to conclusion that it is a false positive. I actually changed the very lines it complains about after discovering and fixing segfault in them. Thanks

rgerhards:
> regarding the now-failing static analyzer, would you mind taking a look at the report?

will have a look at the moment, got distracted yesterday.

To my other question and without digging deeper: how is the new behavior of symlink processing controlled? I did not see any config switch.

jvymazal:
> To my other question and without digging deeper: how is the new behavior of symlink processing controlled? I did not see any config switch.

Well, I did not add any :-)
I tried to keep it in line with original behavior with the addition the code is now actually aware of symlinks and has separate watches and objects for them AND for their targets. I do not see why we should need any new switch for that. If you think there are some cases where current state of PR would break something please bring it up.

rgerhards:
> I tried to keep it in line with original behavior with the addition the code is now actually aware of symlinks and has separate watches and objects for them AND for their targets. I do not see why we should need any new switch for that. If you think there are some cases where current state of PR would break something please bring it up.

mmmhhhh... if there is absolutely no change of behavior vs. previous code, I think we could leave all as is ;-) That's just the philosophical PoV, of course. I'll see that I do some tests if all is kept active and have a deeper look at the new code. 

@davidelang @richm would also appreciate your feedback, as I did not yet have symlinks on my mind.

plugins/imfile/imfile.c
```diff
@@ -702,27 +714,24 @@ detect_updates(fs_edge_t *const edge)
...
-	for(act = edge->active ; act != NULL ; ) {
+	for(act = edge->active ; act != NULL ; act = act->next) {
 		DBGPRINTF("detect_updates checking active obj '%s'\n", act->name);
-		const int r = stat(act->name, &fileInfo);
+		const int r = lstat(act->name, &fileInfo);
 		if(r == -1) { /* object gone away? */
 			DBGPRINTF("object gone away, unlinking: '%s'\n", act->name);
-			act_obj_t *toDel = act;
-			act = act->next;
-			DBGPRINTF("new next act %p\n", act);
```
rgerhards, line 713: 1: I guess the static analyzer is somewhat confused, but isn't the removal of this line (plus the next I flag) the root problem?

plugins/imfile/imfile.c
```diff
@@ -702,27 +714,24 @@ detect_updates(fs_edge_t *const edge)
...
-			act_obj_unlink(toDel);
-			continue;
+			act_obj_unlink(act);
```
rgerhards, line 727: 2: now act is deleted (freed)

plugins/imfile/imfile.c
```diff
@@ -702,27 +714,24 @@ detect_updates(fs_edge_t *const edge)
 {
 	act_obj_t *act;
 	struct stat fileInfo;
+	int restart = 0;
 
-	for(act = edge->active ; act != NULL ; ) {
+	for(act = edge->active ; act != NULL ; act = act->next) {
```
rgerhards, line 722: 3: act in act->next has been freed in line 724 (see comment below).

plugins/imfile/imfile.c
rgerhards: sry, previous line 711 I meant

plugins/imfile/imfile.c
jvymazal: Yes, however, if I unlink it I break the cycle and the processing begins again from the start.

rgerhards: but you actually unlink it in 724?

jvymazal: The problem I see when analyzing the result seems to be that clang seems to ignore [this line ](https://github.com/rsyslog/rsyslog/pull/2749/files#diff-b563a9e8d7f28e380e7fd4dc4467d3e8R969) line which guards all 3 cases it reports.

rgerhards:
2018-07-25 10:18 GMT+02:00 jvymazal <notifications@github.com>:

> *@jvymazal* commented on this pull request.
>
> In plugins/imfile/imfile.c
> <https://github.com/rsyslog/rsyslog/pull/2749#discussion_r205020166>:
>
> >
> -	for(act = edge->active ; act != NULL ; ) {
> +	for(act = edge->active ; act != NULL ; act = act->next) {
>
> The problem I see when analyzing the result seems to be that clang seems
> to ignore this line
> <https://github.com/rsyslog/rsyslog/pull/2749/files#diff-b563a9e8d7f28e380e7fd4dc4467d3e8R969>
> line which guards all 3 cases it reports.
>
>
> Have your read *all* my comments? clang is wrong, but it is confused by a
problem that it does not properly report. The problem itself is present.

plugins/imfile/imfile.c
rgerhards: the problem is not act->name but act itself.

rgerhards: act->next leads to deref of act, which has been freed in line 724.

jvymazal: Yes, but if I free it I never get here again, see comments on below lines.

plugins/imfile/imfile.c
```diff
@@ -702,27 +714,24 @@ detect_updates(fs_edge_t *const edge)
...
 		if(r == -1) { /* object gone away? */
 			DBGPRINTF("object gone away, unlinking: '%s'\n", act->name);
-			act_obj_t *toDel = act;
-			act = act->next;
-			DBGPRINTF("new next act %p\n", act);
-			act_obj_unlink(toDel);
-			continue;
+			act_obj_unlink(act);
+			restart = 1;
+			break;
```
jvymazal, line 729: I had to change a flow now because unlink call can do more thinks due to additional target removals, so if I unlink anything flom list I abort completely and go trough whole function again so I have ensured coherent state of the list.

rgerhards: ok, let me first try to update to clang-6.0 analyzer. Maybe it solves the situation. This also sounds Iike I do need to go more in-depth over the PR. I'll ping you when clang-6 is ready, not sure I will mange this today.

rgerhards:
@jvymazal FYI: I gave changing to clang-6.0 a try, but it found some yet-undetected issues: https://github.com/rsyslog/rsyslog/pull/2875 -- will need to work on them before I can activate.

davidelang:
I see no reason for having a config item for this.

People currently using metadata will keep working, there will just be some extra 
variables that they don't do anything with.

David Lang

rgerhards:
> I see no reason for having a config item for this.
> 
> People currently using metadata will keep working, there will just be some extra
> variables that they don't do anything with.

thx, this clears my concern

@jvymazal can you please rebase, looks like my patch to switch the clang 6 affected your commit. That means there is a good chance it could go through with the next run.

jvymazal:
done

jvymazal:
@rgerhards the only current failure seems unrelated (only kafka test), otherwise the static analyzer looks resolved, thanks!

jvymazal:
@rgerhards I did another rebase as there were some errors from distclean, hope it clears them, do you think that you can manage the code review till 8.37 ?

rgerhards:
Sry, I just did the final merge :-( I cannot do late merges this time as I am out of office next week and so all release testing needs to be done today.

rgerhards:
@jvymazal  sorry I didn't make the 8.37 milestone, but at least now it is merged! Thanks for your persistence!

rgerhards:
@jvymazal thinking about it, I think the doc should also be amended, at least re the new metadata items. Could you have a look at it?

jvymazal:
@rgerhards there are actually no new metadata items that would need documentation, and the behavior regarding original metadata was intentionally left unchanged (with rsyslog/rsyslog-doc/pull/707 actually explaining the behavior), see discussion above. 
What is new is possibility to easily add new metadata field which would contain the actual, absolute name of the file, but that was not done in scope of this.

rgerhards:
@jvymazal  sorry, I am puzzled again. Maybe I should have looked at that aspect in depth.

@davidelang said

> there will just be some extra variables that they don't do anything with.

I had assumed that this PR adds additional metadata, which a user can use but also can ignore. If the PR does not do that, do we expect a follow-up PR to use the new functionality? Otherwise I am back to the question why the change was made at all, as everything then seemed to work as before, just more complicated. I am sure I overlooked something...

plugins/imfile/imfile.c
```diff
@@ -1162,7 +1222,11 @@ enqLine(act_obj_t *const act,
 	msgSetPRI(pMsg, inst->iFacility | inst->iSeverity);
 	MsgSetRuleset(pMsg, inst->pBindRuleset);
 	if(inst->addMetadata) {
-		metadata_values[0] = (const uchar*)act->name;
+		if (act->source_name) {
```
rgerhards, line 1225: mhhh... Isn't that changing the existing behavior?

jvymazal: No, it is not, this is ensuring that what was changed in the internal structures does *not* affect what user has in metadata.

jvymazal:
> Otherwise I am back to the question why the change was made at all, as everything then seemed to work as before, just more complicated.

The original problem was that the structures before were not awware of the symlonks (we relied on filesystem to somehow make it work) which caused that imfile stopped watching rotated files which were loaded via symlink, this should not happen with the new code.

rgerhards:
> The original problem was that the structures before were not aware of the symlinks (we relied on file-system to somehow make it work) which caused that imfile stopped watching rotated files which were loaded via symlink, this should not happen with the new code.

ahhhh! this should probably have went into the commit message, I guess a lot of things would have run smoother ;-) I now see there is an issue ref in the second message, but these scroll away so soon...

The past couple of weeks (?) I was under the impression this is a functional enhancement.

jvymazal:
> The past couple of weeks (?) I was under the impression this is a functional enhancement.

Well, that depend on how you define bugs and features ;-) , as such rsyslog never had any explicit symlink support so the issue could be regarded as feature request rather than bug...

rgerhards:
> Well, that depend on how you define bugs and features ;-) , as such rsyslog never had any explicit symlink support so the issue could be regarded as feature request rather than bug...

ack - I could also have re-read things more often, but I admit I usually look at the commit message and it was missing the fact that the primary target was to enable something that worked strange for other folks. I should have noticed, but I did not - sry.

My friendly suggestion here is just to make the commit message cristal clear, a link to an existing issue is also usually a life saver (at least for me).

That said, all is well, but I wanted to tell what happened on my side ;-)

jvymazal:
> My friendly suggestion here is just to make the commit message cristal clear, a link to an existing issue is also usually a life saver (at least for me).

Ok, will try to remember that for future PRs

davidelang:
I'll also say that it's common for people to be watching long-running issues 
like this in e-mail (and/or just look at the last couple of messages in a long 
thread), so don't assume that people have the entire context from the beginning 
in their head when looking at a message, try to include enough context (or at 
least a pointer to where to refresh themselves) rather than assuming they will 
go through the entire thread.

David Lang

PR #2750, opened 2018-06-01
Put LimitNOFILE in the correct section of the systemd service file

opoplawski:
Noted by log message:
```
usr/lib/systemd/system/rsyslog.service:21: Unknown lvalue 'LimitNOFILE' in section 'Install
```


rgerhards:
thx for the PR, but it is a duplicate of #2740 which I will hopefully merge soon.

Issue #2742, opened 2018-05-29
Queue corruption when omrelp is blocked and rsyslog force-terminated


StrongestNumber9:
### Expected behavior
Queue file works as expected

### Actual behavior
Seems like events in queuefile gets corrupted / duplicated and eventually the whole queuefile vanishes. Ruleset getting malformed is one easily visible example;

```sh
+pszRuleset:1:6:▒▒=▒:
```

### Steps to reproduce the behavior
Bind imrelp to ruleset which has omrelp action with disk queue configured, and omrelp target cannot be reached. 

### Environment
```sh
rsyslogd -v
rsyslogd 8.32.0, compiled with:
  PLATFORM:       x86_64-unknown-linux-gnu
  PLATFORM (lsb_release -d):
  FEATURE_REGEXP:       Yes
  GSSAPI Kerberos 5 support:    No
  FEATURE_DEBUG (debug build, slow code): Yes
  32bit Atomic operations supported: Yes
  64bit Atomic operations supported: Yes
  memory allocator:     system default
  Runtime Instrumentation (slow code): No
  uuid support:       Yes
  systemd support:      No
  Number of Bits in RainerScript integers: 64

See http://www.rsyslog.com for more information.
```

```sh
CentOS Linux release 7.4.1708 (Core)
```

```sh
Linux snip 3.10.0-514.el7.x86_64 #1 SMP Tue Nov 22 16:42:41 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
```

rgerhards:
an important triggering conditions seems to be that rsyslog is force-terminated. So required order of events:

1. rsyslog starts up
    a. omrelp blocks
    b. rsyslog is force-terminated (e.g. by OS script)
2. rsyslog starts up again
    only know the queue corruptions seems visible.

rgerhards:
@StrongestNumber9 have you ever seen this problem on any run but the second? I mean

- initial startup
- initial force-termination
- 2nd startup
  mangled messages here
- 2nd force-termination
- n+1 startup
  no grabled messages
- ...

I mean if you have seen it in any of the n+1 steps?

StrongestNumber9:
If I remember correctly, the broken ruleset error appeared on just one startup. On further restarts I noticed my queuefiles started having duplicated events, some with broken ruleset:

originally 10 events
restart, events from 10 -> 18 (+8 events)
18 -> 24 (+6)
24 -> 28 (+4)
[...]

Issue #1122, opened 2016-08-19
crash on shutdown with nested rulesets


janmejay:
Here is what happens:
- We terminate all rulesets, we do it in a certain order (this is order of cnfDoObj is called, which is config parsing order)
- Its possible to have a scenario similar to this: 
  ```
  ruleset(name="foo") {
  set $.x = re_extract($hostname, '^(._)-[0-9]+$', 0, 1, '_none*');
  ...
  }

  ruleset(name="bar" queue.size="10000" queue.type="LinkedList" queue.workerthreads="2" 
  queue.discardmark="9900") {
  ...
  call foo
  ...
  }
  ```
- Now, termination leads to foo being destroyed before bar, which means statements in foo-ruleset are destroyed, rainerscript functions (funcdata etc) are all destroyed etc.
- Because bar is still not destroyed and we come to it next, if it has some messages in its queue, it tries to process them, which leads to the crash when it calls into foo.

Various solutions are possible:
- Simplest is to check for termination(glbl.GetGlobalInputTermState() etc) before calling into another ruleset (but this is the most naive too, because then ruleset-queue timeouts etc won't work as expected)
- We can re-look at shutdown sequence, where we turn off inputs first and then wait for a single value of timeout (global timeout config) before we destroy rulesets and actions. Once inputs are turned-off, the longer we leave it running, the more messages it will drain out (assuming actions not suspended for some reason). We don't destroy anything in this duration, so messages can go around queue to queue without any problem, except new messages will not enter the system. Along with this, we report per-queue level what number of messages were still present when we actually started destruction. When destruction starts, we use term-state check at every ruleset call and action invocation boundary and discard any messages trying to cross the boundary.  This is not a problem because user has choice to configure quiesce-period for inputs. This does away with queue level drain-timeouts.
- We keep queue-level timeouts for termination but topo-sort by out ruleset inter-dependency. We destroy rulesets in dependent-before-dependency order. (This is the most backward compatibility friendly way by far, I guess?).

Thoughts?

janmejay:
I will fix it, but can't work on it immediately (it'll take me 2 - 3 weeks to get to this). 

I hope that is OK, because the easy workaround is to never define a dependency before dependent.

janmejay:
Also, 3rd seems like the best option to me.

rgerhards:
Topo-Sort does not reliably work, as we may have circular dependencies (which are terminated by ```if``` constructs).

rgerhards:
[...]

PR #2755, opened 2018-06-05
core: misadressing when writing disk queue files

rgerhards:
when writing disk queue files during shutdown, access to freed
memory can occur under these circumstances:

- action A is processing data, but could not complete it
  most importantly, the current in-process batch needs not to
  be totally completed. Most probable cause for this scenario
  is a suspended action in retry mode.
- action A is called from a ruleset RA which
  - does not have a queue assigned
  - where RA is called from a ruleset RO which is bound
    to the input from which the message originated
  - RO must be defined before RA inside the expanded config
- Disk queues (or the disk part of a DA queue) must be utilized
  by A

When re-injecting the unprocessed messages from A into the disk queue, the
name of ruleset RO is accessed (for persisting to disk). However, RO is
already destructed at this point in time.

The patch changes the shutdown processing of rulesets, so that all
shutdown processing is done before any ruleset data is destructed. This
ensures that all data items which potentially need to be accessed
remain valid as long as some part may potentially try to access them.

This follows a the approach used in
   https://github.com/rsyslog/rsyslog/pull/1857
where obviously that part of the problem was not noticed.

see also https://github.com/rsyslog/rsyslog/issues/1122
closes https://github.com/rsyslog/rsyslog/issues/2742

### DIFF
--- a/runtime/ruleset.c
+++ b/runtime/ruleset.c
@@ -832,7 +832,6 @@ CODESTARTobjDestruct(ruleset)
 		parser.DestructParserList(&pThis->pParserLst);
 	}
 	free(pThis->pszName);
-	cnfstmtDestructLst(pThis->root);
 ENDobjDestruct(ruleset)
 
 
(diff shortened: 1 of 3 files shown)


rgerhards:
I have now check the centos6 condition var valgrind errors very much in-depth, even did a clang thread and address sanitizer run on that exact some machine (without any problems). Manually instrumented the module, review debug logs, etc... I finally come to the conclusion that this is a valgrind false positive, as unlikely as it sounds so (pretty old valgrind on that platform, of course).

That means I'll see that I change the CI system to add suppressions for that test (or take some similar action).

PR #2761, opened 2018-06-07
add openssl TLS driver

alorbach:
Initial Implementation of openssl nsd drivers for rsyslog.


rgerhards:
We should add a test for malicious certs with oversize subject names to the testbench. This case was so far missing. I have test data at hand.

rgerhards:
I am merging the openssl driver without in-depth review, as I would like to get it out to the world and hope for feedback, I did a quick review, though, and don't see potential for regressions in non-openssl parts.

I plan to revisit the openssl driver project, including adding additional tests.

PR #2764, opened 2018-06-07
use version=2 in mmkubernetes rulebase files

richm:
We see memory leaks in liblognorm in older versions of
liblognorm unless we explicitly specify version=2


davidelang:
There are also race conditions and other data corruption bugs in the older 
versions of liblognorm and libfastjson.

David Lang

rgerhards:
travis fails as it is unable to install some dependencies ... experience tells this usually clears up within 24hrs. I'll keep trying, else merge with the CI fail

PR #2766, opened 2018-06-11
Fixed empty programname if checked before APP-NAME has been properly copied to TAG when receiving using RFC5424 log lines.

robdewit:
Fixed empty programname if checked before APP-NAME has been properly copied to TAG when receiving using RFC5424 log lines.

Fixes this:
```
template (name="DynamicMessagesFile" type="string" string="/tmp/d/messages-%programname%.log")
*.info if(programname == "xxx") {
        action(type="omfile" File="/tmp/d/test.log")
    } else {
        action(type="omfile" dynaFile="DynamicMessagesFile")
    }
```

Which would produce a file calles messages-.log because programname is empty when receiving RFC5424 log line.

Fixed by singnalling change in TAG when APP-NAME has been parsed and copied to TAG.


robdewit:
Although programname ends up correctly in debug output, the filename in the template is still empty.

Issue #2760, opened 2018-06-06
queued messages are duplicated until queue file is mis-deallocated when output blocked


kortemik:
### Expected behavior
Messages should stay in queue without duplication and queue should not be deallocated before all messages in the queue have been processed
### Actual behavior
Messages are duplicated between restarts when output is blocked, related to #2742 
### Steps to reproduce the behavior
Container image with configuration and binaries provided via email, sources are available as well.
### Environment
- rsyslog version: 
```
rsyslogd 8.32.0, compiled with:
        PLATFORM:                               x86_64-unknown-linux-gnu
        PLATFORM (lsb_release -d):
        FEATURE_REGEXP:                         Yes
        GSSAPI Kerberos 5 support:              No
        FEATURE_DEBUG (debug build, slow code): Yes                                                                                                                                                                                                        
        32bit Atomic operations supported:      Yes                                                                                                                                                                                                        
        64bit Atomic operations supported:      Yes
        memory allocator:                       system default
        Runtime Instrumentation (slow code):    No
        uuid support:                           Yes
        systemd support:                        No
        Number of Bits in RainerScript integers: 64

See http://www.rsyslog.com for more information.
```
- platform:
```
CentOS Linux release 7.4.1708 (Core) 
```
- for configuration questions/issues, include rsyslog.conf and included config files
Provided with-in container image.

rgerhards:
please note: duplication can occur depending on circumstances. So that part is within the spec.

rgerhards:
I possibly have good news: I think I can reproduce a case where message loss occurs during each start. Need to check it's not a bug in the test, but so far looks good. Not sure if it translates to what you see, but I guess that's at least progress.

rgerhards:
commit https://github.com/rgerhards/rsyslog/commit/1ea3be5198aa7f7d1d06055d0e8077b42eab300a most probably is the solution for this problem. I cannot merge at the moment, as the previous patch (that it requires) seems to have run into some CI issues, which I think are actually testbench problems. Still need to check.

In any case, you can apply the patch and run it in your environment. 98% sure it will work - I had it successfully run through Travis.

StrongestNumber9:
First impression is that this patch fixes the problem. 25 events in -> restarts & duplication n times -> 25 events out, clean queue file removal once all got processed.

PR #2767, opened 2018-06-11
core: fix message loss on target unavailibility during shutdown

rgerhards:
Triggering condition:
- action queue in disk mode (or DA)
- batch is being processed by failed action in retry mode
- rsyslog is shut down without resuming action

In these cases messages may be lost by not properly writing them
back to the disk queue.

closes https://github.com/rsyslog/rsyslog/issues/2760

### DIFF
--- a/action.c
+++ b/action.c
@@ -1554,8 +1554,15 @@ processBatchMain(void *__restrict__ const pVoid,
 			/* we do not check error state below, because aborting would be
 			 * more harmful than continuing.
 			 */
-			processMsgMain(pAction, pWti, pBatch->pElem[i].pMsg, &ttNow);
-			batchSetElemState(pBatch, i, BATCH_STATE_COMM);
+			rsRetVal localRet = processMsgMain(pAction, pWti, pBatch->pElem[i].pMsg, &ttNow);
+			DBGPRINTF("processBatchMain: i %d, processMsgMain iRet %d\n", i, localRet);
+			if(   localRet == RS_RET_OK
+			   || localRet == RS_RET_DEFER_COMMIT
+			   || localRet == RS_RET_ACTION_FAILED
+			   || localRet == RS_RET_PREVIOUS_COMMITTED ) {
+				batchSetElemState(pBatch, i, BATCH_STATE_COMM);
+				DBGPRINTF("processBatchMain: i %d, COMM state set\n", i);
+			}
 		}
 	}
 
--- a/runtime/queue.c
+++ b/runtime/queue.c
@@ -1666,6 +1666,7 @@ DeleteProcessedBatch(qqueue_t *pThis, batch_t *pBatch)
 
 	for(i = 0 ; i < pBatch->nElem ; ++i) {
 		pMsg = pBatch->pElem[i].pMsg;
+		DBGPRINTF("DeleteProcessedBatch: etry %d state %d\n", i, pBatch->eltState[i]);
 		if(   pBatch->eltState[i] == BATCH_STATE_RDY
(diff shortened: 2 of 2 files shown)


jvymazal:
Hi @rgerhards , unfortunately I found out that this commit causes double sending of first message put into DA queue in situation when remote TCP rsyslog target is not available. Do you please have any idea what could cause it? Do you want to file an issue for it?

rgerhards:
It sounds like it may be inevitable. We prefer duplication over loss. I *guess* (but only guess!) that rsyslog cannot detect what was actually received at the remote end. So the message gets "unprocessed" flag, which means it is resubmitted.

jvymazal:
Ok, many thanks for info and pointer.

sopos:
Just to clarify. This scenario work as follows:
1. server and client are stopped
2. client is started
3. massages are logged, so they get the DA queue
4. server is started
5. the queue starts to be flushed, while the second logged message in step3 is duplicated

It seems to me quite weird that in this simple scenario a message gets "unprocessed" flag. I would understand If it was some network issue during the transmission.

@rgerhards can you clarify the mechanism of DA queues processing in a nutshell?

rgerhards:
> It seems to me quite weird that in this simple scenario a message gets "unprocessed" flag.

please read this and follow up postings: https://rainer.gerhards.net/2008/04/on-unreliability-of-plain-tcp-syslog.html

sopos:
But this case is not about problem during the communication. It is about failed connect at the beginning and then, after the server gets up, the connection is established and data flows without any disruptions.
I'm do not have any proof of that but I feel this might be some more generic issue around the DA queues.. just consider that.
Otherwise one duplicated messages is definitely better than one lost message and I fully agree RELP should be used where the accurate messages delivery is required.

rgerhards:
@sopos I don't want to re-iterate an 10+ year old discussion, especially as all the findings are very well documented and can be found by following the link I posted.

But let me ask a simple question: how do you think rsyslog can now that this message has been successfully processed and can be dropped?

sopos:
> But let me ask a simple question: how do you think rsyslog can now that this message has been successfully processed and can be dropped?
That's question I do not know the answer to. But if the initial connect failed (server side is down) it must be obvious that none of the messages were delivered. So why there's one delivered twice after the retry when server gets up..

rgerhards:
@sopos I suggest you craft a testbench test that shows the behavior. Then I know exactly what you are talking about and I have something to work with. Please create a separate PR for this test. Let's further discuss findings in that PR.

davidelang:
On Wed, 31 Jul 2019, Rainer Gerhards wrote:

> @sopos I don't want to re-iterate an 10+ year old discussion, especially as all the findings are very well documented and can be found by following the link I posted.
>
> But let me ask a simple question: how do you think rsyslog can now that this message has been successfully processed and can be dropped?

as I am understanding the report:

1. the relp sender is running (the message is in a queue file)

2. the relp receiver starts up

3. the first message gets sent twice

this sounds like there's something in the initial handshake or processing of the 
first message that isn't going right. I see no reason the ack should be failing 
to get through.

am I stating the problem correctly?

David Lang

rgerhards:
This commit deals with send failure conditions. So the scenario does not apply.

sopos:
> as I am understanding the report: 1. the relp sender is running (the message is in a queue file) 2. the relp receiver starts up 3. the first message gets sent twice this sounds like there's something in the initial handshake or processing of the first message that isn't going right. I see no reason the ack should be failing to get through. am I stating the problem correctly? David Lang

Actually I found the issue using omfwd.

I will will try to create a test as Rainer suggested. I will plan to do that in some of the future sprints.

PR #2774, opened 2018-06-13
[WIP] "fromhost" property's case is lowered when using UDP

nhosoi:
When imudp is used for input, the FROMHOST value is all lowered:
`FROMHOST: 'myclient.example.lab', fromhost-ip: '192.168.122.111', HOSTNAME: 'myCLIENT', ...`
while the case is preserved if it is imtcp:
`FROMHOST: 'myCLIENT.example.lab', fromhost-ip: '192.168.122.111', HOSTNAME: 'myCLIENT', ...`
where the host name is defined as "myCLIENT.example.lab myCLIENT" in /etc/hosts.

These 2 behaviours are preferably matched.

My concern is resolveDNS is a common helper function in runtime/msg.c, which is being called not just from getRcvFrom but multiple places.  The rest may expect all lowercase localName.

### DIFF
--- a/plugins/imtcp/imtcp.c
+++ b/plugins/imtcp/imtcp.c
@@ -100,6 +100,7 @@ static struct configSettings_s {
 	int bDisableLFDelim;
 	int discardTruncatedMsg;
 	int bUseFlowControl;
+	int bPreserveCase;
 	uchar *gnutlsPriorityString;
 	uchar *pszStrmDrvrAuthMode;
 	uchar *pszInputName;
@@ -144,6 +145,7 @@ struct modConfData_s {
 	uchar *pszStrmDrvrAuthMode; /* authentication mode to use */
 	struct cnfarray *permittedPeers;
 	sbool configSetViaV2Method;
+	sbool bPreserveCase; /* preserve case of fromhost; true by default */
 };
 
 static modConfData_t *loadModConf = NULL;/* modConf ptr to use for the current load process */
@@ -169,7 +171,8 @@ static struct cnfparamdescr modpdescr[] = {
 	{ "keepalive.probes", eCmdHdlrPositiveInt, 0 },
 	{ "keepalive.time", eCmdHdlrPositiveInt, 0 },
 	{ "keepalive.interval", eCmdHdlrPositiveInt, 0 },
-	{ "gnutlsprioritystring", eCmdHdlrString, 0 }
+	{ "gnutlsprioritystring", eCmdHdlrString, 0 },
+	{ "preservecase", eCmdHdlrBinary, 0 }
 };
 static struct cnfparamblk modpblk =
 	{ CNFPARAMBLK_VERSION,
@@ -375,6 +378,7 @@ addListner(modConfData_t *modConf, instanceConf_t *inst)
 		if(pPermPeersRoot != NULL) {
 			CHKiRet(tcpsrv.SetDrvrPermPeers(pOurTcpsrv, pPermPeersRoot));
 		}
+		CHKiRet(tcpsrv.SetPreserveCase(pOurTcpsrv, modConf->bPreserveCase));
 	}
 
 	/* initialized, now add socket and listener params */
@@ -473,6 +477,7 @@ CODESTARTbeginCnfLoad
 	loadModConf->pszStrmDrvrAuthMode = NULL;
 	loadModConf->permittedPeers = NULL;
 	loadModConf->configSetViaV2Method = 0;
+	loadModConf->bPreserveCase = 1; /* default to true */
 	bLegacyCnfModGlobalsPermitted = 1;
 	/* init legacy config variables */
 	cs.pszStrmDrvrAuthMode = NULL;
@@ -543,6 +548,8 @@ CODESTARTsetModCnf
 			loadModConf->pszStrmDrvrName = (uchar*)es_str2cstr(pvals[i].val.d.estr, NULL);
 		} else if(!strcmp(modpblk.descr[i].name, "permittedpeer")) {
 			loadModConf->permittedPeers = cnfarrayDup(pvals[i].val.d.ar);
+		} else if(!strcmp(modpblk.descr[i].name, "preservecase")) {
+			loadModConf->bPreserveCase = (int) pvals[i].val.d.n;
 		} else {
 			dbgprintf("imtcp: program error, non-handled "
 			  "param '%s' in beginCnfLoad\n", modpblk.descr[i].name);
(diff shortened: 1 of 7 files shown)


davidelang:
hostnames and DNS names are defined to not be case sensitive, so it should be 
just fine to have things squashed to lower case.

if you have "myCLIENT.example.lab myCLIENT" in /etc/hosts, you can ping myclient 
or MYCLIENT or MyClIeNt and it will work.

why should it matter?

David Lang
k

nhosoi:
We got a bug report which does not clearly state why they need to preserve the cases.

This is a provided example.  They are using a machine generated host name.  There could be some software which requires the case information.
```
/etc/hosts: 10.11.22.33 A0043MtStgRtr01
logfile: May 11 00:00:00 a0043mtstgrtr01 34627: ...
```

davidelang:
On Wed, 13 Jun 2018, Noriko Hosoi wrote:

> We got a bug report which does not clearly state why they need to preserve the cases.
>
> This is a provided example.  They are using a machine generated host name.  There could be some software which requires the case information.
> ```
> /etc/hosts: 10.11.22.33 A0043MtStgRtr01
> logfile: May 11 00:00:00 a0043mtstgrtr01 34627: ...
> ```

please try to find more info. Anything that depends on hostname/dns information 
being case sensitive is going to break in all sorts of ugly ways...

nhosoi:
Thank you, @davidelang.  Understood.
> please try to find more info. 

I'm asking them for more information.

rgerhards:
> These 2 behaviours are preferably matched.
> 
> My concern is resolveDNS is a common helper function in runtime/msg.c, which is being called not just from getRcvFrom but multiple places. The rest may expect all lowercase localName.

It's interesting that actually two different values are returned, depending on module. I would say that's not good and also not intended. Maybe we should file a specific issue for that.

rgerhards:
I am not sure we can merge this PR as is: we need to evaluate implications on existing configs. I guess this at least needs to be configurable.

nhosoi:
Thank you for your comments, @rgerhards.  I agree with you.  Changing the behaviour in resolveDNS should be risky, which may break existing software that depends upon it.

If we are to introduce a new option for preserving the case to imudp, we should do the same to imtcp, as well?  (And to maintain the current behaviour, we should set the different default value to each?  It's getting complexer... :)

nhosoi:
Hi @davidelang, Hi @rgerhards,

I've added an option preservecase to imudp and imtcp module for managing the case of FROMHOST value.  Could you please take a look at the latest patch?  The new patch failed the CI tests, though.  It looks to me the failure did not come from the change I made, but I'm not 100% sure.  If you could give me any advice about the test failure, I'd appreciate it, too.

```
Usage:
module(load="imudp" [preservecase="on"|"off"])
module(load="imtdp" [preservecase="on"|"off"])

If preservecase="on", FROMHOST value is handled in the case sensitive manner.
If preservecase="off", FROMHOST value is handled in the case insensitive manner.

To maintain the current behaviour, the default value of preservecase is "on" for imtcp and "off" for imudp.

See also
https://bugzilla.redhat.com/show_bug.cgi?id=1309698
```

rgerhards:
I am currently working heavily on the CI and will handle issues. If some actually need your attention, I let you know.

Please also create a PR in https://github.com/rsyslog/rsyslog-doc for doc update.

rgerhards:
yeah, you need to do a rebase after I have merged https://github.com/rsyslog/rsyslog/pull/2835 -- I'll let you know. That should solve most of the false positives.

rgerhards:
ok, it was already run through. Please do a rebase when convenient. Note that CI may still fail, probably for a couple of days.

plugins/imtcp/imtcp.c
```diff
@@ -170,7 +172,8 @@ static struct cnfparamdescr modpdescr[] = {
 	{ "keepalive.probes", eCmdHdlrPositiveInt, 0 },
 	{ "keepalive.time", eCmdHdlrPositiveInt, 0 },
 	{ "keepalive.interval", eCmdHdlrPositiveInt, 0 },
-	{ "gnutlsprioritystring", eCmdHdlrString, 0 }
+	{ "gnutlsprioritystring", eCmdHdlrString, 0 },
+	{ "preservecase", eCmdHdlrBinary, 0 },
```
rgerhards: the comma needs to go away ;-)

nhosoi: *Shame shame*  Fixed.  Thanks!

nhosoi:
Thanks, @rgerhards!  I've rebased and pushed the patch.  I'm openining the doc pr, next.

runtime/tcpsrv.h
```diff
@@ -177,6 +178,8 @@ BEGINinterface(tcpsrv) /* name must also be changed in ENDinterface macro! */
 	rsRetVal (*SetbSPFramingFix)(tcpsrv_t*, sbool);
 	/* added v19 -- PascalWithopf, 2017-08-08 */
 	rsRetVal (*SetGnutlsPriorityString)(tcpsrv_t*, uchar*);
+	/* added v?? -- Preserve case in fromhost, 2018-07-?? */
+	rsRetVal (*SetPreserveCase)(tcpsrv_t *pThis, int bPreserveCase);
 ENDinterface(tcpsrv)
 #define tcpsrvCURR_IF_VERSION 20 /* increment whenever you change the interface structure! */
```
rgerhards: while this interface is no longer really enforced, the IF_VERSION should be incremented on interface change. So 21 would be the correct value now.

nhosoi: Thank you, @rgerhards!  I've updated tcpsrv.h and rebased.

runtime/tcpsrv.h
```diff
@@ -177,6 +178,8 @@ BEGINinterface(tcpsrv) /* name must also be changed in ENDinterface macro! */
 	rsRetVal (*SetbSPFramingFix)(tcpsrv_t*, sbool);
 	/* added v19 -- PascalWithopf, 2017-08-08 */
 	rsRetVal (*SetGnutlsPriorityString)(tcpsrv_t*, uchar*);
+	/* added v21 -- Preserve case in fromhost, 2018-08-15 */
+	rsRetVal (*SetPreserveCase)(tcpsrv_t *pThis, int bPreserveCase);
 ENDinterface(tcpsrv)
 #define tcpsrvCURR_IF_VERSION 20 /* increment whenever you change the interface structure! */
```
rgerhards: I guess you overlooked this line ;-) ```#define tcpsrvCURR_IF_VERSION 20```

nhosoi:
Ahhh, sorry about it.  Replaced v21 with v20...

rgerhards:
> Ahhh, sorry about it. Replaced v21 with v20...

mhhh... I  still don't see the 21 in the definition... push forgotten?

rgerhards:
> So sorry... I've pushed a new patch...

not at all a problem sh.. happens ;-) Thx for your persistence, I guess we are now pretty close to merge :-)

PR #2775, opened 2018-06-13
imjournal: journal visibility through additional impstats counters

abwaheed:
-- Adds impstats counters for messages read from journal, submitted to main queue, and dropped due to rate-limiting
-- Also, adds counters for journal file rotations and disk usage

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -720,7 +720,7 @@ CODESTARTrunInput
 		 */
 		const int e = sd_journal_get_usage(j, (uint64_t *)&statsCounter.diskUsageBytes);
 		if (e < 0) {
-			LogError(e, RS_RET_ERR, "imjournal: sd_get_usage() failed");
+			LogError(-e, RS_RET_ERR, "imjournal: sd_get_usage() failed");
 		}
 
 		if (readjournal() != RS_RET_OK) {
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -718,8 +718,9 @@ CODESTARTrunInput
 		/*
 		 * update journal disk usage before reading the new message.
 		 */
-		if (sd_journal_get_usage(j, (uint64_t *)&statsCounter.diskUsageBytes) < 0) {
-			LogError(0, RS_RET_ERR, "imjournal: sd_get_usage() failed");
+		const int e = sd_journal_get_usage(j, (uint64_t *)&statsCounter.diskUsageBytes);
+		if (e < 0) {
+			LogError(e, RS_RET_ERR, "imjournal: sd_get_usage() failed");
 		}
 
 		if (readjournal() != RS_RET_OK) {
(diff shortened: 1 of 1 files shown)


davidelang:
Please ad a commit message to this.

David Lang

plugins/imjournal/imjournal.c
```diff
@@ -251,10 +260,26 @@ int sharedJsonProperties)
...
 	}
 
+	discardCount = ratelimiter->missed;
 	CHKiRet(ratelimitAddMsg(ratelimiter, NULL, pMsg));
-	STATSCOUNTER_INC(statsCounter.ctrSubmit, statsCounter.mutCtrSubmit);
 
 finalize_it:
+/*
+ * Since ratelimitAddMsg return value does not provide RS_RET_DISCARDMSG
+ * for all cases, we rely on ratelimiter->missed and ratelimiter->done
```
rgerhards: can you elaborate on this? Is that a bug that needs fixing?

abwaheed: I initially implemented this part based on whether return value is RS_RET_DISCARDMSG and was not seeing discard count change even when it was expected to change. That's when I noticed that ratelimitAddMsg() does not return this value in all cases. 

However, I noticed that this is not true any more. I checked the latest and see it is fixed already:
https://github.com/rsyslog/rsyslog/blob/master/runtime/ratelimit.c#L270

Problem is with my local version that is a few months old. 
I'll update this code. Thanks for catching it.

plugins/imjournal/imjournal.c
```diff
@@ -698,13 +715,20 @@ CODESTARTrunInput
 			continue;
 		}
 
+		/*
+		 * update journal disk usage before reading the new message.
+		 */
+		if (sd_journal_get_usage(j, (uint64_t *)&statsCounter.diskUsageBytes) < 0) {
+			LogError(0, RS_RET_ERR, "imjournal: sd_get_usage() failed");
```
rgerhards: we shoud add the -result of sd_journal_get_usage to errmsg so that we get a better error messag

```
const int e = sd_jour...
....
LogError(e, RS_...)
```

plugins/imjournal/imjournal.c
```diff
@@ -718,8 +718,9 @@ CODESTARTrunInput
 		/*
 		 * update journal disk usage before reading the new message.
 		 */
-		if (sd_journal_get_usage(j, (uint64_t *)&statsCounter.diskUsageBytes) < 0) {
-			LogError(0, RS_RET_ERR, "imjournal: sd_get_usage() failed");
+		const int e = sd_journal_get_usage(j, (uint64_t *)&statsCounter.diskUsageBytes);
+		if (e < 0) {
+			LogError(e, RS_RET_ERR, "imjournal: sd_get_usage() failed");
```
rgerhards: my fault - this needs to be LogError(-e, ...) as of journal doc.

abwaheed:
I changed to -e. Also, you should have permissions to over-write my branch.

rgerhards:
thx again for the PR!

Issue #2789, opened 2018-06-22
ommail documentation examples incorrect


hmoffatt:
### Expected behavior
rsyslog should sent emails with a body following the specified template, when using the example configuration in the documentation: https://www.rsyslog.com/doc/v8-stable/configuration/modules/ommail.html

### Actual behavior
email body is blank.

### Steps to reproduce the behavior
Use the example configuration, log a matching message.

It appears that you need to explicitly set the body template using template="mailBody" on the action clause to make this work.

### Environment
- rsyslog version: 8.34.0
- platform: debian stretch

PascalWithopf:
The default for the parameter enable.body was wrong, so the body was disabled.

You will need to set ```enable.body="on"``` or wait for my pull request to be merged.

PR #2791, opened 2018-06-25
ommail: enable.body parameter default set correctly

PascalWithopf:
When sending a message via ommail the email body was
not sent unless enable.body was specified to "on".
Now the default is "on" and it only needs to be specified
if you don't want the body part to be sent.

Closes https://github.com/rsyslog/rsyslog/issues/2789

### DIFF
--- a/plugins/ommail/ommail.c
+++ b/plugins/ommail/ommail.c
@@ -259,6 +259,7 @@ finalize_it:
 BEGINcreateInstance
 CODESTARTcreateInstance
 	pData->constSubject = NULL;
+	pData->bEnableBody = 1;
 ENDcreateInstance
 
 


davidelang:
On Mon, 25 Jun 2018, Rainer Gerhards wrote:

> rgerhards requested changes on this pull request.
>
> If the previous default was "off" (even if so in error), we should update the doc. Changing the default will break existing deployments, which is against our policy.

what happens if enable.body is on, but no template for the body is provided? If 
that will send out the message with an empty body, it's the same result as 
before.

I would argue that enable.body=off and a template provided is an error to be 
flagged.

The only time it would 'break' a config (assuming it doesn't blow up in the on, 
but no body template specified) is if there is a template for the body and the 
admin was using the default to surpress sending it.

And in that case, the 'breakage' is that there is a body that they don't care 
about, hardly an earth-shattering error.

Now, if rsyslog throws an error if body-on and no template provided, the fix 
would be to make that case send a blank body.

rgerhards:
@davidelang 

> what happens if enable.body is on, but no template for the body is provided?

that's not problem, it's meant like in the other modules: the default template is used, in this case RSYSLOG_Fileformat. The issue really was the wrong default, which did not match the doc. As that default is in place for ~ 10 yrs, I don't think we should change it.

davidelang:
On Tue, 26 Jun 2018, Rainer Gerhards wrote:

> @davidelang
>
>> what happens if enable.body is on, but no template for the body is provided?
>
> that's not problem, it's meant like in the other modules: the default template is used, in this case RSYSLOG_Fileformat. The issue really was the wrong default, which did not match the doc. As that default is in place for ~ 10 yrs, I don't think we should change it.

normally I would agree with you, but in this case I think it's acceptable to 
change the default.

In this case, changing the default will have the effect of making the e-mail 
larger, but is unlikely to change any uses of the e-mail. If people are 
processing the message and only looking at the subject, everything will still 
work.

Someone would have to have automated parsing of the message, with the parsing 
done in a way that breaks if there is a body, before adding a body would hurt 
anything other than message size (which is variable depending on what headers 
are added during processing anyway)

I also don't think that it's going to be very common for people to be using 
ommail and NOT setting a body.

So we are talking a corner case of a corner case that has a risk of breakage, 
with the trade-off being the unexpected default of not sending a message

We could even consider making the default more complicated, if the body.enable 
flag is not explicitly set, and no body.template is specified, don't send a 
message, but if body.enable is not explicitly set and a body.template is 
specified, send it.

The corner case of a corner case of a corner case where someone deliberately 
specifies a body.template and doesn't want the template to be used is getting so 
rare that I think it's worth breaking in order to make the behavior match the 
expected (and documented) behavior.

In the obsolete-legacy config format, you have no way of knowing if a template 
has been explicitly set, in the current action() format, you know what's set. I 
don't see a good reason (other than backwards compatibility) for the body.enable 
flag to exist in the action() format.

David Lang

rgerhards:
The subject only mode was originally made for text paging gateways
(including SMS), which at that time often had problems with the body. But
probably not a concern today...

So I would agree with merging this and let's see what happens...

Rainer

Sent from phone, thus brief.

David Lang <notifications@github.com> schrieb am Di., 26. Juni 2018, 20:40:

Issue #2741, opened 2018-05-28
Asynchronous logging may deadlock


rmetrich:
### Expected behavior

No deadlock when using Asynchronous logging

### Actual behavior

Deadlock of AsyncWriterThread + deadlock of Main thread (after sending SIGHUP, typically done by logrotate).
I wasn't able to reproduce the issue using the regular rsyslog code, but one of Red Hat's customers is hitting this regularly.
To reproduce in my environment, I modified slightly the rsyslog code (see below) to show the window where the issue happens.
Please refer to the commit description.

### Steps to reproduce the behavior

See branch [Issue_1701_Reproducer](https://github.com/rmetrich/rsyslog/tree/Issue_1701_Reproducer) which modifies rsyslog code to very easily reproduce.
Reading #1701, this looks similar, but there is no stack trace in the issue to be sure, and also the issue has been closed.

For now I cannot propose a fix because I don't understand exactly how async writer is supposed to work.

### Environment
- rsyslog version: latest upstream (8.36.0.master)
- platform: Fedora 27, RHEL 7

rgerhards:
fixed by https://github.com/rsyslog/rsyslog/pull/2794 in v8.37

Issue #1701, opened 2017-08-01
rsyslog stops writing logs until restarted


jhenry82:
Apologies in advance for what is going to start out as a vague bug report. I'm very interested in next steps I can provide to help flesh this out, as it is making rsyslog unusable for us.

rsyslog: 8.26
librelp: 1.2.13
CentOS 6.9 x86_64

We have 2 servers that receive logs from about 1000 remote hosts via RELP. The volume is large but not insane, maybe in the dozens of gigabytes per hour range. After a seemingly random amount of time (days to weeks), one of these servers will simply stop writing logs until rsyslog is restarted. Messages continue to be received by RELP and written to disk-assisted queues. But the omfile action that writes them to their final destination seems just infinitely blocked. No errors of any kind are logged by rsyslog when this happens. I do have impstats configured to write to ElasticSearch periodically, and these stats continue to update while the issue is happening.

I monitor open file descriptors and we're only using about 25% of the configured limit, so it shouldn't be that. There's nothing remarkable happening on the box when this happens. System metrics look normal other than memory usage spiking, which seems due to messages starting to pile up in the main queue.

Here is the full config on the affected servers. Is there something obviously boneheaded about the actions or main queue?

rsyslog.conf
```
#################
#### MODULES ####
#################

module(load="imuxsock")  # provides support for local system logging
module(load="imklog")    # provides kernel logging support (previously done by rklogd)
module(load="immark")    # provides --MARK-- message capability
module(load="imfile")    # provides support for tailing local files
module(load="omrelp")    # provides support for RELP protocol

###########################
#### GLOBAL DIRECTIVES ####
###########################

global(
	preserveFQDN="on"
	workDirectory="/var/log/rsyslog_spool"
	umask="0000"
	action.reportSuspensionContinuation="on"
	janitor.interval="1"
)

# Use default timestamp format, but allow app names (syslogtag) longer than 32 characters
# This prevents weird truncated filenames on the log servers, and in Logstash
template(name="LongTagForwardFormat" type="string"
	string="<%PRI%>%TIMESTAMP% %HOSTNAME% %syslogtag%%msg:::sp-if-no-1st-sp%%msg%"
)

# Set the default permissions for all log files.
module(
	load="builtin:omfile"
	dirCreateMode="0755"
	fileCreateMode="0644"
	fileOwner="root"
	fileGroup="root"
	dirOwner="root"
	dirGroup="root"
)

##########################
#### STATS MONITORING ####
##########################
[...]

PR #2794, opened 2018-06-26
Fixed race in asyncWriter

radosroka:
strmFlushInternal and especially doAsyncWriteInternal needs to be called with locked mutex as it is in the rest of code. It changes stream object and that should be done when mutex was correctly locked.

Another problem is that doAsyncWriteInternal containts pthread_cond_wait which specifically requires locked mutex.


radosroka:
Fixes #2741.

radosroka:
Related #1701.

rgerhards:
Travis fails due unavailability of some PPA. Will babysit this.

PR #2795, opened 2018-06-26
omrabbitmq: one pthread mutex per connection

micoq:
A different pthread mutex is created for each connection (action) instead of a single one shared by all connections.
This will improve performance when using multiple concurrent connections to a single (or multiple) RabbitMQ instance(s) (e.g. for load balancing)

### DIFF
--- a/contrib/omrabbitmq/omrabbitmq.c
+++ b/contrib/omrabbitmq/omrabbitmq.c
@@ -55,8 +55,6 @@ MODULE_CNFNAME("omrabbitmq")
  */
 DEF_OMOD_STATIC_DATA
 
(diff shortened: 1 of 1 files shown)


rgerhards:
thx for the PR, much appreciated.

Note to all: this is a contributed module, so I cannot do in-depth review. But patch and reasoning sound convincing to me, so I will merge if CI does not bring up anything.

rgerhards:
Note: the CentOS6 CI failure is caused by some work on the CI system. Will babysit this myself.

rgerhards:
same can happen with the docker slave...

PR #2798, opened 2018-06-26
Don't link tcpflood against -lgcrypt

mbiebl:
While older GnuTLS versions do use gcrypt, newer version no longer do.

From
https://www.gnutls.org/manual/html_node/Cryptographic-Backend.html

"
Cryptographic library layer

The Cryptographic library layer, currently supports only libnettle.
Older versions of GnuTLS used to support libgcrypt, but it was switched
with nettle mainly for performance reasons and secondary because it is
a simpler library to use. In the future other cryptographic libraries
might be supported as well.
"


rgerhards:
docker ubuntu16 slave is experimental, so CI is clean

Issue #2744, opened 2018-06-01
Invalid config statements recreating socket


jvymazal:
### Expected behavior
More explicit error report/failure when this happens. It is very hard to find out what happened in this case. You get basically no hint to what is wrong if you do config typo like the one below.

### Actual behavior
restarting rsyslog recreates /dev/log socket, possibly breaking other input (such as imjournal)

### Steps to reproduce the behavior
-  apply below config
- `stat /dev/log`
- restart rsyslog
- `stat /dev/log` - you see that file is recreated
- `logger test` - message does not get trough

### Environment
- rsyslog version: 8.35.0
- platform: fedora 27
- for configuration questions/issues, include rsyslog.conf and included config files
relevant conf snippet:
`$ModLoad imuxsock`
`$ModLoad imjournal`
`set $var;`

jvymazal:
@rgerhards any ideas on how to make it more clear? Maybe complete abort on such config problems?

jvymazal:
No, they can co-function if configured properly, what I meant is that if you do some kinds of config typos (like the `set $var;` in the example) you break logging and it is very hard to find out what is wrong because rsyslog appears to function.

rgerhards:
there is an ```$AbortOnUncleanConfig``` to make rsyslog abort if something is not right. The core idea - and requirement for many years - was that it is better to keep rsyslog running vs shut it down on all errors. This is what the option controls. See https://www.rsyslog.com/doc/v8-stable/configuration/global/options/rsconf1_abortonuncleanconfig.html

I fear there is no new-style equivalent at the moment.

jvymazal:
Ok, understood, however the new-style equivalent would be nice.

rgerhards:
> however the new-style equivalent would be nice.

definitely!

PR #2803, opened 2018-06-27
add global directive "abortOnUncleanConfig"

rgerhards:
This provides a new-style alternative to $AbortOnUncleanConfig.
Note that a testbench test was changed to the new syntax. Adding
an additional test did not look useful, as the testbench still
sufficiently tests old and new method.

closes https://github.com/rsyslog/rsyslog/issues/2744


rgerhards:
@PascalWithopf could you check how we could best update the doc? I guess there should be some cross-reference between the global() object and the old style definition. Not sure if we already did this in some other case.

PascalWithopf:
@rgerhards I added the documentation for the parameter and a note in the documentation for the legacy parameter that it is also available in the new style.

PR #2806, opened 2018-06-28
testbench: refactor tests using nettester

PascalWithopf:
@rgerhards edit: Some old tests are carried out via the ```nettester``` tool. This was our initial shot at a testbench. While it worked back then, the testbench framework has been much enhanced. These old tests are nowadays very hard to handle, as they miss debug support etc. So it is time to refactor them to new style.

Tests renewed:
timestamp.sh
proprepltest.sh
parsertest.sh


tests/tabescape_dflt.sh
```diff
@@ -1,18 +1,29 @@
...
-  exit 1
-fi
+ruleset(name="ruleset1") {
+	action(type="omfile" file="rsyslog.out.log"
+	       template="outfmt")
+}
+
+'
+. $srcdir/diag.sh startup
+. $srcdir/diag.sh tcpflood -m1 -M "\"<167>Mar  6 16:57:54 172.20.245.8 test: before HT	after HT (do NOT remove TAB!)\""
```
rgerhards, line 18: Is there actually a tab character present? Doesn't look so.

ghost: There is a tab in front of "after". It doesn't look like it but I have confirmed this.

rgerhards:
@PascalWithopf unfortunately the refactored tests do not work correctly. As I know you are away, I will fix them and also add some plumbing. I'll re-write the commit.

Issue #2800, opened 2018-06-27
CI: reduce Travis runtime


rgerhards:
While it is nice to have Travis do checks as early as possible, the runtime is becoming very long and actually this hinders working with PRs. So we should offload some Travis tests to Buildbot. We should try to do this with those tests that are least likely to happen (so that users can still get a good experience in their personal testing via Travis before submitting a PR).

rgerhards:
Basically done, small bits left. Closing this as the improvement is 95% and it would be ok if nothing more happens.

PR #2807, opened 2018-06-28
travis: remove "merge master" test

rgerhards:
We originally merged master into the PR. At that time, github did not
(reliably?) provide a merged code basis. This is now the case, so the
test is kind of unnecessary. It only did some good if master was updated
after PR submission but before Travis started the test. This is pretty
unlikely, so we remove that test - which also is a relief in regard
to Travis runtime.

see also https://github.com/rsyslog/rsyslog/issues/2800


rgerhards:
the docker ubuntu slave is a false positive, due to CI changes

Issue #1323, opened 2016-12-12
report error on mm module with queue


davidelang:
since a queue on a mm module is never useful (the mm module will run, but there will be nothing else that makes use of the result), and we do not support mm modules that have side effects, if a queue is configured on a mm* action() call, the config system should report an error.

rgerhards:
suggestions for text improvement of the accompanying website page is appreciated: https://www.rsyslog.com/mm-no-queue/

PR #2808, opened 2018-06-29
more explicit error msg with message modification mod on queue

rgerhards:
Message modification modules do not work if used with a non-direct queue.
We now make this more explicit in the config parsing error message.

closes https://github.com/rsyslog/rsyslog/issues/1323


davidelang:
Can the config parser detect that there are no actions after a mm module? 
(either because the mm action has a queue, or there is nothing after it in the 
ruleset)? and log a warning that the mm module will have no effect because there 
are no actions following it?

David Lang

rgerhards:
> Can the config parser detect that there are no actions after a mm module?
> (either because the mm action has a queue, or there is nothing after it in the
> ruleset)? and log a warning that the mm module will have no effect because there
> are no actions following it?

Right now it checks for the queue. Checking for actions after it is more work to do. I think let's do that later, the more common case is that a queue is defined. And the message is indeed the new thing and purpose of this PR ;-)

PR #2812, opened 2018-07-02
default rules use container_name_and_id; include rulebase files in dist

richm:
Change the default filename rules to use container_name_and_id
Include the rulebase files in the dist / source tarball


richm:
@davidelang

rgerhards:
@richm @davidelang "bug" label would be correct?

richm:
@rgerhards yes - I'm sure David would agree this is a bug

davidelang:
Yes, there is a default value that doesn't work.

rgerhards:
In the interim I have used:
```
- mmkubernetes bugfix: default rules use container_name_and_id
  also include rulebase files in dist
  Thanks to Richard Megginson for fixing the issue.
```
maybe it's sufficient...

davidelang:
The rulebase file was in the source tree, but not included in the dist packages.

This solves that and gets the rulebase file to the admins who need it.

David Lang

Issue #2753, opened 2018-06-04
I'm getting a build error in /plugins/ommongodb/ommongodb.c 


mmonson:
### Expected behavior
No compile errors on the v8-stable or main branch
### Actual behavior
make[2]: Entering directory '/opt/rsyslog/rsyslog/plugins/ommongodb'
  CC       ommongodb_la-ommongodb.lo
ommongodb.c: In function ‘initMongoDB’:
ommongodb.c:204:3: error: unknown type name ‘mongoc_ssl_opt_t’
   mongoc_ssl_opt_t ssl_opts;
   ^
ommongodb.c:205:31: error: ‘mongoc_ssl_opt_t’ undeclared (first use in this function)
   memset(&ssl_opts, 0, sizeof(mongoc_ssl_opt_t));
                               ^
ommongodb.c:205:31: note: each undeclared identifier is reported only once for each function it appears in
ommongodb.c:206:14: error: request for member ‘pem_file’ in something not a structure or union
      ssl_opts.pem_file = pData->ssl_cert;
              ^
ommongodb.c:207:14: error: request for member ‘ca_file’ in something not a structure or union
      ssl_opts.ca_file = pData->ssl_ca;
              ^
ommongodb.c:208:6: error: implicit declaration of function ‘mongoc_client_set_ssl_opts’ [-Werror=implicit-function-declaration]
      mongoc_client_set_ssl_opts (pData->client, &ssl_opts);

### Steps to reproduce the behavior
1.   download clean repo
2   run autoreconf -fvi
3   make

### Environment
- rsyslog version: 8.35.0
- platform: compiled on Ubuntu 16.04 

This all seems to be due to a missing definition for mongoc_ssl_opt_t

rgerhards:
I guess your version of libmongo-c is too old. Configure should probably check for a minimum version.

rgerhards:
@mmonson did you try with an updated libmongo-c?

rgerhards:
alternatively, please post config.log so that we can check what configure saw

mmonson:
Yes that seems to be working 
I tried running (sudo /usr/sbin/rsyslogd -nd) with the attached rsyslog.conf file and I get this error  I get this same error with the TCP and UDP listeners commented out and with 
$ActionFileDefaultTemplate RSYSLOG_TraditionalFileFormat

2688.264121771:main Q:Reg/w0  : ommongodb.c: ommongodb: ts_gen is 0
[...]

PR #2819, opened 2018-07-04
Remove ?authMechanism=PLAIN to let the mongo library chooses the defa…

jjourdin:
This is a proposal to fix https://github.com/rsyslog/rsyslog/issues/2753.
It removes the hardcoded ?authMechanism=PLAIN

closes https://github.com/rsyslog/rsyslog/issues/2753


rgerhards:
Could you test the patch and it works for you?

jjourdin:
It works for me.
Travis failure seems to be related to omkafka, not to ommongodb.

rgerhards:
> It works for me.

great! That was the confirmation I was looking for ;-)

> Travis failure seems to be related to omkafka, not to ommongodb.

I have restarted. That is actually I a test I watch since a couple of days for occasional failure.

PR #2824, opened 2018-07-06
fix pod_name field match in filename rule - everything up to _

richm:
The pod_name field match in the filename rule must include
everything up to the first _, including dots.


richm:
@davidelang not sure if you are running into this in your kubernetes testing - I have just run into a pod name with `.` in the name which was causing problems with the filename parser lognorm rule - the `.` is a valid character in pod names, so must be accounted for.

PR #2831, opened 2018-07-12
Travis: reduce distcheck tests

rgerhards:
this has been moved to the buildbot infrastructure, where we
do not have the strong runtime limitations. We still need to
run db-related (mysql, ...) distcheck tests as this is not yet
supported in the containers we use on buildbot.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,16 +71,16 @@ matrix:
        services:
          - mysql
          - postgresql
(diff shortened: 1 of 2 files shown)


rgerhards:
freebsd failure is known issue, see https://github.com/rsyslog/rsyslog/issues/2833

rgerhards:
CI issues are issues due to new CI definitions - acceptable.

Issue #2830, opened 2018-07-11
imkmsg builds unclean with newer compilers


rgerhards:
we get a couple of warning, which prevents us from checking this in CI with newer compilers.

currently missing in:
* ubuntu 18.04
* centos 7

PR #2832, opened 2018-07-12
imkmsg bugfix: build warnings on some platforms

rgerhards:
nothing serious, but still breaks CI. Also, we want as-clean-as-possible code

closes https://github.com/rsyslog/rsyslog/issues/2830


rgerhards:
CI failures are acceptable in this case, as I am working on CI

PR #2841, opened 2018-07-16
gcc 8 strict support, including some fixes

rgerhards:


rgerhards:
Fedora 28 fail is OK and expected. LGTM is experiencing issues...

Issue #2842, opened 2018-07-16
lookup tables: potential NULL pointer access


rgerhards:
on Fedora 28 w/ gcc 8 I see:
```
lookup.c: In function 'lookupReadFile':
lookup.c:503:22: error: potential null pointer dereference [-Werror=null-dereference]
    canonicalValueRef = *(uchar**) bsearch(indexes[i].val, pThis->interned_vals,
    ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    pThis->interned_val_count, sizeof(uchar*), bs_arrcmp_str);
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
lookup.c:503:22: error: potential null pointer dereference [-Werror=null-dereference]
lookup.c:539:22: error: potential null pointer dereference [-Werror=null-dereference]
    canonicalValueRef = *(uchar**) bsearch(value, pThis->interned_vals,
    ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    pThis->interned_val_count, sizeof(uchar*), bs_arrcmp_str);
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
lookup.c:539:22: error: potential null pointer dereference [-Werror=null-dereference]
```

It looks valid, even though static analyzers did not yet come up with an alert.

PR #2843, opened 2018-07-16
fix a couple of issues with and detected by gcc 8

rgerhards:
solves a couple of things (see commits for details)

closes https://github.com/rsyslog/rsyslog/issues/2842

### DIFF
--- a/grammar/rainerscript.c
+++ b/grammar/rainerscript.c
@@ -4533,7 +4533,7 @@ constFoldConcat(struct cnfexpr *expr)
 			cnfexprDestruct(expr->r);
 			expr->nodetype = 'S';
 			((struct cnfstringval*)expr)->estr = estr;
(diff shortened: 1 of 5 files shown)


rgerhards:
@janmejay it would be great if you could have a quick glimpse. However, the fix is more or less the detection but not real handling of the corner case. Still better than before, I think.

rgerhards:
@janmejay sry... I put more commits into that branch and did not think that I asked for a quick review. Only the frist commit is relevant in regard to my question

rgerhards:
> This PR looks like it pulled in unrelated changes in other files (it looks like they are all gcc 8 cleanups)

yup, sorry for that, see https://github.com/rsyslog/rsyslog/pull/2843#issuecomment-405240679

I have changed the description. Splitting things of makes little sense as I am about to merge everything soon in any case.

Issue #2564, opened 2018-03-19
test ./imjournal-basic.sh frequenly fails


rgerhards:
currently seen very frequently on fedora buildbot slave.

**recorded failures: http://build.rsyslog.com/testbench-showtest.php?testname=imjournal-basic%.sh**

The failure looks a bit different, here is a frequent case:
```
------------------------------------------------------------
Test: ./imjournal-basic-vg.sh
------------------------------------------------------------
testbench: TZ env var not set, setting it to UTC
testconf.conf is:
     1	$IncludeConfig diag-common.conf
     2	
     3	module(load="../plugins/imjournal/.libs/imjournal")
     4	
     5	template(name="outfmt" type="string" string="%msg%\n")
     6	action(type="omfile" template="outfmt" file="rsyslog.out.log")
==4868== Memcheck, a memory error detector
==4868== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==4868== Using Valgrind-3.13.0 and LibVEX; rerun with -h for copyright info
==4868== Command: ../tools/rsyslogd -C -n -irsyslog.pid -M../runtime/.libs:../.libs -ftestconf.conf
==4868== 
rsyslogd started, start msg not yet seen, pid  4868
rsyslogd started, start msg not yet seen, pid  4868
rsyslogd startup msg seen, pid  4868
startup-vg still running
Written to journal: "TestBenCH-RSYSLog imjournal This is a test message - 1521131225"
imdiag[13500]: mainqueue empty
==4868== 
==4868== HEAP SUMMARY:
==4868==     in use at exit: 24,883 bytes in 56 blocks
==4868==   total heap usage: 4,536,637 allocs, 4,536,581 frees, 2,267,745,613 bytes allocated
==4868== 
==4868== 263 (88 direct, 175 indirect) bytes in 1 blocks are definitely lost in loss record 37 of 42
==4868==    at 0x4C2FA1E: calloc (vg_replace_malloc.c:711)
==4868==    by 0xBC6AC27: nspollConstruct (nspoll.c:93)
==4868==    by 0xBE75C6E: Run (tcpsrv.c:919)
==4868==    by 0x46A260: thrdStarter (threads.c:223)
==4868==    by 0x52D936C: start_thread (in /usr/lib64/libpthread-2.25.so)
==4868==    by 0x6226B4E: clone (in /usr/lib64/libc-2.25.so)
==4868== 
{
   <insert_a_suppression_name_here>
[...]

PR #2851, opened 2018-07-18
Added skip journal load on first run

jvymazal:
When starting rsyslog with imjournal for first time it outputs
an error and plugin does not run because no state file exists yet.
Now it skips the loading and creates state file on first persist.


jvymazal:
@rgerhards we already had this behavior few versions ago but it was accidentally dropped in  #2065 , with this commit I have brought it back while also making it explicit so it stays with future code changes. Current behavior requires you to start rsyslog with imjournal twice before it starts working while it outputs error on first time.

plugins/imjournal/imjournal.c
```diff
@@ -578,6 +578,17 @@ loadJournalState(void)
...
 
+	/* if state file not exists (on very first run), skip */
+	if (access(cs.stateFile, F_OK|R_OK) == -1 && errno == ENOENT) {
+		if (cs.bIgnorePrevious) {
+			/* Seek to the very end of the journal and ignore all older messages. */
+			skipOldMessages();
+		}
+		LogMsg(errno, RS_RET_FILE_NOT_FOUND, LOG_NOTICE, "imjournal: No statefile exists, "
+				"%s will be created (ignore if this is first run)", cs.stateFile);
+		goto finalize_it;
```
rgerhards: please use macro ```FINALIZE``` here (it expands to the same, but is defined to work with the rest of the macros for exception handling).

rgerhards:
> we already had this behavior few versions ago but it was accidentally dropped in #2065 , with this commit I have brought it back while also making it explicit so it stays with future code changes. Current behavior requires you to start rsyslog with imjournal twice before it starts working while it outputs error on first time.

good catch! Could we add a testbench test to *ensure* that it isn't broken again in the future? Or, in other words: why did the existing tests not catch it?  (my fault maybe?)

jvymazal:
If I am looking correctly current imjournal tests (there are only 2 of them) do not specify a statefile in config so have no way to catch this.

rgerhards:
> If I am looking correctly current imjournal tests (there are only 2 of them) do not specify a statefile in config so have no way to catch this.

ah, ok, that makes sense. Could you add a test that does that? I guess you can simply copy one of the others, at least it would detect when it does not start. Maybe a bit more elaborate testing of the state persistence would of course be better ;-)

jvymazal:
Also, when I am going trough the current tests it looks more like it is checking that TESTMSG is present in journal (https://github.com/rsyslog/rsyslog/blob/fd2200f3b55568d9e50e1655bd3196b56e8c2429/tests/imjournal-basic.sh#L26) then in rsyslog.out.log, if that is the case it would not detect even if the module would not be working.

rgerhards:
> Also, when I am going trough the current tests it looks more like it is checking that TESTMSG is present in journal :

yeah, that's right, but is it probably that this message (note the date part)
```
TESTMSG="TestBenCH-RSYSLog imjournal This is a test message - $(date +%s)"
```
is present without the test running?

Anyhow, suggestions on how to make it more bullet proof is highly recommended.

rgerhards:
Doh! The test should go against rsyslog.out.log!

jvymazal:
I have changed all the (now 3) imjournal tests to go against the right file as well as adding one that uses statefile so we have that covered. Unless you have anything else, merge whenever you are ready.

rgerhards:
Oh, I see there were problems with the journal test. Makefile.am quoted https://github.com/rsyslog/rsyslog/issues/2564 - we probably need to check what was going on here. Also looks like the issue of test validity already came up there.

jvymazal:
@rgerhards , I think I solved the problem for now, it is not the most elegant solution however it should give us stable state and it is still much better than having all imjournal tests disabled. From what I was able to find out in some cases journal input just does not work (possibly some permissions error), in imjournal tests I now detect those cases and skip the test. Also if you could please merge it when the rest of CI goes green so I can rebase #2543 onto it and continue working on it with working testbench. Thanks.

rgerhards:
@jvymazal  sry for breaking your commit - I should have reversed merge order but have to admit I did not watch closely enough... Even more thx for your fast action!

jvymazal:
@rgerhards now when finally it all works with the new testbench scripts I have gone trough the results and it looks like there are problems with journald on some of the docker environments (although it appears working journal db itself is empty), others are functional and tests are green now, so if you merge we can close #2564 by this.

rgerhards:
That journal thing makes sense, as there is no systemd et all running in
the containers. Let me work on the image options tomorrow, I think we need
to disable testing on the containers (but can build).

Rainer

Sent from phone, thus brief.


jvymazal <notifications@github.com> schrieb am Di., 24. Juli 2018, 21:35:

> @rgerhards <https://github.com/rgerhards> now when finally it all works
> with the new testbench scripts I have gone trough the results and it looks
> like there are problems with journald on some of the docker environments
> (although it appears working journal db itself is empty), others are
> functional and tests are green now, so if you merge we can close #2564
> <https://github.com/rsyslog/rsyslog/issues/2564> by this.
>

davidelang:
you have to do some extra trickery to get journald working in a container. I've 
set it up, but I had to google for it.

rgerhards:
actually I am not sure I want it that way. It would just be for the journal
tests, and remove much of the simplicity IMHO. we'll probably spawn a VM
for these tests... Or do you think the other way around would be much
better?

Rainer

2018-07-24 21:58 GMT+02:00 David Lang <notifications@github.com>:

davidelang:
a vm will be more reliable as it won't be interacting with the host system.

David Lang

jvymazal:
@rgerhards @davidelang In short, should I remove the journal-check blocks from tests or leave them there?

jvymazal:
> Well... assuming the journal is not present - do we then have `journalctl`? 

That is exactly the problem, we have, even the `systemd-cat` actually exits clean (with 0) so I had to do this explicit check (insert message, check it is there, skip (77) if not), and only do the rsyslog.out check if it is present in journal.

rgerhards:
> That is exactly the problem, we have, even the systemd-cat actually exits clean (with 0) so I had to do this explicit check (insert message, check it is there, skip (77) if not), and only do the rsyslog.out check if it is present in journal.

ah, ok, I hadn't looked at the code yet. I think this is a valid solution. I think if the journalctl check fails, there is no way the test could work. Right? If so, I think the cure is actually quite valid. We just need to make sure we have a (dedicated) builder inside CI that ensure the tests are run.

Are we in agreement?

jvymazal:
> We just need to make sure we have a (dedicated) builder inside CI that ensure the tests are run.

Agree, if all of the non-container runs are enough we should be good (I have gone trough them and it does not look like any of them are skipping the tests).

jvymazal:
@rgerhards If we agreed that current solution is acceptable could you please merge it? I would like to continue work on #2543 for which it would be nice if I could rebase it onto this so I have functional CI and both changes in one place for testing. Many thanks

rgerhards:
thx for the ping - wrangling with CI, so do not check that closely. Pls ping on any other urgent thing as well if I don't act fast enough!

PR #2858, opened 2018-07-20
testbench: kubernetes test cannot recover from failure

rgerhards:
when the test is aborted it is currently not possible to recover the hanging instance, forcing all future tries on that machine to hang until either reboot or manual intervention.

### DIFF
--- a/tests/CI/kill_all_kubernetes_test_server.sh
+++ b/tests/CI/kill_all_kubernetes_test_server.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+# This script shall be executed before the real CI run to make sure
+# the environment is "clean enough". It is not necessary in containers,
+# where we usually get a fresh container for each run.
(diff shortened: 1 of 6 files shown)


rgerhards:
It looks like the kubernetes test server is not always properly shut down.

rgerhards:
@richm does the test server have a capability to detect it cannot bind to the port and shut down gracefully? It's not a full solution, but would at least make clear what is wrong.

rgerhards:
@richm hold on, I am not 100% sure there was a hanging instance, though it looks so. Have instrumented the build to show more info from the test.

rgerhards:
@richm yes, it is a hanging previous instance that causes the issues. We need to think how we can clean up this.

rgerhards:
We could work around this by the timeout command, albeit at the cost of some portability problems.

PR #2878, opened 2018-07-26
fix build issues on armhf systems

rgerhards:
Look like general 32 bit issues, though.


rgerhards:
CI fail is unrelated (open CI issue)

PR #2881, opened 2018-07-26
mmkubernetes: stops working with non-kubernetes container names

richm:
When mmkubernetes encounters a record with a CONTAINER_NAME field,
but the value does not match the rulebase, mmkubernetes returns
an error, and mmkubernetes does not do any further processing
of any records.
The fix is to check the return value of ln_normalize to see if
it is a "hard" error or a "does not match" error.
This also adds a test for pod names with dots in them.


rgerhards:
> buildbot/rsyslog freebsd failure looks like a flake, or at least, unrelated to PR:
> 
> FAIL: daqueue-dirty-shutdown.sh

yeah, unfortunately one of the things I am fighting with.

thanks for the PR!

PR #2906, opened 2018-08-02
write all header metadata to omes for retries

richm:
Write all of the original request metadata fields to `$.omes` for
the retry, if present.  This may include all of the following:
_index, _type, _id, _parent, pipeline
This is in addition to the fields from the response.  If the same
field name exists in the request metadata and the response, the
field from the request will be used, in order to facilitate
retrying the exact same request.

### DIFF
--- a/tests/es-bulk-retry.sh
+++ b/tests/es-bulk-retry.sh
@@ -15,7 +15,7 @@ EOF
 generate_conf
 add_conf '
 module(load="../plugins/impstats/.libs/impstats" interval="1"
-	   log.file="es-stats.log" log.syslog="off" format="cee")
+	   log.file="test-spool/es-stats.log" log.syslog="off" format="cee")
 
 set $.msgnum = field($msg, 58, 2);
 set $.testval = cnum($.msgnum % 2);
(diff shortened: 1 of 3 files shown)


tests/es-bulk-retry.sh
```diff
@@ -59,6 +54,12 @@ ruleset(name="try_es") {
 		}
 		# else fall through to retry operation
 	}
+	if strlen($.omes!_id) > 0 then {
+		set $.es_msg_id = $.omes!_id;
+	} else {
+		# NOTE: depends on rsyslog being compiled with --enable-uuid
```
rgerhards: I think it would be good to check this inside ./tests/Makefile.am, so that the test is not executed if this is not given.

tests/es-bulk-retry.sh
```diff
@@ -77,9 +78,8 @@ if $msg contains "msgnum:" then {
 	call try_es
 }
 
-action(type="omfile" file=`echo $RSYSLOG_OUT_LOG`)
+action(type="omfile" file="'$RSYSLOG_OUT_LOG'")
```
rgerhards, line 81: this does not work, it needs to be as in the original line. We use the backticks feature here: https://www.rsyslog.com/doc/v8-stable/rainerscript/constant_strings.html

I recently added the env var references so that we in the medium term can generate test-specific names and thus support parallel execution.

tests/es-bulk-retry.sh
rgerhards: oh wait.. maybe I was wrong in this instance, let me check...

tests/es-bulk-retry.sh
rgerhards: I thought this may be using shell escape (double quotes), but it isn't . so the file name will actually be '$RSYSLOG_OUT_LOG'. I would expect that CI fails.

rgerhards: my bad, got lost in the quotes myself ;-) Albeit I would still prefer the previous code, as it would be consistent with the other tests. But definitely no hard need to change.

rgerhards: on second thought, your way of doing things I think is actually better. ;-) Let's keep it, I'll see if I sed the rest of the testbench. ;-)

richm: re: quotes - I'm really surprised that
```
`echo $RSYSLOG_OUT_LOG`
```
worked at all - I would have thought that the backticks would have been preserved and not expanded inside the single quotes

re: uuid - I changed the test not to have that dependency, and added a note that it is recommended to use `$uuid` in production

rgerhards: Re Backticks; - they are actually preserved. Rsyslog itself interprets them!

richm:
@rgerhards do you still want me to make the change to use the Backticks?  Or are you ready to lift the "Changes requested" flag?

rgerhards:
@richm thanks again for the PR - and sorry I was not able to merge last week.

PR #2914, opened 2018-08-07
Enhanced symlink rotation support

jvymazal:
Now watching target directory as well to properly catch rotation
of files monitored via symlink.

Also fixed a bug in file move handling, previously state file could be deleted when `node` had more than one `edge`.

### DIFF
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -152,7 +152,7 @@ struct act_obj_s {
 	fs_edge_t *edge;	/* edge which this object belongs to */
 	char *name;		/* full path name of active object */
 	char *basename;		/* only basename */ //TODO: remove when refactoring rename support
-	char *source_name;  /* if this object is target of a symlink, source_name is its name (else NULL) */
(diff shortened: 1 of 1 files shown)


jvymazal:
@rgerhards this is sort of follow-up to #2749 , found few other use-cases where original implementation still was not enough + found the rename bug in the process. The only CI fail is DA-queues which seems like a flake.

rgerhards:
will merge after https://github.com/rsyslog/rsyslog/pull/2927 as I'd like to get coverity clean again.

rgerhards:
Thankfully it my trick to "solve" the coverity false positive worked out ;-) So I feel now even better merging this one here. Thx again!

rgerhards:
@jvymazal I already added this to the ChangeLog w/ previous merge:
```
- bugfix imfile: files which were loaded via symlink were not always followed
  They were stopped watching after being rotated.
```

I think this is sufficient. Or do I need to add some more info?

jvymazal:
@rgerhards yes, I think we can leave it at that. For user it has enough information.

rgerhards:
thx!

PR #2919, opened 2018-08-12
testbench: make some test files dynamic

rgerhards:
this is work torward the ability to run the testbench in parallel


rgerhards:
CI newly initiated just because I squashed the branch (more intelligent than github does). Previous CI run was all-green, so I do not wait for the "new" one to complete.

PR #2920, opened 2018-08-12
testbench: move some cleanup to "make clean" target

rgerhards:
**NOTE: before merge, buildbot VMs should also be changed to call ```make clean``` upon each start of processing.**

To support parallel testbench runs, we need to have dynamic work files.
So deleting work files on each test does not work, especially as we can
no longer assume they are "left-overs" from a failed. tes - they may
actulally (and quite likely) belong to tests that run in parallel.

So the proper solution is to do via "make clean".

closes https://github.com/rsyslog/rsyslog/pull/2916 (which it replaces)

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -1,5 +1,21 @@
 if ENABLE_TESTBENCH
-CLEANFILES = \
+
+CLEANFILES=\
+	*_*.conf \
+	rsyslogd.started work-*.conf rsyslog.random.data \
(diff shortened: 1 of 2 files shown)


rgerhards:
I will close this PR in favor of https://github.com/rsyslog/rsyslog/pull/2919 - originally I thought I add the functionality sooner via this PR, but as it turns out #2919 is also almost ready and has a better solution. So let's not mess with conflicting merges.

PR #2927, opened 2018-08-13
imfile: try to "fix" Coverity false positive

rgerhards:
Coverity seems not to understand realpath() API, but we nevertheless
want to have a clean Coverity report. So we hope this silences it.

CID 1911977


rgerhards:
@jvymazal FYI

PR #2935, opened 2018-08-15
testbench: remove no longer needed .conf files

rgerhards:
they have been replaced by the new config system


rgerhards:
solaris slaves seem to have some connectivty issues atm

rgerhards:
centos6 slave is currently new deployment and does not work correctly yet

PR #2937, opened 2018-08-16
fix imkafka can't parse hostname

snaix:
Fix imkafka can't parse hostname.
Add a option named "ParseHostName"(default is off) in imkafka, to maintain compatibility.
When ParseHostName = "on", a correct $hostname would be parsed.


rgerhards:
It would be great if you could add doc for the new feature in https://github.com/rsyslog/rsyslog-doc

snaix:
Yes, It will be added soon. ;)

rgerhards:
I have squashed the commit to get rid of the corrected initial commit.

PR #2940, opened 2018-08-17
testbench: use new kafka/zookeeper versions, fixed/added tests.

alorbach:
Fixed test configurations (socket timeouts, settings and so on)
Switched testbench to newer kafka/zookeeper versions.
Added new test for multiple imkafka instances.
Renabled sndrcv_kafka_fail.sh testcase.

### DIFF
--- a/tests/sndrcv_kafka-vg-rcvr.sh
+++ b/tests/sndrcv_kafka-vg-rcvr.sh
@@ -2,9 +2,12 @@
 # added 2017-05-03 by alorbach
 # This file is part of the rsyslog project, released under ASL 2.0
 export TESTMESSAGES=100000
+export TESTMESSAGESFULL=100000
 # enable the EXTRA_EXITCHECK only if really needed - otherwise spams the test log
 # too much
-export EXTRA_EXITCHECK=dumpkafkalogs
+# export EXTRA_EXITCHECK=dumpkafkalogs
+echo ===============================================================================
+echo \[sndrcv_kafka-vg-rcvr.sh\]: Create kafka/zookeeper instance and static topic
 . $srcdir/diag.sh download-kafka
 . $srcdir/diag.sh stop-zookeeper
 . $srcdir/diag.sh stop-kafka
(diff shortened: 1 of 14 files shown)


tests/sndrcv_kafka-vg-rcvr.sh
```diff
@@ -1,23 +1,26 @@
 #!/bin/bash
 # added 2017-05-03 by alorbach
 # This file is part of the rsyslog project, released under ASL 2.0
-export TESTMESSAGES=100000
+export TESTMESSAGES=1000
```
rgerhards: I think 1000 messages is far too low - rsyslog buffers this. IIRC I bumped the value to 100000 as otherwise the test pretended to work, but did not do so. you need to have the higher load. So we need to bump it again.

tests/sndrcv_kafka-vg-rcvr.sh
```diff
@@ -76,33 +79,35 @@ action(	name="kafka-fwd"
...
 
-echo Stopping sender instance [imkafka]
+echo \[sndrcv_kafka-vg-rcvr.sh\]: Stopping sender instance [imkafka]
 shutdown_when_empty 2
 wait_shutdown 2
 
-echo Sleep to give rsyslog receiver time to receive data ...
-sleep 5
+echo \[sndrcv_kafka-vg-rcvr.sh\]: Sleep to give rsyslog receiver time to receive data ...
+sleep 20
```
rgerhards: in a follow-up PR we need to fix the synchronization: sleeping isn't really a solution. I also strongly doubt that going from 5 to 20 seconds helps here, at least if the code works correctly. The shutdown processor checks that rsyslog is idle several times before it actually shuts down.

tests/sndrcv_kafka-vg-rcvr.sh
alorbach: In order to get the testcase stable, I would not recommend to use such a high value for the VALLGRIND test.

rgerhards:
This pull request **introduces 1 alert** when merging eb23efac20a82556c6b718b92bd07ca934265ae5 into 670d6661ecb7c1d69b3453f5d674189d2e90af27 - [view on LGTM.com](https://lgtm.com/projects/g/rsyslog/rsyslog/rev/pr-f92826efaf37bded1f2ef0e43f859aaa44404178)

**new alerts:**

* 1 for Empty branch of conditional

---

*Comment posted by [LGTM.com](https://lgtm.com)*

PR #2945, opened 2018-08-19
omprog: improve child process output capture/redirection

### LINKED ISSUE #2787: omprog: stdout/stderr of child process assigned unpredictably
Hi!

A strange behaviour with outputing to stderr from program started by omprog module.
A program code:
```
#include <syslog.h>
#include <stdio.h>
int
main(int argc, char **argv)
{
	fprintf(stderr, "String ERR1");
	syslog(LOG_ERR, "String ERR2");
	fprintf(stderr, "String ERR3");
	return 0;
}
```
### Expected behavior
See in log:
```
Jun 21 17:45:17 legohost prog: String ERR1
Jun 21 17:45:17 legohost prog: String ERR2
Jun 21 17:45:17 legohost prog: String ERR3
```
or
```
Jun 21 17:45:17 legohost prog: String ERR2
```

### Actual behavior
```
Jun 21 17:45:17 legohost prog: String ERR2
Jun 21 17:45:17 legohost ring ERR3
```

Compile specified earlier source code and place in rsyslogd conf something like this:
```
       if ($syslogfacility-text == "auth") then {
               action(type="omprog" binary="/home/user/work/prog/prog")
       }
```
and then execute:
```
logger  -p auth.info test
```

### Environment
- rsyslog version: 8.34.0
- platform: funtoo stable

jsiwrk:
Capture program output using a pipe shared with all child processes,
and write to the file using a dedicated thread. Ensures lines emitted
by the child processes will not be intermingled in the output file if
the lines are less than PIPE_BUF chars long and are written in line-
buffered mode.

Reopen output file on HUP, to support external rotation of the file.

New setting 'fileCreateMode' as in omfile.

With these improvements the 'output' setting should now be usable for
production (it was originally intended only for debugging).

Redirect stdout/stderr of child process to /dev/null when not captured.
Closes #2787

Minor: simplify some test code: 'wait-startup' not needed after
'startup', 'wait-queueempty' not needed before 'shutdown_when_empty'.

### DIFF
--- a/plugins/external/skeletons/python/plugin-with-feedback.py
+++ b/plugins/external/skeletons/python/plugin-with-feedback.py
@@ -6,6 +6,7 @@ To integrate a plugin based on this skeleton with rsyslog, configure an
 'omprog' action like the following:
     action(type="omprog"
         binary="/usr/bin/myplugin.py"
+        output="/var/log/myplugin.log"
         confirmMessages="on"
         ...)
 
@@ -61,20 +62,12 @@ def onInit():
     # Apart from processing the logs received from rsyslog, you want your plugin
     # to be able to report its own logs in some way. This will facilitate
     # diagnosing problems and debugging your code. Here we set up the standard
-    # Python logging system to output the logs to stderr.
+    # Python logging system to output the logs to stderr. In the rsyslog
+    # configuration, you can configure the 'omprog' action to capture the stderr
+    # of your plugin by specifying the action's "output" parameter.
     logging.basicConfig(stream=sys.stderr,
                         level=logging.WARNING,
                         format='%(asctime)s %(levelname)s %(message)s')
-
-    # In the rsyslog configuration, you can configure the 'omprog' action to
-    # capture the stderr of your plugin by specifying the action's "output"
-    # parameter. However, note that this parameter is intended for debugging
-    # purposes and cannot be used on a permanent basis (in particular, it will
-    # not work well if rsyslog decides to launch multiple instances of your
-    # plugin, since this would cause concurrent writes to the output file).
-    # As an alternative to logging to stderr, you can log to a per-process
-    # file, as shown here:
-    # logging.basicConfig(filename="myplugin-{}.log".format(os.getpid()), ...)
 
     # This is an example of a debug log. (Note that for debug logs to be
     # emitted you must set 'level' to logging.DEBUG above.)
--- a/plugins/omprog/omprog.c
+++ b/plugins/omprog/omprog.c
@@ -59,7 +59,23 @@ DEF_OMOD_STATIC_DATA
 #define NO_HUP_FORWARD -1	/* indicates that HUP should NOT be forwarded */
 #define DEFAULT_CLOSE_TIMEOUT_MS 5000
 #define READLINE_BUFFER_SIZE 1024
+#define OUTPUT_CAPTURE_BUFFER_SIZE 4096
 #define MAX_FD_TO_CLOSE 65535
+
+typedef struct outputCaptureCtx {
+	uchar *szFileName;		/* name of file to write the program output to, or NULL */
+	mode_t fCreateMode;		/* output file creation permissions */
+	int bIsRunning;			/* is the output-capture thread running? (if 0, next fields are meaningless) */
+	pthread_t thrdID;		/* ID of the output-capture thread */
+	int fdPipe[2];			/* pipe for capturing the output of the child processes */
+	int fdFile;				/* fd of the output file (-1 if it could not be opened) */
+	int bFileErr;			/* file open error occurred? (to avoid reporting too many errors) */
+	int bReadErr;			/* read error occurred? (to avoid reporting too many errors) */
+	int bWriteErr;			/* write error occurred? (to avoid reporting too many errors) */
+	pthread_mutex_t mutWrite;	/* mutex for reopening the output file on HUP while being written */
+	pthread_mutex_t mutTerm;	/* mutex for signaling the termination of the thread */
+	pthread_cond_t condTerm;	/* condition for signaling the termination of the thread */
+} outputCaptureCtx_t;
 
 typedef struct _instanceData {
 	uchar *szBinary;		/* name of external program to call */
@@ -75,17 +91,15 @@ typedef struct _instanceData {
 	int bSignalOnClose;		/* should send SIGTERM to program before closing pipe? */
 	long lCloseTimeout;		/* how long to wait for program to terminate after closing pipe (ms) */
 	int bKillUnresponsive;	/* should send SIGKILL if closeTimeout is reached? */
-	uchar *szOutputFileName;	/* name of file to write the program output to, or NULL */
-	pthread_mutex_t mut;	/* make sure only one instance is active */
+	outputCaptureCtx_t outputCaptureCtx;	/* settings and state for the output capture thread */
+	pthread_mutex_t mut;	/* make sure only one instance is active (buggy! - see #2813) */
 } instanceData;
 
 typedef struct wrkrInstanceData {
 	instanceData *pData;
 	pid_t pid;			/* pid of currently running child process */
 	int fdPipeOut;		/* fd for sending messages to the program */
-	int fdPipeIn;		/* fd for receiving status messages from the program */
-	int fdPipeErr;		/* fd for receiving error output from the program */
-	int fdOutputFile;	/* fd to write the program output to (-1 if to discard) */
+	int fdPipeIn;		/* fd for receiving status messages from the program, or -1 */
 	int bIsRunning;		/* is program currently running? 0-no, 1-yes */
 } wrkrInstanceData_t;
 
@@ -102,13 +116,14 @@ static struct cnfparamdescr actpdescr[] = {
 	{ "useTransactions", eCmdHdlrBinary, 0 },
 	{ "beginTransactionMark", eCmdHdlrString, 0 },
 	{ "commitTransactionMark", eCmdHdlrString, 0 },
-	{ "output", eCmdHdlrString, 0 },
 	{ "forcesingleinstance", eCmdHdlrBinary, 0 },
 	{ "hup.signal", eCmdHdlrGetWord, 0 },
 	{ "template", eCmdHdlrGetWord, 0 },
 	{ "signalOnClose", eCmdHdlrBinary, 0 },
 	{ "closeTimeout", eCmdHdlrInt, 0 },
-	{ "killUnresponsive", eCmdHdlrBinary, 0 }
+	{ "killUnresponsive", eCmdHdlrBinary, 0 },
+	{ "output", eCmdHdlrString, 0 },
+	{ "fileCreateMode", eCmdHdlrFileCreateMode, 0 }
 };
 
 static struct cnfparamblk actpblk =
@@ -117,53 +132,52 @@ static struct cnfparamblk actpblk =
 	  actpdescr
 	};
 
-/* execute the child process (must be called in child context
- * after fork).
+/* execute the external program (must be called in child context after fork).
  */
 static __attribute__((noreturn)) void
-execBinary(wrkrInstanceData_t *pWrkrData, int fdStdin, int fdStdout, int fdStderr)
+execBinary(const instanceData *pData, int fdStdin, int fdStdout)
 {
-	int i, maxFd, iRet;
+	int fdOutput, maxFd, fd, sigNum;
 	struct sigaction sigAct;
-	sigset_t set;
+	sigset_t sigSet;
 	char errStr[1024];
 
 	if(dup2(fdStdin, STDIN_FILENO) == -1) {
-		DBGPRINTF("omprog: dup() stdin failed\n");
-		/* do some more error handling here? Maybe if the module
-		 * gets some more widespread use...
-		 */
+		goto failed;
 	}
 
-	if(pWrkrData->pData->bConfirmMessages) {
-		/* send message confirmations via stdout */
-		if(dup2(fdStdout, STDOUT_FILENO) == -1) {
-			DBGPRINTF("omprog: dup() stdout failed\n");
-		}
-		/* redirect stderr to the output file, if specified */
-		if (pWrkrData->pData->szOutputFileName != NULL) {
-			if(dup2(fdStderr, STDERR_FILENO) == -1) {
-				DBGPRINTF("omprog: dup() stderr failed\n");
-			}
-		} else {
-			close(fdStderr);
-		}
-	} else if (pWrkrData->pData->szOutputFileName != NULL) {
-		/* redirect both stdout and stderr to the output file */
-		if(dup2(fdStderr, STDOUT_FILENO) == -1) {
-			DBGPRINTF("omprog: dup() stdout failed\n");
-		}
-		if(dup2(fdStderr, STDERR_FILENO) == -1) {
-			DBGPRINTF("omprog: dup() stderr failed\n");
-		}
-		close(fdStdout);
+	if(pData->outputCaptureCtx.bIsRunning) {
+		fdOutput = pData->outputCaptureCtx.fdPipe[1];
 	} else {
-		/* no need to send data to parent via stdout or stderr */
-		close(fdStdout);
-		close(fdStderr);
+		fdOutput = open("/dev/null", O_WRONLY);
+		if(fdOutput == -1) {
+			goto failed;
+		}
 	}
 
-	/* close all file handles the child process doesn't need.
+	if(fdStdout != -1) {
+		/* confirmMessages enabled: redirect stdout to parent via pipe. After
+		 * this point, anything written to the child's stdout will be treated
+		 * by omprog as initialization feedback (see startChild). This
+		 * includes debug messages (DBGPRINTF) when in debug mode. So we
+		 * cannot use DBGPRINTF from this point on, except for error cases.
+		 */
+		if(dup2(fdStdout, STDOUT_FILENO) == -1) {
+			goto failed;
+		}
+	} else {
+		/* confirmMessages disabled: redirect stdout to file or /dev/null */
+		if(dup2(fdOutput, STDOUT_FILENO) == -1) {
+			goto failed;
+		}
+	}
+
+	/* redirect stderr to file or /dev/null */
+	if(dup2(fdOutput, STDERR_FILENO) == -1) {
+		goto failed;
+	}
+
+	/* close the file handles the child process doesn't need (all above STDERR).
 	 * The following way is simple and portable, though not perfect.
 	 * See https://stackoverflow.com/a/918469 for alternatives.
 	 */
@@ -176,16 +190,16 @@ execBinary(wrkrInstanceData_t *pWrkrData, int fdStdin, int fdStdout, int fdStder
 		maxFd -= 10;
 	}
 #	endif
-	for(i = STDERR_FILENO + 1 ; i <= maxFd ; ++i) {
-		close(i);
+	for(fd = STDERR_FILENO + 1 ; fd <= maxFd ; ++fd) {
+		close(fd);
 	}
 
 	/* reset signal handlers to default */
 	memset(&sigAct, 0, sizeof(sigAct));
 	sigemptyset(&sigAct.sa_mask);
 	sigAct.sa_handler = SIG_DFL;
-	for(i = 1 ; i < NSIG ; ++i) {
-		sigaction(i, &sigAct, NULL);
+	for(sigNum = 1 ; sigNum < NSIG ; ++sigNum) {
+		sigaction(sigNum, &sigAct, NULL);
 	}
 
 	/* we need to block SIGINT, otherwise our program is cancelled when we are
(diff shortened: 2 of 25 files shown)


rgerhards:
Thx! FYI: the centos7-5 docker builder seems to fail inside a journal test. Working on that...

jsiwrk:
@rgerhards : just pushed a minor correction (in case you started reviewing).

plugins/external/skeletons/python/plugin-with-feedback.py
```diff
@@ -6,6 +6,7 @@
 'omprog' action like the following:
     action(type="omprog"
         binary="/usr/bin/myplugin.py"
+        output="/var/log/myplugin.log"
```
rgerhards, line 9: This must not be written to /var/log, as CI may not have write access. It should start with the value from env-var ```$RSYSLOG_DYNNAME``` - this is part to support parallel make check.

tests/omprog-close-unresponsive-vg.sh
```diff
@@ -31,9 +31,7 @@ main_queue(
 }
 '
 startup_vg
-. $srcdir/diag.sh wait-startup
```
rgerhards, line 34: do you really want it to run without knowing if it is read to run?

tests/omprog-feedback.sh
```diff
@@ -21,15 +21,14 @@ template(name="outfmt" type="string" string="%msg%\n")
         queue.type="Direct"  # the default; facilitates sync with the child process
         confirmMessages="on"
         useTransactions="off"
-        action.resumeRetryCount="10"
-        action.resumeInterval="1"
+        action.resumeInterval="1"  # retry interval: 1 second
+        action.resumeRetryCount="0" # the default; no need to increase since
```
rgerhards: if the default is meant, it would be best to remove that parameter. This guards against default changes (that cause regeressions).

tests/omprog-output-capture-mt.sh
```diff
@@ -0,0 +1,94 @@
...
+'
+startup
+SECONDS=0
+tcpflood -m$number_of_messages
+
+# Issue some HUP signals to cause the output file to be reopened during
+# writing (not a complete test of this feature, but at least we check it
+# doesn't break the output).
+. $srcdir/diag.sh issue-HUP
+sleep .1
```
rgerhards: msleep prefferred for portability

tests/omprog-output-capture-mt.sh
```diff
@@ -0,0 +1,94 @@
...
+        error_exit 1
+    fi
+done < $RSYSLOG_OUT_LOG
+
+if (( $line_num != $(($number_of_messages * 2)) )); then
+    echo "unexpected number of lines in captured output: $line_num (expected: $(($number_of_messages * 2)))"
+    error_exit 1
+fi
+
+throughput=$(($line_num / $elapsed_seconds))
```
rgerhards: this is a very nice test, but be warned that some of  the CI machines are **really** slow. This must not lead to false positives (we are already plagued too much with them and I am working very hard to reduce them). Maybe this should be put to an optional test (which we than can carry out in a dedicated environment).

tests/omprog-output-capture.sh
```diff
@@ -0,0 +1,62 @@
...
+
+template(name="outfmt" type="string" string="%msg%\n")
+
+:msg, contains, "msgnum:" {
+    action(
+        type="omprog"
+	    binary=`echo $srcdir/testsuites/omprog-output-capture-bin.sh`
+        template="outfmt"
+        name="omprog_action"
+        output=`echo $RSYSLOG_OUT_LOG`
```
rgerhards, line 20: isn't the same used inside the script (haven't double checked!)

tests/omprog-output-capture.sh
```diff
@@ -0,0 +1,62 @@
...
+        fileCreateMode="0644"  # default is 0600
+    )
+}
+'
+startup
+. $srcdir/diag.sh injectmsg 0 10
+shutdown_when_empty
+wait_shutdown
+
+expected_output="[stdout] Starting
```
rgerhards: very minor: I have begin to call this variable ```EXPECTED```

tests/omprog-output-capture.sh
```diff
@@ -0,0 +1,62 @@
...
+[stdout] Received msgnum:00000007:
+[stderr] Received msgnum:00000007:
+[stdout] Received msgnum:00000008:
+[stderr] Received msgnum:00000008:
+[stdout] Received msgnum:00000009:
+[stderr] Received msgnum:00000009:
+[stdout] Terminating normally
+[stderr] Terminating normally"
+
+written_output=$(<$RSYSLOG_OUT_LOG)
```
rgerhards: there is also a cmp function (cmp_equal) inside the testbench plumbing.

plugins/external/skeletons/python/plugin-with-feedback.py
jsiwrk: Note that this is not test code, it is just an example comment in the example Python skeleton (explaining how omprog would be typically configured).

tests/omprog-close-unresponsive-vg.sh
jsiwrk: I saw in diag.sh that 'startup' (and 'startup_vg') already does a 'wait-startup', so I removed this useless 'wait-startup' (when immediately following 'startup') in all tests. Similarly, I removed all 'wait-queueempty' just before 'shutdown_when_empty', for the same reason. Is that correct?

tests/omprog-feedback.sh
jsiwrk: OK. However in this particular case I prefer to leave the line commented-out (rather than delete it) for self-documentation purposes, ok? I have also left queue.type="Direct" (although this is the default) in the tests where this setting is relevant (to avoid someone changing it accidentally...).

tests/omprog-output-capture-mt.sh
jsiwrk: Changed (also in other tests).

tests/omprog-output-capture.sh
jsiwrk: You mean if the invoked program also writes to RSYSLOG_OUT_LOG? No, in this case the program only writes to stdout and stderr.

tests/omprog-output-capture.sh
jsiwrk: Changed together with the next one where applicable.

tests/omprog-output-capture.sh
jsiwrk: You meant 'content-cmp' I assume. Changed where applicable.

Also I have corrected a minor problem in this 'content-cmp' helper function (in diag.sh): when the expected content was multiline, it was not displayed correctly in case of failure (double quotes missing in echo). I have included this minor change in this same PR, if you don't mind.

tests/omprog-output-capture-mt.sh
jsiwrk: True... I included this performance check because my first implementation of the functionality was inefficient and the check helped me refine it. However, I already had to decrease the threshold for the expected throughput (from more than 30.000 lines/sec in my machine to 10.000 lines/sec) just because I observed failures in my Travis CI. Now I realise that with this reduction it doesn't make sense to keep this check, since it would not detect the potential performance degradations it intended to cover.

Besides, it doesn't seem a good idea to include this performance check along with the other purposes of the test: for a more appropriate performance test, HUPs should not be included, the lines emitted by the program should be more realistic (less than 4K of size, which is a lot), and others. Moreover, the precision of the check is very rough (seconds).

So I have removed this performance check from the test. As you say, it would need to be a different test in any case, and this kind of tests require a dedicated environment.

jsiwrk:
Changes applied and rebased to master.

Side note: before the rebase I was able to easily execute all omprog tests with `make check TESTS="omprog-*.sh"`, but now I can't. It seems the makefile now also tries to create the .trs and .log files (?) when invoked with wildcards. Is it possible that this has changed?

tests/omprog-close-unresponsive-vg.sh
rgerhards: ummm... sometimes I should re-read code that has evolved over the years. Quick check reveals you are right - I admit this wasn't on my mind. I did not question the existing tests and took for granted they are right in doing this sequence... Thanks for the wakeup-call!

rgerhards:
> Side note: before the rebase I was able to easily execute all omprog tests with make check TESTS="omprog-*.sh", but now I can't. It seems the makefile now also tries to create the .trs and .log files (?) when invoked with wildcards. Is it possible that this has changed?

I found a bug inside the Makefile, it did not list test extensions, which made it impossible to specify sequence rules (which we need for parallel execution). I wasn't aware that you could use wildcards inside the TESTS variable, I also don't think I have seen this documented as a feaure (but I may be wrong). It's interesting that this now causes problems. The .trs and .log files were always created (under make [dist]check), but with .sh.trs and .sh.log. Sorry if that breaks something, but it was actually a needed bugfix.

Some info: https://www.gnu.org/software/automake/manual/html_node/Parallel-Test-Harness.html

jsiwrk:
The tests seem to expose a hang in FreeBSD and Solaris, I'm investigating the issue. I guess it's related to how the termination of the output-capture thread is handled, something I'm currently reviewing and will likely need changes. But I'm a bit lost since I cannot reproduce the issue on my environment (Linux)... @rgerhards, is there an easy way to obtain a debug output of the tests added in this PR (`omprog-output-capture.sh` and `omprog-output-capture-mt.sh`) in these platforms?

rgerhards:
> The tests seem to expose a hang in FreeBSD and Solaris, I'm investigating the issue. I guess it's related to how the termination of the output-capture thread is handled, something I'm currently reviewing and will likely need changes. But I'm a bit lost since I cannot reproduce the issue on my environment (Linux)... @rgerhards, is there an easy way to obtain a debug output of the tests added in this PR (omprog-output-capture.sh and omprog-output-capture-mt.sh) in these platforms?

sorry, I had a day full of meetings. I have no access to the Solaris box at all, but could get to the freebsd. However, the method that works is that you use the debug plumbing from the testbench. At the start of the test you can set:
```
export RSYSLOG_DEBUG="debug nologfuncflow noprintmutexaction nostdout"
export RSYSLOG_DEBUGLOG="log"
```
and before ```exit_test``` you can do ```cat log``` which emits the full debug log to the CI log. That's how I deal with these kinds of things.

If you look at ``./test/diag.sh`` it also shows some other debug possibilities in it's first 200 lines or so. But you cannot enable it in diag.sh, as this would apply to all tests and a) this takes to long so CI would time out, and b) not all tests work with these settings (there are some that do special things with stdout e.g.).

HTH

tests/omprog-output-capture.sh
```diff
@@ -31,6 +35,7 @@ startup
 shutdown_when_empty
 
 # XXX
+./msleep 30000
```
rgerhards: let me know if that heals the problem - if so, we may need to think about the plumbing. shutdown_when_empty should only return when the queue is really (really!) empty. I can't outrule there is a problem, though ;-)

jsiwrk: I don't think the problem is shutdown_when_empty. These tests do not use the feedback feature (this is on purpose, because we want to capture both stdout and stderr of the program), and therefore are susceptible to timing issues: the queue is empty, but the program has not terminated yet (the messages are buffered in the pipe), or maybe omprog has not terminated writing the captured output yet. However I had put the sleep in the wrong place I think: we must sleep before shutting down, to avoid hitting the shutdown timeout. Let's see with the next commit I just pushed...

However, if this theory is correct, I'm surprised that these timing problems appear only with FreeBSD...

The failures in Solaris are unrelated. I will skip some of the new tests from that platform because of problems with stuff used by the tests (lsof and Python). I haven't studied the causes but I think it isn't worth the effort...

rgerhards: We should increase the shutdown timeouts if this turns out to be true - if that does not help, it'll fail in production, too. As always, it could also be a core engine issue.

Ack on Solaris.

jsiwrk: My theory wasn't correct, since adding the sleeps hasn't solved the problem. In fact, the sleeps shouldn't be necessary, since rsyslog is already waiting for the program to terminate (during 5 seconds, which is the default value of closeTimeout), and it is also waiting for the output capture thread to finish (also during 5 seconds). This should be enough, at least for the `omprog-output-capture.sh` test, which is a very simple test (but is also failing). There must be another (more serious) problem, manifesting only in FreeBSD, and likely related to the output capture thread I added in this PR.

rgerhards:
I'll have a look as soon as I can, but most probably at the weekend at earliest. Please ping me if you haven't heard back by next Wednesday.

jsiwrk:
Note: I have forced the test to fail in all platforms to be able to compare the debug output between them.

jsiwrk:
I finally found the problem on FreeBSD. Apparently, when shutting down, the new omprog output-capture thread I added in this PR was receiving the SIGTERM signal (since this thread is global for the action, it doesn't inherit the [worker threads sigmask](https://github.com/rsyslog/rsyslog/blob/v8.37.0/runtime/wtp.c#L402)), and despite the [signal handler](https://github.com/rsyslog/rsyslog/blob/v8.37.0/tools/rsyslogd.c#L1779) was executed, the [main loop](https://github.com/rsyslog/rsyslog/blob/v8.37.0/tools/rsyslogd.c#L1909) didn't end. The debug output ended consistently with these lines ([deinitAll](https://github.com/rsyslog/rsyslog/blob/v8.37.0/tools/rsyslogd.c#L1938) was not called):
```
...
0597.489489260:main Q:Reg/w0  : wti.c: main Q:Reg/w0: worker IDLE, waiting for work.
...
0598.581395404:imdiag.c       : nsdsel_ptcp.c: --------<NSDSEL_PTCP> calling poll, active fds (2): 7 8 0 
...
0599.086094133:802006800: rsyslogd.c: DoDie called.
0599.086108911:802006800: rsyslogd.c: rsyslogd.c:1787: write failed
```
This caused the shutdown to stop completely. The output of the tests was:
```
ABORT! Timeout waiting on shutdown
Instance is possibly still running and may need
manual cleanup.
```
Blocking all signals for the output-capture thread has solved the problem.

This only happened for FreeBSD, and for a reason I don't understand (?). On the other platforms the fact that the new thread hadn't the signals blocked wasn't problematic. (FTR: I also tried to declare the [bFinished](https://github.com/rsyslog/rsyslog/blob/v8.37.0/tools/rsyslogd.c#L154) flag as `volatile` but didn't work.)

jsiwrk:
The PR is again ready to merge for my part. Rebased to master. (Note: the CI failures for Ubuntu and Suse are unrelated.)

rgerhards:
sry, I had overlooked that you were ready for merge :-( 

Great to hear this, but we may need to work on some of the tests as I have converted the existing tests in the meantime so that the mostly (some work still do to) in a parallel check environment. I think it probably is best to merge the testbench PR first. It will potentially affect your tests, but I think it is easier to fix it that way as I can relatively easily provide info on what needs to be updated, so that you can see what makes most sense from your PoV. I think I already have some open questions/issues.

As a side-note, with the old code I see a problem with signal handling on Solaris: https://build.rsyslog.com/#/builders/73/builds/800/steps/9/logs/test-suite_log

Looks like on only one of the solaris slaves (10x64) the signal is not properly passed. Might be what you have seen, and may be fixed already. For my merge, I'll disable that test on Solaris, but we should re-enable it.

If you are OK with that, I could also add commits to your branch directly, so that I try to fix what I broke and can ask for help only if actually needed. Please let me know how we shall proceed.

Again, my apologies for the extra effort. I try to keep it as low as possible.

rgerhards:
@jsiwrk as far as I could see, omprog-restart-terminated.sh failure was not caused by lsof, but the non-delivery of signals. it least that was it for me...

jsiwrk:
No problem with the changes.

> I could also add commits to your branch directly, so that I try to fix what I broke and can ask for help only if actually needed. Please let me know how we shall proceed.

Whatever you prefer. However, my availability is low these days. I guess you'll advance faster yourself...

> As a side-note, with the old code I see a problem with signal handling on Solaris: 
> Looks like on only one of the solaris slaves (10x64) the signal is not properly passed. Might be what you have seen, and may be fixed already. For my merge, I'll disable that test on Solaris, but we should re-enable it.

This failure is new, and unrelated to this PR. The test doesn't send a signal to the child process (it has not to do with the signal handling inside rsyslog or omprog, it is a direct signal from the test code to the test program). Probably, with the changes you have made, the child process has a too long name (`rsyslog_omprog-restart-terminated-outfile.sh_62653.omprog-restart-terminated-bin.sh`). I bet `pkill` does not find the process because the name exceeds some internal limit in that particular version of Solaris.

>  as far as I could see, omprog-restart-terminated.sh failure was not caused by lsof, but the non-delivery of signals. it least that was it for me...

No, it was caused by lsof not returning the expected output in that version of Solaris (note that, in the output, lsof emits some warnings indicating it cannot gather all information). This caused a later check to fail, when comparing the output of lsof. (Now it fails before reaching that check, because of the above-mentioned failure.)

jsiwrk:
@rgerhards : this weekend I'll probably find some time to contribute. Let me know if you want me to rebase the PR.

rgerhards:
@jsiwrk re solaris, note that I am working on some general plumbing to skip based on platform. If you are faster, I'll adapt my method based on what you have, else it may be useful to wait.

rgerhards:
ok, should have looked first. You already got it, and in a better way. That should not be affected by my work. If it is, I can always adjust during merge.

plugins/omprog/omprog.c
```diff
@@ -176,44 +190,42 @@ execBinary(wrkrInstanceData_t *pWrkrData, int fdStdin, int fdStdout, int fdStder
...
+	 * 'syslog' system call to log the error (we cannot use LogMsg/LogError,
+	 * since these functions add directly to the rsyslog input queue).
+	 */
+	rs_strerror_r(errno, errStr, sizeof(errStr));
+	DBGPRINTF("omprog: failed to execute program '%s': %s\n",
+			pData->szBinary, errStr);
+	openlog("rsyslogd", 0, LOG_SYSLOG);
+	syslog(LOG_ERR, "omprog: failed to execute program '%s': %s\n",
+			pData->szBinary, errStr);
 	exit(1);
```
rgerhards, line 229: we must never do ```exit(1)``` inside a plugin (or even core module other than startup.

The user does not expect rsyslog to die while some subactivity is having issues. I understand it was already there, but we still need to fix that ;-)

plugins/omprog/omprog.c
```diff
@@ -489,35 +408,33 @@ writePipe(wrkrInstanceData_t *pWrkrData, uchar *szMsg)
 	len = strlen((char*)szMsg);
 
 	do {
-		checkProgramOutput(pWrkrData);
 		written = write(pWrkrData->fdPipeOut, ((char*)szMsg) + offset, len - offset);
 		if(written == -1) {
 			if(errno == EPIPE) {
 				DBGPRINTF("omprog: program '%s' terminated, will be restarted\n",
```
rgerhards, line 414: would it be better to expose these messages to the user (via ```LogMsg()```)? Again, nothing new, and if we go that path probably for another PR (so that we can finally merge this one).

plugins/omprog/omprog.c
```diff
@@ -621,9 +537,10 @@ readPipe(wrkrInstanceData_t *pWrkrData)
 		ABORT_FINALIZE(RS_RET_SUSPENDED);
 	}
 
-	iRet = lineToStatusCode(pWrkrData, line);
-	free(line);
+	CHKiRet(lineToStatusCode(pWrkrData, line));
```
rgerhards, line 540: Coverity will bark here, because this is an unnecessary check. So we nee to keep the ```iRet = ...```

plugins/omprog/omprog.c
```diff
@@ -632,7 +549,7 @@ startChild(wrkrInstanceData_t *pWrkrData)
 {
 	DEFiRet;
 
-	assert(pWrkrData->bIsRunning == 0);
+	assert(!pWrkrData->bIsRunning);
```
rgerhards, line 552: any need to remove the more clear check for ```== 0```?

plugins/omprog/omprog.c
```diff
@@ -649,6 +566,213 @@ startChild(wrkrInstanceData_t *pWrkrData)
...
+
+		writeOutputToFile(pCtx, readBuf, lenRead);
+	}
+
+	DBGPRINTF("omprog: all output-capture pipe ends closed, terminating output capture thread\n");
+	pthread_mutex_lock(&pCtx->mutTerm);
+	pCtx->bIsRunning = 0;
+	pthread_cond_signal(&pCtx->condTerm);
+	pthread_mutex_unlock(&pCtx->mutTerm);
+	return NULL;
```
rgerhards, line 689: as it always returns ```NULL```, we should make it void and not ```void *``` I think.

plugins/omprog/omprog.c
rgerhards: I admit it's a style issue, so I would not object if you leave as is. I have often seen folks wonder on the abbreviated case, but no strong feelings here. I guess it's done inconsistently in the codebase anyhow ;-)

rgerhards:
@jsiwrk I think I can (and will) fix the merge conflicts with a follow-up PR. It's a change to the ```wait-startup```

plugins/omprog/omprog.c
jsiwrk: Note this exits the child process. It is correct.

plugins/omprog/omprog.c
jsiwrk: I agree. And in fact I included this change in another PR I've prepared.

rgerhards:
well ... let's do it differently. The PR finally looks good except for the nits. let me handle this during merge. Then it is finally in master and we can look at some of the small things separately.

plugins/omprog/omprog.c
jsiwrk: This code will change again in my next PR (didn't know at that moment that I would need to later refactor it again). Besides, I don't think it will raise a warning since this pattern (CHKiRet as the last sentence) is used in other places.

plugins/omprog/omprog.c
jsiwrk: I made this change mostly because the use was inconsistent within omprog. I personally prefer the direct style since it is a boolean (and this is additionally reinforced by the use of the hungarian notation in this case). But I will also be happy with the other style if you prefer.

plugins/omprog/omprog.c
rgerhards: changed during merge

plugins/omprog/omprog.c
jsiwrk: Note this function is a thread entrypoint and therefore it must return `void *`.

jsiwrk:
@rgerhards : just responded to all your observations. I think the PR can be closed.

rgerhards:
> @rgerhards : just responded to all your observations. I think the PR can be closed.

ok, our work crossed. merged via https://github.com/rsyslog/rsyslog/commit/77399659f6bf2daaeed8cadfdcac787ff8166a5e

plugins/omprog/omprog.c
rgerhards: > Besides, I don't think it will raise a warning since this pattern (CHKiRet as the last sentence) is used in other places.

It does, I remember to have changed for more than one occurrence to make Covertiy happy. Maybe it does not detect all other places. It would be interesting if you could pinpoint some so that I can check.

plugins/omprog/omprog.c
rgerhards: let's keep as you did it - it's inconsistent in any way (and I admit my personal pref is towards this, too, but...)

plugins/omprog/omprog.c
rgerhards: I overlooked this ;-)

rgerhards:
be warned, I have one more testbench PR pending which will require a rebase. But should not require change. I'll try to push it trough quickly, at least as far as omprog tests are affected.

rgerhards:
Ah, I think it's unlikely to affect you.

plugins/omprog/omprog.c
jsiwrk: For example [here](https://github.com/rsyslog/rsyslog/blob/master/plugins/imrelp/imrelp.c#L881). Or in omprog.c, in BEGINmodInit().

It's really an inconvenience, since it forces an asymmetry in how the last line must be written... But I understand it is what it is.

jsiwrk:
I will submit a documentation PR in the near future.

jsiwrk:
> Side note: before the rebase I was able to easily execute all omprog tests with `make check TESTS="omprog-*.sh"`, but now I can't.

FTR: it can be done with this trick:
```
make check TESTS="$(find tests -name 'omprog*.sh' -printf '%P ')"
```

PR #2946, opened 2018-08-19
omprog: fix invalid memory access on partial writes to pipe

jsiwrk:
When sending logs to the program, in case of a partial write to the pipe, invalid data was sent, or an invalid memory access could occur. (A partial write can occur if the syscall is interrupted or the pipe is full.)

I have incidentally found this bug through code inspection (I have not reproduced it). It affects all versions to date.

Ideally there should be a test case for this, but I'm not sure how it could be implemented functionally (?).

Note: the changes from "int" to "ssize_t" are not related to the bug. Just a minor refactor.


rgerhards:
Journal test is having issues, but this is unrelated

rgerhards:
I unfortunately also have no idea how this could reliably be reproduced for a test... I guess this needs to be left without one :-(

Issue #2907, opened 2018-08-03
ommongodb requires mongo-c-driver with SSL


Whissi:
Hi,

you can build mongo-c-driver without SSL. When you do that, building rsyslog with ommongodb will fail with
```
ommongodb.c: In function ‘initMongoDBommongodb.c:204:3: error: unknown type name ‘mongoc_ssl_opt_t   mongoc_ssl_opt_t ssl_opts;
   ^~~~~~~~~~~~~~~~
   mongoc_client_t
ommongodb.c:205:31: error:    memset(&ssl_opts, 0, sizeof(mongoc_ssl_opt_t));
                               ^~~~~~~~~~~~~~~~
                               mongoc_client_t
ommongodb.c:205:31: note: each undeclared identifier is reported only once for each function it appears in
ommongodb.c:206:14: error: request for member ‘pem_file      ssl_opts.pem_file = pData->ssl_cert;
              ^
ommongodb.c:207:14: error: request for member ‘ca_file      ssl_opts.ca_file = pData->ssl_ca;
              ^
ommongodb.c:208:6: error: implicit declaration of function ‘mongoc_client_set_ssl_opts      mongoc_client_set_ssl_opts (pData->client, &ssl_opts);
      ^~~~~~~~~~~~~~~~~~~~~~~~~~
      mongoc_client_set_appname
ommongodb.c:208:6: warning: nested extern declaration of ‘mongoc_client_set_ssl_optscc1: some warnings being treated as errors
```
Please decide if you want to support ommongodb without SSL or not. If you want to support such a configuration, please fix code. If you don't want to support such a configuration, please add configure check for "mongoc_client_set_ssl_opts".

And while you are on that, please update https://github.com/rsyslog/rsyslog/blob/v8.36.0/plugins/ommongodb/ommongodb.c#L5. I.e. header is wrong. Plugin uses original mongo-c-driver.

This is somehow related to issue #2753.

HugoSoszynski:
Hi,

@rgerhards As it has only been a few weeks since I moved to another country for my studies, I do not have much time.

Maybe @jjourdin would like to have a look. 

Sorry for the late response. 


Regards.

jjourdin:
Hi,

I'll fix the code to support ommongodb with and without TLS.

rgerhards:
@HugoSoszynski thanks for looking at this, and all the best for your next effort! :-)

@jjourdin thx, much appreciated!

PR #2950, opened 2018-08-22
ommongodb: fix build issue if mongo-c-driver is not compiled with TLS

jjourdin:
Fix https://github.com/rsyslog/rsyslog/issues/2907

ommongodb now works even if mongoc-driver is not compiled with SSL support.


rgerhards:
merci for the PR! much appreciated!

PR #2952, opened 2018-08-24
Fixed parent name when processing symlinks

jvymazal:
There was a garbage byte left before which could cause errors down the
road.


rgerhards:
@jvymazal for the changelog: could this issue cause harsh consequences, e.g. segfaults? If so, when?

jvymazal:
@rgerhards, segfault or similar is highly unlikely, there is just one garbage byte in the name string, so when hit, it triggers lstat failure and parent is not processed.

rgerhards:
thx!

PR #2958, opened 2018-08-28
testbench: add program to get system time

PascalWithopf:
The new program test_id puts out the system time
up to microseconds.


rgerhards:
actually the intent of this program is to get - in a portable way - an id for a testrun, aka "an unique value"

PR #2970, opened 2018-08-31
lstnPortFileName: abort when file can't be opened

PascalWithopf:
Until now it was tried to write to the file even
if it couldn't be opened. Now the process is aborted.

### DIFF
--- a/plugins/imptcp/imptcp.c
+++ b/plugins/imptcp/imptcp.c
@@ -628,8 +628,9 @@ startupSrv(ptcpsrv_t *pSrv)
 						"error while trying to get socket");
 			}
 			if((fp = fopen((const char*)pSrv->pszLstnPortFileName, "w+")) == NULL) {
-				LogError(errno, NO_ERRCODE, "imptcp: ListenPortFileName: "
+				LogError(errno, RS_RET_IO_ERROR, "imptcp: ListenPortFileName: "
 						"error while trying to open file");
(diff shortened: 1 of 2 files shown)


plugins/imptcp/imptcp.c
```diff
@@ -630,6 +630,7 @@ startupSrv(ptcpsrv_t *pSrv)
 			if((fp = fopen((const char*)pSrv->pszLstnPortFileName, "w+")) == NULL) {
 				LogError(errno, NO_ERRCODE, "imptcp: ListenPortFileName: "
 						"error while trying to open file");
+				ABORT_FINALIZE(RS_RET_NO_ERRCODE);
 			}
```
rgerhards: Please change NO_ERRCODE to RS_RET_IO_ERROR  in all places.

plugins/imptcp/imptcp.c
```diff
@@ -630,6 +630,7 @@ startupSrv(ptcpsrv_t *pSrv)
 			if((fp = fopen((const char*)pSrv->pszLstnPortFileName, "w+")) == NULL) {
 				LogError(errno, NO_ERRCODE, "imptcp: ListenPortFileName: "
```
rgerhards: this should be fixed in all locations ;-)

rgerhards:
Note: this is a fix for a very recently introduced regression

PR #2974, opened 2018-08-31
testbench: Added new kafka tests using kafkacat for better debugging.

alorbach:
Omkafka: 
- Fixed minor issue in omkafka producing wrong kafka timestamps when msgTimestamp was NULL. 
- Setting RD_KAFKA_V_KEY(NULL, 0) in rd_kafka_producev now when KEY is not configured. 
- Fixed minor issue when rsyslog is compiled with --enable-debug and librdkafka is to old. 

Testbench: 
- Using latest kafka server version again.
- Fixed server configuration issues holding the kafka tests back from working. 
- Fixed some config issues in all sndrcv kafka tests. 
- Generating dynamically kafka topics now for each kafka test.
- Testmessages back to 100000 working stable now with log.retention.hours=5000.
- Renabled kafka_multi test which runs a test on 3 kafka/zookeeper instances.

### DIFF
--- a/tests/diag.sh
+++ b/tests/diag.sh
@@ -109,6 +109,8 @@ local0.* ./'${RSYSLOG_DYNNAME}'.HOSTNAME;hostname
 
 
 # begin a new testconfig
+#	2018-09-07:	Incremented inputs.timeout.shutdown to 60000 because kafka tests may not be 
+#			finished under stress otherwise
 # $1 is the instance id, if given
 function generate_conf() {
 	export TCPFLOOD_PORT="$(get_free_port)"
@@ -120,7 +122,7 @@ function generate_conf() {
 		mkdir $RSYSLOG_DYNNAME.spool
 	fi
 	echo 'module(load="../plugins/imdiag/.libs/imdiag")
-global(inputs.timeout.shutdown="10000")
+global(inputs.timeout.shutdown="60000")
 $IMDiagListenPortFileName '$RSYSLOG_DYNNAME.imdiag$1.port'
 $IMDiagServerRun 0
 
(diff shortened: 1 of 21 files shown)


rgerhards:
I think we should go over 50,000 msgs in the not so distant future. Because if everything works, it should work at any data rate - shouldn't it?

tests/diag.sh
```diff
@@ -110,7 +110,7 @@ function generate_conf() {
 	fi
 	echo imdiag running on port $new_port
 	echo "module(load=\"../plugins/imdiag/.libs/imdiag\")
-global(inputs.timeout.shutdown=\"10000\")
+global(inputs.timeout.shutdown=\"60000\")
```
rgerhards: Is this actually necessary? Looks overly excessive. Please provide a reason.

rgerhards:
Assuming that users will use current versions of kafka: is there a hard reason we use an old version inside the testbench? Is there a bug in the new version? If so, we should report and document this. Else we should also use the new version at least in some tests.

tests/diag.sh
alorbach: I had the impression that the input timeout aborts full processing in some cases. Can be set back to 10000 if tests run stable with it.

alorbach:
Yes we can go 50,000 messages in future, but there are problems to be solved on the kafka server side first in order to do this. Which I would go for in near future. This also adresses using newer kafka server versions. There are issues that need to be solved on the kafka server configuration side first in order to get it useable for the testbench.

rgerhards:
> Yes we can go 50,000 messages in future, but there are problems to be solved on the kafka server side first in order to do this. Which I would go for in near future. This also adresses using newer kafka server versions. There are issues that need to be solved on the kafka server configuration side first in order to get it useable for the testbench.

ok, sounds like a "tactical" PR to improve the kafka CI coverage, but not solve all known issues. As long as we have an issue open to actually solve the issues, this IMHO sounds like a good plan.

tests/diag.sh
rgerhards: ok, let's keep it for now. It should actually not have bad side-effects.

rgerhards:
see also https://github.com/edenhill/librdkafka/issues/1961

Issue #2975, opened 2018-08-31
Error message says 'omfile: either the "file" or "dynfile" parameter must be given'


fbicknel:
### Expected behavior
Error message says 'omfile: either the "file" or "dyn**a**file" parameter must be given'
### Actual behavior
Error message says 'omfile: either the "file" or "dynfile" parameter must be given'
#### Setup
Try:
```
template(name="DynoFile" type="string" string="/var/log/remote/%HOSTNAME%/syslog")
ruleset(name="remote") {
    action(type="omfile" dynFile="DynoFile")
}
```
(Label of template irrelevant: hence purposefully and flippantly misspelled.)

Note the spelling of **dynFile**
#### Unexpected Results
`Aug 31 12:01:50 bick-ubtu3 rsyslogd-2211: omfile: either the "file" or "dynfile" parameter must be given [v8.16.0 try http://www.rsyslog.com/e/2211 ]`
#### Change to Fix
```
template(name="DynoFile" type="string" string="/var/log/remote/%HOSTNAME%/syslog")
ruleset(name="remote") {
    action(type="omfile" dynaFile="DynoFile")
}
```
Note the spelling of **dynaFile**
### Environment
- rsyslog version: 8.16.0
- platform: Ubuntu 16.04
- for configuration questions/issues, include rsyslog.conf and included config files

### Why it's important
Because noobs like me spend an hour looking for documentation to explain why this thing doesn't work as expected before finally finding this silly error message led me down the primrose path.

fbicknel:
Pull request [2977](https://github.com/rsyslog/rsyslog/pull/2977) created that fixes this.

rgerhards:
closed via #2977

PR #2977, opened 2018-08-31
WiP: Fix errant error message when dynafile param needed

fbicknel:
Per issue [2975](https://github.com/rsyslog/rsyslog/issues/2975)

Compiled successfully under Ubuntu 16.04, but that's about it.

Dirt simple changes. Not much to see here.

I can't even say with 100% certainty I got everything, as I'm not that familiar with rsyslog... but I did do a search and these were the ones I found. Clearly there are other instances of *dynfile* that aren't malicious. I tried to err on the side of changing as little as possible.

### DIFF
--- a/tools/omfwd.c
+++ b/tools/omfwd.c
@@ -1140,7 +1140,7 @@ CODESTARTnewActInst
 	pvals = nvlstGetParams(lst, &actpblk, NULL);
 	if(pvals == NULL) {
 		LogError(0, RS_RET_MISSING_CNFPARAMS, "omfwd: either the \"file\" or "
-				"\"dynfile\" parameter must be given");
+				"\"dynafile\" parameter must be given");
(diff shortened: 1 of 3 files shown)


fbicknel:
I added `WiP` to the title per the docs because I wanted to be sure I didn't do any harm. I didn't think it would be automatically merged without peer review, but automation ironically scares me.  ;)

rgerhards:
thx for the PR, will review and merge either today or tomorrow

tools/omfwd.c
```diff
@@ -1140,7 +1140,7 @@ CODESTARTnewActInst
 	pvals = nvlstGetParams(lst, &actpblk, NULL);
```
rgerhards, line 1140: good catch! This actually looks like something else is not really correct in existing code... Will investigate.

PR #2982, opened 2018-09-02
omprog: fix output file write lock not always released

jsiwrk:
Fix bug introduced in #2945: output file write lock was not released if
an error occurred writing the file.


jsiwrk:
I made a mistake in my last (just merged) PR. No need to include this fix in the changelog.

Issue #2420, opened 2018-01-20
Deferred messages within a transaction not retried


jsiwrk:
### Expected behavior
If an output plugin confirms individual messages in a transaction with RS_RET_DEFER_COMMIT, and one of the messages in the transaction fails, one would expect that the previous messages are also retried, not only the one that has caused the transaction to fail.

This is what I understand from the documentation:
http://www.rsyslog.com/doc/v8-stable/development/dev_oplugins.html

I assume this is the correct interpretation: otherwise, what would be the difference between confirming messages with RS_RET_OK and RS_RET_DEFER_COMMIT inside a transaction?

### Actual behavior
Deferred messages (confirmed with RS_RET_DEFER_COMMIT) in a transaction are not retried if a later message in the transaction fails (with RS_RET_SUSPENDED). Only the failed message is retried.

Note: if the transaction fails during commit (i.e. the plugin returns RS_RET_SUSPENDED in 'endTransaction'), then the messages in the transaction are correctly retried.

### Steps to reproduce the behavior
Use the automated test 'omprog-transactions-failed-messages.sh' proposed for merge in PR #1992, modifying the test as explained in the TODO comment (to reproduce the issue, simply remove the two lines mentioned in the TODO). 

### Environment
- rsyslog version: 8.32.0
- platform: Ubuntu 16.04

rgerhards:
Just FYI: it's on my todo list, hope to be able to get to it some day...

rgerhards:
This is a hard issue, and I had some AI Agents generate potential fixes just to review their solution passes. I have also generated one with interactive guidance of the AI Agents. I now need to review, and see also which ones are causing defects:

### changes to action processing
- [ ] #5863  - my own try
- [ ] #5819 (Codex)
- [ ] #5859 (Cursor)

### refactoring omhttp to use new transaction interface 
- [ ] #5861 via Cursor/Sonet
- [ ] #5862 via Codex

PR #2983, opened 2018-09-02
omprog: add feedback timeout and keep-alive feature

jsiwrk:
Restart the program if it does not respond within timeout.
New setting 'confirmTimeout' (default 10 seconds).

Allow the program to provide keep-alive feedback when a
message requires long-running processing.

Improve efficiency when reading feedback line (use buffer).

Retry interrupted writes/reads to/from pipe.

New setting 'reportFailures' for reporting error messages
from the program.

Report child termination when writing to pipe.

Minor refactor: renamed writePipe function to sendMessage,
renamed readPipe to readStatus.

### DIFF
--- a/plugins/external/INTERFACE.md
+++ b/plugins/external/INTERFACE.md
@@ -46,9 +46,9 @@ If the plugin writes a line to stdout containing anything else (for example,
 the string `Error: could not connect to the database` followed by a LF), rsyslog
 will consider that the plugin has not been able to process the message. The
 message will be retried later, according to the retry settings configured for
-the `omprog` action (e.g. the `action.resumeInterval` setting). When debugging
-is enabled in rsyslog, the line returned by the plugin will be included in the
-debug logs.
+the `omprog` action (in particular, the `action.resumeInterval` setting). If the
+`reportFailures` flag is set to `on`, rsyslog will also log the returned error
+line.
 
 If the plugin terminates, the message is also considered as non-processed.
 The plugin will later be restarted, and the message retried, according to the
@@ -57,7 +57,10 @@ configured retry settings.
 When starting the plugin, if `confirmMessages` is `on`, rsyslog will also wait
 for the plugin to confirm its initialization. The plugin must write an `OK` line
 to stdout just after starting. If it writes anything else or terminates, rsyslog
-will consider the plugin initialization has failed, and will try to restart it later.
+will consider the plugin initialization has failed, and will try to restart it
+later. It is recommended to ensure that the program is really ready to start
+receiving messages before emitting this first confirmation (for example, by
+checking that a destination database is accessible).
 
 Example of exchanged messages
 -----------------------------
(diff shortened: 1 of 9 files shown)


jsiwrk:
Hope you'll find it interesting. I think a timeout is important for production use of the feedback feature.

rgerhards:
thx for the PR, sounds very interesting!

rgerhards:
It looks like there are at least some unrelated CI failures. I am trying to understand what is wrong there - I admit I currently do not have any idea at all. If all are unrelated, just leave as is, I will see if I can craft a patch that solves the testbench issues when I know what is causing it.

rgerhards:
@jsiwrk do you have an idea where the messages seen at the bottom of e.g. https://build.rsyslog.com/#/builders/77/builds/817 stem back to? It's a mystery to me why they appear at all... I guess this is closely related to why the test hangs.

jsiwrk:
The test failure on FreeBSD was due to `wc` [returning leading whitespace](https://unix.stackexchange.com/questions/205906/extra-space-with-counted-line-number). I've corrected the test to tolerate this.

> do you have an idea where the messages seen at the bottom of e.g. https://build.rsyslog.com/#/builders/77/builds/817 stem back to?

In the case of Solaris, the new test `omprog-feedback-mt.sh` seems to cause a hang. I'm pretty sure it's due to #2356, which [as I noted time ago](https://github.com/rsyslog/rsyslog/pull/1992#issuecomment-377002520), seems an issue that was partially corrected, and is still manifesting on Solaris. I have disabled this test for Solaris with a comment referring that issue. Note we had to do the same with [omprog-transactions-failed-commits.sh](https://github.com/rsyslog/rsyslog/blob/master/tests/omprog-transactions-failed-commits.sh#L12), which apparently also reproduces the same problem. I think an issue should be opened and consider it a bug. When the bug is corrected we will be able to enable both tests again for Solaris. Please note that this bug affects the rsyslog core and does not have to do with omprog.

Note: the failure on Suse seems unrelated. Apparently one of the `imtcp` tests has caused a hang. (One difficulty in analyzing those hangs is that the output only shows the last _passed_ test, but not the test that has actually hung...)

jsiwrk:
> The way the logs are gathered ensures that the tail of the shown log is actually the one that caused the hang,

I understand that, in the following output, `omprog-feedback.sh` has passed correctly (`PASS`), including the shutdown of rsyslog. And that the next test (presumably, `omprog-feedback-mt.sh`, which is the next test listed in the makefile, but the name of this test is not shown) has caused a hang. Is there another possible interpretation?
```
...
PASS: omprog-output-capture.sh
SKIP: omprog-output-capture-mt.sh
PASS: omprog-feedback.sh

command timed out: 300 seconds without output running ['gmake', 'check', 'V=0'], attempting to kill
process killed by signal 9
program finished with exit code -1
elapsedTime=1163.427839
```

rgerhards:
You need to look at step 6! https://build.rsyslog.com/#/builders/81/builds/852

jsiwrk:
Good to know! Thanks for the explanation.

I see in the gathered logs that the test that caused the hang in Solaris (`omprog-feedback-mt.sh`) had almost finished, and that the hang ocurred at shutdown (both in Solaris 10 and 11). With this information, I'm not so sure that it is related #2356. I will try to do some experiments (enable the debug logs, etc.) to see if we can obtain more information.

jsiwrk:
BTW: updated the PR to remove the following binary files I pushed by mistake. Maybe they could be excluded in .gitignore?
```
tests/have_relpSrvSetOversizeMode
tests/test_id
```

jsiwrk:
@rgerhards : I think omprog is currently using the "legacy" transaction interface. (I asked [this question](https://github.com/rsyslog/rsyslog/pull/1992#issuecomment-357460411) time ago but didn't get a response. But I have looked again at [action.c](https://github.com/rsyslog/rsyslog/blob/master/action.c#L1201) and I think I now understand better what that means.)

Could that be related to these Solaris hangs? (Including the hang in [omprog-transactions-failed-commits](https://github.com/rsyslog/rsyslog/pull/1992#issuecomment-377002520), which you didn't comment on.) Moreover: could also this be the cause of #2420 (which didn't get an explanation either)?

As you see you owe me a few answers ;-) Hope you'll find time to answer some day... (Sorry for my insistence, but it seems important.)

Would it be a good idea if I try to refactor omprog to use the "new" transaction interface and see if these problems disappear?

rgerhards:
> As you see you owe me a few answers ;-) Hope you'll find time to answer some day... (Sorry for my insistence, but it seems important.)

Sorry for not replying earlier, but the main reason for that is that I have no good replies yet. Nevertheless I have posted them now. Look at it this way: All in all, I have around 600 to 800 issues that I currently "need" to take care of (not all public; some infrastructure related; even the public just-rsyslog-itself are > 300). This limits my capabilities to address all of them timely. I need to go where it hurts most. Even posting "I am not ready yet" texts causes a lot of time, which I then do not have to actually work on the issues. There is a good reason that I am working on the testbench for several weeks now. However, this even more exhausts my time budget. But I think for some of the very-hard-to-find tests we need better tooling - as you have experience, the issues goes away, so you have no way to reliably reproduce.

Of course I probably do not always set the right priorities. But I do as good as I can. In essence, it boils down to the fact that we have too few great contributors like you. For example, I asked serveral times the  past years for folks who would help with packaging - in the end run I now have done it myself, even though incomplete. Took ~4 weeks so far.

rgerhards:
Re the current failure, I actually have not the slightest idea of why "make distclean" can timeout. Best explanation is testbench machine was too slow (a very seldom cause).

rgerhards:
> Would it be a good idea if I try to refactor omprog to use the "new" transaction interface and see if these problems disappear?

That would probably be useful (if only that we can get rid of the old interface some day).

jsiwrk:
Thanks for the update!

> Sorry for not replying earlier (...)

I understand the reasons perfectly, no worries! Just wanted to make sure we hadn't overlooked an obvious solution to these problems (in case they could be related to the use of the legacy transaction interface). Now I see it's not so easy.

I'll add to my todolist to (some day) try to refactor omprog to use the new transaction interface.

PR #2986, opened 2018-09-03
testbench: use a reliable dynamic imdiag port

rgerhards:
this also requires changes to some tooling. Most importantly, imdiag
now "officially" needs IPv4. The testbench actually always needed that,
but it was not plainly visible.

Also removing some left-over debug output.


rgerhards:
docker-centos6 is experimentally and need not to be considered, thus merging

Issue #323, opened 2015-05-06
No error message on permission denied


friedl:
When having a disk-assisted queue, no error message gets shown when the access permissions to folders/files is denied. Such messages are shown in the debug output, but not when starting rsyslog without debug support.

krawiec:
+1

mostolog:
Related: https://github.com/rsyslog/rsyslog/issues/548

rgerhards:
@PascalWithopf please check if this still exists - I think it is solved in the mean time. If not, please consider fixing it :-)

rgerhards:
@PascalWithopf  does this still exist?

PascalWithopf:
When permission is denied following error is given.
```rsyslogd: error starting up action queue [v8.38.0.master try http://www.rsyslog.com/e/2027 ]```
Followed by a memory leak, detected by Address Sanitizer.

rgerhards:
ok, thx, so the actual error message is missing

PR #2991, opened 2018-09-04
queue.c: add error if queue file can't be accessed

PascalWithopf:
When having a disk-assisted queue without permission
to write to the specified queue file an error will
be put out.

closes https://github.com/rsyslog/rsyslog/issues/323

### DIFF
--- a/runtime/queue.c
+++ b/runtime/queue.c
@@ -782,9 +782,9 @@ qqueueTryLoadPersistedInfo(qqueue_t *pThis)
 	if(stat((char*) pThis->pszQIFNam, &stat_buf) == -1) {
 		if(errno == ENOENT) {
 			DBGOPRINT((obj_t*) pThis, "clean startup, no .qi file found\n");
-			ABORT_FINALIZE(RS_RET_FILE_NOT_FOUND);
 		} else {
-			DBGOPRINT((obj_t*) pThis, "error %d trying to access .qi file\n", errno);
+			LogError(errno, RS_RET_IO_ERROR, "queue: %s: error %d could not access .qi file",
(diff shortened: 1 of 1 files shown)


runtime/queue.c
```diff
@@ -775,16 +775,20 @@ qqueueTryLoadPersistedInfo(qqueue_t *pThis)
...
+	int errNbr;
 
 	ISOBJ_TYPE_assert(pThis, qqueue);
 
 	/* check if the file exists */
 	if(stat((char*) pThis->pszQIFNam, &stat_buf) == -1) {
-		if(errno == ENOENT) {
+		errNbr = errno;
+		if(errNbr == ENOENT) {
+			LogError(0, RS_RET_IO_ERROR, "queue: clean startup, no .qi file found");
```
rgerhards: this is actually not an error, but only a debug status message. It is a normal state. This must not be exposed to the user.

runtime/queue.c
```diff
@@ -775,16 +775,20 @@ qqueueTryLoadPersistedInfo(qqueue_t *pThis)
...
 			DBGOPRINT((obj_t*) pThis, "clean startup, no .qi file found\n");
 			ABORT_FINALIZE(RS_RET_FILE_NOT_FOUND);
 		} else {
-			DBGOPRINT((obj_t*) pThis, "error %d trying to access .qi file\n", errno);
+			LogError(0, RS_RET_IO_ERROR, "queue: error %d could not access .qi file", errNbr);
```
rgerhards: errNbr should be given to LogError, so that it can generate a proper error message including the actual OS error.

Issue #1797, opened 2017-09-27
HTTP output module


imilo:
Unfortunately I can't find a general HTTP output module for Rsyslog. Do you have plans to implement it?

Use case - nowadays more and more custom (micro)services appear, often with a HTTP handler - it would be much easier to send data directly with a built-in module and acknowledging successful receiving from server with some response code like **200**.

Probably I could use **omfwd** with a template which includes some basic HTTP headers, but then there is no chance to know about success on a higher level.

davidelang:
> Unfortunately I can't find a general HTTP output module for Rsyslog. Do you have plans to implement it?

not that I know of, but contributions are welcome :-)

I think the problem is the hidden complexity that would show up once you tried 
to make such a module work with a bunch of services.

you also are going to need persistant connections or something like that or you 
will cripple your performance with all the reconnects.

In a high performance environment, it's going to be ugly enough sending the 
messages one at a time.

imilo:
Thanks for explaining. What about **omelasticsearch**? I see it uses `curl_easy_perform` from the imported library - isn't it almost the same thing?

rgerhards:
Can you specify what exactly the module should do and which parameters it
should support? Without that, I think it is too vague to discuss what is
similar and what not.

Rainer

Am 01.10.2017 15:02 schrieb "imilo" <notifications@github.com>:

imilo:
For example:
```
action(
    type="omhttp"
    ssl="false"
    server="example.com"
    serverport="8080"
    uri="/rsyslog"
    template="httpTemplate"
)
```
.. this sends an event in a POST request to the `http://example.com:8080/rsyslog` -> checks a success; if OK -> sends next event.

But these are just thoughts. I don't know whether **omelasticsearch** holds the connection open, whether it checks a response, etc. If you really say that HTTP output would be very inefficient, and **omelasticsearch** is created somehow in a different way, then OK.

rgerhards:
Do you have any specific need or is this just a general idea?

imilo:
I need to process JSON logs with a custom service, rsyslog easily could take file reading and sending tasks. HTTP handler is the easiest to implement on a server-side, TCP output is not reliable

rgerhards:
ah, OK. Would you be willing to create  a new module omhttp? This could simply be done by
a) copying omelasticsearch
b) removing everything you do not need
c) doing some adjustments for the post
Once I have this PR, I could direct you on how to integrate it into the build system

[...]

PR #2782, opened 2018-06-17
[WIP] http output module for posting logs to REST API

ctramnitz:
initial WIP PR for omhttp as suggested in #1797 
error handling has to be revamped to remove ES specific items

PR #2993, opened 2018-09-05
add contrib module omhttp

PascalWithopf:
closes https://github.com/rsyslog/rsyslog/issues/1797
replaces https://github.com/rsyslog/rsyslog/pull/2782


rgerhards:
core functionality will be merged via https://github.com/rsyslog/rsyslog/pull/2993

darakian:
@PascalWithopf is it possible to get conf file example of using omhttp?
/*auf meine schlechte deutsch*/
Ist es möglich, ein beispiel fur ein conf file mit omhttp haben?

PascalWithopf:
@darakian omhttp is used like the other actions
``action(type="omhttp" ...)``
If you look in the code https://github.com/rsyslog/rsyslog/blob/master/contrib/omhttp/omhttp.c#L143
you can see a list of all parameters you can use, most of them have a self describing name.

The two parameters restpath and dynrestpath work together. In restpath you can specify the path directly or when you set dynrestpath "on" you can specify a template name, and the template will be used for the restpath. That way you can use a dynamic restpath.

davidelang:
On Tue, 20 Nov 2018, Pascal Withopf wrote:

> @darakian omhttp is used like the other actions
> ``action(type="omhttp" ...)``
> If you look in the code https://github.com/rsyslog/rsyslog/blob/master/contrib/omhttp/omhttp.c#L143
> you can see a list of all parameters you can use, most of them have a self describing name.
>
> The two parameters restpath and dynrestpath work together. In restpath you can specify the path directly or when you set dynrestpath "on" you can specify a template name, and the template will be used for the restpath. That way you can use a dynamic restpath.

hmm, every other case where we have an option for a dynamic parameter, it is 
given a different name.

If I could change history, making the system smart enough to figure out if the 
parameter is a template or not would be a good idea (eliminating even the 
dynrestpath=on requirement)

But since we can't change history, I think it would make more sense to make 
dynrestpath be the name of the template to use to be consistant with all other 
similar uses.

and if we do ever make the config parser smart enough to check if the 
file/restpath/etc parameter is the name of a template, then we can do the same 
thing here. (such a change would have a small, but non-zero chance of breaking 
existing configs, so it may require a global option to enable, but in the long 
run would simplify things)

David Lang

darakian:
Many thanks for the help. For anyone that lands here due to a search I had my config in
```
/etc/rsyslog.d/my.conf
```
with the following
```
$template jsonRfc5424Template,"{\"type\":\"syslog\",\"host\":\"%HOSTNAME%\",\"role\":\"api-hello-app\",\"message\":\"<%PRI%>1 %TIMESTAMP:::date-rfc3339% %HOSTNAME% %APP-NAME% %PROCID% %MSGID% %STRUCTURED-DATA% %msg:::json%\"}\n"

module(load="omhttp")
action(type="omhttp" server="127.0.0.1" serverport="8080" restpath="events" dynRestPath="off" usehttps="off" template="jsonRfc5424Template")
```
The template is in place because my log sink requires json formatted logging. One annoyance I came across was. 
usehttps = on/off does not jive with me. usehttps = yes/no does. Minor issue, but counterintuitive.

PR #2996, opened 2018-09-06
build system: fix platform linker options detection

rgerhards:
most importantly, it did not work for AIX, which resulted in problems
starting up rsyslog


rgerhards:
"green CI" is sufficient, red ones are caused by testing CI with kafka settings.

Issue #2960, opened 2018-08-28
imfile module: File reading based on rate limit


dpkkumar01:
I am using imfile module to read my file and sending the logs to kafka. I want to restrict the file reading logic at imfile module based on number of lines or size to maintain my rsyslog resource usage constant. 

whenever i am stopping  and starting the process, it is taking maximum logs and sending to kafka module during the time my system resource utilisation was going to very High which is creating the trouble on my other running applications.

Please suggest on the same.

Thanks..

rgerhards:
you can do rate-limiting at the output, albeit it is limited. However, decent queue size for the worker should solve the problem by itself. Please post full config and impstats log.

dpkkumar01:
@rgerhards ,  I think we should have a control at input side to restrict the entry data instead of making the data in queue at output side. This may break in some case depends on the number of input data.

Please provide your suggestions.....

davidelang:
On Wed, 5 Sep 2018, dpkkumar01 wrote:

> @rgerhards , I think we should have a control at input side to restrict the 
> entry data instead of making the data in queue at output side. This may break 
> in some case depends on the number of input data.

As a general rule, I would disagree with throttling the input, but reading from 
files is a special case where the inputs can appear in _very_ large chunks 
(minutes to hours worth of logs appearing instantly).

We already have limits that say 'only read X messages from a file before moving 
on to the next file' to try and provide some fairness and prevent one file from 
overwelming all other files, it makes sense to me to add additional limits so 
that we can provide some sort of limit/fairness between file input and other 
inputs.

David Lang

rgerhards:
yeah, but you can solve this with a dedicated rule set and proper queue settings for this one. Doesn't look to me like a top priority feature request.

rgerhards:
I just notice I hadn't commented on the queue size, thought I had. It's configured pretty large at 2,000,000 so it's clear that imfile pushes two million log lines before throtteling. I suggest reducing that to whatever is appropriate, maybe 2000? The let's see if any issue at all exists.

davidelang:
On Thu, 6 Sep 2018, Rainer Gerhards wrote:

> yeah, but you can solve this with a dedicated rule set and proper queue 
> settings for this one. Doesn't look to me like a top priority feature request.

can you provide an example of how you would setup this separate queue to rate 
limit how fast the queue outputs logs into the default ruleset?

David Lang

rgerhards:
[...]

PR #2999, opened 2018-09-06
imfile: experimental input throtteling feature

rgerhards:
The new input parameter delay.message has been added. It specifies a delay in microseconds after each line read.
    
closes https://github.com/rsyslog/rsyslog/issues/2960

### DIFF
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -62,6 +62,7 @@
 #include "stringbuf.h"
 #include "ruleset.h"
 #include "ratelimit.h"
+#include "srUtils.h"
 #include "parserif.h"
 
 #include <regex.h>
@@ -128,6 +129,7 @@ struct instanceConf_s {
 	int iFacility;
 	int iSeverity;
 	int readTimeout;
+	unsigned delay_perMsg;
 	sbool bRMStateOnDel;
 	uint8_t readMode;
 	uchar *startRegex;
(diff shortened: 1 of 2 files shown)


plugins/imfile/imfile.c
```diff
@@ -1313,6 +1316,11 @@ enqLine(act_obj_t *const act,
 		metadata_values[1] = file_offset;
 		msgAddMultiMetadata(pMsg, metadata_names, metadata_values, 2);
 	}
+
+	if(inst->delay_perMsg) {
+		srSleep(inst->delay_perMsg % 1000000, inst->delay_perMsg / 1000000);
```
portante, line 1321: Why delay inside `enqLine()`?  Won't that mean the persisted state will be delayed as well?

If we moved this delay to [line 1552](https://github.com/rgerhards/rsyslog/blob/7d5b1c6fbac679bbaf1887bb02c9e0f6c16a50dc/plugins/imfile/imfile.c#L1552), would that be cleaner by keeping `enqLine()` just about queuing, and the handler logic in the same place in the read loop?

Also, have we considered using absolute times instead of relative sleep delays?  Relative sleeps can slide significantly from the target goal.  One of the reasons why POSIX uses absolute times for timed waits, e.g. [`pthread_cond_timedwait()`](http://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_cond_timedwait.html).

PR #3009, opened 2018-09-09
omprog: fix forceSingleInstance flag

### LINKED ISSUE #2468: omprog + disk-assisted queue does not honor workerThreads/forceSingleInstance
### Expected behavior
A single external process always being spawned

### Actual behavior
Multiple processes spawned - one for new messages and one for messages being drained from the DA queue. 

rsyslog configuration
```
module(load="omprog")

if ($syslogfacility-text == 'local0') then
{
  action(
    type="omprog"
    name="slowwrite"
    binary="/root/slowwrite.sh"
    forceSingleInstance="on"
    queue.type="fixedArray"
    queue.size="101"
    queue.filename="slowwrite"
    queue.workerThreads="1"
  )
}
```

omprog script:
```
#!/bin/sh

pid=$$
while read line
do
        rec=`date`
        sleep 3
        write=`date`
        echo "$pid|$rec|$write|$line" >> /tmp/slowwrite.dat
done
```

Execution:
`while true; do logger -p local0.info testing;done`

Watch the PID in the /tmp/slowwrite.dat in combination with ps and/or the work directory where the DA files are (/usr/lib/rsyslog on RHEL7)

### Environment
- rsyslog version: [e.g. 8.31.0 -- see rsyslogd -v]
8.24.0
- platform: [e.g. Ubuntu 16.04]
Redhat Enterprise Linux 7.4

- for configuration questions/issues, include rsyslog.conf:
Included above in the setup.

### LINKED ISSUE #2813: omprog: 'forceSingleInstance' flag not working as intended
The intended purpose of the `forceSingleInstance` flag was to force omprog to start only one instance (process) of the external program. However, all this flag does is serializing the execution of `doAction` between the worker instances (i.e. preventing two worker threads from concurrently executing `doAction`).

See [this conversation](https://github.com/rsyslog/rsyslog/issues/473#issuecomment-392350337) for more details:

@rgerhards said:
> The original intent of forceSingleInstance was (and is) to support program that cannot run in multiple instances. For example, SEC (IIRC), needs to have all data fed to a single instance, else it cannot correlate the data as expected. So it looks like this is a bug, and a bug we should fix.

@jsiwrk said:
> (...) it seems that this feature can be achieved by simply restricting the number of workers to 1 (?).

@rgerhards said:
> Yes, that's right, but other actions on the same queue may want to have more than one worker. So omprog would force everything to one worker without that option.

jsiwrk:
Closes #2813
Closes #2468

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,6 +69,7 @@ cscope.out
 tests/HOSTNAME
 tests/chkseq
 tests/diagtalker
+tests/have_relpSrvSetOversizeMode
 tests/inputfilegen
 tests/minitcpsrv
 tests/msleep
@@ -78,11 +79,13 @@ tests/randomgen
 tests/syslog_caller
 tests/omrelp_dflt_port
 tests/syslog_inject
+tests/test_id
 tests/tcpflood
 tests/uxsockrcvr
 tests/xlate.lkp_tbl
 tests/xlate_1.lkp_tbl
 tests/xlate_array.lkp_tbl
+tests/rstb_*
 .vscode/
 *.sw*
 clang_output_*
--- a/plugins/omprog/omprog.c
+++ b/plugins/omprog/omprog.c
@@ -64,6 +64,13 @@ DEF_OMOD_STATIC_DATA
 #define OUTPUT_CAPTURE_BUFFER_SIZE 4096
 #define MAX_FD_TO_CLOSE 65535
 
+typedef struct childProcessCtx {
+	int bIsRunning;		/* is program currently running? 0-no, 1-yes */
+	pid_t pid;			/* pid of currently running child process */
+	int fdPipeOut;		/* fd for sending messages to the program */
+	int fdPipeIn;		/* fd for receiving status messages from the program, or -1 */
+} childProcessCtx_t;
+
 typedef struct outputCaptureCtx {
 	uchar *szFileName;		/* name of file to write the program output to, or NULL */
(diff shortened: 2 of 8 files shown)


jsiwrk:
I've had to disable the new tests on Solaris [for the same reason](https://github.com/rsyslog/rsyslog/pull/2983#issuecomment-418199907) than in #2983: they cause a hang unrelated to this PR (possibly due to a race condition in the rsyslog core?). This issue seems to affect all omprog tests that start multiple workers. In a future PR I'll try to find out if this problem is related to the legacy transaction interface.

rgerhards:
thx for the PR.

side-note: it would be great if you could put full URLs into the commit instead of #issue-nbr. The reason is that the short form has no meaning outside of github. As such, a more elaborate description would also be good. Github may go away some day (Sourceforge did!) and then we have no way to get the full info if it is not present inside the commit text.

rgerhards:
@jsiwrk just FYI: this is what I wrote up for the ChangeLog: https://github.com/rsyslog/rsyslog/commit/a978072b864324a3a6678660983779b3d2410a1b

In an ideal world, this would match the commit text (I can't update it as it would break the git hash chain on master).

jsiwrk:
Agreed. I normally write more elaborate commit messages but in this case I thought a pointer to the (long) github discussion would suffice. But as you say the commit history needs to be self-descriptive! Taking note... Also taking note of using the full issue URL. (BTW, does github also auto-close issues when referencing them in this way?)

PR #3013, opened 2018-09-11
Imkafka: Added multithreaded support for kafka consumers

alorbach:
Imkafka now creates a worker thread for each configured kafka input. 
Added two tests with 4 and 8 imkafka input sources.

### DIFF
--- a/plugins/imkafka/imkafka.c
+++ b/plugins/imkafka/imkafka.c
@@ -61,6 +61,10 @@ DEFobjCurrIf(ruleset)
 DEFobjCurrIf(glbl)
 DEFobjCurrIf(statsobj)
 
+/* forward references */
+static void * imkafkawrkr(void *myself);
+
+
 struct kafka_params {
 	const char *name;
 	const char *val;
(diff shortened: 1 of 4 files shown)


plugins/imkafka/imkafka.c
```diff
@@ -692,53 +707,53 @@ ENDfreeCnf
...
 		 * against totally hogging the CPU if the users selects a polling interval
 		 * of 0 seconds. It doesn't hurt any other valid scenario. So do not remove.
-		 * rgerhards, 2008-02-14
 		 */
 		if(glbl.GetGlobalInputTermState() == 0)
 			srSleep(0, 100000);
-	} while(glbl.GetGlobalInputTermState() == 0);
+	}
+
+	// TODO: Wait for Worker threads to be finished!
```
rgerhards: I think this comment no longer applies (see `afterRun`)? If so, please remove. Else we need to do the implementation.

tests/imkafka_multi_single.sh
```diff
@@ -0,0 +1,191 @@
...
+
+echo \[sndrcv_kafka.sh\]: stop kafka instances
+. $srcdir/diag.sh stop-kafka
+. $srcdir/diag.sh stop-zookeeper
+
+# Do the final sequence check
+seq_check 1 $TESTMESSAGESFULL -d
+
+echo success
+exit_test
```
rgerhards: The lines needs to be properly terminated (LF)

tests/imkafka_multi_many.sh
```diff
@@ -0,0 +1,155 @@
...
+. $srcdir/diag.sh stop-kafka '.dep_wrk3'
+. $srcdir/diag.sh stop-zookeeper '.dep_wrk1'
+. $srcdir/diag.sh stop-zookeeper '.dep_wrk2'
+. $srcdir/diag.sh stop-zookeeper '.dep_wrk3'
+
+# Do the final sequence check
+seq_check 1 $TESTMESSAGESFULL -d
+
+echo success
+exit_test
```
rgerhards: The lines needs to be properly terminated (LF)

plugins/imkafka/imkafka.c
alorbach: oh indeed, will remove

tests/imkafka_multi_single.sh
alorbach: converted into Unix LF

tests/imkafka_multi_many.sh
alorbach: converted into Unix LF

PR #3014, opened 2018-09-11
imfile: support for endmsg.regex

richm:
This adds support for endmsg.regex.  It is similar to
startmsg.regex except that it matches the line that denotes
the end of the message, rather than the start of the next message.
This is primarily for container log file use cases such as this:

    date stdout P start of message
    date stdout P  middle of message
    date stdout F  end of message

The `F` means this is the line which contains the final part of
the message.  The fully assembled message should be
`start of message middle of message end of message`.
`endmsg.regex="^[^ ]+ stdout F "` will match.

### DIFF
--- a/tests/imfile-endmsg.regex-vg.sh
+++ b/tests/imfile-endmsg.regex-vg.sh
@@ -1,0 +1,109 @@
+#!/bin/bash
+# This is part of the rsyslog testbench, licensed under ASL 2.0
+# This test tests imfile endmsg.regex.
+USE_VALGRIND=true
+echo ======================================================================
+echo [imfile-endmsg.regex.sh]
+. $srcdir/diag.sh check-inotify
+. $srcdir/diag.sh init
+generate_conf
+add_conf '
+module(load="../plugins/imfile/.libs/imfile")
+
+input(type="imfile"
+      File="./'$RSYSLOG_DYNNAME'.crio.input"
+      Tag="file:" addMetadata="on" escapelf="on"
+      endmsg.regex="^[^ ]+ (stdout|stderr) F ")
+
+template(name="outfmt" type="list") {
+  property(name="$!all-json-plain")
+  constant(value=" ")
+  property(name="msg" format="json")
+  constant(value="\n")
+}
+
+if $msg contains "msgnum:" then
+ action(
+   type="omfile"
+   file=`echo $RSYSLOG_OUT_LOG`
+   template="outfmt"
+ )
+'
+if [ "x${USE_VALGRIND:-false}" == "xtrue" ] ; then
+	startup_vg
+else
+	startup
+fi
+if [ -n "${USE_GDB:-}" ] ; then
+	echo attach gdb here
+	sleep 54321 || :
+fi
+
+# write the beginning of the file
+echo 'date stdout P msgnum:0' > $RSYSLOG_DYNNAME.crio.input
+echo 'date stdout F msgnum:1' >> $RSYSLOG_DYNNAME.crio.input
+echo 'date stdout F msgnum:2' >> $RSYSLOG_DYNNAME.crio.input
+
+# sleep a little to give rsyslog a chance to begin processing
+if [ -n "${USE_GDB:-}" ] ; then
+	sleep 54321 || :
+else
+	sleep 1
+fi
+
+echo 'date stdout P msgnum:3' >> $RSYSLOG_DYNNAME.crio.input
+echo 'date stdout P msgnum:4' >> $RSYSLOG_DYNNAME.crio.input
+echo 'date stdout P msgnum:5' >> $RSYSLOG_DYNNAME.crio.input
+
+# give it time to finish
+if [ -n "${USE_GDB:-}" ] ; then
+	sleep 54321 || :
+else
+	sleep 1
+fi
+
+echo 'date stdout F msgnum:6' >> $RSYSLOG_DYNNAME.crio.input
+echo 'date stdout P msgnum:7' >> $RSYSLOG_DYNNAME.crio.input
+
+shutdown_when_empty # shut down rsyslogd when done processing messages
+if [ "x${USE_VALGRIND:-false}" == "xtrue" ] ; then
+	wait_shutdown_vg
+	check_exit_vg
+else
+	wait_shutdown    # we need to wait until rsyslogd is finished!
+fi
+
+# give it time to write the output file
+sleep 1
+
+## check if we have the correct number of messages
+
+NUMLINES=$(wc -l $RSYSLOG_OUT_LOG | awk '{print $1}' 2>/dev/null)
+
+rc=0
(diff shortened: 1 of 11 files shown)


richm:
not sure why the solaris11x64 test is failing - does it have the imfile + wildcard functionality?  It is odd because only the crio file logs are showing up:
```
Error: the expected string 'metadata.*filename.*json[.]input.*msgnum:0.*msgnum:1' not found
Error: the expected string 'metadata.*filename.*json[.]input.*msgnum:2' not found
Error: the expected string 'metadata.*filename.*json[.]input.*msgnum:3.*msgnum:4.*msgnum:5.*msgnum:6' not found
{"metadata":{"filename":"\/export\/home\/buildbot\/rsyslog\/rsyslog_solaris11x64_rsyslog\/build\/tests\/rstb_625992_77b797f4.crio.input","fileoffset":"0"}} date stdout P msgnum:0\\ndate stdout F msgnum:1
{"metadata":{"filename":"\/export\/home\/buildbot\/rsyslog\/rsyslog_solaris11x64_rsyslog\/build\/tests\/rstb_625992_77b797f4.crio.input","fileoffset":"46"}} date stdout F msgnum:2
{"metadata":{"filename":"\/export\/home\/buildbot\/rsyslog\/rsyslog_solaris11x64_rsyslog\/build\/tests\/rstb_625992_77b797f4.crio.input","fileoffset":"69"}} date stdout P msgnum:3\\ndate stdout P msgnum:4\\ndate stdout P msgnum:5\\ndate stdout F msgnum:6
```
not sure what's going on

richm:
docker-ubuntu18-distcheck rsyslog looks like a flake?
```
FAIL: sndrcv_kafka_multi_topics.sh
```

rgerhards:
@richm the recent CI improvements discovered some issues in the testbench itself as well as some potentially long-hidden bugs. This is a bit bad from a CI reliability PoV, but of course it is good that it is caught now. I am working on these issues as a side-line. If you come across something unrelated (now or within the next weeks), just leave it as is. I can then check the actual failed run and maybe get some better info on what needs to be fixed. As I know there are CI-itself issues, I babysit all CI failures and restart where appropriate. I can let you know if I think something really needs your attention. I just usually do not do this if the issue looks clearly related to the PR, e.g. inside a new test or a test that checks the functionality.

Sry for the hassle, but there is no easy answer to improving the overall picture. Some pain associated...

tests/imfile-endmsg.regex-vg.sh
```diff
@@ -0,0 +1,109 @@
...
+}
+
+if $msg contains "msgnum:" then
+ action(
+   type="omfile"
+   file=`echo $RSYSLOG_OUT_LOG`
+   template="outfmt"
+ )
+'
+if [ "x${USE_VALGRIND:-false}" == "xtrue" ] ; then
```
rgerhards, line 32: I will merge this as is, but together with the constant "use_valgrind=true" this is against the testbench plumbing. 

We prefer to have dedicated valgrind tests, which always use valgrind, and those that do not use valgrind. The latter ones can also be run under ASAN control. ASAN and valgrind actually have a set of issue that they do NOT catch both, so running under both makes sense. Also, valgrind considerably changes multithreading properties, so many potential races do not occur when running under valgrind.

We could also change the testbench plumbing so that we do valgrind conditionally. But then we would also need to find a way to run the same test with and without valgrind. I am open to that, and happy to discuss options.

This is not really PR specific, and so I will merge without this being resolved. I would still think it makes sense to discuss the topic.

tests/imfile-endmsg.regex-with-example.sh
```diff
@@ -0,0 +1,152 @@
...
+fi
+
+echo 'date stdout P msgnum:3' >> $RSYSLOG_DYNNAME.crio.input
+echo '{"time":"date", "stream":"stdout", "log":"msgnum:3"}' >> $RSYSLOG_DYNNAME.json.input
+echo 'date stdout P msgnum:4' >> $RSYSLOG_DYNNAME.crio.input
+echo '{"time":"date", "stream":"stdout", "log":"msgnum:4"}' >> $RSYSLOG_DYNNAME.json.input
+echo 'date stdout P msgnum:5' >> $RSYSLOG_DYNNAME.crio.input
+echo '{"time":"date", "stream":"stdout", "log":"msgnum:5"}' >> $RSYSLOG_DYNNAME.json.input
+
+# give it time to finish
```
rgerhards, line 96: if we use this construct more often, it should be moved into the testbench plumbing...

tests/imfile-endmsg.regex.sh
```diff
@@ -0,0 +1,108 @@
...
+		: # ok
+	else
+		echo "Error: the expected string '$string' not found"
+		rc=1
+	fi
+done
+
+if [ $rc -ne 0 ]; then
+    cat $RSYSLOG_OUT_LOG
+    exit 1
```
rgerhards, line 103: this needs to be ```error_exit 1``` - I know that a couple of existing tests has this problem as well. Again, I'll merge and update the test scripts later on (together with the other ones).

rgerhards:
> not sure why the solaris11x64 test is failing - does it have the imfile + wildcard functionality? It is odd because only the crio file logs are showing up:

It has something similiar (Solaris FEN). Test should work, but maybe we hit one of the instabilities (Solaris boxes are at times reeeeaaaaly slow, which could cause issues with current plumbing).

richm:
re: CI flakiness - I totally understand - I have the exact same experience with all of the CI systems I use regularly - it seems an inherent feature of CI is instability :-(

richm:
re: separate valgrind tests - I just really dislike having to maintain two files which are almost identical - I'd rather the files be completely identical except one sets USE_VALGRIND=false and one sets USE_VALGRIND=true - that way, I can do all of the work on the one file, and test it with and without valgrind.  When I'm satisfied with the test, I can just do `cp test.sh test-vg.sh` then edit `test-vg.sh` to use `USE_VALGRIND=true`.

tests/imfile-endmsg.regex.sh
richm: feel free to edit my commits/PRs as you see fit

rgerhards:
@richm 
> > `cp test.sh test-vg.sh` then edit `test-vg.sh` to use `USE_VALGRIND=true`.
> 
> good point... let us think a bit how we could handle that inside the testbench plumbing (and I would like to avoid doing `exit 77` skips as this messes up the test log). Something for 8.39 ;-)

actually, I was too quick to answer. Judging your requirement once again, I think what you want to do is supported by the current testbench. If you look at `diag.sh`, you'll see various commented-out `valgrind` settings (via environment variables). They are in the top 100 lines of that file. If you uncomment them (or set the env var to some value of your linking, all tests will run under valgrind. 

Does that solve your need as well?

richm:
> Does that solve your need as well?

It might - I'll take a look - Thanks!

PR #3017, opened 2018-09-13
Fixed pointers for imptcp counting

jeverakes:


jeverakes:
Fixed error, resubmitting.

PR #3023, opened 2018-09-17
I2938

rgerhards:


rgerhards:
Probably to be replaced by #3026

rgerhards:
replaced by #3026, which is a better solution

PR #3032, opened 2018-09-19
imfile/core: refactor imfile support functions EXPERIMENTAL

rgerhards:
**At this stage, this is an expriment which should help decide if restructuring really makes sense.** But I need to full CI, so this is made a PR.

they are currently in stream.c class, and it is considered
to move them to imfile soley.


rgerhards:
see also #3036

rgerhards:
solved via https://github.com/rsyslog/rsyslog/pull/3036

PR #3011, opened 2018-09-10
testbench: try to reproduce imfile "file no longer monitored" error

rgerhards:
see also https://github.com/rsyslog/rsyslog/issues/1605

PR #3036, opened 2018-09-20
imfile: improve truncation detection

rgerhards:
previously, truncation was only detected at end of file. Especially with
busy files that could cause loss of data and possibly also stall imfile
reading. The new code now also checks during each read. Obviously, there
is some additional overhead associated with that, but this is unavoidable.

It still is highly recommended NOT to turn on "reopenOnTruncate" in imfile.
Note that there are also inherant reliability issues. There is no way to
"fix" these, as they are caused by races between the process(es) who truncate
and rsyslog reading the file. But with the new code, the "problem window"
should be much smaller and, more importantly, imfile should not stall.

see also https://github.com/rsyslog/rsyslog/issues/1605
possibly replaces #3011

### DIFF
--- a/runtime/rsyslog.h
+++ b/runtime/rsyslog.h
@@ -221,9 +221,9 @@ enum rsRetVal_				/** return value. All methods return this if not specified oth
 	/* begin regular error codes */
 	RS_RET_NOT_IMPLEMENTED = -7,	/**< implementation is missing (probably internal error or lazyness ;)) */
 	RS_RET_OUT_OF_MEMORY = -6,	/**< memory allocation failed */
-	RS_RET_PROVIDED_BUFFER_TOO_SMALL = -50,
-/*< the caller provided a buffer, but the called function sees the size of this buffer is too small -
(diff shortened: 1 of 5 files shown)


rgerhards:
We should decide in 8.39 time frame if we want this - or not.

gwohletz:
This seems to be badly broken.  Since we updated to 8.39 we are getting logs infinitely being resent with this message 

syslog:Oct 31 17:36:42 snap1.staging.addepar.com rsyslogd: file 'XXXXX': truncation detected, (cannot seek backward to begin of last block) - re-start reading from beginning: Inappropriate ioctl for device [v8.39.0 try http://www.rsyslog.com/e/51 ]

and file XXXXX has definitely not been truncated.  To make matters worse it seems the package repo has removed the older version...

rgerhards:
thx for the report but pls create an issue for this, I have only by accident seen this in the closed PR. Be sure to include a bit about your environment including file system.

PR #3040, opened 2018-09-21
CI: add valgrind suppressions for -coverage gcc option

rgerhards:


rgerhards:
CI failures are false positives

PR #3041, opened 2018-09-21
omprog: document 'useTransactions' flag as experimental

jsiwrk:
The rsyslog module interface was not designed to support modules that
can act both as transactional and non-transactional according to
a configuration parameter. The old transaction interface (incidentally)
supports this, but it is not supported by the new transaction interface.

Therefore a future change is envisaged regarding this feature, that will
likely break backwards compatibility.


jsiwrk:
As per an email conversation with @rgerhards.

rgerhards:
I am looking at the CI issues, will let you know if you need to check things out.

PR #3043, opened 2018-09-22
CI: customize CodeCov settings

rgerhards:
Initial customization, to be changed as we begin to really use the tool.


rgerhards:
CI fail cannot be related as this just provides config update to CodeCov.

PR #3048, opened 2018-09-23
fix trailing whitespace in Python files

rgerhards:


rgerhards:
failure in CI due to known CI bug

PR #3051, opened 2018-09-23
bugfix imfile: file change was not reliably detected

rgerhards:
A change in the inode was not detected under all circumstances, most importantly not in some logrotate cases.
    
closes https://github.com/rsyslog/rsyslog/issues/1605


jvymazal:
@rgerhards do you please know if the test somewhat reliably fails on rsyslog without this patch?

Thanks

rgerhards:
Don't remember exactly, but I think it did.

PR #3056, opened 2018-09-25
testbench: add very basic ommail test

jgerhards:
not a very good test at the moment, but still better than none at all.


rgerhards:
CI failures are unrelated (kafka)

Issue #3057, opened 2018-09-25
testbench: kafka test instabilities


rgerhards:
unfortunately, we still have them. 

All recorded kafka test failures: http://build.rsyslog.com/testbench-showtest.php?testname=%kafka%.sh

Samples:

* sequence error
  * https://build.rsyslog.com/#/builders/142/builds/1619
  * https://build.rsyslog.com/#/builders/142/builds/1620
  * https://build.rsyslog.com/#/builders/142/builds/1621
  * https://build.rsyslog.com/#/builders/142/builds/1624
  * https://build.rsyslog.com/#/builders/138/builds/762
  ... and many more - looks like it now fails in 90% of the cases

  * https://build.rsyslog.com/#/builders/137/builds/984 - different one, broker die while rsyslog runs
    For things like this, we could check for "broker died" error messages from rsyslog and, if found, skip the test.

* topic creation error
  * https://build.rsyslog.com/#/builders/137/builds/962

rgerhards:
@alorbach `sndrcv_kafka.sh` skipped in CI as it causes lots of false positives

rgerhards:
I disabled kafka tests in most of the buildbot runs, except where absolutely needed. Add a special builder to carry them out (named "... FLACKY").

alorbach:
PR https://github.com/rsyslog/rsyslog/pull/3101 may solve many of thes instability issues we have seen under load. Because under load, the kafka / zookeeper server may not be finished within previous timeframe of 2 seconds, and shutting down zookeeper before kafka resulted in a hanging kafka server instance (Because kafka tries to connect to zookeeper and this blocks the shutdown).

rgerhards:
see also https://github.com/rsyslog/rsyslog/issues/3088

alorbach:
I guess this can be closed by now?

rgerhards:
Not sure, I see a couple of failures: http://build.rsyslog.com/testbench-showtest.php?testname=%kafka%

Exit code 177 is "system error" (so not of our concern) but 1 should be real failures. We need to investigate if they are indeed real or some part of the test is using wrong exit code.

alorbach:
I see what you mean .

rgerhards:
The past couple of weeks seem to indicate we have fixed this by now.

PR #3059, opened 2018-09-26
testbench: restart kafka if brokers do not come up

rgerhards:
kafka tests are plagued by brokers failing to start, which then causes
(false positive) test failures.

We now to detect missing brokers, and do a kafka/zookeepr shutdown and
restart in this case. If that fails again, let's skip the test instead
of failing it - after all, it's not the issue of the to-be-checked code.

see also https://github.com/rsyslog/rsyslog/issues/3057


rgerhards:
CI failures are due to the other kafka instability, so good to go for here.

PR #3061, opened 2018-09-26
testbench: add simple omsnmp test

jgerhards:
currently only checks if module is loaded and checks for
required parameters correctly

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -344,6 +344,10 @@ TESTS +=  \
 endif # HAVE_VALGRIND
 endif # ENABLE_DEFAULT_TESTS
 
+if ENABLE_SNMP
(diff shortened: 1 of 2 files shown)


tests/Makefile.am
```diff
@@ -345,7 +345,12 @@ endif # HAVE_VALGRIND
 endif # ENABLE_DEFAULT_TESTS
 
 
+if ENABLE_SNMP
+TESTS +=  \
+	omsnmp_errmsg_no_params.sh
+endif # if ENABLE_SNMP
 if ENABLE_MAIL
+
```
rgerhards, line 351: I guess the empty line should be in front of the new `if`?

jgerhards: fixed

Issue #3066, opened 2018-09-27
add test for queue encryption


rgerhards:
Actually, we should have two tests:

- [ ] Disk queue with encryption
- [ ] DA queue with encryption (this probably fails, see #2575)

rgerhards:
see also https://github.com/rsyslog/rsyslog/issues/2575

PR #3080, opened 2018-09-28
testbench: add test for encrypted disk queue

rgerhards:
see also https://github.com/rsyslog/rsyslog/issues/3066


rgerhards:
this will be merged voa #3081, which has the same code at it's base. Some fixes were necessary, so we do not  carry forward both PRs.

rgerhards:
merged via https://github.com/rsyslog/rsyslog/pull/3081

PR #3097, opened 2018-09-30
fix bug in my previous commit

rgerhards:
as the kafka tests failed so frequently, I overlooked that I
actually broken one...


rgerhards:
CI failures are unrelated.

PR #3109, opened 2018-10-09
poll() loop. fix mismatch between sock list and pollvector index

jelaas:
There is a mismatch between how the pollfds fd-array is created and how it is read when processing a poll() result.
At creation time the code currently skips lstn->sock if it is -1. This is not done when processing the poll event.
A possibility is also to skip these checks altogether since poll() should ignore fds that have value -1.


rgerhards:
thx for the PR. Does it fix a problem you had? If so, could you describe it for the ChangeLog?

jelaas:
I cannot be certain. We had a problem after upgrade where rsyslog after some time would stop processing at least udp messages. The packets were accounted as buffer errors if I remember correctly.
We backed to 8.36 and everything worked again. I reviewed the changelog and noticed the changes in this area. Reviewed the code and found this problem. Unfortunately we have no easy way of testing.

PR #3121, opened 2018-10-11
cosmetic fix: mangle_qi test tool did not check mandatory arg

rgerhards:


rgerhards:
CodeCov is experimental, thus all go.

PR #3123, opened 2018-10-12
experiment: remove known issue suppression

rgerhards:
this might be solved by https://github.com/rsyslog/rsyslog/pull/3113


rgerhards:
we will give this a try as it looks like this is solved by now

PR #3124, opened 2018-10-12
Bugfix: issue 3105, lmsig-ksils12 

### LINKED ISSUE #3105: lmsig_ksils12: the module skips the signature of the last block in the file in certain conditions
### Expected behaviour
All blocks in the blocks file file must have matching signature element inthe signatures file

### Actual behaviour
In case of async mode and http protocol and certain conditions the last block can lack the 
signature. 

### Steps to reproduce the behaviour
* Configure the module with async mode and http protocol
* Flood the daemon with messages
* During flooding stop or -HUP the module

### Environment
- rsyslog version: 8.38
- platform: centos 7

allanpark:
KSI LS12 module skips signing the last block in the file in some conditions
closes #3105

### DIFF
--- a/runtime/lib_ksils12.c
+++ b/runtime/lib_ksils12.c
@@ -1278,7 +1278,7 @@ static void process_requests(rsksictx ctx, KSI_CTX *ksi_ctx, FILE* outfile) {
 	if (!outfile)
 		goto signing_failed;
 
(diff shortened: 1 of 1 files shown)


rgerhards:
CodeCov is currently experimental, status is OK albeit it shows fail.

PR #3129, opened 2018-10-12
rainerscript: fix parentheses in if clause

jgerhards:
detected by codacy


rgerhards:
This caused regex compile error messages to be incorrect.

PR #3131, opened 2018-10-15
removing result var from context

CodeHarsh:
Fixing https://github.com/rsyslog/rsyslog/issues/3025

### DIFF
--- a/contrib/fmhash/fmhash.c
+++ b/contrib/fmhash/fmhash.c
@@ -158,8 +158,7 @@ hash_wrapper2(struct svar *__restrict__ const sourceVal
 		if (!success) {
 			parser_warnmsg("fmhash: hashXX(string, seed) didn't get a valid 'seed' limit"
 					", defaulting hash value to 0");
-			iRet = RS_RET_ERR;
-			FINALIZE;
+			ABORT_FINALIZE(RS_RET_PARAM_ERROR);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


contrib/fmhash/fmhash.c
```diff
@@ -197,10 +195,9 @@ hash_wrapper3(struct svar *__restrict__ const sourceVal, struct svar *__restrict
 		FINALIZE;
 	}
 
-	CHKiRet((hcontext->hash_wrapper_1_2(sourceVal, seedVal, hcontext)));
-	xhash = hcontext->xhash % mod;
-	DBGPRINTF("fmhash: hashXXmod generated hash-mod %" PRIu64 ".", xhash);
-	hcontext->xhash = xhash;
+	CHKiRet((hcontext->hash_wrapper_1_2(sourceVal, seedVal, hcontext, xhash)));
+	(*xhash) = (*xhash) % mod;
```
rgerhards: cppcheck (via Codacy) has a false positive here, it does not seem to understand that the check in 191 protects again `mod==0`. Nevertheless, we should try to work around this issue. How about using `mod = 1` after line 193?

contrib/fmhash/fmhash.c
```diff
@@ -197,10 +195,9 @@ hash_wrapper3(struct svar *__restrict__ const sourceVal, struct svar *__restrict
 		FINALIZE;
```
rgerhards: These two lines could easier (and more up to the code style) be written as
```
ABORT_FINALIZE(RS_RET_ERR);
```
but probably `RS_RET_PARAM_ERROR` would be more up to the point?

contrib/fmhash/fmhash.c
CodeHarsh: Adding mod =1 made clang analyzer angry. Adding if(mod!=0) did the trick.

rgerhards: > Adding mod =1 made clang analyzer angry. Adding if(mod!=0) did the trick.

oh, the joys of static analyzers... ;-) Thanks for taking care of this. Will merge most probably tomorrow (now  I really need to run out of office ;-)

PR #3145, opened 2018-10-19
tools: add small tool for debug log analysis

rgerhards:


rgerhards:
I think it's not worth adding this...

Issue #3144, opened 2018-10-19
testbench: make test status reporting optional


rgerhards:
the newly introduced test status reporting now always tries to connect to our stats server. This can be considered a privacy invasion and also can wreck our stats. so make this optional and only happen when an explicit URL is given. This can than be set in our CI environment.

PR #3146, opened 2018-10-19
testbench: do test error reporting only if requested

rgerhards:
**NOTE: before merging this, the buildbot config MUST be updated.** Otherwise, we disable stats reporting.

The newly introduced test status reporting now always tries to connect
to our stats server. This can be considered a privacy invasion and
also can wreck our stats. so make this optional and only happen when
an explicit URL is given. This can than be set in our CI environment.

request by setting stats reporting url env var RSYSLOG_STATSURL

closes https://github.com/rsyslog/rsyslog/issues/3144

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,8 @@ compiler:
 
 env:
   global:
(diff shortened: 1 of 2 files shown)


rgerhards:
buildbot definitions have been updated.

rgerhards:
codecov still experimental

Issue #3065, opened 2018-09-27
do at least a small testbench run with debug log enabled


rgerhards:
Prime reason is to check the debug instrumentation code. This should be done for a small, but usefull small sample, e.g. a `imtcp` test or something similar. Single test sufficies. Depending on outcome, we can later check to enhance this scenario.

Note: some of the existing tests cannot be executed in debug mode as this is too slow. Testbench at whole takes around 3 to 4 times longer.

PR #3158, opened 2018-10-22
testbench: add test with debug enabled

jgerhards:
closes https://github.com/rsyslog/rsyslog/issues/3065


tests/mmanon_with_debug.sh
```diff
@@ -0,0 +1,85 @@
...
+ 0.0.0.0.
+ textnoblank0.0.0.0stillnoblank' | cmp - $RSYSLOG_OUT_LOG
+if [ ! $? -eq 0 ]; then
+  echo "invalid response generated, $RSYSLOG_OUT_LOG is:"
+  cat $RSYSLOG_OUT_LOG
+  error_exit  1
+fi;
+if [ ! -e log ]
+then
+    echo "error: file 'log' (Debuglog) not found (should be generated)"
```
rgerhards: `error_exit` is missing
```suggestion
    echo "error: file 'log' (Debuglog) not found (should be generated)"
    error_exit 1
```

tests/mmanon_with_debug.sh
```diff
@@ -0,0 +1,85 @@
...
+ 10.0.0.0.1
+ 0.0.0.0.1
+ 0.0.0.0.
+ textnoblank0.0.0.0stillnoblank' | cmp - $RSYSLOG_OUT_LOG
+if [ ! $? -eq 0 ]; then
+  echo "invalid response generated, $RSYSLOG_OUT_LOG is:"
+  cat $RSYSLOG_OUT_LOG
+  error_exit  1
+fi;
+if [ ! -e log ]
```
rgerhards: I would suggest to place the `then` on the same line, just as in line 76

tests/mmanon_with_debug.sh
jgerhards: fixed

PR #3161, opened 2018-10-23
update shellcheck rules for CodeFactor.io

rgerhards:


rgerhards:
merging early as the PR really only affects CodeFactor.io configuration. All upcoming CI failures, if any, are false positives.

PR #3166, opened 2018-10-23
testbench: try to make omkafka test more stable

rgerhards:
The test often misses messages. One idea is that rsyslog is actually
terminated too early, and thus messages are missing. To proove that idea,
we now explicitely wait for all messages to be processed.

If this does not work out, we should try to run kafkacat inside the
loop - maybe the librdkafka output port of omkafka is too slow.

see also https://github.com/rsyslog/rsyslog/issues/3057


rgerhards:
looks good enough to give it a try and merge to master! @alorbach pls take note

PR #3173, opened 2018-10-24
experiment: work with cppcheck inline suppressions

rgerhards:


rgerhards:
this seems to be no longer necessary in any case...

PR #3181, opened 2018-10-25
fix nit detected by shellcheck

rgerhards:


rgerhards:
codecov is experimental, so CI OK

PR #3188, opened 2018-10-26
cleanup: remove unneeded code

rgerhards:
The syncdemo program was actually that, a demo program to explain
some concepts. It's not really related to the rsyslog code base.


rgerhards:
codecov is still experimental

Issue #2931, opened 2018-08-15
imjournal reloading repeatedly on older CentOS 7.x


jfield:
### Expected behavior
imjournal to reload once per roll

### Actual behavior
imjournal reloading a lot more than once per roll

### Steps to reproduce the behavior
Use imjournal on our 200+ log line per second hosts.
And use a version of CentOS that is no longer current.

I'm mostly documenting this in case someone else runs into it, or in case someone wants to fix it for 7.3.1611 (but really, people should just not run 7.3.1611 anymore!). We're downgrading to 8.36.0-3 and then rolling 8.37+ back out after we upgrade the OS.

### Environment
- rsyslog version: 8.37
- platform: CentOS 7.3.1611

On some higher logging volume 7.3.1611 hosts, we're currently getting spammed with:
```
Aug 14 23:34:03 our-awesome-hostname0000 rsyslogd[25787]: imjournal: journal reloaded... [v8.37.0 try http://www.rsyslog.com/e/0 ]
Aug 14 23:34:03 our-awesome-hostname0000 rsyslogd[25787]: imjournal: journal reloaded... [v8.37.0 try http://www.rsyslog.com/e/0 ]
Aug 14 23:34:03 our-awesome-hostname0000 rsyslogd[25787]: imjournal: journal reloaded... [v8.37.0 try http://www.rsyslog.com/e/0 ]
Aug 14 23:34:03 our-awesome-hostname0000 rsyslogd[25787]: imjournal: journal reloaded... [v8.37.0 try http://www.rsyslog.com/e/0 ]
Aug 14 23:34:03 our-awesome-hostname0000 rsyslogd[25787]: imjournal: journal reloaded... [v8.37.0 try http://www.rsyslog.com/e/0 ]
Aug 14 23:34:03 our-awesome-hostname0000 rsyslogd[25787]: imjournal: journal reloaded... [v8.37.0 try http://www.rsyslog.com/e/0 ]
Aug 14 23:34:03 our-awesome-hostname0000 rsyslogd[25787]: imjournal: journal reloaded... [v8.37.0 try http://www.rsyslog.com/e/0 ]
Aug 14 23:34:03 our-awesome-hostname0000 rsyslogd[25787]: imjournal: journal reloaded... [v8.37.0 try http://www.rsyslog.com/e/0 ]
[...]

PR #3192, opened 2018-10-26
Fix journald excessive reloading behavior

jvymazal:
This is workaround for possible imjournal interaction with systemd
where journal invalidate fix is not present. The code tries to
detect SD_JOURNAL_INVALIDATE loop and not reload after each call.

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -118,7 +118,7 @@ static const char *pidFieldName;	/* read-only after startup */
 static int bPidFallBack;
 static ratelimit_t *ratelimiter = NULL;
 static sd_journal *j;
-static int j_inotify_fd;
+static sbool reloaded = 0;
 static struct {
 	statsobj_t *stats;
 	STATSCOUNTER_DEF(ctrSubmitted, mutCtrSubmitted)
@@ -149,8 +149,6 @@ static rsRetVal openJournal(void) {
 	if ((r = sd_journal_get_fd(j)) < 0) {
 		LogError(-r, RS_RET_IO_ERROR, "imjournal: sd_journal_get_fd() failed");
 		iRet = RS_RET_IO_ERROR;
-	} else {
-		j_inotify_fd = r;
 	}
 	RETiRet;
 }
@@ -160,7 +158,6 @@ static void closeJournal(void) {
 		persistJournalState();
 	}
 	sd_journal_close(j);
-	j_inotify_fd = 0;
 }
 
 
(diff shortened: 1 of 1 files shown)


jvymazal:
This fixes #2931 
It works around journal  rotation detection bug and makes imjournal usable again even for people with systemd without [this](https://github.com/systemd/systemd/pull/5580/commits/09e1b51bf1a07bbc27a2036769860b909b0647e4) fix.

rgerhards:
Thx!

I should probably re-enable the journal tests at least for the PR. Let me do that (hopefully soon). I'll then restart the related CI workers.

jvymazal:
Ok, thanks, I have tried to look at logs you linked in the parent issue, however I am unable to determine what is wrong from them.

rgerhards:
We see some issues again: https://build.rsyslog.com/#/builders/175/builds/660/steps/6/logs/test-suite_log

Not sure if this time it is related to busy log on the CI system...

rgerhards:
I have re-initiated the run

jvymazal:
Ok, I tried to go trough the log file:

On actual test output file there are only two messages (rsyslog start & shutdown), that is strange.
Next, here is what is in the journal on the bot (trimmed to relevant entries), in this order:
```
Oct 26 13:32:11 centos7-5-w3-1vcpu-1gb-fra1-02 rsyslogd[17327]: imfile: on startup file '/var/lib/buildbot/slave/rsyslog_centos7-5/build/tests/rstb_848663_1ee4d6b7.input' does not exist but is configured in static file monitor - this may indicate a misconfiguration. If the file appears at a later time, it will automatically be processed. Reason: No such file or directory [v8.39.0.master]
Oct 26 13:32:11 centos7-5-w3-1vcpu-1gb-fra1-02 rsyslogd[17327]:  [origin software="rsyslogd" swVersion="8.39.0.master" x-pid="17327" x-info="http://www.rsyslog.com"] start
Oct 26 13:32:11 centos7-5-w3-1vcpu-1gb-fra1-02 rsyslogd[17327]:  [origin software="rsyslogd" swVersion="8.39.0.master" x-pid="17327" x-info="http://www.rsyslog.com"] exiting on signal 15.
Oct 26 13:36:05 centos7-5-w3-1vcpu-1gb-fra1-02 imjournal_test[22377]: TestBenCH-RSYSLog imjournal This is a test message - 1540560965 - rstb_110547_944b496a
```
So, as you can see, rsyslog starts, exits almost immediately (within a second) and then, after almost 4 minutes, the tests message finally arrives.

Ant thoughts on this?

rgerhards:
> So, as you can see, rsyslog starts, exits almost immediately (within a second) and then, after almost 4 minutes, the tests message finally arrives.
> 
> Ant thoughts on this?

strange - might this be caused by the journal being too busy?

rgerhards:
@jvymazal FYI: I have re-initiated the test once again; prime reason this time is I want to see if I can gather coverage data. This does not work on the containers, as we do not have journal there running (nor systemd). I also hope the new run might bring up some more interesting info.

rgerhards:
@jvymazal merging this now, it's definitely an improvement. I'll keep an eye on journal test performance and keep you posted if there is anything interesting.

PR #3221, opened 2018-10-30
testbench: grep fails with option error in some tests

rgerhards:


rgerhards:
CodeCov seems to be waay off ;-)

Issue #2211, opened 2017-12-13
cleanup old runtime instrumentation code


rgerhards:
PR https://github.com/rsyslog/rsyslog/pull/2205 removed the ```--enable-rtinst``` configure switch and did some cleanup of the debug code. However, we should do some more agressive cleanup, even though the code is currently mostly dead. It's still not a half hour job, as we need to evaluate which parts we still need, and how they can be best replaced by now-sufficient code.

Actually this is a low priority task.

rgerhards:
a lot of the cleanup has been done in 8.39 timeframe. Some final bits need to be checked.

PR #3222, opened 2018-10-30
cleanup: remove no longer needed --enable-rtinst code

rgerhards:
--enable-rtinst is gone-away since a while, but there were still
some supporting code left. It required careful anaylsis what could
actually be removed. This is now done and the code fully cleaned
up.
    
As a positive side effect, we could eliminate mutex calls inside
the debug system. This means we are more likely to reproduce race
conditions in runs with debugging enabled.
    
closes https://github.com/rsyslog/rsyslog/issues/2211

### DIFF
--- a/action.c
+++ b/action.c
@@ -316,7 +316,7 @@ actionResetQueueParams(void)
 
 	cs.glbliActionResumeRetryCount = 0;		/* I guess it is smart to reset this one, too */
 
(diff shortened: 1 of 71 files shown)


rgerhards:
related doc PR: https://github.com/rsyslog/rsyslog-doc/pull/740

PR #3227, opened 2018-10-30
fix some small issues detected by cppcheck

rgerhards:


rgerhards:
on second thought, this patch indeed is wrong -- and does not make real sense to me (maybe the code changed in between). Closing it, let's see if new warnings pop up and fix them then.

Issue #3006, opened 2018-09-07
`echo $ENV_VAR` causes segfault if $ENV_VAR is not set


dpyro:
### Expected behavior

\`echo $ENV_VAR`` should return an empty string if `$ENV_VAR` is not set

### Actual behavior

Segmentation fault.

```
set $.target = getenv("SYSLOG_TARGET");
```

produces the expected result.

### Steps to reproduce the behavior

```
set $.target = `echo $SYSLOG_TARGET`;
```

### Environment
- rsyslog version: rsyslogd 8.37.0
- platform: CentOS Linux release 7.5.1804 (Core) under docker (centos:7)
- for configuration questions/issues, include rsyslog.conf and included config files

```
set $.target = `echo $SYSLOG_TARGET`
```

julthomas:
Hello,

I believe the bug was recently introduced in commit abe0434 (config: enhance backticks "echo" capability) because strlen() _- but also es_addBuf() -_ is called on the getenv() result that may be NULL if the environment variable does not exist.

That line should be wrapped in an `if (envval != NULL)`
<code>es_addBuf(&estr, envval, strlen(envval));</code>

Julien

rgerhards:
looks like I overlooked the initial posting. @julthomas you are right - are you up for a PR?

rgerhards:
fixed by https://github.com/rsyslog/rsyslog/pull/3237

PR #3237, opened 2018-11-02
config: fix segfault in backticks "echo" expansion of undefined variables

julthomas:
The bug was introduced in commit abe0434 (config: enhance backticks "echo"
capability). The getenv() result passed to strlen() and es_addBuf() may be
NULL if the environment variable does not exist, resulting in a segfault.

This commit fixes Github issue #3006.


rgerhards:
thx for the PR!

CI failures are most probably due to currently ongoing work - will check.

PR #3246, opened 2018-11-05
Fix message timestamps on Solaris

kulikjak:
On Solaris messages don't have their time directly in the raw body but in a separate *log_ctl* structure which is currently not used.

When message is logged and processed, rsyslogd gives it current time because it ignores the actual one. That means that old messages (e.g. from system reboot) get timestamp of processing instead of the reboot itself (it is not a problem for live logging where *now* is used anyway).


rgerhards:

Issue #3104, opened 2018-10-02
testbench: es-bulk-retry test unreliable


alorbach:
Sometimes elasticsearch tests fail like: 
es-bulk-retry.sh 

**recorded failures: http://build.rsyslog.com/testbench-showtest.php?testname=es-bulk-retry.sh**

See buildbot logs: 
* https://build.rsyslog.com/#/builders/138/builds/809/steps/4/logs/test-suite_log
* https://build.rsyslog.com/#/builders/181/builds/256

rgerhards:
The problem persists, we see frequent failures. Just one sample from today: https://build.rsyslog.com/#/builders/138/builds/1032/steps/4/logs/test-suite_log

@richm can you explain what exactly the error is? "assertion failed" is not very informative ;-) I guess the test is racy, but to try to find better syn we need to better understand what exactly fails.

rgerhards:
@richm Good progress! Here we have a case with the new output: https://build.rsyslog.com/#/builders/144/builds/975/steps/3/logs/test-suite_log

I think the most relevant part is:
```
rsyslogd debug: info: trying to cooperatively stop input impstats, timeout 60000 ms
stopping ES with pid 21129
good - no missing or unexpected records were found in Elasticsearch
FAIL: expected 50 successful responses but omelasticsearch stats reported 49
FAIL: The sum of the number of successful responses and bad argument errors and bulk index rejections 217 did not equal the number of requests actually submitted to Elasticsearch 218
```

From that, I would guess that it is "the usual race" we have during shutdown. We should find a way, while rsyslog runs, to check that we got all lines and timeout if not. I'll also see to have a look at the test as soon as a I can.

rgerhards:
yeah, that was my first thought. But it's actually counter values, so we cannot do a simple `grep`. Would it be a solution to not reset counters and then do a `grep` for the exact line? Else we probably need to extend the python a bit, this time probably as real script that we can use in two places. @richm what do you think?

richm:
> Would it be a solution to not reset counters and then do a grep for the exact line?

not sure - how would you reset counters?

rgerhards:
let me check - I thought you had enabled that impstats option...

rgerhards:
@richm my fault - it's not set. So let me see if I can simply `grep` for it...

rgerhards:
[...]

PR #3248, opened 2018-11-05
try to debug es-bulk-retry failures

richm:
Fix es-bulk-retry.sh script and add additional information
when the test fails.
https://github.com/rsyslog/rsyslog/issues/3104

### DIFF
--- a/tests/es-bulk-retry.sh
+++ b/tests/es-bulk-retry.sh
@@ -127,8 +127,12 @@ if [ -f work ] ; then
 	python -c '
 import sys,json
 records = int(sys.argv[1])
+extra_recs = open(sys.argv[2], "w")
+missing_recs = open(sys.argv[3], "w")
 expectedrecs = {}
 rc = 0
+nextra = 0
+nmissing = 0
 for ii in xrange(0, records*2, 2):
 	ss = "{:08}".format(ii)
 	expectedrecs[ss] = ss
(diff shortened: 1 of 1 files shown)


richm:
@rgerhards @alorbach I think I was unable to reproduce the es-bulk-retry test failure.  Going to try a minor change and repush to see if I can re-trigger the CI tests.

rgerhards:
I have dug out the known failures from the database: http://build.rsyslog.com/testbench-showtest.php?testname=es-bulk-retry.sh (note: recording is not yet perfect, so it's a subset).

tests/diag.sh
```diff
@@ -958,7 +958,7 @@ dep_kafka_url=http://www-us.apache.org/dist/kafka/2.0.0/kafka_2.11-2.0.0.tgz
 dep_kafka_cached_file=$dep_cache_dir/kafka_2.11-2.0.0.tgz
 
 if [ -z "$ES_DOWNLOAD" ]; then
-	export ES_DOWNLOAD=elasticsearch-5.6.9.tar.gz
```
rgerhards: do we have a hard reason for the other version or just "this is newer / this is now in widerspread use"? I ask because we intentionally have different versions to cover typical use cases. Also, I cache the ES tarballs inside containers, as somtimes the download takes veeeeery long. So if we change the default, it is wise to update the container def.

rgerhards:
If we know how many lines we expect (or how many with which content), we can change the wait predicate for rsyslog termination. This worked well for many other racy tests. It's what I am doing in #3250 for example: https://github.com/rsyslog/rsyslog/pull/3250/files#diff-a48ab7fa443e450917bf3f12218ab26aR23

rgerhards:
@richm mmmhhh... I tried quite hard to get the tests to fail. Maybe it's related to the newer ES version? I admit I doubt this, but...

otherwise: how about merging this (maybe minus the ES version change) and wait until a problem pops up again. Then we have the instrumentation in place.

What do your think?

rgerhards:
OK, as you can see I have merged. Let's see what happens in the future. @richm I'll let you know when I see a failure, hopefully we then have more insight thx to these changes here.

PR #3259, opened 2018-11-07
ossl nsd driver: Added openssl error stack output into GlblInit

alorbach:
When loading CA, Cert or Keyfile, the OpenSSL error stack will be output into LogError.
Also added 3 tests with invalid certificates/keyfiles for testbench.


alorbach:
Confirmed that testbench integratrion is fine (Y)

PR #3266, opened 2018-11-08
testbench: make es-bulk-retry test more reliable

rgerhards:
We now no longer depend on a fixed 'sleep' command but rather
check the output file for what we expect. This is much more
robust on slow test machines.

We believe this closes the below-mentioned issue. If not, it
should be re-opened.

closes https://github.com/rsyslog/rsyslog/issues/3104


rgerhards:
@richm would be great if you could double-check the PR

richm:
lgtm - thanks!

PR #3276, opened 2018-11-10
buildbot: gather script shall not pickup $RSYSLOG_DYNNAME*.log files

rgerhards:
previously, it also picked test output files if they were left over. For
some test this meant very verbose data could be sent to buildbot master.
We have seen cases with over 20million lines, which made the log view
impossible. This has now been corrected.


rgerhards:
CI fail unrelated

PR #3280, opened 2018-11-11
testbench: modernize some imfile tests, re-enable on FreeBSD

rgerhards:
They did not work on FreeBSD because of an imfile bug which is
now fixed.


rgerhards:
CI fail is unrelated, so ok to merge

Issue #2902, opened 2018-08-02
valgrind tests occasionally fail during shutdown


rgerhards:
This is in rsyslog core. 

**Update analysis in: https://github.com/rsyslog/rsyslog/issues/2902#issuecomment-437300808**

Most probably a new worker thread is scheduled when the worker pool was already shut down. It seems to not be able to actually start, and this will not properly be terminated. Most likely a "harmless" race. Still should be addressed.

In the meantime we should suppress this warning inside CI runs as there is no point in making CI fail for this known issue.

A typical report:
```
==26515== HEAP SUMMARY:
==26515==     in use at exit: 29,685 bytes in 85 blocks
==26515==   total heap usage: 321,712 allocs, 321,628 frees, 98,032,385 bytes allocated
==26515== 
==26515== 640 bytes in 1 blocks are possibly lost in loss record 35 of 41
==26515==    at 0x4C2B9B5: calloc (vg_replace_malloc.c:711)
==26515==    by 0x40128C4: _dl_allocate_tls (in /usr/lib64/ld-2.17.so)
==26515==    by 0x50547FB: pthread_create@@GLIBC_2.2.5 (in /usr/lib64/libpthread-2.17.so)
==26515==    by 0x44B88F: wtpStartWrkr (wtp.c:471)
==26515==    by 0x44B88F: wtpAdviseMaxWorkers (wtp.c:527)
==26515==    by 0x450397: qqueueAdviseMaxWorkers (queue.c:3066)
==26515==    by 0x450397: qqueueEnqMsg (queue.c:3076)
==26515==    by 0x410655: submitMsg2 (rsyslogd.c:1088)
==26515==    by 0x410874: logmsgInternal_doWrite (rsyslogd.c:857)
==26515==    by 0x410978: processImInternal (rsyslogd.c:1667)
==26515==    by 0x40E114: mainloop (rsyslogd.c:1893)
==26515==    by 0x40E114: main (rsyslogd.c:2087)
==26515==
```

rgerhards:
This issue still occurs after #3113 but it seems to occur much less frequently.

rgerhards:
This is another report from Ubuntu 18.04:
```
==28858==    at 0x4C31B25: calloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==28858==    by 0x40134A6: allocate_dtv (dl-tls.c:286)
[...]

PR #3282, opened 2018-11-12
core/wti: changed worker thread termination handling

rgerhards:
while we solved valgrind false positives during shutdown in
2700badfda76c701fa9169c31c9ec4f7993dc0b8, we continue to see
related messages for thread-local-storages. Rather than suppressing
these as well we now refactored the code so that it does not use
detached threads.

This shall improve robustness. In case it does not, we have enhanced
debug messages under CI runs, so potentially still-existing problem
areas should be easier to diagnose.

see also https://github.com/rsyslog/rsyslog/issues/2902


rgerhards:
CI fail is unrelated.

PR #3283, opened 2018-11-12
testbench: fix core file handling when multiple core files exist

rgerhards:
This is primarily a cosmetic fix, as multiple core files should never
exist. The only exception might be parallel builds, but in this case
the whole concept does not work correctly (we may probably see into
that when we finally support parallel builds, which we do not yet).

detected by shellcheck


rgerhards:
CI failure is false positive

PR #3286, opened 2018-11-12
testbench: harden test against slow machines

rgerhards:


rgerhards:
Travis is false positive (problems installing OS packages).

Issue #3294, opened 2018-11-14
omrelp does not (always) emit error message when it suspends itself


rgerhards:
while trying to craft testbench test for https://github.com/rsyslog/rsyslog/issues/3290 I can confirm that omrelp suspends itself, but does not emit a message why.

rgerhards:
core problem is that librelp generates almost no error messages: https://github.com/rsyslog/librelp/issues/130

PR #3295, opened 2018-11-14
bugfix omrelp: emit error messages on suspension

rgerhards:
The module did not emit error messages. Actually, it would have done
    if librelp would emit them (callback is in place), but librelp does
    not do. We expect to fix this at the librelp level as well. But to
    make this work with all versions of librelp, we have added basic
    error messages to omrelp. They may not be totally easy to understand,
    but at least provide the librelp error code (that's the best we can
    do in this situation).
    
see also https://github.com/rsyslog/librelp/issues/130
closes https://github.com/rsyslog/rsyslog/issues/3294


rgerhards:
CI is false positive.

PR #3298, opened 2018-11-14
Https

jsoref:
There are three commits here:
1. Clarify that the trailing `.` in one line is not actually part of the url
1. Use a canonical url for www.rsyslog.com/doc/ (both flavors existed in the repo) -- this will happen if you use a web browser to visit the url anyway
1. Replace `http:` with `https:` for www.rsyslog.com -- this will also happen if you use a web browser to visit the url

There's no good reason to list an http url when the user agent will just be forced to redirect to the https url.


rgerhards:
Thx for the PR. We are moving towards ASL 2.0 as a license. Some of the files that you applied changes to are still under GPLv3+. Do you agree to license **your patch** on these files under ASL 2.0?

jsoref:
Sure

Issue #3039, opened 2018-09-21
random() return negative


snaix:
### Expected behavior
random(20);
return a random number between 0 and 20.
### Actual behavior
return a random number between -20 and 20.
### Steps to reproduce the behavior
set $.rand = random(20);
Export $.rand via template.
### Environment
- rsyslog version: 8.37.0
- platform: centos7

msn:
Also seeing this on centos 7 with rsyslog 8.38. Calling random(100) returns numbers from -100 to 100, which seems somewhat counter intuitive! In theory this could be how it is supposed to work. The documentation could certainly be more clear on the sort of values that are expected to be returned.

rgerhards:
Apologies for the very late reply. This is probably no longer relevant for the original reporter, but I am adding the result for people who find this issue via search while we work through the old backlog. We do not close old issues just because of age.

I checked current main and this appears to be fixed now. random(max) currently takes the absolute value of the random source before applying the modulo, so it should not return negative values for normal positive limits. The existing tests/rscript_random.sh test also sends messages through rsyslog, writes random(10) to an omfile destination, and verifies the generated values match the expected non-negative single-digit range. I re-ran that test successfully on current main.

The current documentation describes random(max) as generating values starting at 0, so the old negative-result behavior is not expected anymore.

Closing this as fixed in current versions. If anyone still sees negative random() values with a current rsyslog version, please open a new issue with the exact version, config snippet, and observed output.

PR #3312, opened 2018-11-22
adding abs in rainscript random impl to be in [0-max] range

CodeHarsh:
New implementation of random uses /dev/urandom for random number generation. This number can be negative. This is the reason we are getting [negative numbers](https://github.com/rsyslog/rsyslog/issues/3039) in random.


CodeHarsh:
@rgerhards Yes its a bug fix. 

I have one question though. Do we really need urandom? 

If I understand it correctly, /dev/urandom is not recommended to be used for very high throughput. Do we actually need a cryptographically strong random number or a pseudo-random is good enough?

rgerhards:
> I have one question though. Do we really need urandom?
> 
> If I understand it correctly, /dev/urandom is not recommended to be used for very high throughput. Do we actually need a cryptographically strong random number or a pseudo-random is good enough?

I am not sure - we do not know what the user uses this data for. And some PRNG are really, really weak. I understand the concern, but I think we do not have enough info on the intended use cases to draw a conclusion. Maybe create a weak random function in addition to the current one?

davidelang:
On Thu, 22 Nov 2018, Harshvardhan Shrivastava wrote:

> Date: Thu, 22 Nov 2018 04:17:49 -0800
> @rgerhards Yes its a bug fix.
>
> I have one question though. Do we really need urandom?
>
> If I understand it correctly, /dev/urandom is not recommended to be used for 
> very high throughput. Do we actually need a cryptographically strong random 
> number or a pseudo-random is good enough?

it's /dev/random that you don't want to rely on for high throughput. 
/dev/urandom exists specificlly for cases where you may want high throughput. 
It's far more secure to use /dev/urandom than to implement your own PRNG.

David Lang

PR #3320, opened 2018-11-30
bugfix: rsyslog may hang during hard cancellation on shutdown

rgerhards:
hard cancellation occurs when an action does not terminate within
timeout; possible reasons are e.g. retry loop or hanging in lib call.

This most probably is a regression from recently introduce thread worker join code.


rgerhards:
CI fail unrelated

PR #3328, opened 2018-12-02
report child process exit status according to config parameter

### LINKED ISSUE #3281: Log child process reaping if processes terminated abnormally or with non-zero exit code
### Desired behavior
Informational logging of reaped child process should be supressed if the process exited normally with an exit code of zero.

### Actual behavior
All reaped child process exits are logged irrespective of whether the process died unexpectedly or exited with a non-zero error code.

Create a rotation script that is invoked once a log file reaches some arbitrary size (e.g. 10000 bytes). Create a logging rule to invoke the rotation script. Once the file is rotated, a "Child XXX has terminated, reaped by main-loop" will be shown in the new log file.

### Environment
```
rsyslogd 8.37.0, compiled with:
	PLATFORM:				arm-tait-linux-gnueabi
	PLATFORM (lsb_release -d):		
	FEATURE_REGEXP:				Yes
	GSSAPI Kerberos 5 support:		No
	FEATURE_DEBUG (debug build, slow code):	No
	32bit Atomic operations supported:	Yes
	64bit Atomic operations supported:	No
	memory allocator:			system default
	Runtime Instrumentation (slow code):	No
	uuid support:				No
	systemd support:			No
```
The platform is a custom OpenEmbedded platform.

## Suggested implementation
Modify the mainloop function to supply a pointer to a status variable in the call to waitpid()  and examine its value on return. Only log a message if:
* `WIFEXITED(status)` returns `false`, or
* `WEXITSTATUS(status)` returns a non-zero value.

jsiwrk:
Add new global setting 'reportChildProcessExits' with possible values
'none|errors|all' (default 'errors'), and new global function
'glblReportChildProcessExit' to report the exit status of a child
process according to the setting.

Invoke the report function whenever rsyslog reaps a child, namely in:
- rsyslogd.c (SIGCHLD signal handler)
- omprog
- mmexternal
- srutils.c (execProg function, invoked from stream.c and omshell)

Remove redundant "reaped by main loop" info log in omprog.

Promote debug message in mmexternal indicating that the child has
terminated prematurely to a warning log, like in omprog.

Closes #3281

### DIFF
--- a/runtime/glbl.c
+++ b/runtime/glbl.c
@@ -31,6 +31,7 @@
 #include <sys/socket.h>
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/wait.h>
 #include <fcntl.h>
 #include <unistd.h>
 #include <pthread.h>
@@ -55,6 +56,10 @@
 #include "net.h"
 #include "rsconf.h"
 #include "queue.h"
+
+#define REPORT_CHILD_PROCESS_EXITS_NONE 0
+#define REPORT_CHILD_PROCESS_EXITS_ERRORS 1
+#define REPORT_CHILD_PROCESS_EXITS_ALL 2
 
 /* some defaults */
 #ifndef DFLT_NETSTRM_DRVR
@@ -90,6 +95,7 @@ static int iMaxLine = 8096;		/* maximum length of a syslog message */
 static uchar * oversizeMsgErrorFile = NULL;		/* File where oversize messages are written to */
 static int oversizeMsgInputMode = 0;	/* Mode which oversize messages will be forwarded */
 static int reportOversizeMsg = 1;	/* shall error messages be generated for oversize messages? */
+static int reportChildProcessExits = REPORT_CHILD_PROCESS_EXITS_ERRORS;
 static int iGnuTLSLoglevel = 0;
 static int iDefPFFamily = PF_UNSPEC;     /* protocol family (IPv4, IPv6 or both) */
 static int bDropMalPTRMsgs = 0;/* Drop messages which have malicious PTR records during DNS lookup */
@@ -168,6 +174,7 @@ static struct cnfparamdescr cnfparamdescr[] = {
 	{ "oversizemsg.errorfile", eCmdHdlrGetWord, 0 },
 	{ "oversizemsg.report", eCmdHdlrBinary, 0 },
 	{ "oversizemsg.input.mode", eCmdHdlrGetWord, 0 },
+	{ "reportchildprocessexits", eCmdHdlrGetWord, 0 },
 	{ "action.reportsuspension", eCmdHdlrBinary, 0 },
 	{ "action.reportsuspensioncontinuation", eCmdHdlrBinary, 0 },
 	{ "parser.controlcharacterescapeprefix", eCmdHdlrGetChar, 0 },
@@ -480,6 +487,25 @@ setOversizeMsgInputMode(const uchar *const mode)
 	RETiRet;
 }
 
+static rsRetVal ATTR_NONNULL()
+setReportChildProcessExits(const uchar *const mode)
+{
+	DEFiRet;
+	if(!strcmp((char*)mode, "none")) {
+		reportChildProcessExits = REPORT_CHILD_PROCESS_EXITS_NONE;
+	} else if(!strcmp((char*)mode, "errors")) {
+		reportChildProcessExits = REPORT_CHILD_PROCESS_EXITS_ERRORS;
+	} else if(!strcmp((char*)mode, "all")) {
+		reportChildProcessExits = REPORT_CHILD_PROCESS_EXITS_ALL;
+	} else {
+		LogError(0, RS_RET_CONF_PARAM_INVLD,
+				"invalid value '%s' for global parameter reportChildProcessExits -- ignored",
+				mode);
+		iRet = RS_RET_CONF_PARAM_INVLD;
+	}
+	RETiRet;
+}
+
 static rsRetVal
 setDisableDNS(int val)
 {
@@ -614,6 +640,42 @@ glblReportOversizeMessage(void)
 {
 	return reportOversizeMsg;
 }
+
+
+/* logs a message indicating that a child process has terminated.
+ * If name != NULL, prints it as the program name.
+ */
+void
+glblReportChildProcessExit(const uchar *name, pid_t pid, int status)
+{
+	DBGPRINTF("waitpid for child %d returned status: %2.2x\n", pid, status);
+
+	if(reportChildProcessExits == REPORT_CHILD_PROCESS_EXITS_NONE
+		|| (reportChildProcessExits == REPORT_CHILD_PROCESS_EXITS_ERRORS
+			&& WIFEXITED(status) && WEXITSTATUS(status) == 0)) {
+		return;
+	}
+
+	if(WIFEXITED(status)) {
+		int severity = WEXITSTATUS(status) == 0 ? LOG_INFO : LOG_WARNING;
+		if(name != NULL) {
+			LogMsg(0, NO_ERRCODE, severity, "program '%s' (pid %d) exited with status %d",
+					name, pid, WEXITSTATUS(status));
+		} else {
+			LogMsg(0, NO_ERRCODE, severity, "child process (pid %d) exited with status %d",
+					pid, WEXITSTATUS(status));
+		}
+	} else if(WIFSIGNALED(status)) {
+		if(name != NULL) {
+			LogMsg(0, NO_ERRCODE, LOG_WARNING, "program '%s' (pid %d) terminated by signal %d",
+					name, pid, WTERMSIG(status));
+		} else {
+			LogMsg(0, NO_ERRCODE, LOG_WARNING, "child process (pid %d) terminated by signal %d",
+					pid, WTERMSIG(status));
+		}
+	}
+}
+
 
 /* set our local domain name. Free previous domain, if it was already set.
  */
@@ -866,6 +928,7 @@ static rsRetVal resetConfigVariables(uchar __attribute__((unused)) *pp, void __a
 	free(oversizeMsgErrorFile);
 	oversizeMsgErrorFile = NULL;
 	oversizeMsgInputMode = glblOversizeMsgInputMode_Accept;
+	reportChildProcessExits = REPORT_CHILD_PROCESS_EXITS_ERRORS;
 	free(pszWorkDir);
 	pszWorkDir = NULL;
 	free((void*)operatingStateFile);
@@ -1076,7 +1139,7 @@ glblProcessCnf(struct cnfobj *o)
 		if(!strcmp(paramblk.descr[i].name, "processinternalmessages")) {
 			bProcessInternalMessages = (int) cnfparamvals[i].val.d.n;
 		} else if(!strcmp(paramblk.descr[i].name, "internal.developeronly.options")) {
-		        glblDevOptions = (uint64_t) cnfparamvals[i].val.d.n;
+			glblDevOptions = (uint64_t) cnfparamvals[i].val.d.n;
 		} else if(!strcmp(paramblk.descr[i].name, "stdlog.channelspec")) {
 #ifndef ENABLE_LIBLOGGING_STDLOG
 			LogError(0, RS_RET_ERR, "rsyslog wasn't "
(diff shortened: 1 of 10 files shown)


runtime/glbl.c
```diff
@@ -1337,31 +1404,31 @@ glblDoneLoadCnf(void)
...
+	        loadConf->globals.gidDropPrivKeepSupplemental = (int) cnfparamvals[i].val.d.n;
 		} else if(!strcmp(paramblk.descr[i].name, "net.acladdhostnameonfail")) {
-		        *(net.pACLAddHostnameOnFail) = (int) cnfparamvals[i].val.d.n;
+	        *(net.pACLAddHostnameOnFail) = (int) cnfparamvals[i].val.d.n;
 		} else if(!strcmp(paramblk.descr[i].name, "net.aclresolvehostname")) {
-		        *(net.pACLDontResolve) = !((int) cnfparamvals[i].val.d.n);
+	        *(net.pACLDontResolve) = !((int) cnfparamvals[i].val.d.n);
 		} else if(!strcmp(paramblk.descr[i].name, "net.enabledns")) {
-		        setDisableDNS(!((int) cnfparamvals[i].val.d.n));
+	        setDisableDNS(!((int) cnfparamvals[i].val.d.n));
```
rgerhards: any specific reason you removed the indention? With it, the code is far more readable.

tests/omprog-restart-terminated-outfile.sh
```diff
@@ -71,7 +71,7 @@ injectmsg 6 1
 injectmsg 7 1
 wait_queueempty
 
-kill -s USR1 $(get_child_pid)
+kill -s KILL $(get_child_pid)
```
rgerhards, line 74: a word of caution: we are moving towards supporting parallel builds, as well es parallel execution of multiple CI runs on the same worker. I think the current `get_child_pid` implementation will not work in that environment. It would probably be good to change this to using child-created pidfiles some time in the future.

tests/omprog-restart-terminated-outfile.sh
```diff
@@ -71,7 +71,7 @@ injectmsg 6 1
...
 ./msleep 100
```
rgerhards, line 75: We should try to avoid sleep calls - past months experience tells they work exceptionally bad when CI is pretty busy. We have lot's of failures because of this. Again, this is nothing specific to the PR, just sharing info.

tests/testsuites/omprog-restart-terminated-bin.sh
```diff
@@ -34,7 +34,7 @@ while [[ -n "$log_line" ]]; do
         # Terminate prematurely by closing pipe, without confirming the message
         echo "Terminating without confirming the last message" >> $outfile
         >&2 echo "[stderr] Terminating without confirming the last message"
-        exit 1
+        exit 0
```
rgerhards, line 37: is this more an `exit 77` (SKIP test)?

runtime/glbl.c
jsiwrk: Actually I corrected the indention... After careful inspection, though, these lines had some spaces instead of tabs. GitHub was displaying a different indention than my editor because of the tab size (8 in GitHub instead of 4 in my editor). I have replaced the spaces with tabs and updated the PR.

BTW, do you have a specific convention for the tab size in which the code is supposed to be viewed? Sometimes there can be differences, as in this case. (Note: people advocate for avoiding the use of tabs in open source projects to avoid this kind of issues, although I personally have no problem with them...)

tests/omprog-restart-terminated-outfile.sh
jsiwrk: If you recall, we already took that into consideration time ago in that test (and similar others). Please see some lines above:
```
# We need a test-specific program name, as the test needs to signal the child process
cp -f $srcdir/testsuites/omprog-restart-terminated-bin.sh $RSYSLOG_DYNNAME.omprog-restart-terminated-bin.sh

# On Solaris 10, the output of ps is truncated for long process names; use /usr/ucb/ps instead:
if [[ $(uname) = "SunOS" && $(uname -r) = "5.10" ]]; then
    function get_child_pid {
        echo $(/usr/ucb/ps -awwx | grep "$RSYSLOG_DYNNAME.[o]mprog-restart-terminated-bin.sh" | awk '{ print $1 }')
    }
else
    function get_child_pid {
        echo $(ps -ef | grep "$RSYSLOG_DYNNAME.[o]mprog-restart-terminated-bin.sh" | awk '{ print $2 }')
    }
fi
```
I.e. the program is renamed prior to spawning it by adding the test runtime random prefix. So in case of a parallel test execution, each test would have its own process name. (Note: because of the long process name, I had to use an alternate way to get the child PID in Solaris 10, as shown above.) Am I missing something?

> It would probably be good to change this to using child-created pidfiles some time in the future.

OK, seems a good idea since that would eliminate the need for the use of ps and also the Solaris special case. The pid file itself would also need to have a random name, though. Taking note of this as a future improvement.

tests/omprog-restart-terminated-outfile.sh
jsiwrk: I know sleeps are not desirable... When I refactored the omprog tests in #2629 I eliminated most of them, and as far as I know the tests have been stable since then (please tell me otherwise). However I left that one (which ensures that the child processes the signal before receiving the next message) to keep the test code simpler, and because it seems very unlikely that the test could break because of this... Taking note of this as a possible future improvement, though.

tests/testsuites/omprog-restart-terminated-bin.sh
jsiwrk: No, this code belongs to a test program started by the test, so the special exit codes for tests don't apply here. (Note: I changed the exit code in that line from 1 to 0 to test that the new global flag does not log normal exits.)

jsiwrk:
The LGTM analysis failure seems not my fault... Please tell me otherwise.

tests/omprog-restart-terminated-outfile.sh
rgerhards: just dispensing info: we have machines where, when under stress, operations take up to 6..7 times longer than I had expected (and anticipated in the tests).

rgerhards:
interestingly, the LGTM failure seems to return with a new push - restarted, let's see what happens next

@jsiwrk do you plan to do any more work on the PR? Else it would be good to update the doc so that we can finally merge :-)

jsiwrk:
OK, will update the docs soon.

Note: I resolved the last conflict using the GitHub UI and I've had a hard time squashing and rebasing the merge commit... Will not use that option again.

rgerhards:
@jsiwrk I think these are valid: https://www.codefactor.io/repository/github/rsyslog/rsyslog/pull/3328

rgerhards:
@jsiwrk the _codecov builder was originally not running omprog tests for some reason I had forgotten. I have re-enabled it, and now I see all sorts of test failures related to omprog tests: http://build.rsyslog.com/testbench-showtest.php?testname=omprog%

Maybe you have an idea? I can of course turn omprog testing off again, but it looks like there indeed is something wrong.

jsiwrk:
> I see all sorts of test failures related to omprog tests: http://build.rsyslog.com/testbench-showtest.php?testname=omprog%

It seems the failures are:
- the failure in `omprog-output-capture-mt.sh` since v8.40, described in #3361 (see there for details)
- failures in multiple valgrind tests, like the following one. The stack trace seems to indicate this is due to a recent change in the core. Also, note that the valgrind tests of other plugins are also detecting this. Therefore it doesn't seem related to omprog or the omprog tests. I've not investigated further.
  ```
    main Q:Reg: enter WrkrExecCleanup
    ==2580== Thread 6 rs:main Q:Reg:
    ==2580== Invalid read of size 8
    ==2580==    at 0x44A30E: wtpWorker (wtp.c:381)
    ==2580==    by 0x5053E24: start_thread (in /usr/lib64/libpthread-2.17.so)
    ==2580==    by 0x5FB7BAC: clone (in /usr/lib64/libc-2.17.so)
    ==2580==  Address 0x8274830 is 304 bytes inside a block of size 312 free'd
    ==2580==    at 0x4C2AD1D: free (vg_replace_malloc.c:530)
    ==2580==    by 0x44A00F: wtpDestruct (wtp.c:179)
    ==2580==    by 0x44F158: qqueueDestruct (queue.c:2737)
    ==2580==    by 0x40E018: deinitAll (rsyslogd.c:2009)
    ==2580==    by 0x40E018: main (rsyslogd.c:2119)
    ==2580==  Block was alloc'd at
    ==2580==    at 0x4C2B9B5: calloc (vg_replace_malloc.c:711)
    ==2580==    by 0x44A3F5: wtpConstruct (wtp.c:116)
    ==2580==    by 0x4504A3: qqueueStart (queue.c:2466)
    ==2580==    by 0x40F1A5: startMainQueue (rsyslogd.c:834)
    ==2580==    by 0x42EB84: activateMainQueue (rsconf.c:808)
    ==2580==    by 0x42EB84: activate (rsconf.c:874)
    ==2580==    by 0x41014B: initAll (rsyslogd.c:1624)
    ==2580==    by 0x40DE97: main (rsyslogd.c:2110)
  ```
- some failures in `omprog-feedback-mt.sh` like the following. There is a core dump and an error indicating a possible "double free". I think it is related to the above error detected by valgrind. Probably this memory error affects high-load tests (like this one) more than others. I've not investigated further.
   ```
    rsyslogd: omprog: error sending message to program: Bad file descriptor [v8.1901.0.master try https://www.rsyslog.com/e/2119 ]
    rsyslogd: omprog: error sending message to program: Bad file descriptor [v8.1901.0.master try https://www.rsyslog.com/e/2119 ]
    rsyslogd: omprog: error sending message to program: Bad file descriptor [v8.1901.0.master try https://www.rsyslog.com/e/2119 ]
    rsyslog debug: omprog_action queue:Reg/w5: enter WrkrExecCleanup
    rsyslog debug: 0x563046a921a0: worker exiting
    double free or corruption (fasttop)
    ./diag.sh: line 495: 24197 Aborted                 (core dumped) LD_PRELOAD= ../tools/rsyslogd -C -n -irstb_817200_3149925e:.pid -M../runtime/.libs:../.libs -frstb_817200_3149925e_.conf
    imdiag[39495]: ./diag.sh: line 726: kill: (24197) - No such process
    13:40:46 wait on shutdown of 24197
    unexpected line count in omprog script output: 9744 (expected: 10000)
   ```

jsiwrk:
> @jsiwrk I think these are valid: https://www.codefactor.io/repository/github/rsyslog/rsyslog/pull/3328

Corrected in PR #3373.

rgerhards:
@jsiwrk I think the valgrind issue has been fixed by https://github.com/rsyslog/rsyslog/pull/3359

runtime/glbl.c
rgerhards: we use the default tab size of 8 (and sorry for the delay, I overlooked the question and have come over it "by accident" now...). The rule we use is "indent by tab" and "indent any follow-up (of an incomplete) line by another tab" - this is always consistent no matter of the tab size. It just get's messi when trying to pretty-print.

Tabs or spaces is sometime religious . while I personally prefer tabs (everybody is free to adjust as according to his personal taste), rsyslog always used tabs (even when I inherited it) and this is why we kept it this way. Re-writing them would wreck git history usefulness (would look like a complete rewrite).

jsiwrk:
The failure of the omprog test with the _codecov builder is being analyzed in issue #3361. I think this PR could be merged (when deemed appropriate) independently of that, since it is unrelated.

rgerhards:
@jsiwrk mhhh... looks like codecov worker still fails, even though I have merged the other patch set :-S

jsiwrk:
Yes, this is expected since the issue has not been corrected yet. Please recall my comments in the patch...

rgerhards:
@jsiwrk sry, been working on a couple of CI issues in parallel... Did you have a chance to see if clang coverage shows the same issues? If not, I can probably give it a try.

jsiwrk:
No, I will probably not be available in the short term, sorry... I would bet it's a problem with the gcc instrumentation, but proving this will require time, since we must craft a minimal program that reproduces the problem (and of course it could also be a problem in the omprog code... but it doesn't seem so). Apart from looking at clang, a provisional workaround that could work would be to use Ubuntu 16.04 (instead of 18.04) in the codecov builder, since I couldn't reproduce the problem on that Ubuntu version. I know that's not acceptable as a permanent solution, but in the meantime...

rgerhards:
@jsiwrk I tried `gcc-8` inside the Ubuntu 18.04 container. This did not help. So if it is a gcc bug, it's a long-standing one. Next is to try clang, which I am pretty positive about...

I would not like to switch to Ubuntu 16.04, as it is hard to get full coverage for the older builds. But we can do (and do!) runs on multiple platforms. Anyhow, it kind of feels strange to fallback to old stuff when not 100% sure why this is needed.

rgerhards:
gcc8 check not required here as it was not present when PR came in (and so I can't trigger it). Also, I know 99.99999% sure it will pass.

rgerhards:
@jsiwrk ¡finally! - with clang, coverage reporting seems to work; ok, I had to disable the helgrind tests, but that is not a real problem. So I am able to finally merge. 

Note, thought, that while it looks like a gcc bug, I am not 100% sure - especially as the bug also occurs on the gcc8. I'll keep an eye on the test and we for sure will notice if some real problem pops up again.

Thanks for all your hard work!

jsiwrk:
Yes, I'm not satisfied either... The problem can (and ideally should) be analyzed more. Right now I cannot dedicate more time to this, but will add it to my todo in case I can in the future... Meanwhile it's great you have managed to work around the issue with clang (or so we hope...).

PR #3330, opened 2018-12-05
imdocker: input plugin enabling pulling container logs from docker

n2yen:
plugin will pull all container logs from a host as well as filling out some basic
 container metadata such as id, name, image, labels.
 * requirements: curl-7.40.0+, for unix_domain_socket option (Docker API listens on a local socket)
 Additionally, plugin will handle the issue of log lines larger than 16KB being split by Docker.

NOTE: may not handle large numbers of containers at the moment due to container polling. Plan to fix this in future. Creating pull request early in order to check CI, get feedback, etc.


davidelang:
You may want to look at the mmkubernetes module, it lets imfile or imjournal 
gather the logs and then adds metadata to them.

If you can leverage that codebase to make your metadata queries be a small 
modification of the existing codebase (or ideally configurable :-) it could be 
better than having two different codebases to do almost the same thing

David Lang

Note, I have not looked at your code yet, just commenting based on your 
description.

n2yen:
Thank you David for your feedback! 
Just to give a bit more background, the motivation for this input plugin in environments where:
-  we can't necessarily depend on access to lots of persistent storage (some cloud environment use-cases in particular). Since the docker API streams the container logs, we don't have to depend on persistent storage.
- imjournal starts to drop log lines.
Also to fix the well known 16K logline problem in Docker [here](https://github.com/moby/moby/issues/34855)

plugins/imdocker/imdocker_unit_tests.h
```diff
@@ -0,0 +1,1671 @@
+
```
rgerhards: We need an ASL 2.0 license header here - or some info on why a different license need to be used.

plugins/imdocker/imdocker.c
```diff
@@ -330,7 +330,7 @@ dockerContLogsBufWrite(docker_cont_logs_buf_t *pThis, const uchar *pdata, size_t
 		uchar *pbuf=NULL;
 		if ((pbuf = realloc(mem->data, mem->len + write_size + 1)) == NULL) {
 			LogError(errno, RS_RET_ERR, "%s() - realloc failed!\n", __FUNCTION__);
-			ABORT_FINALIZE(RS_RET_OUT_OF_MEMORY);
+			return RS_RET_OUT_OF_MEMORY;
```
rgerhards: There is something fishy - what did codacy complain about?

n2yen: codacy complained about 'finalize_t' label was not used - I was pretty mystified about this. Strangely, after fixing some memory leak related issues, these types of errors went away, but left the return statement as is. Any ideas why codacy would complain about that?

rgerhards:
@n2yen In order to make this module really work, we need at least some minimal doc. Otherwise nobody knows about it. Could you add a basic page to https://github.com/rsyslog/rsyslog-doc/tree/master/source/configuration/modules ?

rgerhards:
@n2yen I need to figure out why LGTM seems to continously fail. Will look into that.

rgerhards:
doc PR: https://github.com/rsyslog/rsyslog-doc/pull/758

rgerhards:
@n2yen I see you have added unit tests. Do you have a way to actually execute them as part of CI?

rgerhards:
@n2yen note that I will squash during merge - maybe you want to do it yourself for greater control of the resulting commit?

n2yen:
@rgerhards thanks for the feedback. I'll work on squashing it and also look into integrating the unit tests as part of CI.

rgerhards:
@n2yen FIY: we do currently not do dynamic tests on OSX, just a compile tests. Simply missing know-how. You also ran into some well-known CI flakes (some I actually work on right now). I have restarted CI, but the flakes may reappear.

n2yen:
@rgerhards ok, thanks for the info!  Just to update, I'm running a few CI (clang) issues with curl, and so testing various curl versions...

n2yen:
@rgerhards ok, I'm switching to using the curl master branch to see how it goes. It does seem the unsigned-integer-overflow clang warning (2nd CI compile/run) is problematic in the curl codebase... so far, I've settled on disabling imdocker unit tests in that run for now. Any other ideas?

davidelang:
If curl changes behavior across versions, we need to know about it. We need to 
test with at least the oldest and newest versions work.

I keep hearing about problems with curl changing over time, should we be looking 
a wget or some other program to use instead of curl to get a more stable api to 
use?

David Lang

n2yen:
Hi David,
I think we can minimize the impact (on CI testing) of newer version by testing with a bare bones configuration of curl. In regards to wget, my understanding is a command line tool only (imdocker relies heavily on a small subset of the curl api), but that could be out of date. 
cheers,
Nelson

rgerhards:
@n2yen @davidelang there is a misunderstanding. Here `curl` is used a bit imprecise for *`libcurl`*.

Nevertheless, a plugin must run with all versions of libcurl. What's the actual problem? @n2yen Can you point us to an issue you have with libcurl? AFAIK all other modules work with all versions, maybe with a bit of conditional compilation.

n2yen:
@davidelang @rgerhards thanks for clarifying the misunderstanding. 

In regards to the libcurl issue, it seems later versions fail the _unsigned integer overflow address sanitizer_ runtime check. I was able to resolve it by compiling/configuring a minimal version of libcurl... 
This problem was showing up in the 3rd travis job, which has this check enabled. It passes now with the above tweak.

rgerhards:
Can you (temporarily) undo the tweak so that I can see the actual error? Or do you have a link to it?

Note: buildbot-based CI is currently experiencing issues due to a buildbot bug. But for the current discussion, this is just cosmetic.

n2yen:
@rgerhards done.

rgerhards:
thx - let's look what happens. I notice that you build libcurl from the source instead of relying on the packages. This is very problematic, as we cannot distribute the module in package form. Is there a hard reason to do so? If so, we should see how we can resolve that reason.

n2yen:
yes, because imdocker uses a local socket to make docker api requests, it needs libcurl>=7.40.x. Unfortunately the libcurl4 package on Trusty contains libcurl 7.35.x. 

It is possible to run imdocker using host:port, assuming docker is set up to do so (via host:port), but it's questionable if that would be the preferred configuration. Thus, the thinking behind the hard requirement. What do you recommend?

rgerhards:
> yes, because imdocker uses a local socket to make docker api requests, it needs libcurl>=7.40.x. Unfortunately the libcurl4 package on Trusty contains libcurl 7.35.x.
> 
> It is possible to run imdocker using host:port, assuming docker is set up to do so (via host:port), but it's questionable if that would be the preferred configuration. Thus, the thinking behind the hard requirement. What do you recommend?

Thanks, now I understand much better. If it is a hard requirement,

* we should do a configure check if for the required minimum version - otherwise users trying to rebuild will also run into the same issue
* If Trusty does not fulfill the requirement, I think we should skip build and check on that platform -- there are hopefully ample of others who support it. We cannot build packages on that platform in any case. So if someone builds manually, he should ensure that he installs proper libcurl first. The configure check will help with that. Back-side is that we run no tests on Trusty than - but that shouldn't really matter if we run on more recent distros. I think for a module like this the distro version is not as critical.

Having said that, we should shift our attention the the rest of the CI, what probably means it is time now to enable the module on more platforms (something I need to do).

n2yen:
> > yes, because imdocker uses a local socket to make docker api requests, it needs libcurl>=7.40.x. Unfortunately the libcurl4 package on Trusty contains libcurl 7.35.x.
> > It is possible to run imdocker using host:port, assuming docker is set up to do so (via host:port), but it's questionable if that would be the preferred configuration. Thus, the thinking behind the hard requirement. What do you recommend?
> 
> Thanks, now I understand much better. If it is a hard requirement,
> 
> * we should do a configure check if for the required minimum version - otherwise users trying to rebuild will also run into the same issue
> * If Trusty does not fulfill the requirement, I think we should skip build and check on that platform -- there are hopefully ample of others who support it. We cannot build packages on that platform in any case. So if someone builds manually, he should ensure that he installs proper libcurl first. The configure check will help with that. Back-side is that we run no tests on Trusty than - but that shouldn't really matter if we run on more recent distros. I think for a module like this the distro version is not as critical.
> 
> Having said that, we should shift our attention the the rest of the CI, what probably means it is time now to enable the module on more platforms (something I need to do).

thanks for the feedback! i'll add the configure version check, and disable the checks on trusty distros in the meantime.

rgerhards:
For now, I have updated the Ubuntu 18.04 container to try to build imdocker.

n2yen:
> For now, I have updated the Ubuntu 18.04 container to try to build imdocker.

ok, sounds good.

rgerhards:
> thanks for the feedback! i'll add the configure version check, and disable the checks on trusty distros in the meantime.

Just one thing: I did not see a Travis failure. Did I overlook it?

n2yen:
yes, i think you're right there was no Travis failure. maybe the problem I saw was fixed upstream!

contrib/imdocker/Makefile.am
```diff
@@ -0,0 +1,6 @@
+pkglib_LTLIBRARIES = imdocker.la
+
+imdocker_la_SOURCES = imdocker.c
+imdocker_la_CPPFLAGS = -I$(PTHREADS_CFLAGS) $(RSRT_CFLAGS) $(CURL_CFLAGS) $(LIBLOGGING_STDLOG_CFLAGS)
```
rgerhards: The `-I` is wrong here. I guess this is causing the gcc8 build failure. I have no idea why the others work ;-)

rgerhards:
Testing now also enabled, CI running.

contrib/imdocker/Makefile.am
n2yen: Thanks @rgerhards - updated.

n2yen:
Hi @rgerhards My apologies i just saw this! I'll rebase, and fix the conflict.

n2yen:
Hi @rgerhards, is there any recommendation for changes needed by codecov tool?

n2yen:
Hi @rgerhards, i've added some imdocker end-to-end tests. They work locally, but appear to fail the Travis CI tests for ubuntu 18.  The tests require there to be a docker daemon running inside the test environment... can this be added? Can you point me to the ubuntu 18 images used? I wasn't sure if i saw it reference in this repo.

rgerhards:
We need to look deeper into that. Regularly, the tests are executed **within** the container. I guess we could use Travis for this special case, as it provides VMs with Docker daemon. Today is full for me (mostly), let's see if we get it going during the week. Should not be too much of an issue. 

Core idea:

- instruct Travis to spawn an additonal VM
- add some plumbing so that it calls special test
  * build only rqd components
  * execute imdocker tests only

That should have  minimal overhead with good coverage.

n2yen:
Yes, that sounds good.

n2yen:
Hi @rgerhards, is there anything I can do to help with enabling the docker daemon? I'd be happy to help out if needed.

rgerhards:
I am hit by a flu, more severely than I initially thought. Not sure if I can look into this this week. But in essence it's all Travis stuff that needs to be done.

n2yen:
ok, hope you recover well!

rgerhards:
@n2yen  I am mostly back :-) I could update some defs directly in your branch. Shall I do that?

n2yen:
@rgerhards yes that would be great!

rgerhards:
@n2yen sry for the delay - was heavily working on other PRs. Hope to finally be able to do this here this week.

rgerhards:
ok, it unfortunately looks like I did not yet enable imdocker in any of the CI. So we now need to look at compile problems first. Note that I am probably not able to do all changes today, but those errors you see should be valid.

rgerhards:
@n2yen I have disabled the Armbian machines again. We still have the Ubuntu18 on arm for a compile check, so at least a bit of arm coverage.

It would definitely be a plus if we could go with older libcurl. I will also try to see if I can enable build on LGTM, not sure about there versions.

rgerhards:
@n2yen feel free to drop my change - I have a local copy and can re-commit after your compile fixes. Of course, I am also happy with any other method (like you amending the commit in question).

n2yen:
@rgerhards OK, now that I'm revisiting this, I think i had originally added 'IMDOCKER' option to install.sh in order to conditionally enable imdocker, and imdocker-tests. 
Can it be enabled similarly for other builds like centos7?
 
Regarding testing on Trusty, any way we can skip it, we had a previous discussion on it (hidden in the comments above), and i think we came to the conclusion to test on more recent distros:
 
> * we should do a configure check if for the required minimum version - otherwise users trying to rebuild will also run into the same issue
> * If Trusty does not fulfill the requirement, I think we should skip build and check on that platform -- there are hopefully ample of others who support it. We cannot build packages on that platform in any case. So if someone builds manually, he should ensure that he installs proper libcurl first. The configure check will help with that. Back-side is that we run no tests on Trusty than - but that shouldn't really matter if we run on more recent distros. I think for a module like this the distro version is not as critical.
> 
> Having said that, we should shift our attention the the rest of the CI, what probably means it is time now to enable the module on more platforms (something I need to do).

rgerhards:
The problem here is that most platforms are docker containers.

rgerhards:
This pull request **introduces 2 alerts** when merging fcf2bd3fabd8c147f6d02efb7c1352c3fb4d8ffa into fddfec1349437907a09935f7c4302cc9a8fb6f74 - [view on LGTM.com](https://lgtm.com/projects/g/rsyslog/rsyslog/rev/pr-a42053c2d9cec7bf010236850666d5462ed287ef)

**new alerts:**

* 1 for Commented\-out code
* 1 for Potentially overflowing call to snprintf

---

*Comment posted by [LGTM.com](https://lgtm.com)*

rgerhards:
@n2yen I am giving a couple of CI options a try. Please do not update in the interim. I'll let you know when I am done.

rgerhards:
@n2yen ok, it looks like CentOS 7 is now ready for the imdocker tests. From what I can see, we have real errors inside the log. It is this builder: https://build.rsyslog.com/#/builders/175/builds/1519

I'll standby now and wait for your responses.

rgerhards:
side-note: the imdocker tests seem to have been enabled in some Ubuntu 18.04 based tests. They of course fail, so nothing to be concerned about. I'll disable the tests there soon (needs container rebuild).

n2yen:
Thanks @rgerhards ! I see the outputs . Any chance some of the tests are running in parallel?  It looks like there are outputs from other tests that may clash with validation results.

rgerhards:
Yes, tests are run parallel. We have functions for unique names etc. E.g RSYSLOG_DYNNAME. I can have a look at the tests if you prefer.

n2yen:
I’ll look into how to validate more accurately

n2yen:
Oh that's pretty convenient, sure thing, that would be great!

tests/imdocker-multi-line.sh
```diff
@@ -0,0 +1,51 @@
...
+        GetContainerLogOptions="timestamps=0&follow=1&stdout=1&stderr=0")
+action(type="omfile" template="outfmt"  file="'$RSYSLOG_OUT_LOG'")
+'
+
+#export RS_REDIR=-d
+startup
+NUM_ITEMS=1000
+# launch a docker runtime to generate some logs.
+# These logs started after start-up should get from beginning
+sleep 1
```
rgerhards: FYI: `sleep` doesn't really work, as timing is very very different (and often slow) inside CI.

rgerhards:
ok, now I have a question for you: how are the messages ingested? I have looked at the test, but I don't see that they duplicate to a file. Is it so that imdocker tkes messages from **all** running containers and combines them? That could also become problematic if the machine is running different containers.

contrib/imdocker/imdocker.c
```diff
@@ -0,0 +1,1809 @@
...
+static struct cnfparamdescr modpdescr[] = {
+	{ "apiversionstr", eCmdHdlrString, 0 },
+	{ "dockerapiunixsockaddr", eCmdHdlrString, 0 },
+	{ "dockerapiaddr", eCmdHdlrString, 0 },
+	{ "listcontainersoptions", eCmdHdlrString, 0 },
+	{ "getcontainerlogoptions", eCmdHdlrString, 0 },
+	{ "pollinginterval", eCmdHdlrPositiveInt, 0 },
+	{ "retrievenewlogsfromstart", eCmdHdlrBinary, 0 },
+	{ "trimlineoverbytes", eCmdHdlrPositiveInt, 0 },
+	{ "ruleset", eCmdHdlrString, 0 }
```
rgerhards: parameter "ruleset" is not implemented

rgerhards:
I have updated one test. It's not yet completely done. Pls see commit comment for questions.

davidelang:
On Tue, 26 Feb 2019, Rainer Gerhards wrote:

> ok, now I have a question for you: how are the messages ingested? I have 
> looked at the test, but I don't see that they duplicate to a file. Is it so 
> that imdocker tkes messages from **all** running containers and combines them? 
> That could also become problematic if the machine is running different 
> containers.

It's expected that there are many containers running, there is metadata about 
which container the message is from that is the key to making things work.

This metadata then needs to be translated into container name, etc.

David Lang

contrib/imdocker/imdocker.c
n2yen: yes, it ingests all containers. I think each container needs to be identified using template via container name.  i'm testing something along those lines. will upload soon.

rgerhards:
> It's expected that there are many containers running, there is metadata about which container the message is from that is the key to making things work. This metadata then needs to be translated into container name, etc.

I see. So the module must read **everything** from docker and then sort it in buckets? If so, it's a bit hard to test, but at least we can find ways to do it:
1. generate a unique identifier for each test
2. make the plugin only output data with that identifier
3. force `make check` to not run these tests in parallel (it's unfortunate, but we have other places where we need to do that)

Am I on the right path? If so, what would be the best method to do 2.?

contrib/imdocker/imdocker.c
rgerhards: if we support "ruleset" is not a question of how much it ingests. It makes user's life easier if it does. I came across it when I wanted to split imdocker messages from the rest of the message stream inside the test. Nevertheless, it's not a hard requirement to support multiple rulesets. But than that parameter needs to be removed from the param block as it really does not exist.

rgerhards: "if the module does support ruleset parameter" I wanted to say ;-)

n2yen: > > It's expected that there are many containers running, there is metadata about which container the message is from that is the key to making things work. This metadata then needs to be translated into container name, etc.
> 
> I see. So the module must read **everything** from docker and then sort it in buckets? If so, it's a bit hard to test, but at least we can find ways to do it:
> 
> 1. generate a unique identifier for each test
> 2. make the plugin only output data with that identifier
> 3. force `make check` to not run these tests in parallel (it's unfortunate, but we have other places where we need to do that)
> 
> Am I on the right path? If so, what would be the best method to do 2.?

I have a commit ready which illustrates what David mentioned. 
the idea is to run the container using **--name** option, then output the log line along with the name in order to do a proper count.
should i send it up?

n2yen:
@rgerhards I've merge your changes with some updates made based on method 2... will see if its on the right track, but feel free to update it as well.

rgerhards:
What Happens If two Tests Run parallel? Do both Receive the Same Messages or do they "Steal" each other ones?

davidelang:
On Tue, 26 Feb 2019, Rainer Gerhards wrote:

> What Happens If two Tests Run parallel? Do both Receive the Same Messages or do they "Steal" each other ones?

Just like filenames, the name of the container that you are looking for needs to 
be a parameter for the test.

David Lang

n2yen:
If two tests run in parallel, since both tests would connect to the same docker socket, they should just see any container's messages that are active at that time.

davidelang:
On Tue, 26 Feb 2019, n2yen wrote:

> If two tests run in parallel, since both tests would connect to the same 
> docker socket, they should just see any container's messages that are active 
> at that time.

But for purposes of the test, we want to have a different container for each 
test and have that test only report on logs for that container.

now, if you have two things connecting to the same docker socket, do they both 
see all logs?

David Lang

rgerhards:
AS Long AS they dont steal each other ones, WE can handle it.

David Lang <notifications@github.com> schrieb am Di., 26. Feb. 2019, 20:28:

n2yen:
yes, they both will see all the logs. Each test needs to generate an uniquely named container and only validate those relevant log lines.

n2yen:
@rgerhards , i see the imdocker-basic test has passed in rsyslog-centos7-5!

n2yen:
@rgerhards i think all the imdocker tests are passing now - feel free to modify!

rgerhards:
@n2yen thx! I'll have a look, I think I need to add a bit guarding against timing issues and false negatives. I'll do that in baby-step commits on the basic test. So it would be great if you reviewed the commits to see what I do and why. If you like, you can than apply to other tests (or I will).

rgerhards:
@n2yen I have done a set of updates - see pushed commits. Two issues:

* occasionally a message `{"message":"No such container: c1d5b8b107c76f81bc00b9653d5ed24b351efc917cb5f11bb7ff4bf4258dc1eb"}` is sent - is this the indication of a real problem?
* if I do not slow down the container app (last commit) all(most) all messages are lost - this doesn't look right

n2yen:
> @n2yen I have done a set of updates - see pushed commits. Two issues:
> 
> * occasionally a message `{"message":"No such container: c1d5b8b107c76f81bc00b9653d5ed24b351efc917cb5f11bb7ff4bf4258dc1eb"}` is sent - is this the indication of a real problem?
> * if I do not slow down the container app (last commit) all(most) all messages are lost - this doesn't look right
Hi @rgerhards, thanks for the updates. They look good.  In regards to the last commit and the errors you see, they can be explained:
-  imdocker is API driven so is timing sensitive. Specifically, the containers need to be active, while imdocker is running to pick up the messages.
- The discovery of containers is a polling API call, and for the moment imdocker only works with actively running containers. The test is configured to poll every second, so the test container needs to be running for at least a second to pick up the logs.
- the "No such container" error, is again most likely timing related. A container was discovered, but on the attempt to start receiving messages failed. The likely explanation is the container finished whatever it was doing and is no longer active.

rgerhards:
OK, this sounds like we have some inherent unreliability in this interface. Just to sum up:

* very short running-containers may loose all log messages (use case: container app does immediately error-terminate)
* last messages before container terminates are almost for sure lost

If so, I think we need to clearly tell this in the doc - else I see all those questions in that regard.

This behavior also makes it somewhat hard to make to test independent of timing/speed of CI environment. I need to think a bit about it, probably we need to mount a local dir which we can use to sync between test env inside and outside of container.

rgerhards:
One point:

> * the "No such container" error, is again most likely timing related. A container was discovered, but on the attempt to start receiving messages failed.

Is this really a log message **from** the container - or an **imdocker** error message? I would tend to the latter.

n2yen:
yes, having said that, it should be possible to get containers, inactive or not through the api, i just haven't looked into it yet...

n2yen:
> One point:
> 
> > * the "No such container" error, is again most likely timing related. A container was discovered, but on the attempt to start receiving messages failed.
> 
> Is this really a log message **from** the container - or an **imdocker** error message? I would tend to the latter.

because the log is in json format, i think it is coming from docker... i'll double check.

rgerhards:
> because the log is in json format, i think it is coming from docker... i'll double check.

yeah, I think it comes from docker, but I think it comes from docker API and not as a regular log message from the container?

n2yen:
> > because the log is in json format, i think it is coming from docker... i'll double check.
> 
> yeah, I think it comes from docker, but I think it comes from docker API and not as a regular log message from the container?

Ahh, i see, we should probably avoid publishing these?

rgerhards:
> Ahh, i see, we should probably avoid publishing these?

I'd say it depends. To me, it looks like a message interesting to users, maybe even an error. In this case, we should emit either an rsyslog warning or error message (maybe with the ability to turn this off via config).

We should not, I think, emit is as part of the container's log stream, as that does not seem to be the case.

davidelang:
On Wed, 27 Feb 2019, Rainer Gerhards wrote:

>> Ahh, i see, we should probably avoid publishing these?
>
> I'd say it depends. To me, it looks like a message interesting to users, maybe even an error. In this case, we should emit either an rsyslog warning or error message (maybe with the ability to turn this off via config).
>
> We should not, I think, emit is as part of the container's log stream, as that does not seem to be the case.
>
> @davidelang what do you think?

we do want to process this log, but there are race conditions inherent with 
docker/kubernetes between a container starting/stopping, the log getting passed 
to the log engine (rsyslog in this case) and rsyslog processing the message 
and making the calls to enhance the metadata. Until/unless docker can be 
modified to pass the data in the log message, this will always be the case 
(andis the case for ALL docker log solutions, it's really a badly designed log 
mechanism)

because the metadata lookup fails, this won't populate some variables, and so 
filtering on those variables will fail, but we should not throw those messages 
away, let the admin do that if they want (or filter them into their own output)

David Lang

n2yen:
> On Wed, 27 Feb 2019, Rainer Gerhards wrote: > Ahh, i see, we should probably avoid publishing these? I'd say it depends. To me, it looks like a message interesting to users, maybe even an error. In this case, we should emit either an rsyslog warning or error message (maybe with the ability to turn this off via config). We should not, I think, emit is as part of the container's log stream, as that does not seem to be the case. @davidelang what do you think?
> we do want to process this log, but there are race conditions inherent with docker/kubernetes between a container starting/stopping, the log getting passed to the log engine (rsyslog in this case) and rsyslog processing the message and making the calls to enhance the metadata. Until/unless docker can be modified to pass the data in the log message, this will always be the case (andis the case for ALL docker log solutions, it's really a badly designed log mechanism) because the metadata lookup fails, this won't populate some variables, and so filtering on those variables will fail, but we should not throw those messages away, let the admin do that if they want (or filter them into their own output) David Lang

Yes. However, the docker interface actually does support getting container meta-data of stopped or exited containers, but actually getting logs for stopped containers doesn't appear to work in docker right now (i think this use to work). imdocker is actually configurable to run in that mode, but since the the docker log interface no longer appears to allow access to stopped container logs, its not enabled by default.

Given the above, there will be some timing issues. On this topic, I've been thinking it would be nice to have another (configurable) mode, in which we can marry the features of utilizing kubernetes/docker API with the reliability of persistent logs. One obvious use-case is to get multi-line support on a per container basis.

rgerhards:
ok, question now is how to proceed. @n2yen do you intend to maintain this further once the PR is merge? If so, I would see that we can get the current state into the next release and then continue with a follow-on PR. I admit I am not sure I can actually do it, as I am working on a high-priority issue with imfile and I do not like to merge potentially racy tests - CI is flaky enough without more of them ;-)

n2yen:
@rgerhards yes I can do that. Thanks for your help given the conditions!

n2yen:
BTW, would you like me to rebase the changes?  should i remove the basic test now?

rgerhards:
It's late, but I try to get it in. Did the rebase that is now necessary. But please let us be sure to work on the tests - I may need to disable the others as well, because they are all racy.

rgerhards:
@n2yen damn... I seem to have screwed up during the rebase. Can you `push -f` you local branch. I know I shouldn't do these things in a hurry :-( Sorry for that, but as it looks it's now too late to get this into 8.1903.0. Let's make it into 8.1904.0. Thankfully we have short release cycles...

n2yen:
Hi @rgerhards, would still like me to 'push -f' my local branch?
I've also been working on some minor improvements that *should allow for less racy tests.

rgerhards:
@n2yen thx! Could you do a rebase on the branch? I can also do it and see where I screwed up ;-)

n2yen:
sure.

rgerhards:
mhhh... maybe the cause was different. I am pretty sure I fixed the travis issue. I'll have a new look.

rgerhards:
I now realize what happened: we had **disabled** imdocker in travis. This got somehow lost. Nevertheless, now let's do it ;-). I'll force-push the PR a bit until it finally works.

rgerhards:
@n2yen ok... at least the Travis imdock env seems to work now. But it brings up, I think, a real failure: https://travis-ci.org/rsyslog/rsyslog/jobs/502076130

n2yen:
thanks @rgerhards! yes, the failure seems odd. The message appears to have been received 2x for some reason. I'm not sure I've seen that behavior before - I'll have to investigate it.

rgerhards:
I had a quick look at the test. I see it is not hardened against messages from parallel runs, at least in the config. Have a look in how I did this inside the basic test. But it was a really quick look, I may be totally off ;-)

n2yen:
@rgerhards yup, you're right! updated.

rgerhards:
@n2yen do not wonder: I restart the travis a couple of times. Just want to be sure there are no flakes.

rgerhards:
@n2yen unfortunately it looks like we still have a race. Travis fail occurred after 5 tries. The first four were the VM in question along, the 5th was a full rebuild over all VMs. Not sure if full vs. VM rebuild changes the timing (AFAIK it does not).

n2yen:
hi @rgerhards, made some updates which *should fix the races - all sleeps have now been removed. Also re-enabled imdocker-basic tests

rgerhards:
@n2yen do you think you have some time to check on the race?

rgerhards:
@n2yen sry -outdatd conversation on my screen. I'll let you know if we need to do some more...

rgerhards:
@n2yen I had to rebase to make this mergable again - sry for thrashing your local copy

Note that I also corrected an issue where `--enable-docker` was given for distcheck instead of `--enable-imdocker`.

rgerhards:
I screwed lgmt.yml. Let me find out how to do multi-line (separate PR); when found, I'll fix my mess.

n2yen:
thanks for the heads up @rgerhards, changes look good to me. It looks like lgtm just doesn't like the tab character on multi-lines?

rgerhards:
> thanks for the heads up @rgerhards, changes look good to me. It looks like lgtm just doesn't like the tab character on multi-lines?

I have fixed this via a separate commit, see https://github.com/rsyslog/rsyslog/pull/3581

I'll now rebase and respin CI in the hope to be able to merge soon.

rgerhards:
@n2yen thx for your persistence - finally merged!

n2yen:
@rgerhards thank you!

PR #3345, opened 2018-12-12
mmkubernetes: add support for sslpartialchain for openssl

richm:
This adds support for a new bool parameter `sslpartialchain`

If `"on"`, this will set the OpenSSL certificate store flag
`X509_V_FLAG_PARTIAL_CHAIN`.   This will allow you to verify the Kubernetes API
server cert with only an intermediate CA cert in your local trust store, rather
than having to have the entire intermediate CA + root CA chain in your local
trust store.  See also `man s_client` - the `-partial_chain` flag.
This option is only available if rsyslog was built with support for OpenSSL and
only if the `X509_V_FLAG_PARTIAL_CHAIN` flag is available.  If you attempt to
set this parameter on other platforms, you will get an `INFO` level log
message.  This was done so that you could use the same configuration on
different platforms.


rgerhards:
thx for the great PR, much appreciated!

Issue #1421, opened 2017-02-15
Dynamic key value for omkafka


aalkilani:
Allow dynamic keys using templates, similar to dynaTopic, make available a dynaKey configuration that allows the key to be created via a template. At the very least, have the key take in a variable. As it currently stands the key is simply static for all messages produced.

While similar to dynatopic, I don't see this will need a cache and the limitations that brings in how the dynatopic implementation is done. The topic requiring a cache makes sense, but the key should be handled similar to how the message is.. i.e., not a static string, rather simply part of the datum emitted.

Implementing this is also one step towards resolving #1199

While there are many use cases for this I'm sure, the simplest use case would be appending information RSyslog gets to a Key instead of manipulating the message received as it's being forwarded to Kafka. For example, I could keep the message intact while adding programname, timestamp and hostname to the key for auditing and measuring performance as messages travel through systems.

davidelang:
exposing my ignorance of the kafka protocol, is there a place where it would 
make sense to offer the option of having a template that could be defined to let 
you vary all sorts of things?

what would the rules be about what can and can't be varied between different 
messages sent over the same connection (and even more specifically, sent as a 
single batch)?

David Lang

aalkilani:
For data producers, outside of configuration parameters, Kafka exposes 2 essential things
 - The topic to produce to
 - The datum (ProducerRecord) to produce which is further split into Key and Message

So both the key and message are simply parts of the data being produced and are expected to be something the user chooses on a message by message basis.

Typically, the topic is a little less dynamic and the current implementation works beautifully, in my opinion, to address any dynamic topic routing required which I will add has been great.

mostolog:
Excuse my ignorance, but wouldn't this solve your issue?

```
template(name="topic" type="string" string="whateveryouwant")
...
action(
		name="kafka"
		action.resumeRetryCount="-1"
		action.reportsuspension="on"
		#action.reportSuspensionContinuation="on"
		type="omkafka"
		broker=["kafka:9092"]
		dynatopic="on"
		topic="topic"
		partitions.auto="on"
		dynatopic.cachesize="300"
		template="json"
		errorFile="/logs/rsyslog-kafka.json"
	)
```

rgerhards:
@alorbach isn't that done?

alorbach:
This has not been done yet. 
dynaTopic works different then a dynamic key would work. 
We could use template format to set a dynamic key like we are doing it for the timestamp.

[...]

PR #3350, opened 2018-12-15
omkafka: add support for dynamic keys

ludobrands:
Hi,

we needed a way to force the kafka partition based on the data being pushed. Pretty much the subject of #1421. Here is a patch that allows to use a template to provide the partitioning key.

A new configuration property "dynaKey" is added that, when "on", changes the value of property "key" to a template names instead of a constant value. This is similar in approach to the DynaTopic implementation.  

Our system uses mmjsonparse to extract from the message the property we use as a key. The config we are using  (the relevant part) :
```
module(load="omkafka")
module(load="mmjsonparse")

template(name="json_log" type="list") {
  constant(value="{")
    constant(value="\"@timestamp\":\"")  property(name="timereported" dateFormat="rfc3339")
    constant(value="\",\"tags\":[\"") property(name=".tag") constant(value="\"]")
    constant(value=",\"ctnewlog\": true,")
    property(name="$!all-json" position.from="2")
}

template(name="message_guid_template" type="list") {
  property(name="$!message_guid")
}
...
                action(type="mmjsonparse" cookie="")
                action(type="omkafka"
                        partitions.auto="on"
                        dynaKey="on"
                        key="message_guid_template"
                        template="json_log"
...                
                )
``` 

PR for documentation will follow.

### DIFF
--- a/plugins/omkafka/omkafka.c
+++ b/plugins/omkafka/omkafka.c
@@ -196,6 +196,7 @@ typedef struct s_dynaTopicCacheEntry dynaTopicCacheEntry;
 
 /* Struct for Failed Messages Listitems */
 struct s_failedmsg_entry {
+	uchar* key;
 	uchar* payload;
 	uchar* topicname;
 	SLIST_ENTRY(s_failedmsg_entry) entries;	/*	List. */
@@ -204,6 +205,7 @@ typedef struct s_failedmsg_entry failedmsg_entry;
 
 typedef struct _instanceData {
 	uchar *topic;
+	sbool dynaKey;
 	sbool dynaTopic;
 	dynaTopicCacheEntry **dynCache;
 	pthread_mutex_t mutDynCache;
@@ -257,6 +259,7 @@ static struct cnfparamdescr actpdescr[] = {
 	{ "topic", eCmdHdlrString, CNFPARAM_REQUIRED },
 	{ "dynatopic", eCmdHdlrBinary, 0 },
 	{ "dynatopic.cachesize", eCmdHdlrInt, 0 },
+	{ "dynakey", eCmdHdlrBinary, 0 },
 	{ "partitions.auto", eCmdHdlrBinary, 0 },	/* use librdkafka's automatic partitioning function */
 	{ "partitions.number", eCmdHdlrPositiveInt, 0 },
 	{ "partitions.usefixed", eCmdHdlrNonNegInt, 0 }, /* expert parameter, "nails" partition */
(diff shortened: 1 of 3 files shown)


rgerhards:
Thx for the PR. You have probably seen that CI has raised some issues. It would also be great (and speed up merging) if you could add a test for the new functionality. If not, I'll see that I craft one when I have time and amend it to the PR.

rgerhards:
side-note: as I know we have some flakes inside the CI, I have checked that there exist actual problems with the PR. Just so that you know.

ludobrands:
Down to 1 codestyle issue (tab size is 4 in my IDE) and an error in travis CI that doesn't seem to be linked to my changes. 

No problem to create a test case for this. Had a look a tests/omkafka.sh. As a simple test I imagine the following:
1) create kafka topic with 2 or more partitions
2) set up a rsyslog config with a property value for the key template that is set from an env var
3) set env var to a random value and send M messages to rsyslog. 
4) repeat step 3) N times with different values for the env var 
5) for each partition P in the topic : kafkacat -p P -f '%k' ... | uniq | wc -l. Sum of line counts has to be N. If not, messages with same key where sent to different partitions.  

What do you think? 

Has the docker image everything to run the test?

ludobrands:
Added tests/omkafkadynakey.sh. It uses the 7th digit of the kafka msg as key. So values 0 to 9. The check does a line count on `kafkacat ...-f '%p %k\n' | sort | uniq` which counts the number of partition-key combinations. The count has to equal 10. Sample stdout output of kafkacat:
```
root@a46b5623bd7f:/home/devel/rsyslog/tests# kafkacat -b localhost:29092 -e -C -o beginning -t HNyjt9vf -f '%p %k\n' | sort | uniq
0 4
0 5
0 6
0 7
1 0
1 1
1 2
1 3
1 8
1 9
```
All checks are OK now except for a new one in Codacy/PR introduced by omkafkadynakey.sh. Since those lines come from omkafka.sh and no expansion is expected, I think the failure has to be ignored. 

No clue what yesterday's travis-ci failure was about.

Documentation changes are in https://github.com/rsyslog/rsyslog-doc/pull/759

rgerhards:
@ludobrands The new test is missing in ./tests/Makefile.am, so it is not executed. Can you have a look?

ludobrands:
Yes, that was on purpose. Wanted to share the test methodoly and get some feedback on it before executing it on a system that isn't mine.

rgerhards:
@ludobrands just FYI: CI is currently pretty busy - we may have flakes. I keep an eye on this and let you know if there is a real issue.

rgerhards:
@ludobrands could you please squash the commits and best rebase to master? I did the merge conflict handling, but a squash would change the commit hash and so this would look unmerged. It's cosmetic, I know, but often confusing to folks coming in via google. Sry for not doing the merge somewhat earlier :-(

rgerhards:
@ludobrands just FYI: if yo do not have time to rebase/squash, I'll do it next week,

rgerhards:
@ludobrands 

> if yo do not have time to rebase/squash, I'll do it next week,

done - let's see how CI works out.

rgerhards:
Solaris is false positive, also openCSW workers are now down - irgnoring

Issue #3348, opened 2018-12-14
fix old x.509 tls certs


alorbach:
All certs in testsuites/x.509/ are expired and need to be updated. 
Also check if it is allowed / desired that expired certificates work in testbench?!

PR #3352, opened 2018-12-17
testbench: Updated expired x.509 certs

alorbach:
Also moved old expired certs for future tests (Checking for expired certs and their handling).
they will expire in 100 years (2118) which is properly long enough for rsyslog testbench. 

Closes https://github.com/rsyslog/rsyslog/issues/3348


rgerhards:
@alorbach just to be on the safe side: these are now generated with gnutls tools (so that we still have both gnutls and openssl generated certs)?

alorbach:
@rgerhards yes I used rsyslog articles to generate the certs with the same parameters as before: 
https://www.rsyslog.com/doc/master/tutorials/tls_cert_ca.html
https://www.rsyslog.com/doc/master/tutorials/tls_cert_machine.html

PR #3357, opened 2018-12-19
platform bugfix: HUP did not work reliably on some platforms 

rgerhards:
... at least not on FreeBSD. The problem was that HUP did not awake
the main loop. We now interrupt it explicitly.


rgerhards:
It looks like on FreeBSD we do not get awoken from our `select()` that we use for sleep when a HUP is received. Maybe we are doing something wrong here, maybe we have different semantics.

PR #3367, opened 2018-12-21
testbench: fix issues with parallel test run

rgerhards:
some serialization rules did not work correctly

Note: parallel testing is still experimental and not fully
supported.


rgerhards:
CI fail is unrelated known issue.

PR #3370, opened 2018-12-22
core/stream: always invalidate fd on file close

rgerhards:
while in theory, the fd should immediately be rewritten, in practice
we sometimes see some errors "bad file descriptor" that we cannot
explain. So we clean this up to remove a potential trouble cause.


rgerhards:
CI fail unrelated

PR #3374, opened 2018-12-24
DO NOT MERGE: test to find CI hang

rgerhards:


rgerhards:
we already found the culprit by close testrun log inspection

Issue #3346, opened 2018-12-12
omusrmsg module fails to log to user's tty


Whissi:
Since commit 56ace5e418d149af27586c7c1264fccfbc6badf1, *omusrmsg* is broken:

`strncat()` was replaced with a `memcpy()` function when placing the tty name from the user's utmp (ut_line) entry into a buffer, causing the ttyname to now be missing the *_PATH_DEV* (/dev/) prefix.

memcpy() is not a suitable substitute for strncat() in this context, it is actually replacing the previous added content.

Downstream bug: https://bugs.gentoo.org/673004

PR #3376, opened 2018-12-25
omusrmsg: don't overwrite previous set _PATH_DEV value

Whissi:
Since commit 56ace5e418d149af27586c7c1264fccfbc6badf1, omusrmsg was broken
because "memcpy()" is not a suitable substitute for "strncat()" in this
context, it is actually replacing the previous added content.

Bug: https://bugs.gentoo.org/673004
Closes: https://github.com/rsyslog/rsyslog/issues/3346


rgerhards:
gcc8 check was created after PR, it's not necessary to pass for this PR.

PR #3382, opened 2018-12-28
mmgrok fix: build error w/ gcc-8 due to questionable code construct

rgerhards:
replaced questionable sprintf() by simpler method


rgerhards:
Ci fail unrelated

Issue #2924, opened 2018-08-13
action processing: add ability to SUSPEND an action intentionally


rgerhards:
It has been reported that some TCP receivers exists that accept syslog tcp messages at any rate, even if they do not manage to actually process them. Instead, they silently drop the message. This behavior is not configurable. All in all, it can lead to considerate message loss.

To support such use cases, we need to provide an ability to externally trigger actions suspension and resumption. One proposal is to do this via a configured file which contains the status of the action. Rsyslog shall periodically (configurable) read the file and if it contains SUSPEND suspend the action (and likewise for resume).

selivan:
> Rsyslog shall periodically (configurable) read the file

Just an assumption. May be better to use inotify watch on that file? It's faster then periodic reading file.

rgerhards:
> Rsyslog shall periodically (configurable) read the file
> 
> Just an assumption. May be better to use inotify watch on that file? It's faster then periodic reading file.

good idea, but inotify is

* linux-only, so we cannot use it in general
* racy, so not guaranteed to work (see man page)

PR #3393, opened 2019-01-04
core/action: implement capability to resume/suspend via external file

rgerhards:
It has been reported that some TCP receivers exists that accept syslog tcp
messages at any rate, even if they do not manage to actually process them.
Instead, they silently drop the message. This behavior is not configurable.
All in all, it can lead to considerate message loss.

To support such use cases, we need to provide an ability to externally
trigger actions suspension and resumption.

We do this via a configured file which contains the status of the action.
Rsyslog periodically reads the file and if it contains "SUSPEND", it
suspend the action (and likewise for resume).

closes https://github.com/rsyslog/rsyslog/issues/2924

### DIFF
--- a/action.c
+++ b/action.c
@@ -33,6 +33,7 @@
  *
  * After dequeue, processing is as follows:
  * - processBatchMain
+ * - processMsgMain (direct entry for DIRECT queue!)
  * - ...
  *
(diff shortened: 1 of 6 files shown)


rgerhards:
This is experimental and intentionally non-documented so far. Once it moves into stable/more mainstream, doc will be added,

PR #3400, opened 2019-01-07
omkafka: Add ability to dump librdkafka statistics to a file

pcullen65:
Use statsFile to specify statistics output file; also requires setting statistics.interval.ms confparam to a non-zero value.

### DIFF
--- a/plugins/omkafka/omkafka.c
+++ b/plugins/omkafka/omkafka.c
@@ -174,6 +174,9 @@ typedef struct _instanceData {
 
 	int fdErrFile;		/* error file fd or -1 if not open */
 	pthread_mutex_t mutErrFile;
+	uchar *statsFile;
+	int fdStatsFile;        /* stats file fd or -1 if not open */
+	pthread_mutex_t mutStatsFile;
 	int bIsOpen;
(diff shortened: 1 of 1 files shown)


pcullen65:
I don't believe the two buildbot failure are related to the changes in this PR. As my first PR to rsyslog, please let me know if there's anything else I need to do. Thanks.

rgerhards:
Thx for the PR! The CI failure is indeed unrelated, I'll handle it. Anything else, I let you know after review.

rgerhards:
@pcullen65 It would be great if you could add the relevant doc at https://github.com/rsyslog/rsyslog-doc

rgerhards:
> also requires setting statistics.interval.ms confparam to a non-zero value

I wonder if this is something that the code should check, At a minumum, it should be documented in a way that is expected to not be overlooked ;-)

rgerhards:
I would be great if you could craft a new test based on some of the kafka tests. Understood, though, if too much effort ;-)

pcullen65:
See PR #769 in rsyslog-docs project for corresponding documentation update.

PR #3424, opened 2019-01-16
tcpflood bugfix: tool did not terminate on certificate error

rgerhards:
when tcpflood detected a certificate error, it reported an
error message but did not abort. This could make errors undetectable
during CI runs.


rgerhards:
Travis fail is due to unavailable PPA - acceptable.

PR #3435, opened 2019-01-21
Added missing free of received journal cursor

### LINKED ISSUE #3433: parameter "tmp_cursor" need free after used.
https://github.com/rsyslog/rsyslog/blob/792f1e50e0bc2eac17cf7ee0b9a1a71c4a6b5707/plugins/imjournal/imjournal.c#L624

jvymazal:
closes #3433

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -449,6 +449,7 @@ readjournal(void)
 finalize_it:
 	free(sys_iden_help);
 	free(message);
+	free(c);
 	RETiRet;
 }
 
@@ -470,6 +471,7 @@ persistJournalState(void)
 		}
 	} else {
 		int ret;
+		free(last_cursor);
 		if ((ret = sd_journal_get_cursor(j, &last_cursor))) {
 			LogError(-ret, RS_RET_ERR, "imjournal: sd_journal_get_cursor() failed");
 			ABORT_FINALIZE(RS_RET_ERR);
@@ -630,6 +632,7 @@ loadJournalState(void)
 						iRet = RS_RET_ERR;
 					}
 				}
+				free(tmp_cursor);
 			}
 		} else {
 			LogError(0, RS_RET_IO_ERROR, "imjournal: "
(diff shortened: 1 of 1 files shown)


rgerhards:
@jvymazal FYI: I am adding valgrind-tests for imjournal in #3436 - as usual, it takes longer than expected as there was a lot of wrangling with journal config. Once I am done, I'll re-init the test here. This should cover us at least somewhat in the future (looks like this issue would have been detected).

rgerhards:
@jvymazal new test now active, and it looks like we still have a mem leak. Could you have a look?

Note: I also wrangle with the test environment, as I cannot isolate the journal for pure rsyslog use. Working on that. Right now, this results in false positives. But the current state should indeed show a problem.

jvymazal:
Ok, I found one more possible leak source, will see what CI says now, this is actually great example to test it on, and many thanks for keeping up the effort of journal test automation @rgerhards

rgerhards:
thx for your continuous help!

do you know if we can tell imjournal to read "starting n hours in the past"? That would be good for testbench, but probably also better than the current default of "read everywhere". Here in the testenv the test tried to process 3GB before the timeout hit ;-)

jvymazal:
Well, that is an interesting proposal, it is not trivial and nothing of the sort is currently done in imjournal code, however trough some combination of call I might be able to achieve that. 
Do you think more fine-grained control than current read-all (default) or read only new - `IgnorePreviousMessages`, would be desired?

rgerhards:
Travis is unrelated; working on that.

rgerhards:
> Well, that is an interesting proposal, it is not trivial and nothing of the sort is currently done in imjournal code, however trough some combination of call I might be able to achieve that.
> Do you think more fine-grained control than current read-all (default) or read only new - `IgnorePreviousMessages`, would be desired?

Well, it was just an idea, based on the fact that running the testbench on the real journal is painful. On the other hand, each mode needs to be tested, so a new mode would actually increase test requirements. I still wonder if it would make sense from a users PoV to by default just get the past couple of days instead of pulling in everything that's available. But... if you think it's worth discussing, let's create our own tracker for it. Sry for polluting this one.

PR #3436, opened 2019-01-21
testbench: fix imjournal tests; add imjournal valgrind test

rgerhards:
see also https://github.com/rsyslog/rsyslog/issues/3433


rgerhards:
CI fail in Centos7-5 is intentional - it was the core idea to detect this know bug, so it is a success.

Issue #2706, opened 2018-05-14
remove libloggin-stdlog dependency


rgerhards:
Liblogging-stdlog was introduced to provide a broader ability to send rsyslog internal logs to different sources. However, most distros did not pick up that capability and so instead we do a regular syslog() call. We assume that the actual functionality is never used in practice, so we plan to retire it. That makes building rsyslog from source easier.

#2705 disabled the use of liblogging-stdlog by default during configure. So users (and distros!) could still opt-in to have it enabled if they desire.

With this issue, a couple of releases later, we want to completely remove the functionality. 

If you have a strong reason why you still need it, please post here. If by the assigned milestone we will not see complaints with good reasoning here, we will remove the code.

rgerhards:
This was just done in May 2018  8.36. So it is a bit early to remove it right now. I guess I picked the wrong label when I opened the issues.

rgerhards:
We must note that the testbench also uses this functionality, so there is need for some actual code changes.

rgerhards:
usage inside the testbench is ok, thus closing

PR #3438, opened 2019-01-22
WIP: build system: fix liblogging-stlog check that blocks compilation when…

### LINKED ISSUE #3431: Debian liblogging-stdlog compile error
### Expected behavior
compile rsyslog with --enable-liblogging-stdlog without any error on Debian
### Actual behavior
`./configure --enable-liblogging-stdlog`
End of the output:
```
checking for LIBLOGGING_STDLOG... yes
configure: error: --enable-liblogging-stdlog set but liblogging-stdlog was not found
```
installed packages:
```
dpkg -l |grep liblogging-stdlog
ii  liblogging-stdlog-dev:amd64       1.0.5-2+b2                        amd64        easy to use and lightweight logging library - development files
ii  liblogging-stdlog0:amd64          1.0.5-2+b2                        amd64        easy to use and lightweight logging library
```
stdlog.h:
```
ls -la /usr/include/liblogging/stdlog.h
-rw-r--r-- 1 root root 4165 May  6  2015 /usr/include/liblogging/stdlog.h
```
`./configure --enable-liblogging-stdlog`
### Environment
- rsyslog version: 8.40
- platform: Debian 8, Debian 9
- compile options:
[options.txt](https://github.com/rsyslog/rsyslog/files/2775396/options.txt)

We are using the same Makefile for a while, rsyslog 8.38 and 8.39 can be compiled without any issues.

b0ns:
… lib present

The added check of HAVE_LIBLOGGING_STDLOG, set via AC_DEFINE, fails due to it staying undefined in the configure execution context.

Fixes #3431 - auxiliary var added, to split logic from testbench check and correctly pass the test if library is present. AC_DEFINE untouched due to need for testbench ifdefs.

Bug: https://github.com/rsyslog/rsyslog/issues/3431

Extra: 
Added few details to README.MD
Added Debian section for better coverage.

Note:
As from autoconf documentation: https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Defining-Symbols.html

> By default, AC_OUTPUT places the symbols defined by these macros into the output variable DEFS, which contains an option -Dsymbol=value for each symbol defined. Unlike in Autoconf version 1, there is no variable DEFS defined while configure is running.

In fact by debugging the configure execution the HAVE_LIBLOGGING_STDLOG was resulting undefined in the check at line 1454 of https://github.com/rsyslog/rsyslog/commit/d5ffe6e10eb7f9427c06a72b968b428ca8d6283d#diff-67e997bcfdac55191033d57a16d1408a

This lead to the AC_MSG_ERROR ( added also in the same phase ) that exited the autoconf process blocking compilation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,8 +46,9 @@ Follow the instructions at: https://www.rsyslog.com/doc/build_from_repo.html
 
 In general, you need
 
+* pkg-config
 * libestr
-* liblogging (stdlog component)
+* liblogging (stdlog component, for testbench)
 
 It is best to build these from source.
 
@@ -91,6 +92,24 @@ For KSI, from the Adiscon PPA:
 ```
 sudo apt-get install libksi0 libksi-devel
 ```
+
+#### Debian
+
+```
+sudo apt install build-essential pkg-config libestr-dev libfastjson-dev zlib1g-dev uuid-dev libgcrypt20-dev libcurl4-gnutls-dev zlib1g-dev liblogging-stdlog-dev liblogging-stdlog-dev flex bison
+```
+
+*Note:* For certain libraries version requirements might be higher,
+in that case adding debian backports repositories might help.
+For example installing with apt libfastjson-dev -t stretch-backports.
+
+
+Aditional packages for other modules:
+```
+libdbi-dev libmysqlclient-dev postgresql-client libpq-dev libnet-dev librdkafka-dev libgrok-dev libgrok1 libgrok-dev libpcre3-dev libtokyocabinet-dev libglib2.0-dev libmongo-client-dev
+```
+
+
 
 #### openSUSE 13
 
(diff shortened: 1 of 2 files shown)


rgerhards:
thx for the PR ... just let me check what happend with CodeCov. Obviously it is unrelated...

rgerhards:
> * the bug is there since Nov 03 2018, if nobody noticed till now i guess is because is not really used and the option "--enable-liblogging-stdlog" not enabled beside stdlog being used in the testbench?
> * so what's the real impact of stdlog ? better to use this "enhanced version of syslog() call" ? and is it recommended for standard users?
> * why the PKGcheck for the lib ( with the AC_DEFINE) has to be outside of the classic test on the enable?
> * the comments explain that the check is needed for the testbench and has to be done _anycase_ but why not use standard check and use the enable options for the tests?

see https://github.com/rsyslog/rsyslog/issues/2706 - liblogging-stdlog, while far superior in my PoV, never took off with the distros. So we removed it by default. But the testbench tooling still needs it and we are too busy to change the test programs to do that with their own code. The PLGcheck would possibly be better done in conditional logic, based on testbench and actual module enablement, but so far it didn't really matter - and is another thing that just takes time to do...

rgerhards:
CodeCov is definitely unrelated, it drops due to a new module we needed to enable to handle another PR.

@b0ns if you don't object, I would "just merge" now. Pls let me know.

Issue #3448, opened 2019-01-28
Support RELP+TLS on tcpflood


StrongestNumber9:
```sh
 * -T   transport to use. Currently supported: "udp", "tcp" (default), "tls" (tcp+tls), relp-plain
```
Add support for RELP + TLS. This would be tremendously helpful for testing components that use authenticated imrelp+tls.

rgerhards:
Thx, this is a very good suggestion! I'll see if we can add it to the next release.

PR #3453, opened 2019-01-31
tcpflood: add new transport option relp-tls

PascalWithopf:
Tcpflood can now send messages via relp with tls support.
A test was added aswell.

closes https://github.com/rsyslog/rsyslog/issues/3448


rgerhards:
The failures probably occur due to too-old librelp on some CI workers ... well, that triggers the point that we may need to add conditional compilation w/ autoconf check.

PR #3461, opened 2019-02-06
clickhouse default template fix

gagantrivedi:
STDSQL option added to the default template used in output module of clickhouse


rgerhards:
thx for the fix and sry for the delay - I am hit by a flu. I'll try to craft a new test before I merge, so that I have code that I know that does not work :-)

rgerhards:
thx again for the PR!

PR #3471, opened 2019-02-14
AIX_port: second step.

pduveau:
Hi @rgerhards,
This PR is a huge one. Details are reported in the ChangeLog. Must of usable modules on AIX can now compile and run on this OS. 4 modules are also added to contribution.


rgerhards:
@pduveau thanks for the big contribution. I'll let you wade through the initial fixing round ;-)

I guess the answer is "no", but would it be possible to have a bit more granular commits, e.g. one for each new plugin and so? If it is a lot of work, we can leave as is, but in general it's much nicer to have dedicated commits - which also speeds up review process ;-)

As a side-note, I would still be interested in getting a CI environment for AIX. So far, everything we tried were dead ends (e.g. rented boxes do not support XLC, just GCC, IBM does not want to sponsor the work, so we can only occasionally use their services, .... - all in all very frustrating).

I guess we also need a doc PR for all those new modules :-)

rgerhards:
Note: I intentionally selected the April vs. March milestone as I think I need to have a deep look at the various changes.

pduveau:
Hi @rgerhards,
I have a milestone at end of march. I resigned from my current job and leave it the 29th of march.
I understand the granularity. I can separate my contribution for one PR per new plugins (improg, imtuxedoulog, imbatchreport, pmdb2diag and omrabbitmq) and a global PR for AIX refactoring.
Is this can be OK for you ? So I should start with the refactor for AIX.

rgerhards:
@pduveau 

> I have a milestone at end of march. I resigned from my current job and leave it the 29th of march.

Ah, I see, good to know. So let's try to integrate everything by that deadline.

This also triggers a question. It's not a show stopper, but something the project needs to know: does that also mean that the contribute modules will immediately be orphaned? I mean that nobody maintains them? Again, not a show stopper, but for sure something we should mention in the doc.

> I understand the granularity. I can separate my contribution for one PR per new plugins (improg, imtuxedoulog, imbatchreport, pmdb2diag and omrabbitmq) and a global PR for AIX refactoring.
> Is this can be OK for you ? So I should start with the refactor for AIX.

Different PRs would definitely be useful and help in getting the review done quicker (reviewing ever and ever again such a big monolithic bunch of changes really is no fun and requires extended timeslots to do so).

rgerhards:
direct question
```
core feature: add the ability to forward message through ouput module without
  altering it. omrabbitmq is aware of this new flag.
```
What does that mean? Why can't it be done with a template like this
```
template(name="unaltered" type="string" string="%rawmsg%")
```

pduveau:
Hi @rgerhards,
As the checks are closed to be successful, now I reduce this PR to the AIX port. You will see that even a lot of files has been modified. The modification are mainly to avoid warning on GCC pragmas when using XLC. The test is more on XLC than OS so I changed them to. This allows to use GCC on AIX and XLC on Linux.
Then Next commit on this Branch will be smaller.
B. R.
@pduveau

pduveau:
Hi @rgerhards,

`core  feature: add the ability to forward message through ouput module without
  altering it. omrabbitmq is aware of this new flag. `

With the solution you propose, all the messages are not altered. With the flags you can mix the two behavior with a single output module. For example, we use rsyslog to parse application log files with imfile and other new module (imbatchreport and imtuxedoulog) and use new improg to received system statistics already formatted in json. The message routing is done by rabbitmq using rsyslog tag in amqp routing-key. For us, it is useful : only one collect component deploy on servers.
B.R.
@pduveau

pduveau:
Core PR done. New PR for plugins coming soon

rgerhards:
> With the solution you propose, all the messages are not altered. With the flags you can mix the two behavior with a single output module. For example, we use rsyslog to parse application log files with imfile and other new module (imbatchreport and imtuxedoulog) and use new improg to received system statistics already formatted in json. The message routing is done by rabbitmq using rsyslog tag in amqp routing-key. For us, it is useful : only one collect component deploy on servers.

that's probably something we need to discuss in detail - let's look at the individual module PRs. It probably is a good idea to provide a reasoning up-front and also tell if somebody intends to maintain the module.

rgerhards:
> New PR for plugins coming soon

best make it one per plugin

rgerhards:
It would probably a good idea to do the doc-PRs in parallel to the module code PRs. I need to ask what exactly the modules do in any case, and with the doc this can probably be eased and speeded up.

plugins/imfile/imfile.c
```diff
@@ -316,7 +317,8 @@ static struct cnfparamdescr inppdescr[] = {
 	{ "statefile", eCmdHdlrString, CNFPARAM_DEPRECATED },
 	{ "readtimeout", eCmdHdlrPositiveInt, 0 },
 	{ "freshstarttail", eCmdHdlrBinary, 0},
-	{ "filenotfounderror", eCmdHdlrBinary, 0}
+	{ "filenotfounderror", eCmdHdlrBinary, 0},
+	{ "needparse", eCmdHdlrBinary, 0}
```
rgerhards, line 321: this looks like a feature change, not a port.

runtime/datetime.c
```diff
@@ -103,13 +103,11 @@ timeval2syslogTime(struct timeval *tp, struct syslogTime *t, const int inUTC)
 	struct tm tmBuf;
 	long lBias;
 	time_t secs;
-/* AIXPORT : fix build error : "tm_gmtoff" is not a member of "struct tm"
- *           Choose the HPUX code path, only for this function.
- *           This is achieved by adding a check to _AIX wherever _hpux is checked
+/* AIXPORT : bug fix: manage properly the time zones to calculate bias
```
rgerhards: did the previous method not work or why was it removed?

runtime/datetime.h
```diff
@@ -41,7 +41,7 @@ BEGINinterface(datetime) /* name must also be changed in ENDinterface macro! */
 	void (*getCurrTime)(struct syslogTime *t, time_t *ttSeconds, const int inUTC);
 	rsRetVal (*ParseTIMESTAMP3339)(struct syslogTime *pTime, uchar** ppszTS, int*);
 	rsRetVal (*ParseTIMESTAMP3164)(struct syslogTime *pTime, uchar** pszTS, int*, const int bParseTZ,
-const int bDetectYearAfterTime);
+		const int bDetectYearAfterTime);
```
rgerhards, line 44: indention looks wrong

runtime/datetime.c
pduveau: The previous modification was just to allow build without error but the time zone has not compute as the IBM documentation explains. So it works on partition where time is GMT it worked but when time zone with daylight the computed time in local was buggy.

runtime/msg.c
```diff
@@ -656,6 +658,8 @@ propNameToID(uchar *pName, propid_t *pPropID)
 		*pPropID = PROP_SYS_QHOUR_UTC;
 	} else if(!strcasecmp((char*) pName, "$minute-utc")) {
 		*pPropID = PROP_SYS_MINUTE_UTC;
+	} else if(!strcasecmp((char*) pName, "$wday-utc")) {
```
rgerhards, line 661: functional enhancement (series) -- needs doc update

runtime/msg.h
```diff
@@ -158,6 +158,8 @@ struct msg {
 /* rawmsg does not include a PRI (Solaris!), but PRI is already set correctly in the msg object */
 #define PRESERVE_CASE	0x200
 /* preserve case in fromhost */
+#define KEEP_FORMAT     0x400
```
rgerhards: we need to discuss the need for this - so far, I do not understand why it is useful

runtime/nsd_ptcp.c
```diff
@@ -488,18 +490,18 @@ LstnInit(netstrms_t *pNS, void *pUsr, rsRetVal(*fAddLstn)(void*,netstrm_t*),
...
-			if(r->ai_family == AF_INET6) {
-				isIPv6 = 1;
-				int iOn = 1;
-				if(setsockopt(sock, IPPROTO_IPV6, IPV6_V6ONLY,
-				      (char *)&iOn, sizeof (iOn)) < 0) {
-					close(sock);
-					sock = -1;
-					continue;
-				}
+#ifdef IPV6_V6ONLY
```
rgerhards: we prefer to have the preprocessor indented with the code - else it sometimes gets very badly readable

runtime/nsd_ptcp.c
```diff
@@ -525,23 +527,23 @@ LstnInit(netstrms_t *pNS, void *pUsr, rsRetVal(*fAddLstn)(void*,netstrm_t*),
...
-		#if !defined(_AIX) && !defined(BSD)
-			if(net.should_use_so_bsdcompat()) {
-				if (setsockopt(sock, SOL_SOCKET, SO_BSDCOMPAT,
-						(char *) &on, sizeof(on)) < 0) {
-					LogError(errno, NO_ERRCODE, "TCP setsockopt(BSDCOMPAT)");
-					close(sock);
-					sock = -1;
-					continue;
-				}
+#if !defined(_AIX) && !defined(BSD)
```
rgerhards: indention

runtime/nsd_ptcp.c
```diff
@@ -525,23 +527,23 @@ LstnInit(netstrms_t *pNS, void *pUsr, rsRetVal(*fAddLstn)(void*,netstrm_t*),
...
+				continue;
 			}
-		#endif
+		}
+#endif
 
-	        if( (bind(sock, r->ai_addr, r->ai_addrlen) < 0)
-			#ifndef IPV6_V6ONLY
+		if( (bind(sock, r->ai_addr, r->ai_addrlen) < 0)
+#ifndef IPV6_V6ONLY
```
rgerhards: indention

plugins/imfile/imfile.c
pduveau: Sorry, You are right. I forgot to remove it ad treat it as a feature. Do want me to pull it as a new PR ?

runtime/parser.c
```diff
@@ -498,9 +498,11 @@ SanitizeMsg(smsg_t *pMsg)
 					case '\b':
 						pDst[iDst++] = 'b';
 						break;
+#ifndef __xlc__
```
rgerhards: doesn't that break configs between AIX and non-aix?

runtime/msg.c
pduveau: I know and I will ;-)

runtime/rsyslog.h
```diff
@@ -25,58 +25,65 @@
...
+		 * case.
+		 * Unfortunately ignoring this diagnostic leads to two more info lines
+		 * being emitted where nobody knows what the mean and why they appear :-(
+		 */
+		#pragma GCC diagnostic ignored "-Wjump-misses-init"
+	#endif /* if __GNUC__ >= 8 */
+
+	/* define a couple of attributes to improve cross-platform builds */
+	#if __GNUC__ > 6
+	#define CASE_FALLTHROUGH __attribute__((fallthrough));
```
rgerhards: indention

runtime/nsd_ptcp.c
pduveau: I don't understand this modification. I am not its author.

runtime/rsyslog.h
```diff
@@ -673,7 +685,7 @@ struct actWrkrIParams {
 #  define  __attribute__(x)  /*NOTHING*/
 #endif
 
-#ifndef O_CLOEXEC
+#if !defined(O_CLOEXEC) && !defined(_AIX)
```
rgerhards, line 718: so this works on AIX even if the define does not exist? Sounds strange, pls double-check.

runtime/parser.c
pduveau: The espaced char '\e' is not understood by xlc.

rgerhards:
Indeed, the review is not as bad as I thought. I intentionally did single comments as I thought I would need to stop somewhere in the process. Let's clear the questions, looks like this already looks pretty good to go!

plugins/imfile/imfile.c
rgerhards: just let's document what it does and why it is nedded. I think I already grasped it and if we can agree on the usefulness, no need to craft a dedicated PR for this one.

runtime/datetime.c
rgerhards: OK, so in a nutshell it did not work previously and still does not work - right? I guess it would be helpful to update the comment so that the exact details are known. Makes it easier for others that stumble over it in the future.

runtime/nsd_ptcp.c
rgerhards: We need to clarify this. We also need to make sure that all authors explicitely agree to the contubution. If there are multiples, please make them at least post here that they do. Else we run into license issues. If this is a company contribution (where the authors assigned copyright to the company), it would be good to have a company statement here on file.

runtime/parser.c
rgerhards: ah, ok - maybe we need to do work-around - how about '\x1b`?

runtime/rsyslog.h
pduveau: You are certainly right. I think the problem was due to includes order in source code as rsyslog.h was included just after config.h and before system's ones. So as soon as the include order is adjusted in all source code file the prepocessor test would be rolled back.

runtime/msg.h
pduveau: This flags value allow an input module to inform the output module that this message message should be send as is (rawmsg) instead of using a template. (The json with system metrics as we discuss ealier for example). The core doesn't need to be aware of this and should just transmit it with the message.

rgerhards: This flag actually deeply concerns me -- it  looks like it breaks layers. I do not really understand why it is needed. You can always put in proper templates, so why add a special way to access `rawmsg` property?

rgerhards: suggestion: mabye we should remove it from current PR and add it to the omqmpq updates? In context it may be easier to discuss and we remove one show stopper from this PR here?

davidelang:
On Fri, 15 Feb 2019, Philippe Duveau wrote:

> Hi @rgerhards,
>
> `core  feature: add the ability to forward message through ouput module without
>  altering it. omrabbitmq is aware of this new flag. `
>
> With the solution you propose, all the messages are not altered. With the 
> flags you can mix the two behavior with a single output module. For example, 
> we use rsyslog to parse application log files with imfile and other new module 
> (imbatchreport and imtuxedoulog) and use new improg to received system 
> statistics already formatted in json. The message routing is done by rabbitmq 
> using rsyslog tag in amqp routing-key. For us, it is useful : only one collect 
> component deploy on servers.

the templates are set per action, not per module, so you can easily use one 
template for ont action and a different template for another.

David Lang

davidelang:
On Fri, 15 Feb 2019, Philippe Duveau wrote:

> pduveau commented on this pull request.
>
>
>
>> @@ -158,6 +158,8 @@ struct msg {
> /* rawmsg does not include a PRI (Solaris!), but PRI is already set correctly in the msg object */
> #define PRESERVE_CASE	0x200
> /* preserve case in fromhost */
> +#define KEEP_FORMAT     0x400
>
> This flags value allow an input module to inform the output module that this 
> message message should be send as is (rawmsg) instead of using a template. 
> (The json with system metrics as we discuss ealier for example). The core 
> doesn't need to be aware of this and should just transmit it with the message.

I really don't like adding additional flags for special formatting that can be 
handled by templates (or that templates can be expanded to handle). Adding them 
ends up causing grief when someone wants to do the same thing to a different 
module. And while there are a lot of things in Python tht bug me, I do like the 
idea of one (hopefully obvious) way of getting common things done. Currently all 
output formatting is done via templates.

We have a number of things where we initially couldn't do something, implemented 
a single special case option, and then later implemented a general solution that 
also can handle the special case, but we maintain the special case option for 
backwards compatibility and end up with people fighting to do something that 
needs the general case by using the special case.

David Lang

pduveau:
Ok, my approach with KEEP_FORMAT was wrong. I will change to use two actions. I did not explore the ability of using two actions (I was focused on the number of connections to rabbitmq). I will roll back this.

runtime/datetime.h
pduveau: `const int bDetectYearAfterTime);` is the last argument of `rsRetVal (*ParseTIMESTAMP3164)` why indention is wrong.

davidelang:
On Fri, 15 Feb 2019, Philippe Duveau wrote:

> Ok, my approach with KEEP_FORMAT was wrong. I will change to use two actions. 
> I did not explore the ability of using two actions (I was focused on the 
> number of connections to rabbitmq). I will roll back this.

I had a case where I needed to do different things to different messages, but 
then funnel them through one output action.

what I did was create two templates, and then populate $.message (using the 
exec_template() function) and then used $.message in the template that I used in 
the action

David Lang

rgerhards:
@pduveau 

> Ok, my approach with KEEP_FORMAT was wrong. I will change to use two actions. I did not explore the ability of using two actions (I was focused on the number of connections to rabbitmq). I will roll back this.

I am glad to see it go away. But as @davidelang said, there are probably many ways to achieve what you want to do. I admit I do not understand you need. But I have a strong feeling it can be addressed with either very simple code changes or just config. I would strongly suggest to create a PR for omampq and the let's discuss the requirement in its context. I would strongly advise not to do a lot of development (like second action) before we can have a deeper discussion.

runtime/datetime.c
pduveau: OK, I also made a mistake in my implementation. The final code is simpler. Notice : I don't how it works with hpux as the struct tz is use without to be initialized.

pduveau:
` This also triggers a question. It's not a show stopper, but something the project needs to know: does that also mean that the contribute modules will immediately be orphaned? I mean that nobody maintains them? Again, not a show stopper, but for sure something we should mention in the doc. `
I will transfer knowledge on those modules to another developer inside my company. This is planned on early March.

rgerhards:
@pduveau 
> I will transfer knowledge on those modules to another developer inside my company. This is planned on early March.

This is good news! Would be great to have that person also mentioned inside the doc.

runtime/nsd_ptcp.c
pduveau: I said that I was not the author of this modification because the view below shows on line 493 the pre-processor directive without indent. But in my PR it is not : code and preprocessor are indent together. It seems to be a github issue affecting the view.

pduveau: No problem, I was the author. I usually not indent preprocessor and I change them by reflex. I rolled back the modification.

runtime/nsd_ptcp.c
pduveau: Same comment as previous

runtime/nsd_ptcp.c
pduveau: idem

rgerhards:
non-technical issue: we are moving towards ASL 2.0. You are touching 2 files (action,c imuxsock.c) which are still GPL only. You do trivial changes (like you said in the initial commit ;-)). That would later on be hard to get ACK to move to ASL 2.0. So I will see that I modify the files myself, merge that to master so that the license concern goes away. As a side-note I think I can also improve the situation by using `_Pragma()` instead of `#pragma' - we require C11 in any case.

rgerhards:
@pduveau can you pls rebase to current master and check if it builds successfully on AIX. If so, I think I'll also convert the rest of the pragmas, as the new method looks much nicer.

pduveau:
It compiles successfully without any warning.
Do I have to push after rebase now ?

It's a pity for pragmas. I have done this a year ago on version 8.24.0. But I give up.
Below what I had done in rsyslog.h :
```
#ifdef __GNUC__
# define DO_PRAGMA(x) _Pragma(#x)
# define PRAGMA_GCC_DIAG_IGNORE(x) DO_PRAGMA(GCC diagnostic ignored "-W"#x)
# if __GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ > 4)
#   define PRAGMA_GCC_DIAG_PUSH _Pragma("GCC diagnostic push")
#   define PRAGMA_GCC_DIAG_POP _Pragma("GCC diagnostic pop")
#   define PRAGMA_GCC_DIAG_WARNING(x)
# else
#   define PRAGMA_GCC_DIAG_PUSH
#   define PRAGMA_GCC_DIAG_POP
#   define PRAGMA_GCC_DIAG_WARNING(x) DO_PRAGMA(GCC diagnostic warning "-W"#x)
# endif
#else
# define PRAGMA_GCC_DIAG_PUSH
# define PRAGMA_GCC_DIAG_POP
# define PRAGMA_GCC_DIAG_IGNORE(x)
# define PRAGMA_GCC_DIAG_WARNING(x)
#endif

PRAGMA_GCC_DIAG_IGNORE(declaration-after-statement)
....
PRAGMA_GCC_DIAG_WARNING(declaration-after-statement)

```
This way works fine on AIX and it is flexible as you do not have to modify rsyslog.h each time you need a new pragma. As a modification in this file makes a full recompilation.

rgerhards:
> It's a pity for pragmas. I have done this a year ago on version 8.24.0. But I gave up.

sorry, I do not fully understand. Did you have a problem with this? Actually I thought I would just move exactly into that direction.

rgerhards:
side-note: I see there are still some cosmetic changes in the GPL-only files. I'll see that I can iron them out via a separate PR so that we do not need to take care about them license-wise (simplifies things considerably for me).

runtime/rsyslog.h
rgerhards: I guess we still need to fix this one here.

rgerhards:
> sorry, I do not fully understand. Did you have a problem with this? Actually I thought I would just move exactly into that direction.

@pduveau  sorry, I misread it this morning. Now I see. But I think it's not a problem to define the individual pragmas, this does not change frequently. But you are of course right, the method you describe and that works on AIX is nicer. Still I think the method I began to use is better than the previous one with `#ifdef` all over the place. I think I'll see that I change a couple of more places.

pduveau:
This new method is better than the ifdef for sure. I'll rebase to solve conflicts.

rgerhards:
@pduveau mmhhhh.... it looks like something with the new pragmas does not work - I wonder why my PR went through :-S I'll check, but have a meeting in a couple of minutes, so it may take some time.

rgerhards:
had a quick look - looks like lexer.l had a "bad merge" - it still contained the pragmas, and they are now totally gone. I have on my todo to check if I can use the new macros here as well, will give this a try and let you know.

plugins/imuxsock/imuxsock.c
```diff
@@ -65,6 +65,7 @@
 #include "ratelimit.h"
 
 
+
```
rgerhards: @pduveau would you mind removing this empty line so that I do not get a change entry for this file? It's the GPLv3 issue...

pduveau: Ok, sorry for this

pduveau:
As solaris is running fine, docker makes me crazy. And Codacy/PR seems to hang

rgerhards:
> As solaris is running fine, docker makes me crazy. And Codacy/PR seems to hang

remember: I work on the CI stability - as I said, I'll let you know if you need to check those strange looking case, else I handle it.

pduveau:
Ok I am waiting for news before to rebase the other PRs. I was aware of your great work. It was a joke, no more.

pduveau:
@rgerhards, I would like to know if I rebase now others PR or if I should wait this merge.

rgerhards:
> Ok I am waiting for news before to rebase the other PRs. I was aware of your great work. It was a joke, no more.

TBH I am also going crazy on the false positives. But at least it is improving. Every day something new. E.g. new one solaris env is extremely slow (20min for git pull!) and github since 2 or 3 days loses notifications (that then looks like a CI run hangs). And so on. But again, what I have under my control gets better. I just had hoped it would finally go 100% good, but... ;-)

pduveau:
Welcome to the hell ;-)

rgerhards:
@pduveau you could do me a favor: I don't understand why I have action.c still in "changed files". Could you do one (hopefully final) rebase? If that goes through, I can merge. After that it's probably a good time to do a rebase on the other PRs.

rgerhards:
@pduveau merci beaucoup

rgerhards:
finally! :-D

rgerhards:
ok, I have forgotten to squash ... but so is life ;-) Looks like I was too happy to get over this one :-)

rgerhards:
@pduveau note that the changelog contained entries for the now-separate PRs. I have corrected this. Will add when I merge the PRs.

PR #3473, opened 2019-02-15
Segfault in omclickhouse fixed

gagantrivedi:
Messages which does not contain the word "VALUES" in them were causing
segfault.

ps: let me know if there is any other better way to tackle this.

### DIFF
--- a/plugins/omclickhouse/omclickhouse.c
+++ b/plugins/omclickhouse/omclickhouse.c
@@ -429,6 +429,10 @@ computeBulkMessage(const wrkrInstanceData_t *const pWrkrData,
 	const uchar *const message, char **newMessage)
 {
 	size_t r = 0;
+	if(!strstr((const char*)message, "VALUES")) {
+		return r;
+	}
+
 	if(pWrkrData->batch.nmemb != 0) {
 		*newMessage = strchr(strstr((const char*)message, "VALUES"), '(');
 		r = strlen(*newMessage);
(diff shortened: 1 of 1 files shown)


rgerhards:
sry I didn't mange to look at this PR before the release.

Do you have a sample config that causes the issue? I would like to craft a test for it.

gagantrivedi:
@rgerhards here is the template code which was causing the issue

```
template(name="endpoint" type="list"){{
property(name="msg" field.delimiter="32" field.number="3")
constant(value=" ")
property(name="msg" field.delimiter="32" field.number="4")
}}

set $.endpoint = exec_template("endpoint");
set $.endpointlen = strlen($.endpoint);
set $.totallen = strlen($msg);
set $.userid = field($msg,32,2);
set $.useridlen = strlen($.userid);
set $.data = substring($msg, $.useridlen+$.endpointlen+2, $.totallen);

template(name="access-log" type="list" option.stdsql="on" ){{
constant(value="INSERT INTO rsyslog.SystemEvents (user_id, timestamp, endpoint, data) ")
constant(value="values ( ")

property(name="$.userid" )
constant(value=", ")

property(name="timestamp" dateFormat="unixtimestamp")

constant(value=" ,' ")
property(name="$.endpoint")

constant(value=" ', '")
property(name="$.data")
constant(value="')")
}}

action(type="omclickhouse" Template="access-log" usehttps="off" bulkmode="on" queue.size="1000")            


```

rgerhards:
mhhh - I don't get a segfault but rather the expected error message. Could it be that https://github.com/rsyslog/rsyslog/commit/4b8950434fc054909fb64f4fb3321e72545cc390#diff-9f9e5686b04439b44ad9838c23fa8b4e fixed this here as a side-effect?

rgerhards:
I assume this is fixed in 8.1904.0 - let me know if not.

PR #3477, opened 2019-02-17
Add improg contribution

pduveau:
This module allow to launch a program and consume its output (stdout) lines as messages.


rgerhards:
@jsiwrk based on the great work you do on `omprog`, would you be willing to do a review of this contribution?

rgerhards:
@pduveau I would propose to avoid doc in `README` files, as nobody really will see it. I suggest to move it over to the rsyslog-doc project (and limit `README` to hard facts for developers to avoid duplicate info).

rgerhards:
@pduveau 

*  I neeed to enable the new module in CI, will start with the most important platforms. This may bring up new compile issues (which may be surprising if you don't know the module was not built yet).
* we would really like to see at least some basic tests for the module. You could look at the `omprog` tests, I guess something similar could be used. If in doubt, I'll gladly lend a helping hand.

These comments probably apply to the other module PRs as well. I will not repeat them there.

pduveau:
@jsiwrk, I have rename few instance attribute related to pipes to be more understandable by me. Because Stdout and Stdin is depending of which point of view parent/child. So to be more efficient I preferred to named them direction To/From the parent/child. Also, I preferred a simple acknowledgement mechanism instead of the complexe transaction one. I think transaction has a real sens on outputs but should be just a nice theoretical exercise for inputs with no real needs.

pduveau:
@rgerhards, Thanks for information about your work on CI. Now I understand why so much failure on commits.
For documentation, I will move my contributions' documentation in rsyslog-doc shortly.

rgerhards:
To get rid of the write compile error actually the best thing to do is check the return code and do an error or debug message if it failed. If you really think about it, it's also a good thing to do (I needed to convince myself, too, when I got this the first time).

pduveau:
@rgerhards, for tests of new plugins including omrabbitmq I'll work on them in March.

rgerhards:
> tests of new plugins including omrabbitmq I'll work on them in March.

sounds great! Do you already know what I need to install into the containers so that a test can be carried out? If so, I would begin to add it.

pduveau:
For omrabbitmq, the container should contain RabbitMQ server and its dependencies (erlang, socat). It comes with tools to administer it. Container should install it and start it with default port (5672). I think that most of distribs provide them by default. I don't know for Solaris. We also need curl to get rabbitmq informations

rgerhards:
> Container should install it and start it with default port (5672).

the start must be handled by the testbench, else we need too many ressources. It's easy to do, we do it e.g. for mysql and clickhouse. I can help with that.

pduveau:
@rgerhards, for testbench I'll do my best but I will not be able to build them before early March. I have two others parallel projects to close before end of February. I should not work on rsyslog this week and I have still a last plugin to PR.

rgerhards:
> , for testbench I'll do my best but I will not be able to build them before early March

no problem - I am happy you can look at it! I'll also check if we have remaining real issues and, if so, will let you know.

jsiwrk:
> @jsiwrk based on the great work you do on `omprog`, would you be willing to do a review of this contribution?

Sorry, I am currently too busy for an in-depth review...

However, after a quick glance, it looks good! One thing that could be enhanced: in readChild, read the line using a buffer instead of char by char, for efficiency (time ago I made this optimization in omprog, because the multithreaded tests showed a huge increase of efficiency when doing so, because of the reduction of the number of syscalls).

Another minor detail: the confirmation message in improg is "ACK", whereas in omprog I chose "OK" (also, in omprog, the program can respond with an error, however it is not clear if this could apply to improg: could rsyslog indicate to the program that the message could not be processed?). This is a little inconsistent, but actually I like "ACK" more... So don't know what to say, maybe it's not important to keep both modules consistent to this level (?).

Note however that I do not have experience with input modules... I have not compared the general plumbing with other modules, I assume it's ok.

pduveau:
@jsiwrk, Thanks very much for your pertinent review/comments.
> in readChild, read the line using a buffer instead of char by char

I was thinking to use a buffer. A lake of time made me use this non efficient method. I think to enhance this in a future PR. First, I have to provide test.
> the confirmation message in improg is "ACK" ... in omprog I chose "OK" ... This is a little inconsistent, but actually I like "ACK" more.

I choose ACK driven by usual messaging system (JMS, AMQP, ...) where verbs for this mechanism are ack/nack. I don't think that consistency between those two modules is a real key.

The main objective of this mechanism is to control the flow emitted by a too heavy external program (to avoid flooding). When I conceive it, I was thinking to make this mechanism mandatory but I changed my mind to allow simpler integration (shell script for example).

rgerhards:
@pduveau looks like we have some issues on Solaris - sry, forgot to enable it initially.

rgerhards:
@pduveau compile issue on solaris

rgerhards:
@pduveau I just added at least a very basic dynamic test, which you can also use later on as an example. As it turned out, it also already brought up an issue ;-)

pduveau:
@rgerhards, I saw the issue and patch it (a rookie's one : shame on me).

pduveau:
@rgerhards, I think that this module is ready. I have to switch to AWS Vpc Logs Enhancement. I'll be back to provide tests around the 11th of March.

rgerhards:
@pduveau thx, I'll give it a final check, but also think it is ready.

side-note: I will probably unavailable in the week starting march, 11th,

rgerhards:
it's actually not yet merged because I screwed up again and forgot to do the github squash :-( I've noticed it immediately and so rewrote master branch, it really would be a shame to do this once again with all the fixup commits. I'll submit a new PR...

PR #3476, opened 2019-02-16
Featuring omrabbitmq

pduveau:
- add routing-key template to allow routing depending on the message
- add capability off populating amqp header
- add delivery mode
- add expiration
- add HA capability with two rabbitmq servers

PR #3479, opened 2019-02-18
Contrib_pmdb2diag

pduveau:
This module parse special DB2 Logs to extract data to rsyslog's message structure. It requires that message flags NEEDS_PARSING to be raised by input module. "imfile" propose this enhancement in PR #3471

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -60,6 +60,7 @@ extraction:
           --enable-mmtaghostname
           --enable-imbatchreport
           --enable-imdocker
+          --enable-pmdb2diag
 
 queries:
   - include: cpp/commented-out-code
--- a/ChangeLog
+++ b/ChangeLog
@@ -97,6 +97,7 @@ Version 8.1903.0 (aka 2019.03) 2019-03-05
 - omhttp: rewritten with large feature enhancements
   Many thanks to Gabriel Intrator for this work. Gabriel also has adopted the
   module and plans to support it in the future.
+- pmdb2diag: create parser module for DB2 diag logs
 - TLS subsystem: add support for certless communication
   both openssl and GnuTLS drivers have been updated to support certless
   communications. In this case e.g. Diffie-Helman is used.
--- a/Makefile.am
+++ b/Makefile.am
@@ -320,6 +320,11 @@ if ENABLE_MMTAGHOSTNAME
 SUBDIRS += contrib/mmtaghostname
 endif
 
+# imdb2diag
+if ENABLE_PMDB2DIAG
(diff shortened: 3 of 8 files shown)


rgerhards:
I have updated CI to compile this module at least on some platforms. The compile CI failures are valid, the rest need a restart. But this happens automatically when you fix the compile issues, so I spare the resources.

rgerhards:
@PascalWithopf it looks like Codacy found a valid point. The ubuntu18 is a flake, but one that I hopefully fixed this morning. I do not restart as you need to re-push anways.

pduveau:
Same problem as #3476

rgerhards:
@pduveau compile issue on solaris

pduveau:
@rgerhards, compilation issue is solved now an existing test failed

rgerhards:
Thx, reinitiated.

pduveau:
Hi @rgerhards, can you check this error.

rgerhards:
> Hi @rgerhards, could you check this error.

The solaris workers were offline for quite a while, probably due to some machine issues. But it looks like they have come online again. I've restarted.

Also, I am back now - so let's see that we add tests and get the PRs through.

rgerhards:
The Fedora 28 is an interesting flake: its based on very fast port reuse by the OS. Probably this happens when CI is very busy. We have no cure against this right now, requires more dev work...

rgerhards:
@pduveau I can provide you with access credentials for the buildbot admin interface. That would enable you to restart individual workers yourself. I guess the overload is sometimes caused by the full CI running quickly after each other. Buildbot cannot cancel old instance, unfortunately.

Not sure if the credentials make sense at this point in time. If you think yes, please send me private mail.

pduveau:
@rgerhards It should be better to provide an 'operator' level access with just the 'relaunch' button enabled. If this is to heavy to allow this level, I'm interesed in having this feature open to me so my private email is philippe.duveau@free.fr. And I'll not use this access to do more than relaunch individual check. The heavy usage of CI are very painfull during the day. It is the reason why I launched PR check during the night (01:00 to 10:00).

rgerhards:
> The heavy usage of CI are very painfull during the day. It is the reason why I launched PR check during the night (01:00 to 10:00).

yup - but on the other hand it brings up issues otherwise only seen in production. The bad thing is flakes, which I am working on. But that seems to be a continuous/eternal project...

I'll send you mail in a couple of minutes.

rgerhards:
The fedora 29 "flake" is currently under investigation. It may not be a flake ;-)

That said, it looks like we are very close to merge. I'll have a look again later today or tomorrow.

PR #3482, opened 2019-02-18
contrib_imtuxedoulog

pduveau:
Tuxedo ULOG input module


rgerhards:
I admit I am a bit concerned because this sounds like some duplication of imfile, but I think it is acceptable as contributed module. @davidelang what do you think?

rgerhards:
I have updated CI to compile this module at least on some platforms. The compile CI failures are valid, the rest need a restart. But this happens automatically when you fix the compile issues, so I spare the resources.

pduveau:
@rgerhards, 
> I admit I am a bit concerned because this sounds like some duplication of imfile, but I think it is acceptable as contributed module. @davidelang what do you think?
As I explained in the documentation, it is a duplication of polling mode of imfile. Except that the log lines read from the file need to have the filename to construct the timestamp (you have the time (hhmmss.ms) in the log and the date is in the filename (ulog.MMDDYY). So a parser could not get this information. And if you use a glob to open the file the module will open many unusefull files (all days in directory) but only the current date is usefull.
I tried to handle this in imfile : the result was unmaintainable. So as I had to mange some particularities, I decided to manage everything about ULOGs in one module and parse the log in the input module to avoid two modules which can be used separately.

rgerhards:
> As I explained in the documentation, it is a duplication of polling mode of imfile. Except that the log lines read from the file need to have the filename to construct the timestamp (you have the time (hhmmss.ms) in the log and the date is in the filename (ulog.MMDDYY). So a parser could not get this information. And if you use a glob to open the file the module will open many unusefull files (all days in directory) but only the current date is usefull.
> I tried to handle this in imfile : the result was unmaintainable. So as I had to mange some particularities, I decided to manage everything about ULOGs in one module and parse the log in the input module to avoid two modules which can be used separately.

sorry, I should have explained. I read all this and without it I would have objected. I see the dependency on the file name's date and that sounds really strange.

pduveau:
Tuxedo is very strange ;-)

davidelang:
so from the discussion here, if imfile could match only some files, not all files to watch, then this module could be eliminated as the filename metadata could be used to extract the date portion, is this correct?

or would we need to also enhance the date functions to manipulate dates correctly?

pduveau:
The problem is the filename is changing everyday. This could AWFULLY be handled by notify operating system using glob filename definition. But this is not working at all in a polling configuration (AIX notify system is not integrated with imfile). So imfile can NOT be a candidate on AIX.

rgerhards:
@pduveau could you add that information to a `README` file? I think it would be useful for later developers, as it shows the design decision and also ways to improve. And `README` this time is the right place as the info is targeted towards developers.

@davidelang In the light of all this, I think the right thing to do now is merge - after all, it's contrib.

pduveau:
Ok, I'll add this discussion summary to the README.

pduveau:
@rgerhards, As I understood, as long as I don't provide test the codecov report will be bad. Am I right ?

rgerhards:
@pduveau 

> as long as I don't provide test the codecov report will be bad. Am I right ?

yupp, so this looks good to go. I'll do a quick review soon.

rgerhards:
static analyzer was not yet enabled for this module -- will change and re-initiate...

pduveau:
Just a last commit to make "static analyzer" happy.

pduveau:
Same problem as in #3476

rgerhards:
@pduveau it would be great if you could do a rebase

rgerhards:
@pduveau can we get some tests here that actually run the code?

pduveau:
@rgerhards, there are tests with 83% coverage. One positive with all messages variance and one with error message

rgerhards:
mhhh... sry, I probably overlooked `tests/imtuxedoulog_data.sh`. This is the only test other than the load tests, right? It's fine if so, as it seems to cover much. I just wonder why I overlooked it...

rgerhards:
Let's give it a final soon with updated CI.

rgerhards:
@pduveau Thanks again for the contribution and your hard work to help getting it in shape for merge!

side-note: I'll try to work on the other outstanding PRs tomorrow.

PR #3491, opened 2019-02-20
testbench: modernize test and hopefully make more robust on slow CI

rgerhards:


rgerhards:
FreeBSD is actually "green", but github seems to have missed status notification twice in row. So merging now.

PR #3492, opened 2019-02-20
cleanup: use less verbose C11 method to use pragmas

rgerhards:


rgerhards:
all "green", notification seem to have been lost

PR #3496, opened 2019-02-20
On read timeouts, guard against derefrencing uninitialized act->pStrm…

n2yen:
…. This appears to

occur for active objects that relate to directories or symlinks


n2yen:
Hi @rgerhards @alorbach, noticed this issue appears to be related to act->pStrm being uninitialized for directories and/or symlinks in the active object list. 
Issue # 3474. Would this be a proper fix?

rgerhards:
sounds interesting, thx for the PR. Pls give us a bit to evaluate.

PR #3497, opened 2019-02-21
lexer.l: change to PRAGMA macros

rgerhards:

### DIFF
--- a/grammar/lexer.l
+++ b/grammar/lexer.l
@@ -31,10 +31,6 @@
 %top{
 #ifndef __clang_analyzer__ /* this is not really our code */
 #include "config.h"
-#if !defined(__GCC__)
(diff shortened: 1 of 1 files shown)


pduveau:
Hi @rgerhards, I don't think that moving rsyslog.h above system includes is a good idea. I think you should move all includes in %top block and put the pragmas at the end of the block as well.

pduveau:
Ok as soons as this RP is merged. I'll rebase again.

rgerhards:
did this as side-activity while in the meeting. Will merge when CI is through. Thx for the suggestion!

rgerhards:
@pduveau merged - some Solaris is missing in CI, but one is through so I think it was OK to merge. Solaris workers are right now veeeery slow. I guess OpenCSW is having issues.

PR #3503, opened 2019-02-22
imfile: Added new option statefilealwaysremove

alorbach:
imfile: Added new option statefilealwaysremove 

Due problems with logrotate and INOTIFY, we added an option 
to control how statefiles for moved files are handeled. 

By default "statefilealwaysremove" will be off. But if "on"
the statefile for moved files will also be removed. 

Otherwise when using logrotate for example, imfile could stop monitoring 
rotated files. This problem only happens in INOTIFY mode. 

testbench: Added new tests for logrotate and multiple rotates.

### DIFF
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -130,8 +130,9 @@ struct instanceConf_s {
 	uchar *pszFileBaseName;
 	uchar *pszTag;
 	size_t lenTag;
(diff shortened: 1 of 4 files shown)


rgerhards:
This will most probably be replaced by https://github.com/rsyslog/rsyslog/pull/3521

rgerhards:
replaced by #3521

PR #3504, opened 2019-02-22
ksi components: workaround to make compile with newer ksi library

rgerhards:
it looks like new ksi library has deprecated some APIs, which now
leads to build failure if we do not permit the diagnostic.


rgerhards:
@vii5ard are you aware of some KSI lib changes that cause the failures I work around here?

rgerhards:
got a better solution!

PR #3514, opened 2019-02-24
contrib imbatchreport

pduveau:
This input module manage batches' reports : complete file as a single log.
The objective is to use rsyslog as a single collector to a central logs system (ElasticSearch, ...)


davidelang:
how does this differ from imfile?

David Lang
k

pduveau:
Documentation is coming. You'll find the answer in.

rgerhards:
doc: https://github.com/rsyslog/rsyslog-doc/pull/785

rgerhards:
@pduveau enabling module in CI runs

davidelang:
Reading this, it seems to me that a couple tweaks to imfile capabilities would 
give the equivalent capabilities

1. add a readmode that's 'entire file'

   there is a race condition here, how do you know the entire file is written and you aren't reading a partial file?

   If the file is small enough, the write may be atomic, but you can't count on it.

   Possibly have it only read files that are at least 1 second old??

then you parse the file with a liblognorm parser (I believe this was added to 
imfile in the last year, am I correct). Can this parser populate the standard 
properties?

2. add an option to liblognorm to process what it's passed as one entity instead 
of breaking it down by newlines. We've needed the ability to handle multi-line 
messages for a while. extend %whitespace% to optionally include linefeed and 
newline and add %linefeed% and %newline% items so rulefiles don't have to 
distinguish between newlines in the message and newlines in the file format.

3. add an 'errorfile' type capability to imfile that will write the message to a 
different file if parsing fails (including message too large), and if the 
readmode is 'entire file', rename the file (note, the regex should apply to the 
full path, not just the basename)

There are a number of special case items in imbatchreport that will be limits in 
the future (structured data in only the last 150 characters for example), let's 
try to avoid those.

David Lang

pduveau:
Please David, keep imfile code simple to be maintainable by common developers. As we are usual to say this case : "It is not because you have a hammer that all your problems are nails". You should use the right tool according to characteristics of the need. I think imfile has to handle fs notify mechanisms for various OS and be a precision screwdriver not a hammer.
It is exactly like databases each database technology has its usages. Few editors tried to handle all usages in one database making them complex to operate with a huge knowledge/training for DBA.

If I understand correctly your proposal, the modifications concerns :
- adapting core/stream to support a new mode, 
- adapting liblognorm to support white spaces reduction,
- writing a new parser to extract structured data.
At least the 2 first adaptations are very risky (a real bugs' provider) compared to the need I would like to handle.

Also using a unique error file do not correspond to admins needs, because the problem will be generally corrected by adapting the message size, rename back the rejected files to be finally collected after restarting rsyslog.

davidelang:
On Mon, 25 Feb 2019, Philippe Duveau wrote:

> Please David, keep imfile code simple to be maintainable by common developers.

at the same time, we want to avoid duplication of effort, and especially having 
multiple modules that do close to the same thing, but with one having one 
feature and another having a different feature (over time, the 'missing' 
features will get added until the modules get closer to duplicates)

> If I understand correctly your proposal, the modifications concerns :
> - adapting core/stream to support a new mode,

This isn't that hard (I added two modes, another mode has been added since).
The hard part is how you define that the file is complete and ready to be read, 
reading to EOF is trivial.

> - adapting liblognorm to support white spaces reduction,

That's already there, %whitespace% matches any number of whitespaces.
The 'hard' partof the liblognorm change is to make it able to stop treating a 
newline as the end of the message.

> - writing a new parser to extract structured data.

This should be extracted with the standard tools for parsing messages 
(liblognorm) rather than a special parser that only matches your current use 
case. Even if we end up adding imbatchreport, we should not have a hard-coded 
parser for your single use case (unless you can make the case that that 
particular message format is extremely common)

> At least the 2 first adaptations are very risky (a real bugs' provider) compared to the need I would like to handle.

they are both things that are useful to other people in other situations.

> Also using a unique error file do not correspond to admins needs, because the 
> problem will be generally corrected by adapting the message size, rename back 
> the rejected files to be finally collected after restarting rsyslog.

That's fine, I see the use case for this and think it's worth supporting. Over 
time we keep adding error file support to more and more modules, adding it to 
imfile doesn't seem like a bad idea.

If you are in 'whole file' mode, then the errorfile can reasonably be a rename 
of the file (just make sure that it's not going to be read in as well or you end 
up with a loop)

Rainer is the one who makes the final decision here, and he knows the 
maintinance overhead of a little-used contrib module better than I do. I just 
think that we can solve your problem in a way that ends up better supported over 
the long run, and is something that helps others as well.

David Lang

pduveau:
I forgot to say that like the precedent module for tuxedo, this module need to use glob to identify files to manage. This mechanism is not provide by imfile in polling mode (mandatory for AIX). On AIX, imfile can not be a candidate.

rgerhards:
> This mechanism is not provide by imfile in polling mode (mandatory for AIX).

That was the state roughly 1,5 yrs ago. Today, all modes are available, not matter if polling, FEN or inotify.

That said, I personally think it makes sense to merge imbatch right now just like it is. As far as I see, it's an AIX only module. Imfile is already pretty complicated, and I admit I prefer to keep it as simple as possible, at least for border cases - which I think this is. If there is real interest in imbatch, we can always incorporate it's functionality into imfile. After all, imbatch is contrib, so we do not guarantee anything about it.

rgerhards:
@pduveau I just noticed we also have not enabled LGTM. We need to do this via `.lgtm.yml` in the root directory. Could you update this file in all PRs?

rgerhards:
@pduveau I have added a commit to add LGTM check

rgerhards:
@pduveau could you pls do a rebase?

pduveau:
Hi @rgerhards, Could you check "docker-ubuntu18-san rsyslog".

rgerhards:
> Could you check "docker-ubuntu18-san rsyslog".

seems to be a flake, I have restarted it. Will you add tests during the week? I'd like to merge this with the tests included.

pduveau:
Hi @rgerhards, All the tests are provided for the modules imbatchreport, mmtaghostname, pmdb2diag and imtuxedoulog I think you can merge those modules. I will soon provide tests for improg.
I'm doing a hard work for omrabbitmq. I got a lot a erratic problems with "docker-ubuntu18-san rsyslog" I hope that the restart have solved my problems.

rgerhards:
> Hi @rgerhards, All the tests are provided for the modules imbatchreport, mmtaghostname, pmdb2diag and imtuxedoulog I think you can merge those modules. I will soon provide tests for improg.

thx, I didn't realize this. I've just had a very quick look, I see some old testbench plumbing. I'll add a commit to your PR with some slight changes. I do a single commit so that you can easily see. You should squash it afterwards.

rgerhards:
@pduveau commit pushed: https://github.com/rsyslog/rsyslog/pull/3514/commits/b3050719674ab03a6833b6cd04cd743a92896c36

rgerhards:
fixed merge confilict

rgerhards:
@pduveau will somebody work on these PRs in the future?

pduveau:
@rgerhards, this PR is complete for me except squashing the commits.

PR #3515, opened 2019-02-24
Drop messages that were rejected due to 'RD_KAFKA_RESP_ERR_MSG_SIZE_T…

n2yen:
…OO_LARGE' error.

### DIFF
--- a/plugins/omkafka/omkafka.c
+++ b/plugins/omkafka/omkafka.c
@@ -804,7 +804,9 @@ writeKafka(instanceData *const pData,  uchar *const key, uchar *const msg,
 		updateKafkaFailureCounts(msg_kafka_response);
 
 		/* Put into kafka queue, again if configured! */
-		if (pData->bResubmitOnFailure && b_do_resubmit) {
+		if (pData->bResubmitOnFailure &&
+				b_do_resubmit &&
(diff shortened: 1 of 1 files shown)


n2yen:
Ran into an issue where kafka processing rates would drop significantly. The idea is to avoid continually resubmit messages that are too large, such as per kafka configuration, "message.max.bytes".

rgerhards:
sounds like a good idea! would you mind updating the doc as well?

n2yen:
Will do!

fivesheep:
Actually, it's more than just this error. in kafka, there's a concept called retriable exceptions and a handful non-retriable exceptions which msg too large is one of them. It's better to include all the non-retriable ones if possible

n2yen:
hi @fivesheep @rgerhards, 
Would you be opening a separate PR to handle all non-retry-able exceptions? 
We would need to carefully identify them from all the possible rd kafka error returns...

rgerhards:
@n2yen 

> Would you be open to having a separate PR to handle all non-retry-able exceptions?
> We would need to carefully identify them from all the possible rd kafka error returns...

sorry I overlooked this. For sure I appreciate all PRs that help improve the state of rsyslog and it's modules!

PR #3519, opened 2019-02-27
Enlarged msg offset types for bigger structured messages

jvymazal:
using a large enough (dozens of kBs) structured message
it is possible to overflow the signed short type which leads
to rsyslog crash.

This is possible because with setting higher maxMessageSize  one can overflow the small signed type which the core dumps during additional message processing.

### DIFF
--- a/runtime/msg.c
+++ b/runtime/msg.c
@@ -839,7 +839,7 @@ msgBaseConstruct(smsg_t **ppThis)
 	pM->iFacility = LOG_INVLD;
 	pM->iLenPROGNAME = -1;
 	pM->offAfterPRI = 0;
-	pM->offMSG = -1;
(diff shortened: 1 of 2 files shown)


jvymazal:
@rgerhards could you please look into whether this way of fixing the problem is adequate? 
Thanks

rgerhards:
give me a bit .. very busy with... Will see to get it into next release.

jvymazal:
Ok, many thanks

jvymazal:
As far as I found the negative value is not either used or checked anywhere, moreover it makes sense to me to have variables used in memcpy() as unsigned type as sort-of another layer of crash prevention.

rgerhards:
thx for the triple-check ;-)

PR #3522, opened 2019-03-01
Allow header values to have spaces

nbrownus:
Allows spaces in `headervalue`s for the `omhttp` module


rgerhards:
@nbrownus I would say that's a bugfix, do you agree?

nbrownus:
Certainly, spaces are valid in header values.

PR #3525, opened 2019-03-01
contrib_mmnull

pduveau:
This module allows to force hostname after parsing to the localhostname of rsyslog and/or add a tag messages received from input modules without tag attributes.


rgerhards:
I think the name `mmnull` should be changed to something more intuitive - maybe `mmtaghostname`. Slso not good, I know. But "null" implies it does nothing (think `pmnull` which does exactly this: no parsing).

pduveau:
OK

PR #3526, opened 2019-03-01
contrib_mmtaghostname

pduveau:
This module allows to force hostname after parsing to the localhostname of rsyslog and/or add a tag to messages received from input modules without tag parameter.


rgerhards:
@pduveau could you pls do a rebase?

PR #3535, opened 2019-03-06
WiP: rsyslog doesn't respond to TLS close_notify 

Arnix:
do a check with larger buffer and reset on >= 0 bytes read .

### DIFF
--- a/runtime/nsd_ptcp.c
+++ b/runtime/nsd_ptcp.c
@@ -896,12 +896,12 @@ CheckConnection(nsd_t *pNsd)
 {
 	DEFiRet;
 	int rc;
-	char msgbuf[1]; /* dummy */
+	char msgbuf[32]; /* dummy */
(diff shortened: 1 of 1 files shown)


rgerhards:
While I don't have a clear indication *why* this should work, reports exists it *actually works*. So I think I'll go ahead and merge ... and let's see if the problem persists ;-)

markdascher:
Agree that this will probably do the trick. It's still not entirely ideal since the check only occurs when there are logs to send, but it's definitely a step in the right direction.

Not sure if the increased buffer size is needed, but it can't hurt.

rgerhards:
let's see if problems re-occur. BTW: did you try with gnutls only or also with openssl?

markdascher:
My testing was only with GnuTLS so far.

rgerhards:
It would be great if you could give openssl a spin w/ 8.1903.0. The driver name is `ossl` vs `gtls`.

PR #3557, opened 2019-03-20
omelasticsearch: add support for rebindinterval

richm:
Adds support for a new parameter `rebindinterval`
This parameter tells omelasticsearch to close the connection and
reconnect to Elasticsearch after this many operations have been
submitted.  The default value `-1` means that omelasticsearch will
not reconnect.  A value greater than `-1` tells omelasticsearch,
after this many operations have been submitted to Elasticsearch,
to drop the connection and establish a new connection.  This is
useful when rsyslog connects to multiple Elasticsearch nodes
through a router or load balancer, and you need to periodically drop
and reestablish connections to help the router balance the
connections.  Use the counter `rebinds` to monitor the
number of times this has happened.

### DIFF
--- a/tests/es-basic.sh
+++ b/tests/es-basic.sh
@@ -4,7 +4,14 @@
 export ES_DOWNLOAD=elasticsearch-6.0.0.tar.gz
 export ES_PORT=19200
 export NUMMESSAGES=1000 # 1000 is sufficient, as this test is pretty slow
-export QUEUE_EMPTY_CHECK_FUNC=es_shutdown_empty_check
+REBIND_INTERVAL=100 # should be enough to run several times for $NUMMESSAGES
+
+queue_empty_check() {
+	es_shutdown_empty_check && \
+	content_check --regex '"name": "omelasticsearch".*"submitted": '$NUMMESSAGES \
+		$RSYSLOG_DYNNAME.spool/omelasticsearch-stats.log
+}
+export QUEUE_EMPTY_CHECK_FUNC=queue_empty_check
 
 download_elasticsearch
 prepare_elasticsearch
@@ -16,6 +23,8 @@ add_conf '
 template(name="tpl" type="string"
 	 string="{\"msgnum\":\"%msg:F,58:2%\"}")
 
+module(load="../plugins/impstats/.libs/impstats" interval="1"
+	   log.file="'"$RSYSLOG_DYNNAME.spool"'/omelasticsearch-stats.log" log.syslog="off" format="cee")
 module(load="../plugins/omelasticsearch/.libs/omelasticsearch")
 
 if $msg contains "msgnum:" then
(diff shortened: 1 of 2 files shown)


davidelang:
Our other output modules call this "rebindInterval", could you use the same 
name?

David Lang

rgerhards:
I think the `es-basic.sh` test is now racy again. It now depends on proper timing of stats emission. We should try to make sure we have the correct number and change the termination check to check that as well.

richm:
> we need to solve the test raciness.

try the latest commit

rgerhards:
> try the latest commit

It's not a question of trying - reviewing the test code, I do not see that the appearance of the stats record is checked before rsyslog is terminated. I'll have a closer look and will possibly sumbit a commit that introduces that.

To avoid racynes, we change the testbench to check that the outcome is actually achieve before we terminate. This solved many of the races and flakes we previously had.

tests/es-basic.sh
```diff
@@ -16,21 +17,60 @@ add_conf '
...
 	       server="127.0.0.1"
 	       serverport="'$ES_PORT'"
 	       template="tpl"
-	       searchIndex="rsyslog_testbench")
+	       searchIndex="rsyslog_testbench"
+	       rebindinterval="'$REBIND_INTERVAL'")
 '
 startup
 injectmsg  0 $NUMMESSAGES
+wait_queueempty
```
richm: Wait for all records to be processed by rsyslog

tests/es-basic.sh
```diff
@@ -16,21 +17,60 @@ add_conf '
...
+# now, wait for $NUMMESSAGES to be reported in stats file e.g.
+# Fri Mar 29 02:29:04 2019: @cee: { "name": "omelasticsearch", "origin": "omelasticsearch", "submitted": 1000, ...
+wait_content '"name": "omelasticsearch".*"submitted": '$NUMMESSAGES $RSYSLOG_DYNNAME.spool/omelasticsearch-stats.log
```
richm: Wait for the stats file to report that omelasticsearch has processed all of the records

then shutdown rsyslog

@rgerhards does this not solve the race, because now the test waits until rsyslogd has a chance to write out the stats file?

rgerhards: line 37 should solve the issue - If you agree, I'll add a commit which combines this via the new method used for this kind - it's a bit faster.

actually
```
export QUEUE_EMPTY_CHECK_FUNC=es_shutdown_empty_check
```
is already part of it. But it's easier to show in code than english text ;-)

richm: > If you agree, I'll add a commit which combines this via the new method used for this kind

yes, please do

rgerhards: will do soon (finishing up things before switching)

rgerhards:
pushed https://github.com/rsyslog/rsyslog/pull/3557/commits/e0c056d285d305db5dba987466acab3a62daf598, feel free to squash. Commit comment says it all.

richm:
thanks! - squashed

PR #3560, opened 2019-03-22
omamqp: fix build errors

rgerhards:
They occur on some, newer, platforms.


rgerhards:
Fedora29 was unrelated issue - this is being worked on.

Issue #3561, opened 2019-03-22
openssl component test fails on Fedora 29


rgerhards:
It looks like there are some problems with the openssl component. This needs to be investigated.

sample: https://build.rsyslog.com/#/builders/197/builds/4

container for reproduction: rsyslog/rsyslog_dev_base_fedora:29

Issue #1503, opened 2017-04-05
rsyslog doesn't respond to TLS close_notify


markdascher:
We've noticed that rsyslog, when sending to a remote server, never reads from a TLS socket after the initial connection establishment. For example, let's say that rsyslog is connected to a receiver that closes idle connections after 15 minutes. Here's what happens:

* rsyslog connects to remote, negotiates TLS, and sends a few messages. `netstat -t` shows that Recv-Q and Send-Q are both empty. (So far, so good.)
* After 15 minutes of inactivity, remote sends a TLS close_notify, followed by a TCP FIN, but rsyslog never responds to either one. In fact, `netstat -t` shows Recv-Q has 54 bytes waiting to be read, which is the size of the TLS close_notify. It also shows the connection is in a CLOSE_WAIT state, which means that the TCP connection is half-closed. Remote has already stopped listening, so any further sends will be lost.
* This condition remains unchanged, until the next time rsyslog has a message to send. Since the TCP connection is still half-open, the send succeeds, and we see one more data segment being transmitted.
* Best case scenario now: Remote responds with a TCP RST, forcing the TCP connection to close. `netstat -t` no longer lists the connection at all. We've lost at least one message so far, but the next send will fail, allowing rsyslog to reconnect. Relatively painless.
* Worst case scenario: Remote responds with a TCP RST, which is intercepted by an intermediate packet filter/firewall for some reason, and dropped. The OS continues with TCP retransmissions, but without any active response, this may take hours/days to finally kill the TCP connection. `netstat -t` shows Send-Q gradually fill up, as more messages are sent. I imagine rsyslog would eventually stall entirely, or if this target has its own queue, just that thread would stall.

I've poked through the code and found that [nsd_gtls](https://github.com/rsyslog/rsyslog/blob/master/runtime/nsd_gtls.c#L1425) includes a CheckConnection function, which defers to [nsd_ptcp](https://github.com/rsyslog/rsyslog/blob/master/runtime/nsd_ptcp.c#L809)'s function of the same name. But all it does is a 1-byte peek. Interestingly, this would actually work perfectly if the remote end simply closed the TCP connection, without sending a TLS close_notify first. But the fact that there's 54 bytes waiting in Recv-Q prevents the recv from ever returning zero. Skipping the close_notify would be inconsistent with [Section 4.4 of RFC5425](https://tools.ietf.org/html/rfc5425#section-4.4), though.

[...]

Issue #1231, opened 2016-10-30
rsyslog hangs when TLS close_notify goes unanswered


markdascher:
When shutting down after forwarding logs to a TLS target, rsyslog sends a TLS **close_notify**, then waits for a **close_notify** in response.  If the remote end never answers, rsyslog hangs until the remote end closes the TCP connection.  This behavior is compliant with the syslog spec, but a timeout would be helpful.

Repro steps:
* Configure a dummy TLS server using **openssl**.  Use **nc** as a proxy, to prevent rsyslog from seeing the TCP connection close.  (`openssl s_server` won't respond with a **close_notify**, but will close the TCP connection.)
```
openssl req -x509 -newkey rsa:2048 -keyout server.pem -out cert.pem -days 365 -nodes
sleep inf | openssl s_server -key server.pem -cert cert.pem -quiet &
mkfifo back
nc -l 6514 <back | nc localhost 4433 >back &
```
* Configure rsyslog to forward logs to the dummy TLS server.
```
$DefaultNetstreamDriverCAFile /home/.../cert.pem
$ActionSendStreamDriver gtls
$ActionSendStreamDriverMode 1
$ActionSendStreamDriverAuthMode anon
syslog.=info @@localhost:6514
```
* Start rsyslog, and then use `logger` to send messages.  They'll appear on STDOUT.
* Try to kill rsyslog: `sudo pkill -TERM rsyslog`

Affected code:
* https://github.com/rsyslog/rsyslog/blob/master/runtime/nsd_gtls.c#L1149-L1152

Easy fix:
* Change both instances of `GNUTLS_SHUT_RDWR` to `GNUTLS_SHUT_WR`.
* This may technically be vulnerable to a truncation attack, but so is the existing code.  (Since `gnuRet` isn't fully checked, rsyslog happily proceeds without hearing a **close_notify**, as long as the TCP connection is closed.)  On the other hand, syslog receivers don't normally send anything to the transmitter, making the idea of a truncation attack somewhat meaningless.

Hard fix:
[...]

PR #3565, opened 2019-03-25
openssl: Added better error handling in osslRecordRecv

alorbach:
Handling of SSL_ERROR_SYSCALL has been hardened.

Also fixed SSL Shutdown handling in tcpflood (openssl code).
If SSL_Shutdown returns error, we call SSL_read as described in
the documentation to do a bidirectional shutdown.

Closes https://github.com/rsyslog/rsyslog/issues/3561

### DIFF
--- a/runtime/nsd_ossl.c
+++ b/runtime/nsd_ossl.c
@@ -196,17 +196,18 @@ void osslLastSSLErrorMsg(int ret, SSL *ssl, int severity, const char* pszCallSou
 	int iSSLErr = 0;
 	if (ssl == NULL) {
 		/* Output Error Info*/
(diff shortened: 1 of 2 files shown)


rgerhards:
> If SSL_Shutdown returns error, we call SSL_read as described in
> the documentation to do a bidirectional shutdown.

@alorbach pls have a look at #1503  and #1231 - doesn't this solve that situation as well (for openssl)?

alorbach:
@rgerhards: As the shutdown handling is corrected for SSL_ERROR_SYSCALL and other unexpected errors in osslRecordRecv, this could properly fix #1503 and #1231 as well.

rgerhards:
ok, than I also ask the folks in #1231, #1503 to check the new version once merged

PR #3566, opened 2019-03-25
Fetching journal cursor only for valid journal

jvymazal:
The sd_journal_get_cursor() got called regradless of previous
retcodes from other jorunal calls which flooded logs with journald
errors. Now skipping the call in case of previous journal call
non-zero result.

Also, making WorkAroundJournalBug true by default, as there were no
confirmed performance regressions for a quite long time.

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -135,7 +135,7 @@ static char *last_cursor = NULL;
 
 #define J_PROCESS_PERIOD 1024  /* Call sd_journal_process() every 1,024 records */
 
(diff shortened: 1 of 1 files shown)


jvymazal:
@rgerhards just a note, code in `CODESTARTafterRun` macro is executed after `CODESTARTfreeCnf`?

jvymazal:
Ok, finally found the last extra free(), we should be good, so merge whenever you are ready. BTW the imjournal CI enabled on the 7.5 CentOS builder seems to work just right, as it failed till I sorted the memory mismatch, thanks to keep it running.

PR #3568, opened 2019-03-26
testbench: enable ompipe test on Solaris again and modernize it

rgerhards:


rgerhards:
Looks like the problem no longer exists...

Issue #3570, opened 2019-03-26
Some tests are failing on Solaris SPARC


kulikjak:
Hi,

since 8.1903.0, we have four tests fail on SPARC version of Solaris. Those tests are:

- diskqueue.sh
- rsf_getenv.sh
- queue-persist.sh
- rcvr_fail_restore.sh

Since all of them seem to have the same problem, I debugged **diskqueue.sh** only.

All tests fail during `seq_check` and the reason is, that output log contains `00` lines (sometimes only those, sometimes there are also correct lines like e.g. `00000012`). Since `seq_check` expects them to be strictly increasing, it fails (but that is not the problem).

I added a debug print into tcpfood right before send to check whether problem is there, but messages seems to be ok:

    <167>Mar  1 01:00:00 172.20.245.8 tag msgnum:00000000:
    <167>Mar  1 01:00:00 172.20.245.8 tag msgnum:00000001:
    ....

It looks a little bit like rsyslog sometimes saves only first two digits and incorrectly cuts out rest, but that is just my wild guess. Sadly I don't know where to look / what to try next.

All tests are ok on intel. Also, we did not saw this before 8.1903.0.

kulikjak:
Oh, I forgot to attach full logs:

[diskqueue.log](https://github.com/rsyslog/rsyslog/files/3008661/diskqueue.log)
[example.out.log](https://github.com/rsyslog/rsyslog/files/3008664/example.out.log)

mbiebl:
I think I ran into the same issue on Debian, see e.g.
https://buildd.debian.org/status/fetch.php?pkg=rsyslog&arch=sparc64&ver=8.1903.0-2&stamp=1553601459&raw=0
(it also failed on various other architectures like mips, s390x, ppc64, ...)

A git bisect pointed at 1255a67fdec2fc44cd49b6ea8c463f4319910812 as the first faulty commit and it helped reverting that commit which I did in https://salsa.debian.org/debian/rsyslog/commit/976a79dfef385a11981965e11a3fe510ed17524c

The result looks much better now
https://buildd.debian.org/status/package.php?p=rsyslog&suite=experimental

Could you try if reverting 1255a67fdec2fc44cd49b6ea8c463f4319910812 works for you as well?

rgerhards:
[...]

PR #3574, opened 2019-03-28
try to fix regression from 1255a67fdec2fc44cd49b6ea8c463f4319910812

rgerhards:
closes https://github.com/rsyslog/rsyslog/issues/3570


jvymazal:
Thanks for the effort, you were much faster then me in getting to solve this.

rgerhards:
> Thanks for the effort, you were much faster then me in getting to solve this.

at least sometimes I am faster ;) I had to fix the commit message, so let's give CI another run.

Issue #3573, opened 2019-03-28
Wrong semantics in omelasticsearch createMsgFromRequest()


snaix:
### Expected behavior
As createMsgFromRequest() comment said in plugins/omelasticsearch.c :

>request string looks like this:
>"{\"create\":{\"_index\": \"rsyslog_testbench\",\"_type\":\"test-type\",
>\"_id\":\"FAEAFC0D17C847DA8BD6F47BC5B3800A\"}}\n
>{\"msgnum\":\"x00000000\",\"viaq_msg_id\":\"FAEAFC0D17C847DA8BD6F47BC5B3800A\"}\n"
>**we don't want the meta header, only the data part**
>start = first \n + 1
>end = last \n 

An expected behavior is rsyslog can get rawmsg as data part.

### Actual behavior
Only message format that contains "message" filed can be extracted from request, but not for custom message format.
If there is no message field, all request text will be filled for rawmsg.
It's confusing when use omelasticsearch retry. user can not get a correct rawmsg from rsyslog, and could lead a data lost because a wrong json message(contain meta header) would be sent to ES.

### Environment
The latest code shows that this problem still exists.

davidelang:
can you please show a message written with the RSYSLOG_DebugFormat that has the 
problem you are describing?

I am not understanding the issue.

David Lang

snaix:
> can you please show a message written with the RSYSLOG_DebugFormat that has the problem you are describing? I am not understanding the issue. David Lang

For instance:
I have a ruleset A for a normal omelasticsearch action, and another ruleset B for retryRuleset when  omelasticsearch failed for 429 status .
Obviously, the original, failed message  from ruleset A is needed in ruleset B .
Current, ruleset B extract the original message from elasticsearch reqeuset like:

> "{"create":{"_index": "rsyslog_testbench","_type":"test-type",
> "_id":"FAEAFC0D17C847DA8BD6F47BC5B3800A"}}\n
> {"msgnum":"x00000000","viaq_msg_id":"FAEAFC0D17C847DA8BD6F47BC5B3800A"}\n"

The original message is the data part, we needed:
> {"msgnum":"x00000000","viaq_msg_id":"FAEAFC0D17C847DA8BD6F47BC5B3800A"}\n"

But only data part contains "message" field can be extracted, or omelasticsearch will keep the whole request in rawmsg.
so we get a rawmsg like this:
> "{"create":{"_index": "rsyslog_testbench","_type":"test-type",
[...]

PR #3577, opened 2019-03-29
omelasticsearch - retry: set rawmsg to data from original request

richm:
https://github.com/rsyslog/rsyslog/issues/3573
Previously, when constructing the message to submit for a retry
for an original request, if the original request did not contain
the field `message`, the system property `rawmsg` was set to
the entire metadata + data from the original request.  This was
causing problems with Elasticsearch.  This patch changes
the code so that the `rawmsg` will be set to only the data part
of the original request if there is no `message` field.

### DIFF
--- a/plugins/omelasticsearch/omelasticsearch.c
+++ b/plugins/omelasticsearch/omelasticsearch.c
@@ -1138,7 +1138,8 @@ createMsgFromRequest(const char *request, context *ctx, smsg_t **msg, fjson_obje
 		const size_t msgLen = (size_t)json_object_get_string_len(jo_msg);
 		MsgSetRawMsg(*msg, rawmsg, msgLen);
 	} else {
-		MsgSetRawMsg(*msg, request, strlen(request));
+		/* use entire data part of request as rawmsg */
+		MsgSetRawMsg(*msg, datastart, datalen);
 	}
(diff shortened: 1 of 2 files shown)


davidelang:
rawmsg should always be what arrived over the wire, which includes any 
headers/metadata that's part of the message.

David Lang

richm:
@davidelang @snaix Then I'm not sure what to do in this case - @snaix - is the problem that there is an embedded newline in the rawmsg e.g. it looks like this? `{"metadata":.}\n{"message":....}`?  And that causes problems?  I guess I need more information about the actual failure.  @snaix perhaps you could provide your rsyslog config that exhibits the problem?

rgerhards:
> @davidelang @snaix Then I'm not sure what to do in this case - @snaix - is the problem that there is an embedded newline in the rawmsg e.g. it looks like this? `{"metadata":.}\n{"message":....}`? And that causes problems? I guess I need more information about the actual failure. @snaix perhaps you could provide your rsyslog config that exhibits the problem?

Holding merge until this is settled. @snaix  @davidelang pls comment.

rgerhards:
If I don't hear otherwise by Monday, April 8th 2019, I'll merge.

PR #3589, opened 2019-04-04
Revert "WiP: rsyslog doesn't respond to TLS close_notify "

### LINKED ISSUE #3587: openssl: check for random SSL_ERROR_SYSCALL errors on fedora 29
When build machines are on load, openssl tests appear to fail random with SSL_ERROR_SYSCALL on fedory 29, sample buildlog:
https://build.rsyslog.com/#/builders/197/builds/99/steps/4/logs/test-suite_log

rgerhards:
Reverts rsyslog/rsyslog#3535

This actually caused a lot of problems, including https://github.com/rsyslog/rsyslog/issues/3587

The main issue is that the errno check seems actually be disabled. This might look like it worked, because it will frequently close connection. This "fixes" the original problem because we get lots and lots of new connections.

fixes #3587

### DIFF
--- a/runtime/nsd_ptcp.c
+++ b/runtime/nsd_ptcp.c
@@ -896,12 +896,12 @@ CheckConnection(nsd_t *pNsd)
 {
 	DEFiRet;
 	int rc;
-	char msgbuf[32]; /* dummy */
+	char msgbuf[1]; /* dummy */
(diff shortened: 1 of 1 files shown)


rgerhards:
CI fail is unrelated, so ready to merge

PR #3598, opened 2019-04-05
testbench: Add missing omhttp tests to EXTRA_DIST

Whissi:
Looks like nobody tried to run tests since last release against release tarball :-0


rgerhards:
> Looks like nobody tried to run tests since last release against release tarball :-0

This must be another problem - the testbench runs `make distcheck`. I'll keep this open to investigate...

rgerhards:
root cause found and fixed in #3604 
thx again for this PR!

Issue #3599, opened 2019-04-05
omhttp fails tests with Python 3


Whissi:
All omhttp tests are failing on Gentoo. Example output:
```
testbench: TZ env var not set, setting it to UTC
------------------------------------------------------------
20:13:59[0]  Test: ./omhttp-basic.sh
------------------------------------------------------------
Started omhttp test server with args -p 46977  with pid 36048
config rstb_317676_dbd2dc75_.conf is:
     1	module(load="../plugins/imdiag/.libs/imdiag")
     2	global(inputs.timeout.shutdown="60000"
     3	       default.action.queue.timeoutshutdown="20000"
     4	       default.action.queue.timeoutEnqueue="20000")
     5	$MainmsgQueueTimeoutEnqueue 20000 # use legacy-style so that we can override if needed
     6	$IMDiagListenPortFileName rstb_317676_dbd2dc75.imdiag.port
     7	$IMDiagServerRun 0
     8	$IMDiagAbortTimeout 580
     9	
    10	:syslogtag, contains, "rsyslogd"  ./rstb_317676_dbd2dc75.started
    11	###### end of testbench instrumentation part, test conf follows:
    12	
    13	template(name="tpl" type="string"
    14		 string="{\"msgnum\":\"%msg:F,58:2%\"}")
    15	
    16	module(load="../contrib/omhttp/.libs/omhttp")
    17	
    18	if $msg contains "msgnum:" then
    19		action(
    20			# Payload
    21			name="my_http_action"
    22			type="omhttp"
    23			errorfile="rstb_317676_dbd2dc75/omhttp.error.log"
    24			template="tpl"
    25	
    26			server="localhost"
    27			serverport="46977"
    28			restpath="my/endpoint"
    29			batch="off"
    30	
    31			# Auth
    32			usehttps="off"
    33	    )
rsyslogd: NOTE: RSYSLOG_DEBUG_TIMEOUTS_TO_STDERR activated
main Q:Reg: worker start requested, num workers currently 0
main Q:Reg: wrkr start initiated with state 0, num workers now 1
rsyslog debug: main Q:Reg: worker 0x7f2787c43b80 started
rsyslog debug: main Q:Reg: started with state 3, num workers now 1
20:13:59[0]  rstb_317676_dbd2dc75:.pid found, pid 36056
20:13:59[0]  rsyslogd startup msg seen, pid  36056
imdiag port: 35629
injecting 10000 messages
[...]

PR #3600, opened 2019-04-05
tests: omhttp: add Python 3 compatibility

Whissi:
Fixes: https://github.com/rsyslog/rsyslog/issues/3599

### DIFF
--- a/tests/diag.sh
+++ b/tests/diag.sh
@@ -1334,7 +1334,7 @@ get_inode() {
 		printf 'FAIL: file "%s" does not exist in get_inode\n' "$1"
 		error_exit 100
 	fi
-	python -c 'import os; import stat; print os.lstat("'$1'")[stat.ST_INO]'
+	python -c 'import os; import stat; print(os.lstat("'$1'")[stat.ST_INO])'
 }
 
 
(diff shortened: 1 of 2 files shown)


rgerhards:
@Whissi Thx for the PR. To detect these things in the future immediately: can you suggest how I need to change one of the existing environments (or create a new one) so that Python 3 is default? I know I could use gentoo, but you know I don't know how to set it up ;-) -- or is it easier with containers?

Whissi:
We do have https://github.com/gentoo/gentoo-docker-images.
You still have to switch Python default to Py3 I think and we only publish "stable" stage3 images yet (so you don't get latest bits).

But you should be able to do the same with Debian/Ubuntu for example:

1. Make sure Python 3 is installed, `apt install python3`
2. Switch default Python using `update-alternatives --config python` (you have to 'install' alternatives first)

Keep in mind that this might break the system if packages like apt will only work with system's original Python version. In this case I recommend to do all apt calls first (update, install deps...) and just before starting to build rsyslog, manually change `/usr/bin/python` symlink.

rgerhards:
ok, changing the default does not work - it breaks buildbot. I wonder why it does, because they support python 3, but I guess that doesn't really help ;-)

rgerhards:
@Whissi will merge now as release is immanent. Nevertheless, would you be willing to help us create a CI container based on the gentoo-provided one? When we do it together, it should not be too hard. We probably need to a new issue if we take that route.

Issue #3334, opened 2018-12-09
emit at least a warning message for ultra-large queues


rgerhards:
We see error reports from users who have configured excessively large queues and receive an OOM condition. So we should at least warn if a queue is configured very large (e.g. greater than 100,000 messages).

For an example see https://github.com/rsyslog/rsyslog/issues/3314

davidelang:
One reason for this is that people have no idea how much space a message takes 
in a queue.

How hard would it be to add per-queue stats about memory used? (current and 
peak, just like size and maxsize)

David Lang

rgerhards:
Pretty hard, we would need to count each and every  allocation plus guess the malloc overhead. More complicated with libs like json.

Issue #3314, opened 2018-11-23
memory leak


amuhametov:
Hi.
Got memory leaks with kafka.
<img width="1066" alt="screenshot 2018-11-23 at 15 40 44" src="https://user-images.githubusercontent.com/652693/48944029-27cfc380-ef36-11e8-9e5a-53b00a060dfa.png">

```
[Wed Nov 21 10:42:11 2018] rsyslogd invoked oom-killer: gfp_mask=0x201da, order=0, oom_score_adj=0
[Wed Nov 21 10:42:11 2018] rsyslogd cpuset=/ mems_allowed=0-1
[Wed Nov 21 10:42:11 2018] CPU: 7 PID: 29241 Comm: rsyslogd Kdump: loaded Not tainted 3.10.0-862.6.3.el7.x86_64 #1
[Wed Nov 21 10:42:11 2018] Hardware name: ETegro Technologies Hyperion RS125 G4/ETRS125G4, BIOS S2MS3C01 04/24/2014
[Wed Nov 21 10:42:11 2018] Call Trace:
[Wed Nov 21 10:42:11 2018]  [<ffffffff94f0e80e>] dump_stack+0x19/0x1b
[Wed Nov 21 10:42:11 2018]  [<ffffffff94f0a190>] dump_header+0x90/0x229
[Wed Nov 21 10:42:11 2018]  [<ffffffff948f7ed2>] ? ktime_get_ts64+0x52/0xf0
[Wed Nov 21 10:42:11 2018]  [<ffffffff9494e8df>] ? delayacct_end+0x8f/0xb0
[Wed Nov 21 10:42:11 2018]  [<ffffffff949980c4>] oom_kill_process+0x254/0x3d0
[Wed Nov 21 10:42:11 2018]  [<ffffffff94997b6d>] ? oom_unkillable_task+0xcd/0x120
[Wed Nov 21 10:42:11 2018]  [<ffffffff94997c16>] ? find_lock_task_mm+0x56/0xc0
[Wed Nov 21 10:42:11 2018]  [<ffffffff94998906>] out_of_memory+0x4b6/0x4f0
[Wed Nov 21 10:42:11 2018]  [<ffffffff9499f224>] __alloc_pages_nodemask+0xaa4/0xbb0
[Wed Nov 21 10:42:11 2018]  [<ffffffff949e8ce8>] alloc_pages_current+0x98/0x110
[...]

PR #3630, opened 2019-04-18
core: emit a warning message for ultra-large queue size definitions

rgerhards:
We see error reports from users who have configured excessively large queues
and receive an OOM condition or other problems.

With that patcjh we generate a warnomg message if a queue is configured very
large. "Very large" is defined to be in excess of 500000 messages.

see also https://github.com/rsyslog/rsyslog/issues/3314
closes https://github.com/rsyslog/rsyslog/issues/3334

### DIFF
--- a/runtime/queue.c
+++ b/runtime/queue.c
@@ -82,6 +82,8 @@ DEFobjCurrIf(statsobj)
 unsigned int iOverallQueueSize = 0;
 #endif
 
+#define OVERSIZE_QUEUE_WATERMARK 500000 /* when is a queue considered to be "overly large"? */
+
 /* overridable default values (via global config) */
 int actq_dflt_toQShutdown = 10;		/* queue shutdown */
 int actq_dflt_toActShutdown = 1000;	/* action shutdown (in phase 2) */
@@ -3250,6 +3252,12 @@ qqueueApplyCnfParam(qqueue_t *pThis, struct nvlst *lst)
 			}
 		} else if(!strcmp(pblk.descr[i].name, "queue.size")) {
 			pThis->iMaxQueueSize = pvals[i].val.d.n;
+			if(pThis->iMaxQueueSize > OVERSIZE_QUEUE_WATERMARK) {
+				parser_errmsg("queue.size=%d is very large - is this "
+					"really intended? More info at "
+					"https://www.rsyslog.com/avoid-overly-large-in-memory-queues/",
+					pThis->iMaxQueueSize);
+			}
 		} else if(!strcmp(pblk.descr[i].name, "queue.dequeuebatchsize")) {
 			pThis->iDeqBatchSize = pvals[i].val.d.n;
 		} else if(!strcmp(pblk.descr[i].name, "queue.mindequeuebatchsize")) {
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -209,6 +209,7 @@ TESTS +=  \
 	tcp_forwarding_tpl.sh \
 	tcp_forwarding_dflt_tpl.sh \
 	tcp_forwarding_retries.sh \
+	queue_errmsg-oversize.sh \
 	queue-minbatch.sh \
 	queue-minbatch-queuefull.sh \
 	arrayqueue.sh \
@@ -1449,6 +1450,7 @@ EXTRA_DIST= \
 	tcp_forwarding_ns_tpl.sh \
 	tcp_forwarding_dflt_tpl.sh \
 	tcp_forwarding_retries.sh \
+	queue_errmsg-oversize.sh \
 	queue-minbatch.sh \
 	queue-minbatch-queuefull.sh \
 	killrsyslog.sh \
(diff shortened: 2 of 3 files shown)


davidelang:
It would be good if we could make the error message say something along the 
lines of "this queue size would take up >XXGB of ram if it fills up" rather than 
just something like "this is an unsually large queue size, are you sure?"

( I know the queue size isn't deterministic as it depends on what is parsed into 
variables, but if we assume that every message is $maxmessagesize long and fills 
up the queues, that should give us a reasonable baseline to warn from )

David Lang

rgerhards:
> ( I know the queue size isn't deterministic as it depends on what is parsed into variables, but if we assume that every message is $maxmessagesize long and fills up the queues, that should give us a reasonable baseline to warn from )

This may be totally of the real value, so I am hesitant to put any KiB value in here. I had just written an article that I intended to link to, but while saving the server made it disappear (grrr....)

rgerhards:
oh... the posting made it to the database. Here is the first version: https://www.rsyslog.com/avoid-overly-large-in-memory-queues/ Text improvements are welcome (I can also generate an account for you if you like ;-)).

davidelang:
On Thu, 18 Apr 2019, Rainer Gerhards wrote:

>> ( I know the queue size isn't deterministic as it depends on what is parsed into variables, but if we assume that every message is $maxmessagesize long and fills up the queues, that should give us a reasonable baseline to warn from )
>
> This may be totally of the real value, so I am hesitant to put any KiB value in here. I had just written an article that I intended to link to, but while saving the server made it disappear (grrr....)

what's in the message object? going from memory (several years old)

* msg (up to $maxmessagesize)
* rawmsg (up to $maxmessagesize)
* standard properties (is a couple hundred bytes reasonable?)
* a bit of other overhead (roll into a constant with the standard properties)
* variables (as three json strings, which could get very large)

we can't predict the variables, but we should be able to say that if every 
message were $maxmessagesize, the queue will take up X Gb of ram, and could take 
more depending on your config and what parsing has happened at the time the 
message is put into the queue.

some sort of guidelines along these lines would be helpful to put into the 
article (if not the message)

David Lang

davidelang:
On Thu, 18 Apr 2019, Rainer Gerhards wrote:

> oh... the posting made it to the database. Here is the first version: 
> https://www.rsyslog.com/avoid-overly-large-in-memory-queues/ Text improvements 
> are welcome (I can also generate an account for you if you like ;-)).

probably a good idea to make me an account.

David Lang

rgerhards:
@davidelang account creation is on the way (see mail).

I am merging as-is, but am open to further changes. I just doubt that a message like *"your queue could use 2000MiB or more"* is really that helpful. Do you think it is? Again, it's a very incorrect number for a regular log message mix...

Nicheurity:
We need ability to use very large queue size along with 'global(AbortOnUncleanConfig="on")'.
Before upgrading to  8.1905, there was no problem and the very large memory queue are intended. 

If this is a _warning_ message for overly-large-in-memory-queues, then I wouldn't expect rsyslog to fail to start, showing errors with 'config is not clean' message. Setting the global abort on unclean config to off introduces too much risk that syntax errors might not be caught and running in an unstable state. 

It is necessary to prevent rsyslog from starting from when there are legitimate errors, and prevent false alarms when configured with very large memory queues.

Please consider _veryLarge.queue="on"_ or some other parameter that would suppress the related 'warning/error' messages, as these custom configurations are setup as intended. 

This purposed update can resolve both issues:
- 1> Remove false positive errors for intended very large queues. The errors are seen during startup and configuration validation, but should show the config is 'ok'.
- 2> Reduce risk of running on dirty configurations that include very large memory queues, by fixing compatibility with abort on unclean config = on.

davidelang:
I think we've already had a patch to reduce this from an error to a warning.

David Lang

Nicheurity:
Thank you for the prompt response - when will the patch be packaged (rpm)? 

Also, I think it would be very helpful to have rsyslog assist with recommended queue sizing as described in the comments. Having dedicated syslog servers with a lot of hardware resources it is challenging to determine sizing for all the rulesets, according to impstats/dynstats for actual utilization (increase and reduction). Maybe this can generated from a -N5 config validation run-level?

davidelang:
On Sun, 9 Jun 2019, Nicheurity wrote:

> Thank you for the prompt response - when will the patch be packaged (rpm)?

it will be in the 8.1907 release

> Also, I think it would be very helpful to have rsyslog assist with recommended 
> queue sizing as described in the comments. Having dedicated syslog servers 
> with a lot of hardware resources it is challenging to determine sizing for all 
> the rulesets, according to impstats/dynstats for actual utilization (increase 
> and reduction). Maybe this can generated from a -N5 config validation 
> run-level?

part of the problem is that the memory used depends a lot on how much you make 
use of variables, that can multiply your memory usage by 2-3x easily (and not 
hard to go much higher)

David Lang

PR #3640, opened 2019-04-30
fix cosmetic coverity leak report (CID 207601)

rgerhards:
While in theory this is a leak, in practice it's extremely unlikely
and irrelevant. Nevertheless, let's fix it ;-)


rgerhards:
Not sure why CodeCov reports just 20% - seems false positive...

Issue #1385, opened 2017-02-02
check that queue same file name is given only for one queue


rgerhards:
If the same name is specified for multiple queues, the queue files will become corrupted. Currently this is not checked, and a check should be added for obvious reasons. This should err-out during config parsing.

davidelang:
we need a similar check for actions with the same name. It doesn't corrupt 
things if they have the same name, but it does mess up the stats.

rgerhards:
Right - could you add an issue?

Sent from phone, thus brief.

Am 02.02.2017 10:44 vorm. schrieb "davidelang" <notifications@github.com>:

PR #3644, opened 2019-05-02
config processing: check disk queue file is unique

rgerhards:
If the same name is specified for multiple queues, the queue files
will become corrupted. This commit adds a check during config parsing.
If duplicate names are detected the config parser errors out and the
related object is not created.

Note: this may look to a change-of-behaviour to some users. However,
this never worked and it was pure luck that these users did not run
into big problems (e.g. DA queues were never going to disk at the
same time). So it is acceptable to error out in this hard error case.

closes https://github.com/rsyslog/rsyslog/issues/1385

### DIFF
--- a/runtime/queue.c
+++ b/runtime/queue.c
@@ -149,6 +149,14 @@ static struct cnfparamblk pblk =
 	  cnfpdescr
 	};
 
+/* support to detect duplicate queue file names */
+struct queue_filename {
+	struct queue_filename *next;
+	const char *dirname;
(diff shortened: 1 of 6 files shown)


rgerhards:
note: also flagged as "bug" as one might consider not emitting an error on a fatal condition could be considered a bug.

wadey:
I think this change introduced a segfault in https://github.com/rsyslog/rsyslog/issues/3681

Issue #3639, opened 2019-04-29
RFE: Mention external lookup table use on reload and restart


atc0005:
### Expected behavior

rsyslog mentions that it is referencing a lookup table on reload and restart.

### Actual behavior

rsyslog mentions external lookup table file(s) upon reload only.

### Steps to reproduce the behavior

Here is a reference to the lookup table being used when reloading rsyslog:

```
2019-04-29T11:34:41.842584-05:00 sawmill1 rsyslogd: [origin software="rsyslogd" swVersion="8.1904.0" x-pid="841" x-info="https://www.rsyslog.com"] rsyslogd was HUPed
2019-04-29T11:34:41.842837-05:00 sawmill1 rsyslogd: lookup table 'ip_hostname_mappings' reloaded from file '/etc/rsyslog/common/ip_hostname_mappings.json' [v8.1904.0 try https://www.rsyslog.com/e/0 ]
```

A useful enhancement (e.g., for troubleshooting purposes) would be to mention use of such files when reloading (as is already being done) *and* when restarting rsyslog (not currently being done).

### Environment
- rsyslog version: 8.1904.0
- platform: Ubuntu 16.04

rgerhards:
well... I think it is absolutely clear that when you stop and start rsyslog it will load the lookup table on start. Do we **really** need to log a message each time we start?

atc0005:
> Do we **really** need to log a message each time we start?

I don't think there is a *need* for this, but I believe it could be a useful enhancement.This would provide an indication (aside from the results of using the file) that the file was indeed read and that no issues were encountered. In other words, this would be a useful "checkbox" of sorts to use when troubleshooting to confirm that not having a mapping available for use by rsyslog is not due to a case of the file not being read.

rgerhards:
@davidelang @mbiebl @Whissi @richm what do you think?

davidelang:
logging on reload makes sense as you are changing the behavior (new file)

At startup, it's much less clear. The "Unix Way" is to say nothing when things are working, only log when something has gone wrong, so a failed load of the file needs to be logged. I'm far less confident that a successful load needs to be logged. If it is logged, it should be at a very log severity (debug)

I don't see that logging it hurts much, we are only talking about one line at startup.

If we do log such successful events at startup, are there other things we will need to log as well? (how far does this slippery slope go?)

so I would lean towards not logging at startup, but wouldn't mind if it did.

davidelang:
[...]

PR #3650, opened 2019-05-07
new global config parameter "internalmsg.severity"

rgerhards:
permits to specify a severity filter for internal message. Only
messages with this severity level or more severe are logged.

Orignally this was done in rsyslog.conf as usual: you can filter
rsyslog messages on severity, just like any other. But with systemd,
we now emit primarily to the journal, and this is outside of rsyslog's
rule engine and so regular filters do not apply (at least in regard
to the journal). Logging to journal is good, because finally
folks begin to see the messages (traditional distro configs discard
them, for whatever is the reason).

This commit implements a global setting for a severity-based filter
for internal messages, before submitted to journal. So it's not 100%
of what rsyslog can do, but at least some way to customize.

see also https://github.com/rsyslog/rsyslog/issues/3639

### DIFF
--- a/runtime/glbl.c
+++ b/runtime/glbl.c
@@ -82,6 +82,8 @@ stdlog_channel_t stdlog_hdl = NULL;	/* handle to be used for stdlog */
 #endif
 
 static struct cnfobj *mainqCnfObj = NULL;/* main queue object, to be used later in startup sequence */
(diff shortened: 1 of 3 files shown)


rgerhards:
CodeCov is acceptable

Issue #3667, opened 2019-05-16
Leading space is lost on message when using RFC5424, SD is set and message starts with space


StrongestNumber9:
### Expected behavior
Space is reserved when structured data is used and message starts with space

### Actual behavior
Leading space is lost as seen in the 'sd-sp' part
```sh
# jq -r '.msg' /debug.log
nosd-nosp
sd-nosp
 nosd-sp
sd-sp

# cat debug.log
{ "msg": "nosd-nosp", "rawmsg": "<13>1 2019-05-15T11:21:57+03:00 domain.tld tag - - - nosd-nosp", "timereported": "2019-05-15T11:21:57+03:00", "hostname": "domain.tld", "syslogtag": "tag", "inputname": "imrelp", "fromhost": "localhost", "fromhost-ip": "127.0.0.1", "pri": "13", "syslogfacility": "1", "syslogseverity": "5", "timegenerated": "2019-05-16T09:53:11.986104+03:00", "programname": "tag", "protocol-version": "1", "structured-data": "-", "app-name": "tag", "procid": "-", "msgid": "-", "uuid": null, "$!": null }
{ "msg": "sd-nosp", "rawmsg": "<13>1 2019-05-15T11:21:57+03:00 domain.tld tag - - [abc@123 a=\"b\"] sd-nosp", "timereported": "2019-05-15T11:21:57+03:00", "hostname": "domain.tld", "syslogtag": "tag", "inputname": "imrelp", "fromhost": "localhost", "fromhost-ip": "127.0.0.1", "pri": "13", "syslogfacility": "1", "syslogseverity": "5", "timegenerated": "2019-05-16T09:53:12.006722+03:00", "programname": "tag", "protocol-version": "1", "structured-data": "[abc@123 a=\"b\"]", "app-name": "tag", "procid": "-", "msgid": "-", "uuid": null, "$!": null }
[...]

PR #3672, opened 2019-05-17
testbench: test for leading space issue in RFC 5424 parser

rgerhards:
see also https://github.com/rsyslog/rsyslog/pull/3671
see also https://github.com/rsyslog/rsyslog/pull/3667


rgerhards:
problem was confirmed with unpatched version of rsyslog. current master is patched, so this PR should succeed.

@StrongestNumber9 thx!

Issue #3662, opened 2019-05-14
 Duplicate logs appear when the log is printed continuously


wsp1991:
### Expected behavior
logs not duplicate
### Actual behavior
"testcases 743343" log duplicate
May 03 17:43:10 localhost.localdomain root[23110]: print by testcases 743341
May 03 17:43:10 localhost.localdomain root[23123]: print by testcases 743342
May 03 17:43:10 localhost.localdomain root[23156]: print by testcases 743343
May 03 17:43:10 localhost.localdomain root[23156]: print by testcases 743343
May 03 17:43:10 localhost.localdomain root[23159]: print by testcases 743344
May 03 17:43:10 localhost.localdomain root[23160]: print by testcases 743345
May 03 17:43:10 localhost.localdomain root[23163]: print by testcases 743346
May 03 17:43:10 localhost.localdomain root[23168]: print by testcases 743347
May 03 17:43:11 localhost.localdomain root[23179]: print by testcases 743348
May 03 17:43:11 localhost.localdomain root[23188]: print by testcases 743349
### Steps to reproduce the behavior
Execute the following script；
test code(shell)
while true
do
         logger -p local7.info "print by testcases $i'
         ((i=i+1))
done

### Environment
- rsyslog version: 
- platform: 
- for configuration questions/issues, include rsyslog.conf and included config files

wsp1991:
“Fetching journal cursor only for valid journal” this patch has a issue

		if (cs.stateFile) {
		/* If we have locally saved cursor there is no need to read it from state file */
		if (cs.bWorkAroundJournalBug && last_cursor)
		{
			if (sd_journal_seek_cursor(j, last_cursor) != 0) {------>**Here should get the next position**
				LogError(0, RS_RET_ERR, "imjournal: "
					"couldn't seek to cursor `%s'\n", last_cursor);
			iRet = RS_RET_ERR;
			}
		}
		else if (cs.stateFile) {

rgerhards:
@jvymazal could you have a look?

rgerhards:
@wsp1991 rsyslog version? platform?

jvymazal:
I will verify this, looks like with next imjournal update I will need to craft some test for duplicates.

rgerhards:
@jvymazal thx a lot!

rgerhards:
[...]

PR #3675, opened 2019-05-20
Advance cursor after reloading journal

jvymazal:
This fixes situation when journal was preloaded from previously
saved cursor but was not advanced to next entry so reading begun
from last message which was therefore duplicated.

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -545,8 +545,10 @@ pollJournal(void)
 			if (sd_journal_seek_cursor(j, last_cursor) != 0) {
 				LogError(0, RS_RET_ERR, "imjournal: "
 					"couldn't seek to cursor `%s'\n", last_cursor);
-			iRet = RS_RET_ERR;
+				iRet = RS_RET_ERR;
 			}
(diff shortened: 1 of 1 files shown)


jvymazal:
closes #3662

jvymazal:
Ok, found out the issue, the journal actually needs TWO sd_journal_next() calls after re-opening one to position the jorunal at cursor and second to actually advance the entry (to unread one in our case)

jvymazal:
@rgerhards do you think we can get it to 1906 please?

rgerhards:
> do you think we can get it to 1906 please?

It's technically impossible - we have no 1906 release ;-) We release every 6 weeks, so sometimes we skip a month. Next release is July, 9th... and by that I see that @friedl accidently "created" 1906 on the web site.

jvymazal:
oh, ok, 1907 it is then. Thanks for info.

rgerhards:
Web site is now fixed. Good we noticed it :-) I think the "skip some month" theme is sometimes confusing, but I neither want to go to monthly nor to bi-monthly releases, so that's the best compromise we have.

Issue #3681, opened 2019-05-28
Segfault on 8.1905.0 with multiple queues with filename set


wadey:
### Expected behavior

No segfault

### Actual behavior

Segfault, valgrind stack trace:

```
Invalid read of size 8
   at 0x1815EF: checkUniqueDiskFile (queue.c:3276)
   by 0x1823E0: qqueueApplyCnfParam (queue.c:3406)
   by 0x11F62E: createMainQueue (rsyslogd.c:804)
   by 0x14FDCA: activateMainQueue (rsconf.c:809)
   by 0x14FF22: activate (rsconf.c:878)
   by 0x120E77: initAll (rsyslogd.c:1623)
   by 0x1219FF: main (rsyslogd.c:2120)
 Address 0x6d08d10 is 16 bytes inside a block of size 24 free'd
   at 0x4C30D3B: free (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
   by 0x179935: qqueueDoneLoadCnf (queue.c:187)
   by 0x14F5EC: tellCoreConfigLoadDone (rsconf.c:638)
   by 0x151863: load (rsconf.c:1344)
   by 0x120BBC: initAll (rsyslogd.c:1541)
   by 0x1219FF: main (rsyslogd.c:2120)
 Block was alloc'd at
   at 0x4C31B25: calloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
   by 0x181677: checkUniqueDiskFile (queue.c:3287)
   by 0x1823E0: qqueueApplyCnfParam (queue.c:3406)
   by 0x18BB16: actionConstructFinalize (action.c:568)
   by 0x190265: addAction (action.c:2119)
   by 0x19051A: actionNewInst (action.c:2196)
   by 0x143FB3: cnfstmtNewAct (rainerscript.c:4420)
   by 0x132A40: yyparse (grammar.y:191)
   by 0x15174D: load (rsconf.c:1318)
   by 0x120BBC: initAll (rsyslogd.c:1541)
   by 0x1219FF: main (rsyslogd.c:2120)
```

I think this was introduced by https://github.com/rsyslog/rsyslog/pull/3644

### Steps to reproduce the behavior

Create an `action` with `queue.filename` set, and then set `queue.filename` on `main_queue` as well.

### Environment
- rsyslog version: 8.1905.0
- platform: Ubuntu 18.04

davidelang:
[...]

PR #3688, opened 2019-06-03
core bugfix: segfault on startup depending on queue file names

rgerhards:
rsyslog will segfault on startup when a main queue file name has
been set and at least on other queue contains a file name. This
was cased by too-early freeing config error-detection data
structures. It is a regression caused by commit e22fb205a3.

Thanks to Wade Simmons for reporting this issue and providing
detailled analysis. That greatly helps fixing it quickly.

closes https://github.com/rsyslog/rsyslog/issues/3681

### DIFF
--- a/runtime/rsconf.c
+++ b/runtime/rsconf.c
@@ -635,7 +635,6 @@ static inline rsRetVal
 tellCoreConfigLoadDone(void)
 {
 	DBGPRINTF("telling rsyslog core that config load for %p is done\n", loadConf);
-	qqueueDoneLoadCnf();
 	return glblDoneLoadCnf();
 }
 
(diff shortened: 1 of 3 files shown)


andvgal:
@rgerhards it seems not all suggested changes have been implemented: https://github.com/rsyslog/rsyslog/issues/3681#issuecomment-497972617

Please ensure `qqueueDoneLoadCnf()` sets the global variable to NULL after free to avoid any possible segfaults in the future.

rgerhards:
> Please ensure `qqueueDoneLoadCnf()` sets the global variable to NULL after free to avoid any possible segfaults in the future.

I do not believe in the "everything must be set to NULL" paradigm. It is better to segfault in this case than not detect the problem.

Issue #3684, opened 2019-05-31
Lots of missing files in 8.1905 release tarball


Whissi:
```
make[4]: Entering directory '/var/tmp/portage/app-admin/rsyslog-8.1905.0/work/rsyslog-8.1905.0/tests'
make[4]: Nothing to be done for 'imbatchreport_errmsg_no_params.log'.
make[4]: Nothing to be done for 'imbatchreport_errmsg_glob_not_regular.log'.
make[4]: Nothing to be done for 'imbatchreport_errmsg_glob_dir_fake.log'.
make[4]: Nothing to be done for 'imbatchreport_errmsg_glob_dir_not_dir.log'.
make[4]: Nothing to be done for 'imbatchreport_errmsg_regex.match.reject.log'.
make[4]: Nothing to be done for 'imbatchreport_errmsg_regex.match.rename.log'.
make[4]: Nothing to be done for 'imbatchreport_errmsg_regex.nomatch.log'.
make[4]: Nothing to be done for 'imbatchreport_errmsg_not_supported1.log'.
make[4]: Nothing to be done for 'imbatchreport_errmsg_not_supported2.log'.
make[4]: Nothing to be done for 'imbatchreport_errmsg_not_supported3.log'.
make[4]: Nothing to be done for 'imbatchreport_errmsg_delete_params.log'.
make[4]: Nothing to be done for 'imbatchreport_errmsg_rename_params.log'.
make[4]: Nothing to be done for 'imbatchreport_delete_success.log'.
make[4]: Nothing to be done for 'imbatchreport_delete_structdata.log'.
make[4]: Nothing to be done for 'imbatchreport_rename_success.log'.
make[4]: Nothing to be done for 'imbatchreport_delete_toolarge.log'.
make[4]: Nothing to be done for 'imbatchreport_rename_toolarge.log'.
make[4]: Nothing to be done for 'mmtaghostname_tag.log'.
make[4]: Nothing to be done for 'mmtaghostname_server.log'.
make[4]: Nothing to be done for 'imtuxedoulog_errmsg_no_params.log'.
make[4]: Nothing to be done for 'imtuxedoulog_data.log'.
make[4]: Leaving directory '/var/tmp/portage/app-admin/rsyslog-8.1905.0/work/rsyslog-8.1905.0/tests'
fatal: making test-suite.log: failed to create imbatchreport_errmsg_no_params.trs
fatal: making test-suite.log: failed to create imbatchreport_errmsg_no_params.log
[...]

PR #3690, opened 2019-06-04
testbench: some modules were excluded in "make distcheck"

rgerhards:
see also https://github.com/rsyslog/rsyslog/issues/3684


rgerhards:
CodeCov false positive due to incorrect `.codecov.yml` (now fixed in master branch)

PR #3697, opened 2019-06-07
omamqp1 - port to latest api, add tests

richm:
This brings omamqp1 up-to-date with the latest qpid-proton-c
api version.  This also adds a test for the plugin, to test
the basic functionality.  The test requires the user to
install qdrouterd and the python qpid-proton library in order
to use the simple_recv.py test program.


richm:
@kgiusti fyi

tests/omamqp1-basic.sh
```diff
@@ -0,0 +1,77 @@
...
+qdrouterd > $RSYSLOG_DYNNAME.spool/qdrouterd.log 2>&1 & qdrouterdpid=$!
+sleep 5 # give qdrouterd a chance to start up and listen
+# have to start reader before writer - like a pipe - for the client, NUMMESSAGES could
+# be much less than the number of records if batching is used - for the client, it means
+# batches, not records - but there can't be more batches than records
+amqp_simple_recv $AMQP_URL amq.rsyslogtest $NUMMESSAGES > $RSYSLOG_DYNNAME.spool/amqp_simple_recv.out 2>&1 &
+#export RSYSLOG_DEBUG=debug
+#export RSYSLOG_DEBUGLOG=/tmp/rsyslog.debug.log
+startup
+if [ "${USE_VALGRIND:-}" = YES ] ; then
```
rgerhards, line 37: This should not be necessary, as the testbench framework handles this. Do you have seen some cases where a problem occurred? If so, I'd like to fix that in the plumbing.

tests/omamqp1-basic.sh
```diff
@@ -0,0 +1,77 @@
...
+inp = file(sys.argv[1],"r").read()
+last = 0
+idx = inp.find("msgnum:",last)
+while idx > -1:
+    msgstr = inp[(idx+7):(idx+15)]
+    print msgstr
+    last = idx+16
+    idx = inp.find("msgnum:",last)
+' $RSYSLOG_DYNNAME.spool/amqp_simple_recv.out > $RSYSLOG_OUT_LOG
+if seq_check  0 $(( NUMMESSAGES - 1 )) ; then
```
rgerhards: line 73-77 could simply be replaced by
```
seq_check
```
if I don't overlook anything. I am trying to keep the individual test file as simple as possible. So we should change that if there is no hard reason to keep as currently is.

tests/omamqp1-basic.sh
richm: Yes, the problem is that you do not get a full valgrind stack trace with the full symbols for leak detection because the module is unloaded before valgrind can report the leaks.

I'm not sure how you want to handle this.  I'm not sure if it is ok to change all valgrind tests to not unload the modules at shutdown.  I know there is an option to run valgrind without the leak detection, but tests would have to explicitly use that flag in order to skip leak checking, and therefore unload the modules.

tests/omamqp1-basic.sh
richm: I'll change it

tests/omamqp1-basic.sh
rgerhards: > Yes, the problem is that you do not get a full valgrind stack trace with the full symbols for leak detection because the module is unloaded before valgrind can report the leaks.

actually, that was not the point of my question ;-) That `if` guards a `sleep 5` which I think is not necessary. But I'll also check how the rest of the testbench handles module unload. I think it is automatically handled in valgrind tests.

richm: ah, I see - yes, it seems that rsyslog needed a little longer to start up when using valgrind, otherwise, the test would not pass

rgerhards: OK, good to know. Looks like a problem with the testbench plumbing. I'll merge the PR with that sleep and then will try to reproduce the issue with a separate PR.

PR #3700, opened 2019-06-12
testbench: check that valgrind runs start up correctly

rgerhards:


rgerhards:
@richm I don't see any indication the `sleep` in valgrind case is required. So I would remove it and keep an eye on the test to see if a problem occurs. Agree?

richm:
> @richm I don't see any indication the `sleep` in valgrind case is required. So I would remove it and keep an eye on the test to see if a problem occurs. Agree?

Yes, that's fine.

PR #3704, opened 2019-06-14
WiP: issue-1503 handle TLS close notify from peer

Arnix:
Check if a close-notify has been received, if so shutdown the SSL connection and return an indication that the peer has close the connection. 

Note to reviewer, the logging done is probably not correct. Please advice.

### DIFF
--- a/tests/omamqp1-basic.sh
+++ b/tests/omamqp1-basic.sh
@@ -34,9 +34,6 @@ amqp_simple_recv $AMQP_URL amq.rsyslogtest $NUMMESSAGES > $RSYSLOG_DYNNAME.spool
 #export RSYSLOG_DEBUG=debug
 #export RSYSLOG_DEBUGLOG=/tmp/rsyslog.debug.log
 startup
-if [ "${USE_VALGRIND:-}" = YES ] ; then
-	sleep 5 # give server a little extra time to start up
-fi
 if [ -n "${USE_GDB:-}" ] ; then
     echo attach gdb here
     sleep 54321 || :


Arnix:
this was breaking, giving it some more work

Arnix:
Made another go at it. I do not love the solution, but it might have an embryo for a better solution. I have tested it on our setup and it works, now see if the tests are breaking or not in the CI build.

Arnix:
@rgerhards - can you give some guidance on how to figure out what goes wrong in "buildbot/rsyslog docker-fedora29"?

rgerhards:
> @rgerhards - can you give some guidance on how to figure out what goes wrong in "buildbot/rsyslog docker-fedora29"?

that was an unrelated flake - we unfortunately have them from time to time, especially if CI gets busy. I have restarted the worker.

Arnix:
@rgerhards - I see that this failed again, with other targets now. Any magic comment I can issue to trigger a new build, or do I have to commit a change?

rgerhards:
I have restarted the ones in question. You can do a full rebuild simply via
```
$ git commit --amend
$ git push -f
```
This creates a new commit hash (without changing anything) and thus triggers a new build. But be careful: the more quickly you do this, the more pressure is on the CI machines and with increasing pressure false positive rates increase. We are working on that, but are obviously not yet at a full solution.

Arnix:
@rgerhards - I have tried a number of times to get through all the tests, but with no complete success. I am not sure if you are interested in looking at this PR. As I said in the initial comment, it is not the perfect solution, but it do address the problem with a TLS close notify sent from the remote receiver. In short, I check the underlying tcp socket if there are any data available for consumption. If not, return with success. If data is available, a SSL_peek is done to check for an error situation. If SSL_get_error indicates that a close notify is received, return with abort. I am uncertain if the SSL_shutdown should be done here or not. The step with checking the underlying TCP socket is not elegant, but since the socket is blocking, there is no way (from what I have found) to do a check on SSL level. There is a code line that attempts to set the BIO, wrapping the TCP socket, to non-blocking, but as far as I can see from documentation that cannot possibly work.

rgerhards:
I have re-initiated the tests a couple of times. So the failure looks valid. Did you run the test in your local environment? Does it work there?

Arnix:
@rgerhards - I am not able to find these tests. Looking at the logs I get few clues on where to look.

rgerhards:
It's all in./tests

PR #3707, opened 2019-06-17
check where we need to fix alignment

rgerhards:


rgerhards:
fixes in https://github.com/rsyslog/rsyslog/pull/4072

PR #3724, opened 2019-06-26
WIP: handle TLS close-notify, verify build first

Arnix:

### DIFF
--- a/runtime/nsd_ossl.c
+++ b/runtime/nsd_ossl.c
@@ -1304,6 +1304,8 @@ CheckConnection(nsd_t __attribute__((unused)) *pNsd)
 {
 	nsd_ossl_t *pThis = (nsd_ossl_t*) pNsd;
 	ISOBJ_TYPE_assert(pThis, nsd_ossl);
+	nsd_ptcp_t *pThat = (nsd_ptcp_t *) pThis->pTcp;
+	ISOBJ_TYPE_assert(pThat, nsd_ptcp);
 
 	dbgprintf("CheckConnection for %p\n", pNsd);
 	return nsd_ptcp.CheckConnection(pThis->pTcp);


Arnix:
@rgerhards - so I started from scratch. Apologies for that, but I got lost somewhere. This time, I have made a minimal commit first, that simply assigns a pointer and pushed to trigger the build. And it still fails. I am simply not able to sort these issues out myself. So the question back to you is: are you willing to help out and have a look at the submission of the problem that is reported on close-notify handling, which I believe I have a decent sketch for a solution, or should I just stop wasting mine and your time on this? I am happy to supply the full patch, but I cannot solve the build problems. I am simply not fluent enough on your entire setup and do not have time.  Thanks you, @Arnix

rgerhards:
> So the question back to you is: are you willing to help out and have a look at the submission of the problem that is reported on close-notify handling, which I believe I have a decent sketch for a solution, or should I just stop wasting mine and your time on this? I am happy to supply the full patch, but I cannot solve the build problems

well... I haven't worked on TLS for quite a while - @alorbach has crafted much of the current commits (including the complete openssl subsystem). So he's probably to one to look for detail issue.

Other than that, the last PR did not have build problems, but problems during runtime. The test results looked like they were caused by the new code. Thus I recommended to check if you could reproduce the issue locally. Having this ability makes troubleshooting much easier.

I admit that, to be helpful at the detail level, I would need to invest quite a bit to come back to solid understanding of the TLS subsystem. Right now, I cannot do that.

I would recommend the following:

1. check if the current CI fails are flakes or look real
2. if real, check if you can reproduce locally. That would at least enable much better debugging
3. if you can test locally and do not see the issue or if we need to rely on CI only, let's check if @alorbach can have a look

We need some persistence with the issue as I know that @alorbach, for personal reasons, will have only very limited time for open source work in the next 2 to 3 month.

rgerhards:
side-note: given the current patch, this looks like a flake, but as I said, I investigate.

rgerhards:
@Arnix FYI: Fedora30 was a flake, probably the rest also. I am using this opportunity to slighly improve CI, so the failure state will stay for a short while. I'll let you know when I am done.

rgerhards:
The Travis failure is correct given the current state of the code - check yourself...

Arnix:
@rgerhards - I will push this again, then when I see the build is stable I will apply the actual patch and let you know. Probably not until next week. 
Thanks!

rgerhards:
CI will have flakes next week as well. While we work hard on this, there is little we can do against it. I babysit PRs, as I did with the last one. When I come to the conclusion it is not flake (like last time), I'll let you know. You don't need to care too much about the interim failures (well, most contributors have a quick look, as very often it is not a flake in early PR development).

rgerhards:
@Arnix still interest in pursuing this further?

Arnix:
Hi

Yes, I am. But entering summer vacation now. So no capacity before August.
Feel free to close the PR if this is too late. I can open a new one when I
am back.

Have a nice summer,
Arild

rgerhards:
@Arnix thx! I will close the PR just to keep the noise level low. Feel free to open a new one whenever you are ready.

PR #3725, opened 2019-06-27
omclickhouse: fix segfault on omclickhouse batchmode

wdjwxh:
omclickhouse use ` strchr(strstr(message,"VALUES ") ` when building the batch sql. But the message may not contains `VALUES`, then it segfault.  
This commit fix that . It will continue build the batch sql, and then report errors by clickhouse's return message.

### DIFF
--- a/plugins/omclickhouse/omclickhouse.c
+++ b/plugins/omclickhouse/omclickhouse.c
@@ -431,8 +431,9 @@ computeBulkMessage(const wrkrInstanceData_t *const pWrkrData,
 	const uchar *const message, char **newMessage)
 {
 	size_t r = 0;
-	if(pWrkrData->batch.nmemb != 0) {
-		*newMessage = strchr(strstr((const char*)message, "VALUES"), '(');
+	char *v;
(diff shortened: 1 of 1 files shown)


plugins/omclickhouse/omclickhouse.c
```diff
@@ -431,8 +431,9 @@ computeBulkMessage(const wrkrInstanceData_t *const pWrkrData,
 	const uchar *const message, char **newMessage)
 {
 	size_t r = 0;
-	if(pWrkrData->batch.nmemb != 0) {
-		*newMessage = strchr(strstr((const char*)message, "VALUES"), '(');
+	char *v;
+	if(pWrkrData->batch.nmemb != 0 && (v = strstr((const char *)message, "VALUES")) != NULL) {
+		*newMessage = strchr(v, '(');
```
rgerhards: thx for the PR. But while we are at it, I think this is an additional potential NULL pointer reference in the case `'('` is not present. Would you like to add the check here as well?

wdjwxh: ok, soon

wdjwxh: @rgerhards  done.

Issue #3694, opened 2019-06-06
[ enhancement ] Turning off UseSpecialParser for FreeBSD imuxsock


zeising:
### Summary
FreeBSD switched logging to use rfc5424 behaviour by defalt on logs arriving on log sockets and elsewhere. This seem to confuse rsyslog parsing, and the result is that rsyslog fails to parse out things like program name. This can be fixed by turning off the special parser, using `useSpecialParser=off`. This should probably be made the default for FreeBSD (or handled in another way, if appropriate)
Some more information is available in this mail thread:
http://rsyslog-users.1305293.n2.nabble.com/imuxsock-needs-UseSpecialParser-off-quot-to-parse-var-run-log-correctly-on-FreeBSD-td7594964.html

### Expected behavior
Proper parsing of things like programname, syslogtag, and so on. See this debug log with useSpecialParser turned off:
```
Debug line with all properties:
FROMHOST: 'test', fromhost-ip: '127.0.0.1', HOSTNAME: 'test.daemonic.se', PRI: 133,
syslogtag 'blah[35820]', programname: 'blah', APP-NAME: 'blah', PROCID: '35820', MSGID: '-',
TIMESTAMP: 'Jun  6 11:32:34', STRUCTURED-DATA: '-',
msg: 'blahblah'
escaped msg: 'blahblah'
inputname: imuxsock rawmsg: '<133>1 2019-06-06T11:32:34.666+02:00 test.daemonic.se blah 35820 - - blahblah'
$!:
$.:
$/:
```

### Actual behavior
Here is a snippet of a log (from logger), with using the `RSYSLOG_DebugFormat` template

```
Debug line with all properties:
FROMHOST: 'test', fromhost-ip: '127.0.0.1', HOSTNAME: 'test', PRI: 133,
syslogtag '1', programname: '1', APP-NAME: '1', PROCID: '-', MSGID: '-',
TIMESTAMP: 'Jun  6 10:49:22', STRUCTURED-DATA: '-',
msg: ' 2019-06-06T10:49:22.311548+02:00 test.daemonic.se blah 1065 - - blahblah'
escaped msg: ' 2019-06-06T10:49:22.311548+02:00 test.daemonic.se blah 1065 - -
blahblah'
inputname: imuxsock rawmsg: '<133>1 2019-06-06T10:49:22.311548+02:00
test.daemonic.se blah 1065 - - blahblah'
$!:
$.:
$/:
```
Note how programname and other things are not set correctly.
[...]

PR #3733, opened 2019-07-01
imuxsock: support FreeBSD 12 out of the box

rgerhards:
FreeBSD 12 uses RFC5424 on the system log socket by default. This
format is not supported by the special parser used in imuxsock.
Thus for FreeBSD the default needs to be changed to use the
regular parser chain by default. That is all this commit does.

Note: the code was merged untested, as we currently do not have
a matching CI environment. If it does not work, please report.

closes https://github.com/rsyslog/rsyslog/issues/3694

### DIFF
--- a/plugins/imuxsock/imuxsock.c
+++ b/plugins/imuxsock/imuxsock.c
@@ -44,6 +44,9 @@
 #ifdef HAVE_LIBSYSTEMD
 #	include <systemd/sd-daemon.h>
 #endif
+#if defined(__FreeBSD__)
+	#include <sys/param.h>
+#endif
 #include "rsyslog.h"
 #include "dirty.h"
 #include "cfsysline.h"
@@ -173,6 +176,11 @@ static int nfd = 1; /* number of active unix sockets  (socket 0 is always reserv
 			socket, even if it is not enabled. */
 static int sd_fds = 0;			/* number of systemd activated sockets */
 
+#if (defined(__FreeBSD__) && (__FreeBSD_version >= 1200061))
+	#define DFLT_bUseSpecialParser 0
+#else
+	#define DFLT_bUseSpecialParser 1
+#endif
 #define DFLT_bCreatePath 0
 #define DFLT_ratelimitInterval 0
 #define DFLT_ratelimitBurst 200
@@ -320,7 +328,7 @@ createInstance(instanceConf_t **pinst)
 	inst->ratelimitBurst = DFLT_ratelimitBurst;
 	inst->ratelimitSeverity = DFLT_ratelimitSeverity;
 	inst->bUseFlowCtl = 0;
-	inst->bUseSpecialParser = 1;
+	inst->bUseSpecialParser = DFLT_bUseSpecialParser;
 	inst->bParseHost = UNSET;
 	inst->bIgnoreTimestamp = 1;
 	inst->bCreatePath = DFLT_bCreatePath;
@@ -1240,7 +1248,7 @@ CODESTARTbeginCnfLoad
 	pModConf->bAnnotateSysSock = 0;
 	pModConf->bParseTrusted = 0;
 	pModConf->bParseHost = UNSET;
-	pModConf->bUseSpecialParser = 1;
+	pModConf->bUseSpecialParser = DFLT_bUseSpecialParser;
 	/* if we do not process internal messages, we will see messages
 	 * from ourselves, and so we need to permit this.
 	 */


zeising:
I'll give it a spin and see what happens.

plugins/imuxsock/imuxsock.c
```diff
@@ -1240,7 +1240,11 @@ CODESTARTbeginCnfLoad
 	pModConf->bAnnotateSysSock = 0;
 	pModConf->bParseTrusted = 0;
 	pModConf->bParseHost = UNSET;
-	pModConf->bUseSpecialParser = 1;
+	#if (defined(__FreeBSD__) && (__FreeBSD__ >= 1200000))
```
zeising: This is not right. Either, you want to check if `__FreeBSD__ is >= 12` (this is defined by the compiler), or you want `__FreeBSD_version` from `<sys/param.h>`. This can give more fine grained version check. In this case, the precise version to check against is `1200061` , which is when this changed went into what then became FreeBSD 12.0.

rgerhards: thx - sticking with the compiler as I understand that all stable FreeBSD 12 have RFC5424 on the system log socket. Right?

zeising: All stable FreeBSD 12 should have RFC5424, yes. The recommended way is generally to use `__FreeBSD_version`, since the value of __FreeBSD__ is dependent on the system you build on, not the system you target, as I understand it.

zeising:
There is something else going on here. Applying the patch patch on top of rsyslog 8.1905.0 does not help the situation. However, adding UseSpecialParser to the imuxsock config makes the syslog messages look as they should.

This is the conf I'm testing with, UseSpecialParser=off can be toggled on and off.
```
module(load="immark")
module(load="imuxsock"
        SysSock.Use="off"
        SysSock.FlowControl="on"
)
module(load="imklog")

input(type="imuxsock"
        Socket="/var/run/log"
#       UseSpecialParser="off"
)

input(type="imuxsock"
        Socket="/var/run/logpriv"
#       UseSpecialParser="off"
)

*.* action(type="omfile" file="/var/log/all.log.debug" template="RSYSLOG_DebugFormat")
*.* /var/log/all.log
```

I have a test VM and can provide logs.


When digging into this, I noticed there is also a local (to the FreeBSD ports/package system) attempt at fixing this, in a similar way but using configure to select the behavior. It basically toggles `pModConf->bUseSpecialParser` to 0 or 1 by default depending on the configure option. This fix has been there for some time, and is in the package I initially discovered this issue with. For details, see the following:
https://github.com/freebsd/freebsd-ports/blob/master/sysutils/rsyslog8/files/patch-plugins_imuxsock_imuxsock.c
https://github.com/freebsd/freebsd-ports/blob/master/sysutils/rsyslog8/files/patch-configure.ac
https://github.com/freebsd/freebsd-ports/blob/master/sysutils/rsyslog8/files/patch-config.h.in

I was unfortunately not aware of this before tonight.

rgerhards:
very interesting... I'll check the flow on this variable.

rgerhards:
@zeising I admit I do not understand why the ports patch does not work. Mine indeed had a missing bit which could cause it to fail. I'll do a fixup of my patch, making it kind of a hybrid between the ports patch and my initial one. If that still does not work, we need to debug on your VM. Probably I add some debug messages in the new patch so that we can do that easily.

rgerhards:
Doh! My predicate check was inverted... Still wonder why the ports version does not work.

zeising:
The issue is a bit strange. I even tried to force `pModConf->bUseSpecialParser` to the other value, in case the preprocessor stuff was broken.
I kind of ran out of time last night, I'll poke around more tonight.
Thanks for all the help!

rgerhards:
@zeising thx for *your* help! I have updated the PR once again. Please give it a try. 

It also contains debug string, so if it still does not work correctly, we can create a debug log and see what the internal values are. I'll clean up that debugging stuff once we know it works.

side-note: if you have a script to setup a plain VM so that it can build rsyslog, that would be most helpful. I could use that to give the CI worker a try.

zeising:
> @zeising thx for _your_ help! I have updated the PR once again. Please give it a try.
> 
> It also contains debug string, so if it still does not work correctly, we can create a debug log and see what the internal values are. I'll clean up that debugging stuff once we know it works.

I'll give it a spin tonight.

> side-note: if you have a script to setup a plain VM so that it can build rsyslog, that would be most helpful. I could use that to give the CI worker a try.

I'll have to give this a think. Since this was a one-off I did it the manual way, but it should be possible. I'll get back to you on this one.

rgerhards:
@zeising FYI: CI fail is due to codestyle and rooted in the to-be-removed debug messages. So basically CI is "green".

plugins/imuxsock/imuxsock.c
```diff
@@ -173,6 +176,11 @@ static int nfd = 1; /* number of active unix sockets  (socket 0 is always reserv
 			socket, even if it is not enabled. */
 static int sd_fds = 0;			/* number of systemd activated sockets */
 
+#if (defined(__FreeBSD__) && (__FreeBSD_version >= 1200061))
+	#define DFLT_bUseSpecialParser 1
+#else
+	#define DFLT_bUseSpecialParser 0
```
zeising, line 180: Inverting these, setting DFLT_bUseSpecialParser to 0 in the FreeBSD case, seem to solve the issue.

This is the output from running rsyslogd with debug enabled.
```
3709.314145803:main thread    : imuxsock.c: imuxsock: allocating memory for 2 listeners
3709.314154565:main thread    : ratelimit.c: ratelimit:imuxsock:new ratelimiter:bReduceRepeatMsgs 0
3709.314273893:main thread    : imuxsock.c: imuxsock: set listener[1].bUseSpecialParser to 0 (default 0)
3709.314280628:main thread    : ratelimit.c: ratelimit:imuxsock:new ratelimiter:bReduceRepeatMsgs 0
3709.314287006:main thread    : imuxsock.c: imuxsock: set listener[2].bUseSpecialParser to 0 (default 0)
3709.314478218:main thread    : imuxsock.c: imuxsock: Opened UNIX socket '/var/run/log' (fd 3).
3709.314637005:main thread    : imuxsock.c: imuxsock: Opened UNIX socket '/var/run/logpriv' (fd 4).
```

rgerhards: > Inverting these, setting DFLT_bUseSpecialParser to 0 in the FreeBSD case, seem to solve the issue.

yes, of course - made the same mistake twice ;-) I'll fix and then merge.

zeising: Sounds good.

zeising: Thank you for helping out and fixing this!

PR #3739, opened 2019-07-04
fix failpath on actionTryRemoveHardErrorsFromBatch

kortemik:
fixes failpath on actionTryRemoveHardErrorsFromBatch happens only when

actionCommit () 
-> actionTryCommit => FAIL
-> actionCommit somewhat unhappy 
-> actionTryRemoveHardErrorsFromBatch() // contains memcpy which did work with non allocated area and non-element size alligned copy
-> actionCommit unhappy, has still uncommited messages.
-> actionTryCommit // would pass the rogue memory range to module

### DIFF
--- a/action.c
+++ b/action.c
@@ -1408,7 +1408,7 @@ done:
 
 static rsRetVal
 actionTryRemoveHardErrorsFromBatch(action_t *__restrict__ const pThis, wti_t *__restrict__ const pWti,
-	actWrkrIParams_t *const new_iparams, unsigned *new_nMsgs)
+	actWrkrIParams_t **new_iparams, unsigned *new_nMsgs)
 {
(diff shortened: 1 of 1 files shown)


kortemik:
@rgerhards  i really hope it's finally now ok, but shouldn't we add also CHKiRet on the call to actionTryRemoveHardErrorsFromBatch to handle the possible oom case as well? I think the oom case might be actually relevant at this point.

rgerhards:
more later but in case you did not see the announcement on twitter: I am working on tumbleweed CI worker, it probably errs out invalidly.

kortemik:
I need to subscribe you on twitter!

kortemik:
well seems like it needs a bit more work around itself

PR #3740, opened 2019-07-04
core bugfix: incorrect error message on duplicate module load

rgerhards:
A Null-pointer was passed to printf instead of the module name.
On some platforms this may lead to a segfault. On most platforms
printf check's for NULL pointers and uses the string "(null)"
instead.

In any case, the module name is missing from the error message.


rgerhards:
PR structure validation is acceptable in this special case (doing unrelated nit-fixes with a single PR)

PR #3743, opened 2019-07-11
iminternal: suppress mutex double-unlock

tblume:
If there is a burst of log messages during a time when rsyslog is unable
to output (either during log rotation, an out-of-space condition, or
some other similar condition), rsyslog can SEGFAULT due to a mutex
double-unlock.


rgerhards:
thx for the PR!

PR #3745, opened 2019-07-12
EXPERIMENTAL core: potential message loss during tx when suspended

rgerhards:


rgerhards:
replaced by https://github.com/rsyslog/rsyslog/pull/3755

PR #3747, opened 2019-07-15
build system: support cross-platform build for mysql/mariadb

rgerhards:
rsyslog fails to cross build from source, because it uses mysql_config
and mysql_config is unfixably broken for cross compilation. It would be
better to use pkg-config. The attached patch makes rsyslog try
pkg-config first and fall back to mysql_config.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -759,29 +759,31 @@ AC_ARG_ENABLE(mysql,
          esac],
         [enable_mysql=no]
 )
-if test "x$enable_mysql" = "xyes"; then
-  AC_CHECK_PROG(
-    [MYSQL_CONFIG],
-    [mysql_config],
-    [mysql_config],
-    [no],,
-  )
-  if test "x${MYSQL_CONFIG}" = "xno"; then
-    AC_MSG_FAILURE([mysql_config not found - usually a package named mysql-dev, libmysql-dev or similar, is missing - install it to fix this issue])
-  fi
-  AC_CHECK_LIB(
-    [mysqlclient],
-    [mysql_init],
-    [MYSQL_CFLAGS=`$MYSQL_CONFIG --cflags`
-     MYSQL_LIBS=`$MYSQL_CONFIG --libs`
-    ],
-    [AC_MSG_FAILURE([MySQL library is missing])],
-    [`$MYSQL_CONFIG --libs`]
-  )
+AS_IF([test "x$enable_mysql" = "xyes"],[
+  PKG_CHECK_MODULES([MYSQL],[mysqlclient],,[
+    AC_CHECK_PROG(
+      [MYSQL_CONFIG],
+      [mysql_config],
+      [mysql_config],
+      [no],,
+    )
+    AS_IF([test "x${MYSQL_CONFIG}" = "xno"],[
+      AC_MSG_FAILURE([mysql_config not found - usually a package named mysql-dev, libmysql-dev or similar, is missing - install it to fix this issue])
+    ])
+    MYSQL_CFLAGS=`$MYSQL_CONFIG --cflags`
+    MYSQL_LIBS=`$MYSQL_CONFIG --libs`
+  ])
   AC_MSG_CHECKING(if we have mysql_library_init)
   save_CFLAGS="$CFLAGS"
   CFLAGS="$CFLAGS $MYSQL_CFLAGS"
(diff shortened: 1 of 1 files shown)


rgerhards:
did not yet hear back from original contributor - de-assigning milestone, but keep the PR open a bit longer.

rgerhards:
It looks like this patch causes CFLAGS to be damaged in some cases. From config.log:
```
configure:21259: result: no
configure:21296: checking for libgcrypt-config
configure:21314: found /usr/local/bin/libgcrypt-config
configure:21327: result: /usr/local/bin/libgcrypt-config
configure:21341: checking for gcry_cipher_open in -lgcrypt
configure:21367: gcc -std=gnu99 -o conftest -g -O2   conftest.c -lgcrypt -I/usr/local/include
-L/usr/local/lib -lgcrypt -lgpg-error
        	 -lmysqlclient -lpthread  >&5
/usr/local/bin/ld: cannot find -lmysqlclient
collect2: error: ld returned 1 exit status
```

@mbiebl can you have a look?

rgerhards:
@helmutg any interest in fixing this?

mbiebl:
The problem afaics is that `-lmysqlclient` is added to `LIBS`. Maybe LIBS is not properly restored?

mbiebl:
The generated configure code looks like this:
```
fi
  { $as_echo "$as_me:${as_lineno-$LINENO}: checking for mysql_init in -lmysqlclient" >&5
$as_echo_n "checking for mysql_init in -lmysqlclient... " >&6; }
if ${ac_cv_lib_mysqlclient_mysql_init+:} false; then :
  $as_echo_n "(cached) " >&6
else
  ac_check_lib_save_LIBS=$LIBS
LIBS="-lmysqlclient $MYSQL_LIBS
   $LIBS"
```
That doesn't look right

mbiebl:
I think the problematic part is, that PKG_CHECK_MODULES itself does mangling of LIBS/CFLAGS and we run the AC_TRY_LINK inside the PKG_CHECK_MODULES macro

mbiebl:
Maybe we need to pull the save_LIBS="$LIBS" (and CFLAGS) outside the PKG_CHECK_MODULES macro

rgerhards:
> we run the AC_TRY_LINK inside the PKG_CHECK_MODULES macro

maybe I am wrong, but it doesn't looks so to me. Nevertheless, I have move the saving of LIBS somewhat above. Let's see how this works out.

mbiebl:
might very well be that I lost count of opening and closing braces.

mbiebl:
Ah, wait a sec

```
  AC_CHECK_LIB(
    [mysqlclient],
    [mysql_init],
    ,
    [AC_MSG_FAILURE([MySQL library is missing])],
    [$MYSQL_LIBS]
  )
  AC_MSG_CHECKING(if we have mysql_library_init)
  save_CFLAGS="$CFLAGS"
  CFLAGS="$CFLAGS $MYSQL_CFLAGS"
  save_LIBS="$LIBS"
  LIBS="$LIBS $MYSQL_LIBS"
  AC_TRY_LINK(
    [#include <mysql.h>
     #include <stdio.h>],
    [mysql_library_init(0, NULL, NULL)],
    [have_mysql_library_init=yes],
    [have_mysql_library_init=no])
  CFLAGS="$save_CFLAGS"
  LIBS="$save_LIBS"
```
Doesn't AC_CHECK_LIB add the libraries it checks for to LIBS?
I.e. we save save_LIBS too late and we need to do it before AC_CHECK_LIB?

mbiebl:
yup, we don't define a action-if-found method for AC_CHECK_LIB, so the default is to add the library to LIBS. https://www.gnu.org/software/autoconf/manual/autoconf-2.66/html_node/Libraries.html

rgerhards:
> yup, we don't define a action-if-found method for AC_CHECK_LIB, so the default is to add the library to LIBS. https://www.gnu.org/software/autoconf/manual/autoconf-2.66/html_node/Libraries.html

thx for the swift action. Do you think the "move up" of saving the libs is sufficient/good or should we use another solution?

mbiebl:
Let's see how the builds go, but if they succeed, I'd say it's good as-is.

rgerhards:
yup, works. So I'll merge. @mbiebl thx again for the help!

PR #3751, opened 2019-07-17
properly propagate TCP Streamdriver.Mode setting

jvymazal:
If StreamDriver.mode is not set per action the module setting was not
inherited which could cause messages meant to be encrypted to be
present in plaintext on wire.

### DIFF
--- a/tools/omfwd.c
+++ b/tools/omfwd.c
@@ -1100,6 +1100,9 @@ initTCP(wrkrInstanceData_t *pWrkrData)
 		CHKiRet(tcpclt.SetFraming(pWrkrData->pTCPClt, pData->tcp_framing));
 		CHKiRet(tcpclt.SetFramingDelimiter(pWrkrData->pTCPClt, pData->tcp_framingDelimiter));
 		CHKiRet(tcpclt.SetRebindInterval(pWrkrData->pTCPClt, pData->iRebindInterval));
+		if (cs.iStrmDrvrMode) {
+			pData->iStrmDrvrMode = cs.iStrmDrvrMode;
+		}
 	}
 finalize_it:
 	RETiRet;


jvymazal:
Ok, this probably comes from me not properly understanding how the legacy system works. Do you please remember correct way of setting this with legacy syntax?

rgerhards:
> If StreamDriver.mode is not set per action the module setting was not
> inherited

I just checked the code, there is not per-module default for the stream driver mode.

The proper way to implement this is to add a new module default to the module param block, see here https://github.com/rsyslog/rsyslog/blob/master/tools/omfwd.c#L158

Then, this needs to be stored and used as default if the action does not specify something different. The right place to do this is when setting the input defaults: https://github.com/rsyslog/rsyslog/blob/master/tools/omfwd.c#L1162

While doing so, I would suggest that default values for other TLS parameters can be set, because a default really only makes sense if all params can be set (they belong all together).

rgerhards:
> Ok, this probably comes from me not properly understanding how the legacy system works. Do you please remember correct way of setting this with legacy syntax?

very ugly. In a nutshell, you specify the params in front of the action or input definition. The input than takes whatever is set before it. Some inputs/actions reset some params to defaults, some do not. Some params are acutally globaly, most are not. That mess was the reason we ditched it ;-) Full reference probably in an old v3 doc set.

Please note that we do NOT permit any changes to the legacy system any longer. So the only way to create a default is by using new syntax. Else we would break existing configs.

jvymazal:
Ok, just to clear it up, my current understanding is that there is no actual problem when using new syntax and we do not want to mess with the old one so the potential changes would only affect new config?

rgerhards:
> Ok, just to clear it up, my current understanding is that there is no actual problem when using new syntax and we do not want to mess with the old one so the potential changes would only affect new config?

Yes: changes to obsolete legacy format will either
* break existing configs
* introduce new features
New features in obsolete legacy format are bad, because it is extremely confusing and we do not want to promote this for any new config.

New config format is always fine and co-exists with legacy format.

In regard to this PR it is important to understand that no defaults for TLS inputs currently exist - neither in obsolete legacy format nor in new/current style. If we want to have them, we need to enhance the module. Enhancements must be done in new style. How this can be done I described above.

Does that clear up the situation?

jvymazal:
Yes, I just wanted to make sure there is no actual bug in code which would need fixing. In this case I am fine with cancelling this PR as the problem will have to be solved via different configuration/documentation update on our side.

Many thanks for patience and explanation.

rgerhards:
double-checked: there are actually some defaults, search for "TLS" on https://www.rsyslog.com/doc/v8-stable/rainerscript/global.html -- I overlooked them because they are part of the rsyslog global object.

There are also some obsolete legacy format settings for these global ones. Just FYI.

PR #3755, opened 2019-07-18
core/action: special handling for suspending action via ext file

rgerhards:
We try to reduce potential problems during transaction commit be ensuring
that the action is only tried to be suspended at begin of transaction.
This also simplifies the code a bit and also reduces performance
impact of that "suspension via external state file" feature.


rgerhards:
CodeCov report looks strange - probably a baseline run errored out. Thus ignoring this issue.

PR #3762, opened 2019-07-23
Add "fsync" option for imjournal

jvymazal:
The new option makes possible to force physical write of stateFile
to persistent storage, ensuring we do not lose/duplicate messages
in case of hard crash or power loss.

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -24,6 +24,7 @@
 #include "config.h"
 #include "rsyslog.h"
 #include <stdio.h>
+#include <dirent.h>
 #include <assert.h>
 #include <string.h>
 #include <stdarg.h>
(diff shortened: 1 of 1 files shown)


jvymazal:
Creating doc-pr now, @rgerhards can you get this to 8.1908 ? (to know which version should I put to docs as starting)

Thanks

rgerhards:
> can you get this to 8.1908

yup!

rgerhards:
doc: https://github.com/rsyslog/rsyslog-doc/pull/836

rgerhards:
sry, I was too quick to merge :-( -- the new option will not work because uit is missing in the `modpdescr` table - so the config system does not know about it.

Shall I fix or will you do? Both fine to me.

jvymazal:
Sorry, overlooked that bit, I will add it, the same branch is ok?

rgerhards:
perfect, thx! I should have spotted this before merge ;-)

PR #3764, opened 2019-07-24
Option SocketType added into imuxsock

### LINKED ISSUE #3711: Add option for using imuxsock in SOCK_STREAM
currently imuxsock only supports SOCK_DGRAM would be nice with a config option to support SOCK_STREAM as well

ZoltanFridrich:
closes #3711

### DIFF
--- a/plugins/imuxsock/imuxsock.c
+++ b/plugins/imuxsock/imuxsock.c
@@ -41,6 +41,7 @@
 #include <sys/stat.h>
 #include <sys/un.h>
 #include <sys/socket.h>
+#include <pthread.h>
 #ifdef HAVE_LIBSYSTEMD
 #	include <systemd/sd-daemon.h>
 #endif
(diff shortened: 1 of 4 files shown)


ZoltanFridrich:
resolves issue #3711

rgerhards:
Thx for the PR. Would it be possible to add a test at least for the sock_stream type? It should submit multiple messages inside a connection and probably do at least two connections. I think it is vital to proof that this use case works and continues to do so as the code evolves in the future.

rgerhards:

ZoltanFridrich:
I will make a PR for the documentation ASAP, the test will take me some time though.

rgerhards:
> the test will take me some time though.

if you need help with the test, pls let me know. I suggest to start with existing tests. Please create non-root tests, as otherwise we cannot run them during regular CI.

ZoltanFridrich:
There is a PR for documentation #837

ZoltanFridrich:
There is actually much more work then just creating a different type of socket. Because stream sockets need to actually listen and accept all incoming connections I had to move the listening endless loop into a different thread. Because of so many complications I decided that it would be probably simpliest to manage those open connections separately in one synchronized container.

This is just a concept (I didnt test it yet), can you please review my code @rgerhards @jvymazal to check if this solution would be acceptable?

rgerhards:
TSAN worker currently always fails - I have cancelled and will restart it once the core issue is solved. That worker is experimental.

ZoltanFridrich:
OK the test is ready

rgerhards:
It looks like there are some real issues in CI.

rgerhards:
@ZoltanFridrich any interest in finishing this PR?

ZoltanFridrich:
Yes ofc, I just didnt realise there is something wrong with the PR.

rgerhards:
I am cleaning up my post-covid stack of PRs (yes, I know, still late, but...). My apologies for the late action.

If anyone is still interested in this PR, pls let me know. I can than help with get it going.

Issue #3798, opened 2019-08-07
imptcp: data race


rgerhards:
When running test `imptcp_conndrop.sh` under LLVM TSAN, we see some data races. An example can be found at the end of this issue.

Initial verification shows that more than one worker may be processing the same session concurrently. While this can increase performance thanks to multiple `recv` operations in place, I think it may cause subtle use-afer-free bugs when the session is closed and a parallel worker is still sitting in `recv`. I have seen this behavior during my testing.

```
WARNING: ThreadSanitizer: data race (pid=36272)
  Write of size 8 at 0x7b18000008d0 by thread T5:
    #0 DataRcvd /home/rger/proj/rsyslog/plugins/imptcp/imptcp.c:1298:26 (imptcp.so+0x7d63)
    #1 sessActivity /home/rger/proj/rsyslog/plugins/imptcp/imptcp.c:1892 (imptcp.so+0x7d63)
    #2 processWorkItem /home/rger/proj/rsyslog/plugins/imptcp/imptcp.c:1937 (imptcp.so+0x7d63)
    #3 wrkr /home/rger/proj/rsyslog/plugins/imptcp/imptcp.c:2091:4 (imptcp.so+0x6e75)

  Previous write of size 8 at 0x7b18000008d0 by thread T3:
    #0 DataRcvd /home/rger/proj/rsyslog/plugins/imptcp/imptcp.c:1298:26 (imptcp.so+0x7d63)
    #1 sessActivity /home/rger/proj/rsyslog/plugins/imptcp/imptcp.c:1892 (imptcp.so+0x7d63)
    #2 processWorkItem /home/rger/proj/rsyslog/plugins/imptcp/imptcp.c:1937 (imptcp.so+0x7d63)
    #3 enqueueIoWork /home/rger/proj/rsyslog/plugins/imptcp/imptcp.c:2020:3 (imptcp.so+0x3d62)
    #4 processWorkSet /home/rger/proj/rsyslog/plugins/imptcp/imptcp.c:2048 (imptcp.so+0x3d62)
    #5 runInput /home/rger/proj/rsyslog/plugins/imptcp/imptcp.c:2400 (imptcp.so+0x3d62)
    #6 thrdStarter /home/rger/proj/rsyslog/runtime/../threads.c:239:9 (rsyslogd+0x548326)

  Location is heap block of size 96 at 0x7b18000008a0 allocated by main thread:
    #0 calloc <null> (rsyslogd+0x42e08a)
[...]

PR #3799, opened 2019-08-07
testbench: disable TSAN for imptcp until further clarification

rgerhards:
see also https://github.com/rsyslog/rsyslog/issues/3798


rgerhards:
PR no longer necessary as we fix the root cause via https://github.com/rsyslog/rsyslog/pull/3797

PR #3803, opened 2019-08-09
travis: libmongoc-dev does not install all needed

rgerhards:
there seem to be packaging changes in ubuntu or some other repos
we use on travis. This should fix the issue.


rgerhards:
This is only a travis change. So any other CI failures are obviously false positives. Thus merging early - other PRs are waiting for this here.

PR #3796, opened 2019-08-07
omelasticsearch bugfix: potential segmentation fault

nhosoi:
When duplicated configurations are accidentally loaded, rsyslog eliminates extra duplicated instances.  Omelasticsearch has a linked list to manage multiple instances.  When a configuration instance is freed in freeInstance, the instance should be removed from the list, as well.

PR #3809, opened 2019-08-12
testbench: add tests for omelasticsearch error configs

rgerhards:
These test test errors described in
   https://github.com/rsyslog/rsyslog/pull/3796


rgerhards:
Check with unfixed code went well: CI aborted due to errors - this is how it should work!
Will restart once the patch is merged.

PR #3828, opened 2019-08-15
testbench: experimentally increase queue empty check period

rgerhards:
The test frequently failed in a way that suggests rsyslog terminated too
early. We try if we can improve the situation by prolonging the empty
check. While there are better solutions for this special case, there are
some cases where we do no have them. So it is good to know if this
method works.


rgerhards:
CI fails are obviously false positives - good to merge

Issue #3727, opened 2019-06-28
imtcp: add input() parameters for NSD driver options


skupkosk:
================== Maintainer edit ==================
See below: this actually is about **adding additional input parameters** rather then moving parameters. Relevant comment with detail info: https://github.com/rsyslog/rsyslog/issues/3727#issuecomment-524750746
================ End Maintainer edit ==================

### Expected behavior
Be able to use imtcp module for both, clean and TLS inputs. The similar way it is used on omfwd actions. Every omfwd action can use their own StreamDriver options.

### Actual behavior
The module is loaded once with NSD preloaded without possibility to change in input() calls.
At the moment the only workaround is the availability of imptcp which can be used for plain TCP.

davidelang:
we don't want to remove the tls settings from the module() call for backwards 
compatibility reasons

But we do need to make it possible to set them on the input() call. There are 
just too may reasons to need to have different settings for differen inputs 
(encrypted, not encrypted, different certs on different inputs, different auth 
modes on different ports, etc)

David Lang

davidelang:
another possible options would be to allow a module to be loaded multiple times, 
with different names (the internals would have to end up in their own namespaces 
to avoid conflicts)

David Lang

skupkosk:
Would prefer the way of not needing loading module multiple times.

The side-or-main motivation could be the ability to throw away one of imtcp or imptcp modules which are just doing the same.

davidelang:
as I understood it, the 'p' in imptcp was origionally for 'parallel', in that it 
allowed more than one thread to work on the input from one connection. This is a 
big performance win if you have a lot of logs from one source, but you can't do 
encryption in parallel like that, so imtcp will need to stay around.

I agree that loading the module multiple times would not be ideal, but if it 
could be done with less development effort, it may be a win.

David Lang

rgerhards:
Loading a module multiple times is not possible nor desirable - it would require a complete redesign. This should be changed as follows:

* make module parameters act as defaults
* add input parameters

Actually, I thought this was already done, but it doesn't look so. Out of my head I do not know if it is a smaller or larger task. Tentatively assigning milestone, let's see if someone has time to look at it.

rgerhards:
I had a look at the code. It's unfortunately not a small job. The server is currently written in a way that assigns the parameter at the global server basis, not the individual listener. I'll further evaluate what needs to be changed to support that.

rgerhards:
[...]

PR #3845, opened 2019-08-27
Stricter GnuTLS operation

jvymazal:
This commit adds to new flags which can be set to allow
1) checking of extendedKeyUsage certificate field
2) stricter checking of certificate name/adresses

### DIFF
--- a/runtime/nsd_gtls.c
+++ b/runtime/nsd_gtls.c
@@ -1004,6 +1004,7 @@ gtlsChkPeerName(nsd_gtls_t *pThis, gnutls_x509_crt_t *pCert)
 	int iAltName;
 	size_t szAltNameLen;
 	int bFoundPositiveMatch;
+	int bHaveSAN = 0;
 	cstr_t *pStr = NULL;
 	cstr_t *pstrCN = NULL;
 	int gnuRet;
@@ -1023,6 +1024,7 @@ gtlsChkPeerName(nsd_gtls_t *pThis, gnutls_x509_crt_t *pCert)
 		if(gnuRet < 0)
 			break;
 		else if(gnuRet == GNUTLS_SAN_DNSNAME) {
+			bHaveSAN = 1;
 			dbgprintf("subject alt dnsName: '%s'\n", szAltName);
 			snprintf((char*)lnBuf, sizeof(lnBuf), "DNSname: %s; ", szAltName);
 			CHKiRet(rsCStrAppendStr(pStr, lnBuf));
@@ -1032,8 +1034,8 @@ gtlsChkPeerName(nsd_gtls_t *pThis, gnutls_x509_crt_t *pCert)
 		++iAltName;
 	}
 
-	if(!bFoundPositiveMatch) {
-		/* if we did not succeed so far, we try the CN part of the DN... */
+	/* Check also CN only if not configured per stricter RFC 6125 or no SAN present*/
+	if(!bFoundPositiveMatch && (!pThis->bSANpriority || !bHaveSAN)) {
 		CHKiRet(gtlsGetCN(pCert, &pstrCN));
 		if(pstrCN != NULL) { /* NULL if there was no CN present */
 			dbgprintf("gtls now checking auth for CN '%s'\n", cstrGetSzStrNoNULL(pstrCN));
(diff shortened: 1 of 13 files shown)


jvymazal:
@rgerhards please check this from a design perspective, if it is ok I will create a doc PR to complement this.

Thanks

rgerhards:
Not done full review yet, but it looks CI is unhappy for a reason. Also we need to add support for this also to the openssl driver.

rgerhards:
@jvymazal question for you: I have some refactoring due in https://github.com/rsyslog/rsyslog/issues/3727 - originally thought to have a first try today, but it interferes with your PR. I think the parameter passing interface will become much cleaner when the refactoring is done. Do you prefer me to hold this work or would you be willing to adapt to changes I potentially do? In any way, I'll wait for the CI outcome, maybe I can merge soon :-)

jvymazal:
Any idea how many changes would it require? After few more details it looks like current state might actually work, I plan to do some functional testing today and if it is good honestly I would prefer to merge it and do the changes in next PR (perhaps along with OSSL implementation - for that I first need to study how ossl alternative to these gnutls checks actually looks).

rgerhards:
If we are close to merge, I'll hold my work. The base idea in the refactoring is to ensure that all parameters can be set on a per listener basis in contrast to the module level right now. This requires a refactoring of all the interfaces. A rough outline can be found here: https://github.com/rsyslog/rsyslog/issues/3727#issuecomment-525705318

runtime/nsd_gtls.h
```diff
@@ -99,4 +104,11 @@ rsRetVal gtlsRecordRecv(nsd_gtls_t *pThis);
 #define GTLS_ANON_PRIO "NORMAL:+ANON-DH:+ANON-ECDH:+COMP-ALL"
 #endif
 
+#if GNUTLS_VERSION_MAJOR > 3 || (GNUTLS_VERSION_MAJOR == 3 && GNUTLS_VERSION_MINOR >=3)
+#define EXTENDED_CERT_CHECK_AVAILABLE
+#endif
+
+#define GTLS_KP_SERVER "1.3.6.1.5.5.7.3.1"
```
ueno: Why not use `GNUTLS_KP_TLS_WWW_SERVER` and `GNUTLS_KP_TLS_WWW_CLIENT` directly?

jvymazal: Is it identical functionality-wise? If yes, I will change it, that way it will be more readable for others.

jvymazal: Ok, found it at:
https://github.com/gnutls/gnutls/blob/27ec82e781b3960693cd4550440ab7c1664553b8/lib/includes/gnutls/x509.h#L74
will update PR

jvymazal:
@rgerhards, from initial testing it seems that we are good, so if it looks good to you as well we should be ready to merge.

runtime/nsd_gtls.c
```diff
@@ -1154,8 +1156,23 @@ gtlsChkPeerCertValidity(nsd_gtls_t *pThis)
...
+			data.data = (uchar *)GNUTLS_KP_TLS_WWW_SERVER;
+		} else { /* server mode */
+			data.data = (uchar *)GNUTLS_KP_TLS_WWW_CLIENT;
+		}
+		data.size = ustrlen(data.data);
+		CHKgnutls(gnutls_certificate_verify_peers(pThis->sess, &data, 1, &stateCert));
+	}
+#endif
 
 	if(stateCert & GNUTLS_CERT_INVALID) {
```
ueno, line 1177: Perhaps it might make sense to check `GNUTLS_CERT_PURPOSE_MISMATCH` below for diagnostic purposes?

jvymazal: good catch, will add it

rgerhards:
looks like we need a configure check...

jvymazal:
Oh, I see now, while the core checking functionality was introduced in 3.3.0, the error reporting was added on only in 3.4.0, will update the ifdef macro

tammar96:
@rgerhards How can I enable check for extendedkeyusage in config file? I am not able to find it in the code. Is it something like `$InputTCPServerStreamDriverModeCheckExtendedKeyPurpose 1` ?

davidelang:
new features are not going to get legacy config options, so it's more likely to 
be a module load time parameter

David Lang

rgerhards:
There is a doc PR mentioned in this thread. See there: https://github.com/rsyslog/rsyslog-doc/pull/842 -- I merged this, so the current master branch doc up on rsyslog.com should also hold it.

PR #3869, opened 2019-09-19
fix transactional on remove HardErrors

kortemik:
fixes failpath on actionTryRemoveHardErrorsFromBatch happens only when

actionCommit ()
-> actionTryCommit => FAIL
-> actionCommit somewhat unhappy
-> actionTryRemoveHardErrorsFromBatch() // contains memcpy which did work with non allocated area and non-element size alligned copy
-> actionCommit unhappy, has still uncommited messages.
-> actionTryCommit // would pass the rogue memory range to module

### DIFF
--- a/action.c
+++ b/action.c
@@ -1495,7 +1495,7 @@ actionTryRemoveHardErrorsFromBatch(action_t *__restrict__ const pThis, wti_t *__
 			sizeof(actWrkrIParams_t) * pThis->iNumTpls);
 		ret = actionTryCommit(pThis, pWti, oneParamSet, 1);
 		if(ret == RS_RET_SUSPENDED) {
-			memcpy(new_iparams + *new_nMsgs, &oneParamSet,
+			memcpy(new_iparams + (*new_nMsgs * pThis->iNumTpls), &oneParamSet,
 				sizeof(actWrkrIParams_t) * pThis->iNumTpls);
 			++(*new_nMsgs);
 		} else if(ret != RS_RET_OK) {
@@ -1515,7 +1515,7 @@ actionCommit(action_t *__restrict__ const pThis, wti_t *__restrict__ const pWti)
 {
 	actWrkrInfo_t *const wrkrInfo = &(pWti->actWrkrInfo[pThis->iActionNbr]);
 	/* Variables that permit us to override the batch of messages */
-	unsigned nMsgs;
+	unsigned nMsgs = 0;
 	actWrkrIParams_t *iparams = NULL;
 	int needfree_iparams = 0; // work-around for clang static analyzer false positive
 	DEFiRet;


kortemik:
@rgerhards I think this one was quite hard to find bug. It's affecting the return codes of omfile if dynafile is used in rare cases. Mergeable?

rgerhards:
@kortemik great find - still a bit puzzled that the standard allocator does create unaligned address on stack. However, can't we fix this by specifying proper alignment attribute on the stack var? I really can't believe memcpy will not work on stack data.

kortemik:
I think the root cause is mixing of CONF_OMOD_NUMSTRINGS_MAXSIZE with dynamically allocated transactional part of the union in actWrkrInfo_t (runtime/wti.h#L54) so that's why I chose to use malloc as the size is not really known until runtime. The union may be also the root cause why compiler chose such an allocation scheme.

kortemik:
in any case, the more alarming issue here is the destination addressing of memcpy which was fixed by adding the template amount into calculation, omfile uses 2 different templates so it caused troubles for failing messages when we tried to implement queues to those with omfile-hardened:
```
new_iparams + (*new_nMsgs * pThis->iNumTpls)
```
However there still is an issue that queues get corrupt sometimes, but now at least the module gets retry with proper content in memory.

action.c
```diff
@@ -1484,24 +1484,27 @@ actionTryRemoveHardErrorsFromBatch(action_t *__restrict__ const pThis, wti_t *__
...
 	*new_nMsgs = 0;
 	for(unsigned i = 0 ; i < nMsgs ; ++i) {
 		setActionResumeInRow(pWti, pThis, 0); // make sure we do not trigger OK-as-SUSPEND handling
-		memcpy(&oneParamSet, &actParam(wrkrInfo->p.tx.iparams, pThis->iNumTpls, i, 0),
+		memcpy(oneParamSet, &actParam(wrkrInfo->p.tx.iparams, pThis->iNumTpls, i, 0),
 			sizeof(actWrkrIParams_t) * pThis->iNumTpls);
 		ret = actionTryCommit(pThis, pWti, oneParamSet, 1);
 		if(ret == RS_RET_SUSPENDED) {
-			memcpy(new_iparams + *new_nMsgs, &oneParamSet,
+			memcpy(new_iparams + (*new_nMsgs * pThis->iNumTpls), oneParamSet,
```
rgerhards: I tend to think this is the real bug and it stems back to the invalid not-use of the param access macro (which the new code mimics).

rgerhards:
@kortemik I think we do not need to go away from the stack alloc as it is - it's all within well defined C standards. Would you be willing to verify a candidate patch with just the minimal change?

kortemik:
@rgerhards of course! stack is faster too, and if the static amount of templates is intended for transactional om-modules then it's ok to leave it as-is.

rgerhards:
> , and if the static amount of templates is intended for transactional om-modules then it's ok to leave it as-is.

the max number of templates is a fixed constant inside rsyslog. No plugin can have more than this max.

rgerhards:
@kortemik my proposed patch is obviously wrong - CI already told me ;-) It would still be good if you could make just the minimal change to your PR and verify that it addresses your problem.

action.c
kortemik: i agree, now the PR is modified just to include this and one trivial initialization fix.

action.c
```diff
@@ -1515,7 +1515,7 @@ actionCommit(action_t *__restrict__ const pThis, wti_t *__restrict__ const pWti)
 {
 	actWrkrInfo_t *const wrkrInfo = &(pWti->actWrkrInfo[pThis->iActionNbr]);
 	/* Variables that permit us to override the batch of messages */
-	unsigned nMsgs;
```
rgerhards, line 1518: do you have any specific need for this other than cosmetic? I ask because the variable is always set before being used.

action.c
rgerhards: thx a lot, much appreciated!

action.c
kortemik: code quality checks failed some of the builds when it was without initializer, i think freebsd handled warnings as errors and failed the build without this, so i had to add it.

rgerhards: thx, so it looks like some compiler is doing false positives (like in the case 2 lines below).

Issue #3872, opened 2019-09-24
msg.c: check and correct default for msg offset value. 


alorbach:
Commit https://github.com/rsyslog/rsyslog/commit/1255a67fdec2fc44cd49b6ea8c463f4319910812 changed the default for pM->offMSG from -1 to 0. 
Commit https://github.com/rsyslog/rsyslog/commit/73506acff23f144d277fa811aa60bf4a290b9875 again reversed the changed from 0 to -1, but to change properly was unintentional. 

Change the default back from -1 to 0 now, as a default of -1 can cause unexpected results if offMSG is not set before used.

alorbach:
Closed by 
https://github.com/rsyslog/rsyslog/pull/5794

PR #3873, opened 2019-09-24
msg.c: set default for offMSG back to 0.

alorbach:
closes: https://github.com/rsyslog/rsyslog/issues/3872

Commit 1255a67 changed the default for pM->offMSG from -1 to 0.
Commit 73506ac again reversed the changed from 0 to -1, but to change properly was unintentional.

This PR changes the default back from -1 to 0 now, as a default of -1 can cause unexpected results if offMSG is not set before used.


alorbach:
New test imudp-pmdb2diag-offset.sh from https://github.com/rsyslog/rsyslog/pull/3876 won't fail in this PR.

rgerhards:
final conclusion: we do not need to do this; the problem this PR tries to address was a bug in a different component.

PR #3876, opened 2019-09-25
testbench: Added test for imudp and pmdb2diag to trigger offMSG bug

alorbach:
This test should trigger a FAIL in current master branch if pmdb2diag is used and compiled with clang and address sanitizer is being used. See also: https://github.com/rsyslog/rsyslog/pull/3873

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -1319,7 +1319,8 @@ endif
 
 if ENABLE_PMDB2DIAG
 TESTS += \
-	pmdb2diag_parse.sh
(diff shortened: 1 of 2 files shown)


alorbach:
test fails as supposed to here:
https://build.rsyslog.com/#/builders/138/builds/2696/steps/4/logs/stdio
https://build.rsyslog.com/#/builders/138/builds/2696/steps/4/logs/test-suite_log

rgerhards:
finally not merging as this bug was in a different region and the proposed check does not even work now.

PR #3878, opened 2019-09-26
plugin code restructuring, added remote option

jvymazal:
Decomposed ReadJournal() a bit, also now coupling journald
variables in one struct, added few warning messages and debug
prints to help with bug hunts in future, also got rid of two
needless journald calls. WorkAroundJournalBug now deprecated.
Added option to pull journald records from outside local machine.

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -81,8 +81,9 @@ static struct configSettings_s {
 	int iDfltFacility;
 	int bUseJnlPID;
 	char *usePid;
-	int bWorkAroundJournalBug;
+	int bWorkAroundJournalBug; /* deprecated, left for backwards compatibility only */
 	int bFsync;
+	int bRemote;
 } cs;
 
 static rsRetVal facilityHdlr(uchar **pp, void *pVal);
@@ -100,7 +101,8 @@ static struct cnfparamdescr modpdescr[] = {
 	{ "usepidfromsystem", eCmdHdlrBinary, 0 },
 	{ "usepid", eCmdHdlrString, 0 },
 	{ "workaroundjournalbug", eCmdHdlrBinary, 0 },
-	{ "fsync", eCmdHdlrBinary, 0 }
+	{ "fsync", eCmdHdlrBinary, 0 },
+	{ "remote", eCmdHdlrBinary, 0 }
 };
 static struct cnfparamblk modpblk =
 	{ CNFPARAMBLK_VERSION,
@@ -120,8 +122,6 @@ static prop_t *pLocalHostIP = NULL;	/* a pseudo-constant propterty for 127.0.0.1
 static const char *pidFieldName;	/* read-only after startup */
 static int bPidFallBack;
 static ratelimit_t *ratelimiter = NULL;
-static sd_journal *j;
-static sbool reloaded = 0;
 static struct {
 	statsobj_t *stats;
 	STATSCOUNTER_DEF(ctrSubmitted, mutCtrSubmitted)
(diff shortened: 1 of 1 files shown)


jvymazal:
In scope of this I found two sd_journal_*() calls which were remnant from old code so those are removed.

Also this closes #3707

jvymazal:
@rgerhards  I am going trough CI and finding strange reports of leaks associated with JSON objects. I have not changed anything in JSON reading code except taking it whole to separate function for better readability. Do you please know how are the fjson_object supposed to be freed?

Thanks

rgerhards:
Soin boarding plane. Some json*put() function.

rgerhards:
@jvymazal did my brief message help?

jvymazal:
@rgerhards if you could just briefly look on second commit (I will remove/squish it later), I cannot understand how just moving the json-related code to separate func with no actual code changes can introduce quite massive (test reported over 20MB) memory leak.

Any pointers much appreciated,

rgerhards:
will have a look - might not happen today, as I need to prepare s/t that a co-worker needs in a presentation tomorrow ;-)

rgerhards:
I had a quick look nevertheless - I agree it looks strange.  Did the problem consistently appear?

jvymazal:
Yes, I tried several version of `json*put()` and it made no difference on the result.

rgerhards:
The leak goes away 

a) when the code is in separate function
b) it is merged into the main function

I did the review suspecting b), but now see that I have possibly mis-read your posting. In any case it looks pretty strange.

jvymazal:
The leak appeared when I separated the code (1st commit), when I merge it back (testing 2nd commit) it goes away, however that I cannot imagine how separating the code which has no ties to rest of original function can cause it.

rgerhards:
> The leak appeared when I separated the code (1st commit), when I merge it back (testing 2nd commit) it goes away, however that I cannot imagine how separating the code which has no ties to rest of original function can cause it.

full ack. Could you reproduce the issue locally? With valgrind and with clang address sanitizer?

jvymazal:
With valgrind output seems to roughly correspond to CI run.

rgerhards:
@jvymazal sry, still been too busy today. Will retry tommorrow, but have a small 1hr time slot :( Next week is hopefully much better (weekend also pretty full...)

rgerhards:
@jvymazal pls have a look: when I run the commit in CI, I see some aborts, e.g. https://build.rsyslog.com/#/builders/175/builds/2376

I just removed the "testing commit". Later in the log I also see a mem leak being reported, but I guess that's just a side-effect of the misadressing. Did you see the same?

jvymazal:
ok, I have made little tweak, now the misaddressing can be clearly seen, looks like I have overdone the memory guards.

jvymazal:
Now it looks much better, unless you see some issues with the changes we should be good to merge.

rgerhards:
I agree, just let's wait for the CI flake (for cosmetic reasons). side-note: I have just addressed the most annoying flake ;-)

rgerhards:
@jvymazal I don't know what is going on here, but the commit introduces a regression that CI did not detect in the initial run. Now it fails consistently, see https://build.rsyslog.com/#/builders/175/builds/2399/steps/5/logs/stdio

 I ran git bisect and it points to this commit. Looks like a problem with state file name generation.

plugins/imjournal/imjournal.c
```diff
@@ -615,6 +673,9 @@ loadJournalState(void)
 	int r;
 	FILE *r_sf;
 
+	DBGPRINTF("Loading journal position, at head? %d, reloaded? %d\n",
+			  journalContext.atHead, journalContext.reloaded);
+
 	if (cs.stateFile[0] != '/') {
```
rgerhards, line 679: This seems now to be called multiple times and continuously adds the work dir if it is not an absolute path. The modification of the config variable here looks strange in any case.

rgerhards: The more general problem is that the `cs.` obsolete legacy structure is "abused" for the new-style config. This makes things pretty complicated. Would best be refactored as well.

rgerhards:
I guess we did not see the regression in the first runs because the journal was not sufficiently busy to cause multiple calls to loadJournalState(). PR #3902 should fix the issue.

plugins/imjournal/imjournal.c
jvymazal: Thanks for suggestion, will try to not forget it in next commit.

PR #3879, opened 2019-09-26
print also ratelimiter parameters

jvymazal:
this could help users to find out exactly how is ratelimiter set
when inspecting logs so then can easier adjust configuration


rgerhards:
CodeCov is false positive, so good to merge.

PR #3896, opened 2019-10-14
Test to support bug tracking for imjournal-refactoring PR 3878

rgerhards:
testing PR to check for issues seen in #3878


rgerhards:
no longer needed, did what it was created for. PR #3878 is now fixed.

PR #3902, opened 2019-10-16
imjournal: fix regression

rgerhards:
commit 78976a9bc059 introduced a regression that caused writing
the journal state file to fail. This happens when the state file
is given as relative file name and the working directory is also
a relative path. This situation is very uncommon. So most deployments
will never experience it. We discovered the issue during CI runs
where the trigger condition is given. Note that it also takes
multiple times of loading the journal to actually see the bug.

see also https://github.com/rsyslog/rsyslog/pull/3878


jvymazal:
Good job at spotting this.

Issue #3853, opened 2019-08-30
Python 2 removal (in Debian sid/bullseye)


mbiebl:
Bug-Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=938417

Copied verbatim from the downstream bug report:

> Python2 becomes end-of-live upstream, and Debian aims to remove
> Python2 from the distribution, as discussed in
> https://lists.debian.org/debian-python/2019/07/msg00080.html
> 
> Your package either build-depends, depends on Python2, or uses Python2
> in the autopkg tests.  Please stop using Python2, and fix this issue
> by one of the following actions.
> 
> - Convert your Package to Python3. This is the preferred option.  In
>   case you are providing a Python module foo, please consider dropping
>   the python-foo package, and only build a python3-foo package.  Please
>   don't drop Python2 modules, which still have reverse dependencies,
>   just document them.
>   
>   This is the preferred option.
> 
> - If the package is dead upstream, cannot be converted or maintained
>   in Debian, it should be removed from the distribution.  If the
>   package still has reverse dependencies, raise the severity to
>   "serious" and document the reverse dependencies with the BTS affects
>   command.  If the package has no reverse dependencies, confirm that
>   the package can be removed, reassign this issue to ftp.debian.org,
>   make sure that the bug priority is set to normal and retitle the
>   issue to "RM: PKG -- removal triggered by the Python2 removal".
> 
> - If the package has still many users (popcon >= 300), or is needed to
>   build another package which cannot be removed, document that by
>   adding the "py2keep" user tag (not replacing the py2remove tag),
>   using the debian-python@lists.debian.org user.  Also any
>   dependencies on an unversioned python package (python, python-dev)
>   must not be used, same with the python shebang.  These have to be
>   replaced by python2/python2.7 dependencies and shebang.
> 
>   This is the least preferred option.
> 
> If the conversion or removal needs action on another package first,
> please document the blocking by using the BTS affects command, like
> 
>   affects <bug number of blocking py2removal bug> + src:rsyslog
> 
> If there is no py2removal bug for that reverse-dependency, please file
> a bug on this package (similar to this bug report).
> 
> If there are questions, please refer to the wiki page for the removal:
> https://wiki.debian.org/Python/2Removal, or ask for help on IRC
> #debian-python, or the debian-python@lists.debian.org mailing list.

Afaics, this only affects the test-suite.

Whissi:
[...]

PR #3911, opened 2019-10-17
INVLD: testbench bugfix: use of python 2 in testbench

rgerhards:
see also https://github.com/rsyslog/rsyslog/issues/3853


rgerhards:
A partial solution will possibly be applied via https://github.com/rsyslog/rsyslog/pull/3984 and than this PR closed. Remaining problems to be addressed in new PR.

Issue #3885, opened 2019-10-01
asynwr_deadlock2 fails on s390x, sparc64, hppa, ppc64, mips, powerpc


mbiebl:
Version: 8.1908.0

Since 8.1908.0, rsyslog fails to build on s390x on Debian, specifically it fails during `make check` when running the test-suite.

```
FAIL: asynwr_deadlock2
======================

testbench: TZ env var not set, setting it to UTC
------------------------------------------------------------
12:32:33[0]  Test: ./asynwr_deadlock2.sh
------------------------------------------------------------
config rstb_171544_e7e00756_.conf is:
     1	module(load="../plugins/imdiag/.libs/imdiag")
     2	global(inputs.timeout.shutdown="60000"
     3	       default.action.queue.timeoutshutdown="20000"
     4	       default.action.queue.timeoutEnqueue="20000")
     5	$MainmsgQueueTimeoutEnqueue 20000 # use legacy-style so that we can override if needed
     6	$IMDiagListenPortFileName rstb_171544_e7e00756.imdiag.port
     7	$IMDiagServerRun 0
     8	$IMDiagAbortTimeout 580
     9	
    10	:syslogtag, contains, "rsyslogd"  ./rstb_171544_e7e00756.started
    11	###### end of testbench instrumentation part, test conf follows:
    12	
    13	$ModLoad ../plugins/imtcp/.libs/imtcp
    14	$MainMsgQueueTimeoutShutdown 10000
    15	$InputTCPServerRun 45093
    16	
    17	$template outfmt,"%msg:F,58:3%,%msg:F,58:4%,%msg:F,58:5%\n"
    18	$template dynfile,"rstb_171544_e7e00756.%msg:F,58:2%.log" # use multiple dynafiles
    19	
    20	$OMFileFlushOnTXEnd on
    21	$OMFileFlushInterval 10
    22	$OMFileIOBufferSize 10k
    23	$OMFileAsyncWriting on
    24	$DynaFileCacheSize 4
    25	local0.* ?dynfile;outfmt
rsyslogd: NOTE: RSYSLOG_DEBUG_TIMEOUTS_TO_STDERR activated
rsyslogd: info: imdiag does 20 empty checks due to CI_SHUTDOWN_QUEUE_EMPTY_CHECKS
main Q:Reg: worker start requested, num workers currently 0
main Q:Reg: wrkr start initiated with state 0, num workers now 1
[...]

PR #3921, opened 2019-10-22
WiP: omfile bugfix: stream class does not close re-opened fd

rgerhards:
depending on timing the stream class may need to re-open an fd
during desctruction of the strm object. This fd was never closed.
Trigger:
- async writing
- dynafiles

see also https://github.com/rsyslog/rsyslog/issues/3885


rgerhards:
fixed via #3922

PR #3922, opened 2019-10-23
omfile bugfix: file handle leak

rgerhards:
The stream class does not close re-opened file descriptors.
This lead to leaking file handles and ultimately to the inability
to open any files/sockets/etc as rsyslog ran out of handles.

The bug was depending on timing. This involed different OS
thread scheduler timing as well as workload. The bug was more
common under the following conditions:

- async writing of files
- dynafiles
- not commiting file data at end of transaction

However it could be triggerred under other conditions as well.

The refactoring done in 8.1908 increased the likelyhood of
experienceing this bug. But it was not a real regression, the new
code was valid, but changed the timing so that the race was more
likely.

Thanks to Michael Biebl for reporting this bug and helping to
analyze it.

closes https://github.com/rsyslog/rsyslog/issues/3885

**probably replaces #3921**

### DIFF
--- a/runtime/stream.c
+++ b/runtime/stream.c
@@ -448,6 +448,19 @@ strmWaitAsyncWriterDone(strm_t *pThis)
 	}
 }
 
+/* stop the writer thread (we MUST be runnnig asynchronously when this method
+ * is called!). Note that the mutex must be locked! -- rgerhards, 2009-07-06
+ */
+static void
+stopWriter(strm_t *const pThis)
+{
+	pThis->bStopWriter = 1;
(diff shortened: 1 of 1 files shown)


mbiebl:
@rgerhards I've pulled this patch into Debian 
https://salsa.debian.org/debian/rsyslog/commit/67a5f0ed7f80ff1eddc03a2d6c2ffdd56346d9a9

The results look good so far
https://buildd.debian.org/status/logs.php?pkg=rsyslog&ver=8.1910.0-2&suite=sid

rgerhards:
@mbiebl Thx! I've just merged this as final fix.

PR #3927, opened 2019-10-24
testbench: add test for async dynafile writer

rgerhards:
This models a problem from practice which was pretty hard
to reproduce.

Note: this  **MUST** initially fail without the proper patch for #3686. 

see also https://github.com/rsyslog/rsyslog/issues/3885


rgerhards:
good - without patch, the test fails. Now waiting until patch is merged (may take some time).

PR #3932, opened 2019-10-25
MMDARWIN:: Optimisations, new parameters, update to protocol header

frikilax:
#Optimisations
- use permanent worker-dependent buffers to avoid malloc/free for each entry
- move socket structures to worker data, remove global mutex
- add log lines for parameters and general workflow
- don't send body if empty/incomplete (see new parameters)
- don't close/reopen socket every time -> let session open or create new every X entry (see new parameters)
- clean up code

#New parameters
- added 'send_partial', to let mmdarwin send body if not all fields were retrieved, or not
  default false = only send complete bodies
- added 'socket_max_use' to open new session every X packet, useful for some versions of Darwin (prior to 1.1)
  default is 0 = do not open new session/keep only one

#Update to protocol
- added 'evt_id' to the darwin header (Darwin v1+ compatibility)

### DIFF
--- a/contrib/mmdarwin/mmdarwin.c
+++ b/contrib/mmdarwin/mmdarwin.c
@@ -37,6 +37,8 @@
 #include <string.h>
 #include <sys/un.h>
 #include <sys/socket.h>
+#include <uuid/uuid.h>
+#include <json.h>
 
 #include "protocol.h" /* custom file written for Darwin */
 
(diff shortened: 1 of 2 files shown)


rgerhards:
@frikilax just in case you have not seen:

* some commented-out code needs to be removed
* static analyzer issue http://ubuntu16.rsyslog.com/19-10-28_20-59-02/2019-10-28-205951-8998-1/ needs to be addressed
* I did not check further, there may be more things revealed by CI (but may also be some flakes)

Issue #3784, opened 2019-08-01
imjournal plugin truncates messages to 64K regardless of MaxMessageSize or journald's LineMax


rwdalpe:
### Expected behavior

If journald and rsyslog have both been configured to allow for messages longer than 64K and the imjournal module is being used, then rsyslog should receive the full content of messages longer than 64K when they're read from the journal (up to the $MaxMessageSize limit).

### Actual behavior

Any message from the journal longer than 64K is truncated to 64K by imjournal, regardless of configuration.

### Steps to reproduce the behavior

1. Configure journald to allow for long lines. Sample `/etc/systemd/journald.conf`:

```
[Journal]
LineMax=100K
```

2. Configure rsyslog to use imjournal and allow for similarly large messages:

```
$MaxMessageSize 102400 
...
$ModLoad imjournal
...
$IMJournalStateFile imjournal.state
# Make sure the systemd unit that you want to use for testing is logging to disk so we can see results. Here's a catch-all as an example:
#*.*;auth,authpriv.none        -/var/log/syslog
```

3. Have any systemd unit that logs to journald produce a log line around 100K in size
4. Check the output file on disk and see that it was truncated to 64K

### Environment
- rsyslog version: 8.24.0
- platform: Amazon Linux 2
- systemd version: 219

### Notes

I think this is happening because of [how imjournal calls sd_journal_get_data](https://github.com/rsyslog/rsyslog/blob/master/plugins/imjournal/imjournal.c#L312).

As described at https://www.freedesktop.org/software/systemd/man/sd_journal_get_data.html :

>Also note that, by default, data fields larger than 64K might get truncated to 64K. This threshold may be changed and turned off with sd_journal_set_data_threshold() (see below).

Maybe a fix is as straightforward as calling sd_journal_set_data_threshold with the value of the $MaxMessageSize when the journal is opened.

davidelang:
8.24 is about 2.5 years old, there have been some pretty significant 
modificatiosn to imjournal over that time. Could you please check that the 
problem still exists with a current version? (at least 8.1904)

David Lang

rgerhards:
@jvymazal can you pls have a look

jvymazal:
will for sure needs verification but in theory this sound about right

jvymazal:
> 8.24 is about 2.5 years old, there have been some pretty significant modificatiosn to imjournal over that time. Could you please check that the problem still exists with a current version? (at least 8.1904) David Lang

[...]

PR #3936, opened 2019-10-30
#3784: set the journal data threshold to MaxMessageSize

rwdalpe:
When data is read from the journal using sd_journal_get_data it may be
truncated to a certain threshold (64K by default).

If the rsyslog MaxMessageSize is larger than the threshold, there's a
chance rsyslog will receive incomplete messages from the journal.
Empirically, this appears to happen reliably when XZ compression is
used by journald. Systems where journald uses LZ4 compression do not
appear to suffer this issue reliably--if at all.

This change sets the threshold to the MaxMessageSize when the
journal is opened.


rwdalpe:
Whoops, my merge went a bit haywire when bringing in the latest changes. I'll fix this up.

Issue #3939, opened 2019-10-31
unable to configure tls1.2 using openssl 1.0.2t


tapakund:
### Expected behavior
I have building rsyslog version 8.1910.0 with the below configs:
```
./configure \
    --prefix=%{_prefix} \
    --enable-relp \
    --enable-gnutls\
    --enable-imfile \
    --enable-imjournal \
    --enable-impstats \
    --enable-imptcp \
    --enable-openssl
```
Below is the rsyslog config i am adding to disable TLS1 and TLS1.1.
```
module( load="imtcp"
        StreamDriver.Name="ossl"
        StreamDriver.Mode="1"
        StreamDriver.AuthMode="anon"
        gnutlsPriorityString="Protocol=ALL,-SSLv2,-TLSv1.0,-TLSv1.1
        MinProtocol=TLSv1.2"
       )
input(type="imtcp" port="1514")
```
So, i am getting error saying Openssl is to old for setting gnutlsPriorityString.

But as per this documentation: https://www.rsyslog.com/doc/v8-stable/configuration/modules/imtcp.html

It says:

gnutlsPriorityString
type	default	mandatory	obsolete legacy directive
string	none	no	none
New in version 8.29.0.

This strings setting is used to configure driver specific properties. Historically, the setting was only meant for gnutls driver. However with version v8.1905.0 and higher, the setting can also be used to set openssl configuration commands.

For GNUTls, the setting specifies the TLS session’s handshake algorithms and options. These strings are intended as a user-specified override of the library defaults. If this parameter is NULL, the default settings are used. More information about priority Strings here

For OpenSSL, the setting can be used to pass configuration commands to openssl libray. OpenSSL Version 1.0.2 or higher is required for this feature. A list of possible commands and their valid values can be found in the documentation:

I  am using Openssl 1.0.2t version.

root@tkundu-photon-1 [ /WS/photon-1.0/photon/stage/RPMS/x86_64 ]# openssl version
OpenSSL 1.0.2t-fips  10 Sep 2019
root@tkundu-photon-1 [ /WS/photon-1.0/photon/stage/RPMS/x86_64 ]#

So, i modified this line to make sure this code is executed with openssl 1.0.2t.
https://github.com/rsyslog/rsyslog/blob/1771f861203f67b57c340884c40c880714fc2786/runtime/nsd_ossl.c#L1795

### Actual behavior

After doing this change, no TLS version support i could see in the tcp port selected in imtcp.

### Environment
[...]

PR #3944, opened 2019-11-04
ossl driver: fix wrong OpenSSL Version check (OPENSSL_VERSION_NUMBER)

alorbach:
Fixed OpenSSL Version check in:
- SetGnutlsPriorityString function in nsd_ossl.c
- initTLS() function tcpflood.c

See https://www.openssl.org/docs/man1.1.0/man3/OPENSSL_VERSION_NUMBER.html
for more.

closes https://github.com/rsyslog/rsyslog/issues/3939


rgerhards:
CodeCov definitely looks like a false positive, so merging.

PR #3953, opened 2019-11-08
Report access error only on files

jvymazal:
Leaving reporting access for directories in DEBUG only as rsyslog
goes all the way to root and paths higher in filesystem are
usually blocked by selinux policy rules.


jvymazal:
is it possible to include this in 8.1911.0 please?

rgerhards:
will try - I currently do not understand why it seems to consistenly fail the same CI check.

rgerhards:
ubuntu 18 distcheck fail is a false positive.

PR #3962, opened 2019-11-13
relp: Add support setting openssl configuration commands.

alorbach:
Add new configuration parameter tls.tlscfgcmd to omrelp and imrelp.
(Using relpSrvSetTlsConfigCmd and relpCltSetTlsConfigCmd)

OpenSSL Version 1.0.2 or higher is required for this feature.
A list of possible commands and their valid values can be found in the
documentation: https://www.openssl.org/docs/man1.0.2/man3/SSL_CONF_cmd.html

The setting can be single or multiline, each configuration command is
separated by linefeed (n). Command and value are separated by
equal sign (=). Here are a few samples:

tls.tlscfgcmd="Protocol=ALL,-SSLv2,-SSLv3,-TLSv1,-TLSv1.2"

tls.tlscfgcmd="Protocol=ALL,-SSLv2,-SSLv3,-TLSv1
MinProtocol=TLSv1.2"

Add to new testcases for librelp and tlscfgcmd.

### DIFF
--- a/plugins/imrelp/imrelp.c
+++ b/plugins/imrelp/imrelp.c
@@ -96,6 +96,9 @@ struct instanceConf_s {
 	uchar *caCertFile;
 	uchar *myCertFile;
 	uchar *myPrivKeyFile;
+#if defined(HAVE_RELPENGINESETTLSCFGCMD)
+	uchar *tlscfgcmd;
+#endif
 	int iKeepAliveIntvl;
 	int iKeepAliveProbes;
 	int iKeepAliveTime;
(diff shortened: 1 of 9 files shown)


plugins/imrelp/imrelp.c
```diff
@@ -618,6 +631,10 @@ CODESTARTnewInpInst
 			} else {
 				fclose(fp);
 			}
+#if defined(HAVE_RELPENGINESETTLSCFGCMD)
+		} else if(!strcmp(inppblk.descr[i].name, "tls.tlscfgcmd")) {
```
rgerhards, line 634: this should be changed so that the parameter is not ignored if it is not supported - that surprises users. This is done for parameter "address" just a couple of lines above. I also suggest to indent the preprocessor as done there (easier to read IMHO).

alorbach: Should be fixed now.

PR #3967, opened 2019-11-15
Enhance mmutf8fix handling of incorrect UTF-8 sequences

turchanov:
1. Invalid utf8 detection didn't handle 3 and 4-byte overlong encodings (2 byte overlong econdings were handled explicitly by rejection E0 and E1 start bytes). Unified checks for overlong encodings.
2. Surrogates U+D800..U+DFFF are not valid codepoints (Unicode Standard, D92)
3. Replacement of characters in invalid 3 or 4-bytes encodings was too eager. It must not replace bytes which are valid UTF-8 sequences. For example, in [0xE0 0xC2 0xA7] sequence the 0xC2 is invalid as a continuation byte, but it starts a valid UTF8 symbol [0xC2 0xA7]. That is, with current
code processing the sequence will result in "???" but the correct result is "?§" (provided that the replacement character is "?").
4. Various tests for UTF-8 invalid/valid sequences.

### DIFF
--- a/plugins/mmutf8fix/mmutf8fix.c
+++ b/plugins/mmutf8fix/mmutf8fix.c
@@ -206,16 +206,16 @@ doCC(instanceData *pData, uchar *msg, int lenMsg)
 
 /* fix an invalid multibyte sequence */
 static void
-fixInvldMBSeq(instanceData *pData, uchar *msg, int lenMsg, int strtIdx, int *endIdx, int8_t seqLen)
+fixInvldMBSeq(instanceData *pData, uchar *msg, int lenMsg, int strtIdx, int cnt)
 {
-	int i;
+	int i, endIdx;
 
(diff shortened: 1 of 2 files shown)


turchanov:
Hmm.... 
```
buildbot/rsyslog docker-ubuntu18-distcheck rsyslog
buildbot/rsyslog docker-ubuntu18-tsan rsyslog
```
Fail these tests:
FAIL: libdbi-asyn.sh
FAIL: mysql-asyn.sh

@rgerhards, looks like the cause is outside of the domain of my changes. What should I do?

rgerhards:
> looks like the cause is outside of the domain of my changes. What should I do?

These seem to be flakes - I'll take care of them. Thx for the PR and checking it goes through well. Should your attention actually be needed, I'll let you know.

Issue #3969, opened 2019-11-16
imjournal: why is the delay 10s on error retry？


wsp1991:
### Steps to reproduce the behavior
in imjournal.c
```
static void
tryRecover(void) {
        LogMsg(0, RS_RET_OK, LOG_INFO, "imjournal: trying to recover from unexpected "
                "journal error");
        STATSCOUNTER_INC(statsCounter.ctrRecoveryAttempts, statsCounter.mutCtrRecoveryAttempts);
        closeJournal();
        srSleep(10, 0); // do not hammer machine with too-frequent retries
        openJournal();
}
```
here, srSleep(10, 0);Why is the delay 10 s？Can it be shortened or removed？
if produce log fastly and journald buffer set small, delay 10s, Logs will be lost.
### Environment
- rsyslog version: 
- platform: 
- for configuration questions/issues, include rsyslog.conf and included config files

rgerhards:
good question - it somehow looks like 10s was not the real intent...

PR #3970, opened 2019-11-16
imjournal bugfix: do not wait too long on recovery try

rgerhards:
When trying to recover journal errors, imjournal waited a hardcoded
period of 10s between tries. This was pretty long and could lead to
loss of journal data.

This commit adjust it to 100ms, which should still be fully sufficient
to prevent the journal from "hammering" the CPU.

It may be worth considering to make this setting configurable - but
let's first see if there is real demand to actually do that.

closes https://github.com/rsyslog/rsyslog/issues/3969

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -759,7 +759,7 @@ tryRecover(void) {
 	LogMsg(0, RS_RET_OK, LOG_INFO, "imjournal: trying to recover from journal error");
 	STATSCOUNTER_INC(statsCounter.ctrRecoveryAttempts, statsCounter.mutCtrRecoveryAttempts);
 	closeJournal();
-	srSleep(10, 0);	// do not hammer machine with too-frequent retries
+	srSleep(0, 200000);	// do not hammer machine with too-frequent retries
 	openJournal();
 }
 


wsp1991:
@rgerhards What do you mean about journal "hammering" the CPU？What's the situation？

rgerhards:
> What do you mean about journal "hammering" the CPU？What's the situation？

very frequent retries which cause 100% cpu utilization -- this has happened due to journal bugs in the past and given our experience we are sure it will happen in the future as well.

wsp1991:
@rgerhards I remove 10s delay；produce log fastly，if  execute tryRecover function，find Log discontinuity，some log loss.

rgerhards:
> @rgerhards I remove 10s delay；produce log fastly，if execute tryRecover function，find Log discontinuity，some log loss.

yeah, I think this is because the function tries to recover problems that the journal had. @jvymazal can you shed some light on such situations?

jvymazal:
I only now noticed it is actually a 10s sleep, I think in the original conversation (long, long time ago...) we talked about something like 10ms. Nonetheless the function is basically an error handler and should never be called in normal journald operation, only when some non-typical error happens.

jvymazal:
> @rgerhards I remove 10s delay；produce log fastly，if execute tryRecover function，find Log discontinuity，some log loss.

Not sure how exactly you are executing the `tryRecover()`, however some message loss is kind of expected there. The function was put there only to cover some cases of basically broken journal where it at least  allow us to continue logging future messages, it is bit like lightweight plugin reboot.

rgerhards:
> I only now noticed it is actually a 10s sleep, I think in the original conversation (long, long time ago...) we talked about something like 10ms.

Same with me ... I may have mistakenly applied it. 10s sounds way too long. This is now fixed thankfully.

PR #3980, opened 2019-11-20
omkafka: ensure math lib is bound to executable

rgerhards:
librdkafka needs it, and it seems not to be present inside compiled
binary. This is a try to integrate it via omkafka.

Note: I need to merge this in order to use the regular package build
process. If it does not work out, I will undo it via another commit.


rgerhards:
fail is a false positive, so OK to merge

PR #3981, opened 2019-11-20
ratelimit: increase rate limit interval parameter

3chas3:
The burst parameter in the ratelimit was increased to an unsigned int
but the interval was as an unsigned short. While it may be unusual,
there is possibly a chance to need to represent an interval longer than
about 3/4 of a day.

While here, go through and normalize all the various incarnations of
rate limiting to be explicitly unsigned int for the burst and interval.


rgerhards:
@3chas3 Thx for the PR!

FYI: CI is most probably some flakes - we are working on these issues. I'll let you know if something really needs your attention.

PR #3983, opened 2019-11-21
testbench: add some debugging instrumentation

rgerhards:
We have a long-standing issue with the mysql tests. We can't find the
root cause with special test runs. This commit now adds some debug
info. The idea is to merge this and run it in all upcoming tests, so
that we can hopefully get a sufficiently large sample that we can
address the problem.


rgerhards:
CI is false positive

PR #3432, opened 2019-01-20
New module IMPCAP & ommongodb improvement

KGuillemot:
#### Add new modules : 
 - **impcap** : Capture packets from network interfaces and play with them in rsyslog as JSON objects
 - **mmcapture** : Extract payloads of ipcap captured packets and store them on disk

#### Improve ommongodb : 
 - Let ommongodb module works event if mongo-c-driver is not compiled with SSL support.
 - Add parsing of "date" or "time" field as rfc3339 to insert as ISODate in MongoDB

PR #3987, opened 2019-11-22
OMMONGODB :: Add other supported formats for 'time' and 'date' fields

frikilax:
This replaces PR #3432 ommongodb part.

All credits goes to @KGuillemot.


rgerhards:
@frikilax do we need some doc update?

frikilax:
> @frikilax do we need some doc update?

@KGuillemot ?

frikilax:
> > @frikilax do we need some doc update?
> 
> I would say no, but I would like @KGuillemot to validate this

OK confirmed, this does not require a doc update

rgerhards:
> OK confirmed, this does not require a doc update

thx, so we are ready for merge :-)

PR #3993, opened 2019-11-26
core bugfix: action on retry mangles messages

rgerhards:
**submiting this PR primarily to do a full CI run and see if I overlook something**

When a failed action goes into retry, template content is rendered
invalid if the action uses more than 1 template.

This is a minimal fix based on Mikko Kortelainen's somewhat more
elaborate patch. All credits belong to Mikko who analyzed and fixed
this hard-to-find issue.

see also https://github.com/rsyslog/rsyslog/pull/3869


rgerhards:
did not work out - #3869 will provide the solution

PR #3994, opened 2019-11-26
omsnmp: Add support for new option "snmpv1dynsource"

alorbach:
If set, the source field from SNMPv1 trap can be overwritten
with a template, default is "%fromhost-ip%". The content should be a
valid IPv4 Address that can be passed to inet_addr(). If the content
is not a valid IPv4 Address, the source will not be set.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -179,6 +179,10 @@ if ENABLE_OMHTTP
 SUBDIRS += contrib/omhttp
 endif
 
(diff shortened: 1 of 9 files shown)


rgerhards:

alorbach:
Fixed DISTCHECK_CONFIGURE_FLAGS

alorbach:
Removed references to cs.pszSnmpV1DynSource

PR #4000, opened 2019-11-29
Handle journal invalidate

3chas3:
The journal API will sometimes issue an SD_JOURNAL_INVALIDATE
which results in a repositioning of the cursor. From the man page:

    Programs only interested in a strictly sequential stream
    of log data may treat SD_JOURNAL_INVALIDATE the same way as
    SD_JOURNAL_APPEND, thus ignoring any changes to the log view
    earlier than the old end of the log stream.

If we have configured IgnorePreviousMessages, we want to treat the
invalidate event as an append, and wind to the end of the message
stream.

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -619,6 +619,10 @@ handleRotation(void)
 	else if (cs.stateFile) {
 		iRet = loadJournalState();
 	}
+	else if (cs.bIgnorePrevious) {
+		/* Seek to the very end of the journal and ignore all older messages. */
+		skipOldMessages();
+	}
 	journalContext.reloaded = 1;
 
 finalize_it:


rgerhards:
@jvymazal can you pls have a quick look and let me know what you think - thx!

jvymazal:
Looks good, this is actually really edge case (in non-error conditions should never happen), which I forgot about during the rewrite.

3chas3:
> Looks good, this is actually really edge case (in non-error conditions should never happen), which I forgot about during the rewrite.

It's not exactly an edge case. systemd seems to signal invalidate on a fresh boot when rsyslog first starts. This is "correct" behavior, your application may not have seen any of the journal and should be informed of the total contents. However, we don't want that behavior because we are trying to emulate some older behavior in which syslog always gets the tail of the journal whenever it attaches.

We could switch to saving the journal pointer but that's something for another day.

rgerhards:
> We could switch to saving the journal pointer but that's something for another day.

This is actually done. That said, it now sounds to me like the consequences of this patch are probably not desired (loss of otherwise received messages). @jvymazal what do you say?

jvymazal:
Actually the consequences are desired, @3chas3  it is not an edge case in the sense of SD_JOURNAL_INVALIDATE event, that is indeed normal, expected and can be quite frequent. The proposed patch is an edge case in the sense that it is an `else` to `if` clause that should be always trough outside of some error scenarios, that is normally we always have some cursor saved so this will not have an effect. However if we from some reason (loaded garbage cursor, some systemd-side problem etc. - these were experienced before) don't have any cursor than we need this for `IgnorePreviousMessages` to correctly function (I do not know how widely used that option really is). 

Now looking at the code it could be structured/commented bit better so what I have just written is clear from the code, @rgerhards if you agree I will do a followup PR after merging this so it is clear which code is "main expected path" and what is more of error handling.

rgerhards:
> @rgerhards if you agree I will do a followup PR after merging this so it is clear which code is "main expected path" and what is more of error handling.

That would be great!

I'll merge today, but am at the moment not at a machine where I can maintain the ChangeLog alongside the merge, so I merge when I am in order to avoid forgetting the update ;-)

Issue #3364, opened 2018-12-20
TLS Network Drivers: Add setting permitExpiredCerts


alorbach:
Add new Setting "permitExpiredCerts" to control if expired certificates are allowed to be used. 
Implement into ossl and gtls driver.

PR #3366, opened 2018-12-21
TLS Network Drivers: Add setting permitExpiredCerts

alorbach:
The new Option controls how expired certificates are handeled.

For imtcp, the setting is: 
streamdriver.permitexpiredcerts

For omfwd, the setting is: 
streamdriver.permitexpiredcerts

It can have one of the following values:
on = Expired certificates are allowed
off = Expired certificates are not allowed
warn = Expired certificates are allowed but warning will be logged (Default due legacy support)

PR #4029, opened 2019-12-11
Changed default for permitExpiredCerts to "off"

jvymazal:
This is to be conssitent with rsyslog's prior behavior where
expired certs were automatically rejected.
This behavior changed with fixing #3364

### DIFF
--- a/runtime/nsd_gtls.c
+++ b/runtime/nsd_gtls.c
@@ -1461,16 +1461,16 @@ SetPermitExpiredCerts(nsd_t *pNsd, uchar *mode)
 	nsd_gtls_t *pThis = (nsd_gtls_t*) pNsd;
 
 	ISOBJ_TYPE_assert((pThis), nsd_gtls);
-	/* default is set to warn! */
-	if(mode == NULL || !strcasecmp((char*)mode, "warn")) {
-		pThis->permitExpiredCerts = GTLS_EXPIRED_WARN;
-	} else if(!strcasecmp((char*) mode, "off")) {
+	/* default is set to off! */
+	if(mode == NULL || !strcasecmp((char*)mode, "off")) {
 		pThis->permitExpiredCerts = GTLS_EXPIRED_DENY;
+	} else if(!strcasecmp((char*) mode, "warn")) {
+		pThis->permitExpiredCerts = GTLS_EXPIRED_WARN;
 	} else if(!strcasecmp((char*) mode, "on")) {
 		pThis->permitExpiredCerts = GTLS_EXPIRED_PERMIT;
 	} else {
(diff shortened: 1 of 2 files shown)


rgerhards:
Note static analyzer currently fails due to pre-existing issues and us moving to a newer version of the analyzer. I am fixing up issues in https://github.com/rsyslog/rsyslog/pull/4025. Will restart this PR here when done.

rgerhards:
Was there a reason for changing the behavior?

Sent from phone, thus brief.

Andre Lorbach <notifications@github.com> schrieb am Do., 12. Dez. 2019,
10:26:

jvymazal:
> Was there a reason for changing the behavior? Sent from phone, thus brief. 

Could not find any, thus filed this for backwards compatibility

alorbach:
According to this, legacy behaviour is indeed IGNORE (Warn): 
https://github.com/rsyslog/rsyslog/pull/3366

jvymazal:
> According to this, legacy behaviour is indeed IGNORE (Warn):
> #3366

Was the original default documented somewhere? According to my testing between 8.37 and 8.1911 and also looking at #3366 diff the original behavior is to abort connection on expired certs.

alorbach:
> > According to this, legacy behaviour is indeed IGNORE (Warn):
> > #3366
> 
> Was the original default documented somewhere? According to my testing between 8.37 and 8.1911 and also looking at #3366 diff the original behavior is to abort connection on expired certs.

As far as I know , the bejavior was not documented before.

rgerhards:
@alorbach I am puzzled now - you say the legacy was "permit and warn", @jvymazal says it was "reject" (and he tested this). @alorbach can you double-check your claim?

I am hesitant to merge this after such a long period, but if the default were actually really "permit and warn", I think we cannot merge. So I would like to clarify before merge.

jvymazal:
@rgerhards aside from having tested this before, [this part of original diff](https://github.com/rsyslog/rsyslog/pull/3366/files#diff-59bd0cab14a644c83915727cb9398d57L1112) shows that previously rsyslog just aborted upon detected expiration.

rgerhards:
I'll now merge with special info in ChangeLog - especially as @alorbach did not chime in.

The problem I see is that the default was changed for a long time. I may be forced to change back if we receive multiple complaints.

rgerhards:
@jvymazal just for your information, this is the ChangeLog text I added: https://github.com/rsyslog/rsyslog/commit/9325f3c49321e1b31ccb359ca6df49aa793e66cd

Thanks again for your persistence!

jvymazal:
@rgerhards, thanks, I am glad it is sorted out after all.

Issue #4036, opened 2019-12-14
Feature request: allow TLS SNI to be set manually


jfcantu:
Currently in the GnuTLS NSD implementation, when rsyslog is forwarding traffic to a remote syslog server, it sets the TLS SNI to always be the hostname of the remote server if the remote server is specified by host and not by IP.

It would be helpful if the SNI could be configured manually. For our use case, the remote server is actually a TLS proxy that relays traffic to an actual service, and uses the SNI to determine what service is the intended target.

For this to work, we need to be able to specify the TLS SNI, since the hostname configured for rsyslog is the hostname of the TLS proxy and not the syslog service-specific hostname.

https://github.com/rsyslog/rsyslog/blob/e1e904c67700ebf6fcd5c98292f5ce3abb0388d3/runtime/nsd_gtls.c#L2003

rgerhards:
If this is changed, it needs to work with all drivers. As such changes to openssl driver are also necessary.

jfcantu:
Took a stab at it - I'm no C expert, but it seems to work with both drivers.

rgerhards:
Apologies for the very late reply. This is probably no longer relevant for the original reporter, but I am adding the answer for people who find this issue through search. We do not lightly close old issues just because of age; we are working through the backlog and closing them when they have been implemented or superseded.

This feature is implemented in current rsyslog. `omfwd` has `StreamDriver.RemoteSNI`, and the current netstream/TLS driver interface carries that setting down to both GnuTLS and OpenSSL. The code now has `SetDrvrRemoteSNI` / `SetRemoteSNI` plumbing in the netstream layer and explicit remote-SNI handling in `nsd_gtls.c` and `nsd_ossl.c`; if the setting is present it is used instead of deriving SNI from the target hostname.

Closing this as implemented. If there is a current case where `StreamDriver.RemoteSNI` does not work with a specific TLS driver, please open a new issue with the exact rsyslog version, `omfwd` action, stream driver, and TLS handshake evidence.

PR #4037, opened 2019-12-16
omfwd: Allow configurable TLS SNI

jfcantu:
PR for issue #4036.

This branch adds a StreamDriverRemoteSNI parameter to the GnuTLS and OpenSSL stream drivers for `omfwd`, which allows a custom SNI to be sent to the server during the TLS handshake process.

Tested this out by connecting to `gnutls-serv` in debug mode and verifying the "server name" shown in output, and it seems to behave as expected with both drivers, with and without the parameter specified.


rgerhards:
If possible, it would be good if we could have a test for the functionality.

runtime/nsd_ossl.c
```diff
@@ -1710,13 +1756,17 @@ Connect(nsd_t *pNsd, int family, uchar *port, uchar *host, char *device)
 		ABORT_FINALIZE(RS_RET_NO_ERRCODE);
 	}
 
+	CHKiRet(SetServerNameIfPresent(pThis, host));
```
alorbach: Question: Why does this need to be checked if if "pThis->remoteSNI == NULL?
Like in nsd_gtls.c line 2022.

jfcantu: This is actually a slight change to `nsd_ossl`, which currently behaves differently than `nsd_gtls` (if I'm reading the code right.) The change made in PR #1393 to set the SNI based on the hostname, was only made to `nsd_gtls` - so currently, `nsd_ossl` actually has no SNI support at all. I'll update the PR to indicate that this is actually also a feature add to `nsd_ossl`.

But if I'm understanding your question right: the reason to call `SetServerNameIfPresent` even if `pThis->remoteSNI == NULL`, is to set a default SNI (the value of the `target` parameter to `omfwd`, if it's not an IP address) if no SNI is specified in configuration.

jfcantu:
Thanks for the feedback! I'll add testcases later today.

runtime/nsd_ossl.c
alorbach: Thanks for the explanation (Y).

jfcantu:
Slight tangent: is there a plan to upgrade the development Docker images to include Python 3 (especially since we're less than two weeks from Python 2.7's EOL?)

I'm still working on the test cases, but the hangup is I need to write a TLS mini-server that reports the SNI it receives when `omfwd` connects to it. This would be easiest for me to do in Python, but on CentOS 7 at least, the latest Python version available in the distro's repository (2.7.5) has deprecated TLS functionality.

(I'm working through Plan B right now, which is writing the mini-server in C, but... I'm not as skilled with C, so it's taking me a bit longer.)

tests/diag.sh
```diff
@@ -2325,6 +2325,27 @@ snmp_stop_trapreceiver() {
...
+	"./$1_sni_server" "$2" "$srcdir/tls-certs/cert.pem" "$srcdir/tls-certs/key.pem" 1>"$RSYSLOG_DYNNAME.sni-server.stdout" &
+	echo "$!" >"$RSYSLOG_DYNNAME.sni-server.pid"
+}
+
+omfwd_sni_check() {
+    # Args: 1=sni
+	# 1 is the SNI we're expecting (can be omitted to confirm no SNI is sent)
+	sni=${1:-""}
+
+	sleep 5
```
rgerhards: sleep should be avoided except absolutely necessary - CI can be very slow and using fixed timing almost always leads to false positives

jfcantu: Ah, okay. Sorry for the delay - holidays etc. - I'm still working on figuring out why the tests are failing on certain distros.

rgerhards: I am currently working on the CI. Due to this, there is a higher rate of flakes than usual.  That said, you also need to rebase to current master. Otherwise we are no longer able to do CI runs (see merge conflicts below).

jfcantu:
I think the test suite is about as good as I can make it right now - for some reason, there's a handful of environments where rsyslog can't connect to the micro-server that writes the TLS SNI to the console, and I'm not sure why.

rgerhards:
@jfcantu I will investigate

rgerhards:
@jfcantu Due to my sluggishness, this PR unfortunately went way out of sync with the current codebase. My apologies. I can see if I can rebase it, but it would probably be better to extract the changes and integrate them into the current codebase. I will look at this, and I can commit with the proper --author applied.

rgerhards:
@jfcantu I created a rebased version with some fixes and adjustments for current code base in https://github.com/rsyslog/rsyslog/pull/6452. Once this is ready-for-merge, I'll see if I can push it to your repo and then will merge it. Else I'll merge "my" version, which still has the correct author (you),

Again, thx for your contribution, even though it took years...

rgerhards:
@jfcantu pushed the rebased version. I expect this is ready  to merge.

Issue #4041, opened 2019-12-17
Rsyslogd aborts when writing to disk queue from multiple workers


miztake:
### Expected behavior
Rsyslogd does not abort.

### Actual behavior
Rsyslogd aborts.

### Environment
Some action queues were defined for each rule with the same filter.

   /etc/rsyslog.conf
   --------
    $ActionQueueType LinkedList
    $ActionQueueFileName queue_fwd1
    $ActionResumeRetryCount -1
    $ActionQueueSaveOnShutdown on
    *.info;mail.none;authpriv.none;cron.none        @@11.11.11.11:514

    $ActionQueueType LinkedList
    $ActionQueueFileName queue_fwd2
    $ActionResumeRetryCount -1
    $ActionQueueSaveOnShutdown on
    *.info;mail.none;authpriv.none;cron.none        @@22.22.22.22:514

    $ActionQueueType LinkedList
    $ActionQueueFileName queue_fwd3
    $ActionResumeRetryCount -1
    $ActionQueueSaveOnShutdown on
    *.info;mail.none;authpriv.none;cron.none        @@33.33.33.33:514
   --------

Because the destination does not exist, the queue is full.
When multiple action queues became full at the same time and tried to write to disk at the same time, the json library terminated abnormally.

  backtrace
  --------
    Program received signal SIGABRT, Aborted.
    [Switching to Thread 0x7fffe4ff9700 (LWP 11752)]
    0x00007ffff69c1207 in raise () from /lib64/libc.so.6
    (gdb) bt
    #0  0x00007ffff69c1207 in raise () from /lib64/libc.so.6
    #1  0x00007ffff69c28f8 in abort () from /lib64/libc.so.6
    #2  0x00007ffff6a03cc7 in __libc_message () from /lib64/libc.so.6
    #3  0x00007ffff6a0a574 in malloc_printerr () from /lib64/libc.so.6
    #4  0x00007ffff6a0ffa3 in _int_realloc () from /lib64/libc.so.6
    #5  0x00007ffff6a10e42 in realloc () from /lib64/libc.so.6
    #6  0x00007ffff71959f9 in printbuf_extend (p=p@entry=0x7fffc80008f0, min_size=<optimized out>) at printbuf.c:79
    #7  0x00007ffff7195b20 in printbuf_memappend_no_nul (p=p@entry=0x7fffc80008f0, buf=0x7fffe8083348 "16472", size=5)
        at printbuf.c:105
[...]

PR #4042, opened 2019-12-17
Fix race condition related to libfastjson when using DA queue

miztake:
Rsyslogd aborts when writing to disk queue from multiple workers simultaneously.
It is assumed that libfastjson is not thread-safe.
Resolve libfastjson race condition when writing to disk queue.

see also https://github.com/rsyslog/rsyslog/issues/4041


miztake:
Hi, all,

Do you have any comments on this PR?

rgerhards:
> Do you have any comments on this PR?

no, but I would like to check first if this is a bug in libfastjson

rgerhards:
I'll merge now to meet the next scheduled stable release. I'll add another issue tracker to check if this is "just" a work-around or a permanent solution.

PR #4044, opened 2019-12-18
fix indention

rgerhards:
detected by clang 10


rgerhards:
static analyzer error is pre-existing

PR #4045, opened 2019-12-18
fix misleading/invalid indention

rgerhards:
detected by clang 10


rgerhards:
static analyzer is known pre-existing problem

PR #4061, opened 2019-12-20
Do not close /dev/urandom file descriptor after fork on Linux systems

simonjhaggett:
This patch updates prepareBackground() in tools/rsyslogd.c to check that an open
file descriptor is not being used for random number generation before closing it
as an unnecessary open file. This fixes an issue on Linux systems where the file
descriptor obtained for /dev/urandom by seedRandomNumber() in runtime/srutils.c
was being closed after the fork. This could be observed in procfs, where
/proc/<PID>/fd/ would show no open descriptors to /dev/urandom in the forked
process. With this fix applied, procfs confirms that the forked rsyslog process
does retain a valid file descriptor to /dev/urandom.

I found that this issue led to rsyslog intermittently hanging during seedIV()
in runtime/libgcry.c. After the fork, the closed file descriptor number tended
to get re-assigned. randomNumber() would then read from an incorrect (although
still valid) file descriptor, and could block (depending on the state of that
file descriptor). This gave rise to the intermittent hang that I observed.

Signed-off-by: Simon Haggett <simon.haggett@gmail.com>

### DIFF
--- a/runtime/srUtils.h
+++ b/runtime/srUtils.h
@@ -96,6 +96,7 @@ int getSubString(uchar **ppSrc,  char *pDst, size_t DstSize, char cSep);
 rsRetVal getFileSize(uchar *pszName, off_t *pSize);
 int containsGlobWildcard(char *str);
 void seedRandomNumber(void);
+int isRandomNumberFd(int fd);
 #define MAX_RANDOM_NUMBER RAND_MAX
 long int randomNumber(void);
 long long currentTimeMills(void);
--- a/runtime/srutils.c
+++ b/runtime/srutils.c
@@ -725,6 +725,11 @@ void seedRandomNumber(void)
 	}
 }
 
+int isRandomNumberFd(const int fd)
+{
+	return fd >= 0 && fd == fdURandom;
+}
+
 long int randomNumber(void)
 {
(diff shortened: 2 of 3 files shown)


rgerhards:
IIRC an open FD can be used to escape the chroot environment. Unfortunately I do not know the details any longer. If so,  we probably need to look for another fix (re-open after backgrounding?).

simonjhaggett:
> IIRC an open FD can be used to escape the chroot environment. Unfortunately I do not know the details any longer. If so, we probably need to look for another fix (re-open after backgrounding?).

Ah, sorry I hadn't realized that. Re-opening /dev/urandom after backgrounding should also work, I'll look at raising a new PR with that fix.

skupkosk:
Nice reading with tool on escaping chroots: https://github.com/earthquake/chw00t

simonjhaggett:
> Nice reading with tool on escaping chroots: https://github.com/earthquake/chw00t

Thanks, that's useful reading.

I've now raised https://github.com/rsyslog/rsyslog/pull/4064 with an alternative fix (reopening /dev/urandom in the child).

rgerhards:
replaced by #4064

PR #4064, opened 2019-12-23
Reopen /dev/urandom file descriptor after fork on Linux systems

simonjhaggett:
This patch updates prepareBackground() in tools/rsyslogd.c to reopen any file
descriptors used for random number generation in the child process. This fixes
an issue on Linux systems where the file descriptor obtained for /dev/urandom
by seedRandomNumber() in runtime/srutils.c was left closed after the fork. This
could be observed in procfs, where /proc/fd/ would show no open descriptors to
/dev/urandom in the forked process. /dev/urandom is reopened as the child may be
be operating in a jail, and so should not continue to use file descriptors from
outside the jail (i.e. inherited from the parent process).

I found that this issue led to rsyslog intermittently hanging during seedIV()
in runtime/libgcry.c. After the fork, the closed file descriptor number tended
to get re-assigned. randomNumber() would then read from an incorrect (although
still valid) file descriptor, and could block (depending on the state of that
file descriptor). This gave rise to the intermittent hang that I observed.

This PR provides an alternative fix to that given in https://github.com/rsyslog/rsyslog/pull/4061.

Signed-off-by: Simon Haggett <simon.haggett@gmail.com>

### DIFF
--- a/runtime/srutils.c
+++ b/runtime/srutils.c
@@ -717,12 +717,26 @@ static long int randomInsecureNumber(void)
 static int fdURandom = -1;
 void seedRandomNumber(void)
 {
+	if(fdURandom >= 0) {
+		/* Already opened. */
+		return;
+	}
 	fdURandom = open("/dev/urandom", O_RDONLY);
 	if(fdURandom == -1) {
 		LogError(errno, RS_RET_IO_ERROR, "failed to seed random number generation,"
 			" will use fallback (open urandom failed)");
 		seedRandomInsecureNumber();
 	}
+}
+
+void seedRandomNumberForChild(void)
+{
+	/* The file descriptor inherited from our parent will have been closed after
+	 * the fork. Discard this and call seedRandomNumber() to open /dev/urandom
+	 * again.
+	 */
+	fdURandom = -1;
+	seedRandomNumber();
 }
 
 long int randomNumber(void)
@@ -743,6 +757,11 @@ long int randomNumber(void)
 void seedRandomNumber(void)
 {
 	seedRandomInsecureNumber();
+}
+
+void seedRandomNumberForChild(void)
+{
+	seedRandomNumber();
 }
 
 long int randomNumber(void)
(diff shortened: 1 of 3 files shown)


runtime/srutils.c
```diff
@@ -745,6 +754,10 @@ void seedRandomNumber(void)
 	seedRandomInsecureNumber();
 }
 
+void closeRandomNumberFd(void)
```
rgerhards: This looks like it (and all calls) can simply be removed?

simonjhaggett: Yes, it's not strictly needed since the file descriptor would already be included in the list closed by `prepareBackground()`. However, I was concerned that `seedRandomNumber()` could be changed in future to only open `/dev/urandom` if `fdURandom` was negative. Which would then re-introduce the bug.

I could remove `closeRandomNumberFd()` and instead add a comment to `seedRandomNumber()` warning of this, if you'd prefer?

rgerhards: > if `fdURandom` was negative

as we know about this problem, I think it would make sense to set it to -1 on close. That way we could actually do the -1 check.

side-note: think about refactoring. If we would leave the code as-is, it would sooner or later be removed because it "obviously" is unnecessary.

simonjhaggett: OK, so perhaps the following would be a neater fix:

* Have `seedRandomNumber()` only open `/dev/urandom` if  `fdURandom` is negative.
* Add a new `seedRandomNumberForChild()` function, which simply sets `fdURandom` to -1 and calls `seedRandomNumber()` again.
* Have `prepareBackground()` call `seedRandomNumberForChild()` after closing file descriptors.

That would address my concerns, and means we wouldn't have `closeRandomNumberFd()` (which as you say could be removed by future refactoring work). If this sounds reasonable to you, then I'll update the PR to implement this today.

rgerhards: sounds good to me!

simonjhaggett: Excellent, thanks. I've now updated this PR.

Issue #4020, opened 2019-12-09
Increase queue size on shutdown if action queue is full


StrongestNumber9:
We think that increasing queue size on shutdown when queues are full is useful feature to prevent losing any messages and to further ensure reliablity of the system. [Reference discussion](https://github.com/rsyslog/rsyslog/issues/3941#issuecomment-549765813)

rgerhards:
@StrongestNumber9 I think #4069 will do the trick. Could you test in your environment?

Issue #3941, opened 2019-10-31
Imrelp discards events and acks them OK even if omfwd can't forward them and queue is full


StrongestNumber9:
### Expected behavior
Imrelp gets event, runs it through rulesets and acks NOT OK if queues are full, informing the client that your data couldn't be dealt with.

### Actual behavior
Imrelp gets event, runs it through rulesets and acks OK even if omfwd queue is full and target is not answering. This causes silent loss of events by rate of roughly 2 seconds per event. In this case omfwd doesn't respect action.resumeInterval and action.resumeRetryCount settings.

### Steps to reproduce the behavior
rsyslog.conf
```sh
$LocalHostName test
$AbortOnUncleanConfig on
$PreserveFQDN on

module(load="imrelp")

global(
    workDirectory="/"
    maxMessageSize="256k"
)
main_queue(queue.type="Direct")
input(type="imrelp" port="601" ruleset="incoming" MaxDataSize="256k")

ruleset(name="incoming") {
    action
    (
        type="omfwd"
        name="omfwd"
        RebindInterval="10000"
        protocol="tcp"
        target="127.0.0.1"
        port="10514"
        queue.size="1000"
        queue.type="Disk"
        queue.checkpointinterval="65536"
        queue.filename="queue"
        queue.maxfilesize="32m"
        queue.saveonshutdown="on"
        queue.syncqueuefiles="off"
        queue.maxdiskspace="128m"
        ResendLastMSGOnReconnect="on"
        action.ResumeInterval="60"
        action.resumeRetryCount="-1"
    )
}
```

Commands executed
```sh
[root@example /]# rsyslogd -f /rsyslog.conf -N9
rsyslogd: version 8.38.0, config validation run (level 9), master config /rsyslog.conf
rsyslogd: End of config validation run. Bye.

[root@example /]# tcpflood -T relp-plain -t 127.0.0.1 -p 601 -m 1000 -M "Filling the queue"
00001 open connections
starting run 1
Sending 1000 messages.
00001000 messages sent
runtime: 0.416
00001 close connections
End of tcpflood Run

[root@example /]# grep 'pszRawMsg' /queue.00000001 | wc -l
1000

[root@example /]# tcpflood -T relp-plain -t 127.0.0.1 -p 601 -m 10 -M "Should not get accepted"
00001 open connections
starting run 1
Sending 10 messages.
00000010 messages sent
runtime: 0.001
00001 close connections
End of tcpflood Run

[...]

PR #4069, opened 2019-12-27
queue: permit ability to double size at shutdown

rgerhards:
This prevents message loss due to "queue full" when re-enqueueing data
under quite exotic settings.

see also https://github.com/rsyslog/rsyslog/issues/3941#issuecomment-549765813
closes https://github.com/rsyslog/rsyslog/issues/4020


rgerhards:
@StrongestNumber9 Can you provide some feedback if this patch was useful? If so, I would merge.

StrongestNumber9:
Hi @rgerhards 

Seems like the patch is useful.

PR #4080, opened 2020-01-01
testbench: rough fix for clickhouse 2019 test data problem

rgerhards:
not a real fix, but to get us going: clickhouse test had hardcoded
year; needs to be removed soon. Right now, just pusing it to 2020.


rgerhards:
test fails are false positives

PR #4085, opened 2020-01-02
testbench: update to current 7.5 version of ElasticSearch

rgerhards:
This also restructures the tests order a bit.


rgerhards:
We already have moved to even newer ES versions while this PR was stalled/forgotten.

PR #4089, opened 2020-01-03
testbench: make imdocker tests more robust in regard to timing

rgerhards:
some have problems on heavily loaded (CI) machines. This commit solves
that problem.


rgerhards:
solaris failure is a known issue which is being addressed in #4095

Issue #4035, opened 2019-12-13
SSL_set_verify_depth restricts certificate verification


amarkelov:
### Expected behavior
Unlimited depth. 

### Actual behavior
If client is issued TLS certificate by CA from a different branch than the server's and the depth is greater than the hardcoded limit, TLS connection fails.

### Steps to reproduce the behavior
Create root CA with two branches of sub-CAs of different depth greater than 2 for the CA that issues certificate for client for example. TLS handshake will fail because root CA is higher up the chain.

### Environment
- rsyslog version: 
    $ rpm -qa | grep rsyslog
    rsyslog-openssl-8.1901.0-1.el7.x86_64
    rsyslog-8.1901.0-1.el7.x86_64
    rsyslog-gnutls-8.1901.0-1.el7.x86_64
    $ rsyslogd -v
    rsyslogd  8.1901.0 (aka 2019.01) compiled with:
        PLATFORM:               x86_64-redhat-linux-gnu
        PLATFORM (lsb_release -d):      
        FEATURE_REGEXP:             Yes
        GSSAPI Kerberos 5 support:      Yes
        FEATURE_DEBUG (debug build, slow code): No
        32bit Atomic operations supported:  Yes
        64bit Atomic operations supported:  Yes
        memory allocator:           system default
        Runtime Instrumentation (slow code):    No
        uuid support:               Yes
        systemd support:            Yes
        Number of Bits in RainerScript integers: 64

    See https://www.rsyslog.com for more information.
- platform: CentOS 7.5
- for configuration questions/issues, include rsyslog.conf and included config files

### Suggested fix

Do not limit the verification depth or make it an optional parameter in configuration file:

    diff --git a/runtime/nsd_ossl.c b/runtime/nsd_ossl.c
    index 4f8dd845a..76cc132fa 100644
    --- a/runtime/nsd_ossl.c
    +++ b/runtime/nsd_ossl.c
    @@ -626,7 +626,6 @@ osslInitSession(nsd_ossl_t *pThis) /* , nsd_ossl_t *pServer) */
                    dbgprintf("osslInitSession: enable certificate checking (Mode=%d)\n", pThis->authMode);
                    /* Enable certificate valid checking */
                    SSL_set_verify(pThis->ssl, SSL_VERIFY_PEER|SSL_VERIFY_FAIL_IF_NO_PEER_CERT, verify_callback);
    -               SSL_set_verify_depth(pThis->ssl, 2);
            }
     
[...]

PR #4091, opened 2020-01-03
ossl tls: Added support to configure certificate verify depth

alorbach:
Support added in omfwd as instance parameter:
streamdriver.TlsVerifyDepth
Support added in imtcp as module parameter:
streamdriver.TlsVerifyDepth

Can be 2 or higher.
Support added into ossl driver

Parameter
streamdriver.TlsVerifyDepth

closes: https://github.com/rsyslog/rsyslog/issues/4035

### DIFF
--- a/plugins/imtcp/imtcp.c
+++ b/plugins/imtcp/imtcp.c
@@ -63,6 +63,7 @@
 #include "ruleset.h"
 #include "rainerscript.h"
 #include "net.h" /* for permittedPeers, may be removed when this is removed */
+#include "parserif.h"
 
 MODULE_TYPE_INPUT
 MODULE_TYPE_NOKEEP
@@ -133,6 +134,7 @@ struct modConfData_s {
 	int iStrmDrvrMode; /* mode for stream driver, driver-dependent (0 mostly means plain tcp) */
 	int iStrmDrvrExtendedCertCheck; /* verify also purpose OID in certificate extended field */
 	int iStrmDrvrSANPreference; /* ignore CN when any SAN set */
+	int iStrmTlsVerifyDepth; /**< Verify Depth for certificate chains */
 	int iAddtlFrameDelim; /* addtl frame delimiter, e.g. for netscreen, default none */
 	int maxFrameSize;
 	int bSuppOctetFram;
@@ -174,6 +176,7 @@ static struct cnfparamdescr modpdescr[] = {
 	{ "streamdriver.name", eCmdHdlrString, 0 },
 	{ "streamdriver.CheckExtendedKeyPurpose", eCmdHdlrBinary, 0 },
 	{ "streamdriver.PrioritizeSAN", eCmdHdlrBinary, 0 },
+	{ "streamdriver.TlsVerifyDepth", eCmdHdlrPositiveInt, 0 },
 	{ "permittedpeer", eCmdHdlrArray, 0 },
 	{ "keepalive", eCmdHdlrBinary, 0 },
 	{ "keepalive.probes", eCmdHdlrPositiveInt, 0 },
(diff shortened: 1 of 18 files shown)


runtime/tcpsrv.c
```diff
@@ -1000,6 +1000,7 @@ BEGINobjConstruct(tcpsrv) /* be sure to specify the object type also in END macr
 	pThis->bUseFlowControl = 1;
 	pThis->pszDrvrName = NULL;
 	pThis->bPreserveCase = 1; /* preserve case in fromhost; default to true. */
+	pThis->DrvrTlsVerifyDepth = 25;
```
rgerhards: This looks inconsistent to the other default. I would suggest to leave the default at 0, as this avoids the API call and so the library default is used (whatever it is). I think this is the least confusing option for users.

tools/omfwd.c
```diff
@@ -85,6 +85,8 @@ typedef struct _instanceData {
 	int iStrmDrvrMode;
 	int iStrmDrvrExtendedCertCheck; /* verify also purpose OID in certificate extended field */
 	int iStrmDrvrSANPreference; /* ignore CN when any SAN set */
+#	define OMFWD_VERIFY_DEPTH 100
```
rgerhards: 0 makes a better default (see other comment)

plugins/imtcp/imtcp.c
```diff
@@ -133,6 +134,8 @@ struct modConfData_s {
 	int iStrmDrvrMode; /* mode for stream driver, driver-dependent (0 mostly means plain tcp) */
 	int iStrmDrvrExtendedCertCheck; /* verify also purpose OID in certificate extended field */
 	int iStrmDrvrSANPreference; /* ignore CN when any SAN set */
+#	define IMTCP_VERIFY_DEPTH 100
```
rgerhards: 0 makes a better default (see other comment)

alorbach: yes this can be removed now, as default is handeled in nsd_ossl.c now

runtime/nsd_ossl.c
```diff
@@ -78,6 +78,7 @@ DEFobjCurrIf(nsd_ptcp)
 	#define RSYSLOG_BIO_number_read(SSLBIO) SSLBIO->num
 	#define RSYSLOG_BIO_number_written(SSLBIO) SSLBIO->num
 #endif
+#define OSSL_VERIFY_DEPTH 100
```
rgerhards: I guess this can be removed?

alorbach: We need the default here as we set the verify_depth default here for the existing API call which was hardcoded to 2 before.

alorbach: Removed need of OSSL_VERIFY_DEPTH , it should be fine now

rgerhards:
doc: https://github.com/rsyslog/rsyslog-doc/pull/878

tests/imtcp-tls-ossl-basic-verifydepth.sh
```diff
@@ -0,0 +1,33 @@
...
+global(	defaultNetstreamDriverCAFile="'$srcdir/tls-certs/ca.pem'"
+	defaultNetstreamDriverCertFile="'$srcdir/tls-certs/cert.pem'"
+	defaultNetstreamDriverKeyFile="'$srcdir/tls-certs/key.pem'"
+)
+module(	load="../plugins/imtcp/.libs/imtcp"
+	StreamDriver.Name="ossl"
+	StreamDriver.TlsVerifyDepth="5" 
+	StreamDriver.Mode="1"
+	StreamDriver.AuthMode="x509/certvalid" )
+input(	type="imtcp"
```
rgerhards: Please update the test plumbing - this is the old, racy version.  I have just fixed similar tests here: https://github.com/rsyslog/rsyslog/pull/4102/files

Pls also have all input parameters on one line - this simplifies future changes via `sed`. If there are very many parameters, at least the `type` and `port` should be in one line and in this order.

alorbach: done

rgerhards:
Note to myself: fix test after merge

alorbach:
> I'll do the fix after merge

Ok thanks

PR #4095, opened 2020-01-06
Solaris

dago:
These changes are required to allow compilation on Solaris 10 with GCC.

### DIFF
--- a/compat/strndup.c
+++ b/compat/strndup.c
@@ -21,15 +21,20 @@
 #include "config.h"
 #ifndef HAVE_STRNDUP
 
+
 #include <stdlib.h>
 #include <string.h>
+
+extern char *strndup(const char *s, size_t n);
+
 char *
 strndup(const char *s, size_t n)
 {
 	const size_t len = strlen(s);
+	char *new_s;
 	if(len <= n)
 		return strdup(s);
-	char *const new_s = malloc(n+1);
+	new_s = malloc(n+1);
 	if(new_s == NULL)
 		return NULL;
 	memcpy(new_s, s, n);
--- a/plugins/mmexternal/mmexternal.c
+++ b/plugins/mmexternal/mmexternal.c
@@ -472,8 +472,8 @@ callExtProg(wrkrInstanceData_t *__restrict__ const pWrkrData, smsg_t *__restrict
 			switch(errno) {
 			case EPIPE:
 				LogMsg(0, RS_RET_ERR_WRITE_PIPE, LOG_WARNING,
-						"mmexternal: program '%s' (pid %d) terminated; will be restarted",
-						pWrkrData->pData->szBinary, pWrkrData->pid);
+						"mmexternal: program '%s' (pid %ld) terminated; will be restarted",
+						pWrkrData->pData->szBinary, (long) pWrkrData->pid);
 				CHKiRet(cleanup(pWrkrData));
 				CHKiRet(tryRestart(pWrkrData));
 				writeOffset = 0;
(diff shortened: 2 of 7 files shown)


rgerhards:
@dago I am working on CI, thus a couple of flakes...

I just wonder why we do not get build errors on the Solaris 10 workers without this patch. Do you have an explanation?

dago:
The warnings are only activated when compiling with GCC, so when you compile with GCC on other platforms the code path is not used and if you compile with Sun Studio in Solaris the warnings are not activated. The CI only uses Sun Studio to build on Solaris.

dago:
You can have a look at the OpenCSW CI buildbot which shows the first errors: https://buildfarm.opencsw.org/buildbot/builders/rsyslog-solaris10-i386/builds/1292/steps/shell_2/logs/stdio

rgerhards:
> You can have a look at the OpenCSW CI buildbot which shows the first errors: https://buildfarm.opencsw.org/buildbot/builders/rsyslog-solaris10-i386/builds/1292/steps/shell_2/logs/stdio

mhhh... this is strange. Have a look at https://build.rsyslog.com/#/builders/73/builds/2980
Steps 12+ use gcc. Do we need to fix something in our CI?

dago:
It looks like that after the autoreconf not everything is rebuilt. I'd suggest to add something like `gmake distclean` and then cleanly configure again (or just use different targets for Sun Studio and GCC).

rgerhards:
> It looks like that after the autoreconf not everything is rebuilt. I'd suggest to add something like `gmake distclean` and then cleanly configure again (or just use different targets for Sun Studio and GCC).

Doh! I overlooked that there is no `clean` at all. Will fix CI (when done with the other CI work currently going on) and merge only after that - so that I have a test case to know for sure this works ;-)

Thx!

rgerhards:
looks like there is a small glitch in building: https://build.rsyslog.com/#/builders/73/builds/2996/steps/15/logs/stdio

dago:
Ah yes, I wrapped that with an #ifdef but missed the commit:
```
diff --git a/runtime/rsyslog.h b/runtime/rsyslog.h
index dd256a8..85d4d30 100644
--- a/runtime/rsyslog.h
+++ b/runtime/rsyslog.h
@@ -198,7 +198,9 @@
 #define        LOG_LPR         (6<<3)  /* line printer subsystem */
 #define        LOG_NEWS        (7<<3)  /* network news subsystem */
 #define        LOG_UUCP        (8<<3)  /* UUCP subsystem */
+#if !defined(LOG_CRON)
 #define        LOG_CRON        (9<<3)  /* clock daemon */
+#endif
 #define        LOG_AUTHPRIV    (10<<3) /* security/authorization messages (private) */
 #define        LOG_FTP         (11<<3) /* ftp daemon */
 #if defined(_AIX)              /* AIXPORT : These are necessary for AIX */
```

rgerhards:
> Ah yes, I wrapped that with an #ifdef but missed the commit:

should I amend a commit to fix this? Wanted to ask before going ahead...

dago:
I just added the commit, was waiting to be pushed anyway :-)

rgerhards:
looks like we have another one: https://github.com/rsyslog/rsyslog/pull/4100 ;-)

rgerhards:
finally ;-)

dago:
Thanks for merging this so quickly! OpenCSW buildbot is now also clean: https://buildfarm.opencsw.org/buildbot/waterfall?category=rsyslog

PR #4100, opened 2020-01-07
omprog bugfix: unclean data type conversions

rgerhards:
breaks compilation e.g. on Solaris 10 with gcc


rgerhards:
Solaris failure is known, see #4095

PR #4105, opened 2020-01-08
testbench: modernize some tests

rgerhards:
This also makes port use more reliable.


rgerhards:
TSAN is known false positive.

PR #4108, opened 2020-01-09
satisfy-gcc-flag-fno-common

tblume:
Porting advice for gcc 10:

A common mistake in C is omitting <code>extern</code> when declaring a global
variable in a header file.  If the header is included by several files it
results in multiple definitions of the same variable.  In previous GCC versions
this error is ignored.  GCC 10 defaults to <code>-fno-common</code>, which
means a linker error will now be reported. To fix this, use <code>extern</code>
in header files when declaring global variables, and ensure each global is
defined in exactly one C file. As a workaround, legacy C code can be compiled
with -fcommon.

### DIFF
--- a/tools/omfile.h
+++ b/tools/omfile.h
@@ -32,7 +32,7 @@ rsRetVal modInitFile(int iIFVersRequested __attribute__((unused)), int *ipIFVers
  * other way to have the functionality (well, one way would be to go through the
  * globals, but that seems not yet justified. -- rgerhards, 2010-03-01
  */
-uchar	*pszFileDfltTplName;
+extern uchar	*pszFileDfltTplName;
 #endif /* #ifndef OMFILE_H_INCLUDED */
 /* vi:set ai:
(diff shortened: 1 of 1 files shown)


rgerhards:
thx for the PR. Q: is gcc 10 sufficiently stable? I had the impression "not yet". If it is, I would add it to our CI compile checks.

marxin:
> thx for the PR. Q: is gcc 10 sufficiently stable? I had the impression "not yet". If it is, I would add it to our CI compile checks.

It's still under development. But one can test it with any stable GCC (9.2.0) with `-fno-common`.

rgerhards:
> But one can test it with any stable GCC (9.2.0) with `-fno-common`.

Thx - will add. Until then, I keep the PR open, so that I get a known failure.

rgerhards:
I'll merge now to meet the next scheduled release - will test with a specific testing PR later on.

PR #4109, opened 2020-01-09
Dynafile + async KSI module bugfix

taavi-valjaots:
Bugfix: KSI module + dynafile in asynchronous mode fixed. KSI module kept just 1 signature file in focus and some of files were not closed. There was possibility that multiple KSI signatures of separate log files were stored in a single signature file. Also the block timeout was measured for one log file.


rgerhards:
Tere @taavi-valjaots thx for the PR. Please have a look at CI results but be warned that we are currently restructuring the system and have some flakes. So any failure not related to the KSI components you can ignore ;-)

PR #4119, opened 2020-01-13
tooling: pass additional environment vars to dev container

rgerhards:


rgerhards:
CI fails are (mostly) false positive, TSAN is a bit in question. In order to be able to check and fix potentially new issues, I need to merge this PR, as it provides the base for all other PRs to work with the new systems. Should an error be present, I'll fix that later on.

PR #4129, opened 2020-01-17
tcpsrv bugfix: build problem regression

rgerhards:


rgerhards:
CI failures are known false positives being addressed in https://github.com/rsyslog/rsyslog/pull/4118

PR #4133, opened 2020-01-20
testbench: improvements

rgerhards:

PR #4135, opened 2020-01-20
core bugfix: misadressing in pselect()

rgerhards:
regression from 723f6fdfa

### DIFF
--- a/tools/rsyslogd.c
+++ b/tools/rsyslogd.c
@@ -1915,7 +1915,7 @@ wait_timeout(const sigset_t *sigmask)
 		}
 	}
 #else
-	pselect(1, NULL, NULL, NULL, &tvSelectTimeout, sigmask);
+	pselect(0, NULL, NULL, NULL, &tvSelectTimeout, sigmask);
 #endif /* AIXPORT : SRC end */
 }
 


rgerhards:
Solaris CI fail is a known issue addressed in https://github.com/rsyslog/rsyslog/pull/4133, so "all go"

rgerhards:
@Frogging101 it would be good if you could have a look. I did not try to fix the issue for AIX as I do not know if it was intentional there. I don't think so, but I cannot test.

Frogging101:
I think that should be `pselect(0, ...)` in both cases. I do not know why I wrote `1` there.

How did this bug manifest? Did it crash?

rgerhards:
SAN (the LLMV sanitizers) found it, but only only occasionally. I guess it could lead to a crash.

Thx for the comment, I'll now also fix the AIX part.

rgerhards:
@Frogging101 mmhhh... I still see issues occasionally: https://build.rsyslog.com/#/builders/211/builds/324/steps/4/logs/test-suite_log - do you have an idea of what this could be? it happens on multiple platforms, so it cannot be a bad lib...

Frogging101:
I've looked over this, and I can't see anything wrong with the way we're 
using pselect. I'll look again later when I have time.

Thanks for running these tests.

rgerhards:
@Frogging101 I now tend to think that this is a false positive in valgrind or, better said, a known issue inside the lib/kernel. I just did not yet find a good confirmation for that. If we can find, it is reasonable to suppress the error.

PR #4136, opened 2020-01-20
ompipe: add action parameter tryResumeReopen

julthomas:
Hello Rainer,

Sometimes we need to reopen a pipe after an ompipe action gets suspended. Sending an HUP signal to rsyslog does the job but requires an interaction with rsyslog. The patch adds support for a new boolean option, tryResumeReopen, for the ompipe action. It mimics what an HUP signal would do.

We have issues with ompipe in the following context:
- ompipe writes to a Nagios 4 command fifo
- Nagios gets restarted
- The rsyslog ompipe action suspends but never resumes, even with action.resumeinterval and action.resumeintervalmax set to small values, and action.resumeretrycount set to -1. I believe the select() call always returns 0.

Using this patch works for us. It adds option tryResumeReopen and keeps the current behavior by default ("off" by default). I am not sure if it is properly designed, would you consider including it?

Best regards,
Julien

### DIFF
--- a/tools/ompipe.c
+++ b/tools/ompipe.c
@@ -70,6 +70,7 @@ typedef struct _instanceData {
 	short	fd;		/* pipe descriptor for (current) pipe */
 	pthread_mutex_t mutWrite; /* guard against multiple instances writing to same pipe */
 	sbool	bHadError;	/* did we already have/report an error on this pipe? */
+	sbool	bTryResumeReopen;	/* should we attempt to reopen the pipe on action resume? */
 } instanceData;
 
 typedef struct wrkrInstanceData {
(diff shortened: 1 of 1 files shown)


rgerhards:
Thx for the PR, much appreciated. Please note: CI is currently a bit flaky as I work on it. Any real issue I let you know.

rgerhards:

julthomas:
Hello Rainer,

Proposed pull-request for the documentation:
rsyslog/rsyslog-doc#881

Regards,
Julien

julthomas:
I have a doubt about the mutex lock on the close and -1 reset on the file descriptor.
It is not taken as well in doHUP, what do you think?
Julien

rgerhards:
> I have a doubt about the mutex lock on the close and -1 reset on the file descriptor.
> It is not taken as well in doHUP, what do you think?

sry for the delay ... it's an exceptional time... I reviewed the comment and originally, on quick look, it seemed to be a question about this PR. That was why I didn't merge quickly. Now I see it is a related but different issue. I think you are right, will try a patch. That also means I can finally merge this PR :-)

Thanks again for your help and patience.

PR #4137, opened 2020-01-20
Enhancement: default type for elasticsearch >=7 compatability

maierru:
https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html
```
Elasticsearch 7.x
 Specifying types in requests is deprecated. ....
Note that in 7.0, _doc is a permanent part of the path, 
and represents the endpoint name rather than the document type.
```


rgerhards:
@maierru This looks like it is incompatible with previous versions of ES.

rgerhards:
No interest in making it compatible?

maierru:
@rgerhards resend pr when test on different versions.
Close just to not annoying in list of active pr

PR #4142, opened 2020-01-21
MMDBLOOKUP : Fix no entry found in database

KGuillemot:
Check found_entry attribute to avoid wrong output when entry is not in database

### DIFF
--- a/plugins/mmdblookup/mmdblookup.c
+++ b/plugins/mmdblookup/mmdblookup.c
@@ -362,6 +362,11 @@ CODESTARTdoAction
 		dbgprintf("Got an error from the maxminddb library: %s\n", MMDB_strerror(mmdb_err));
 		ABORT_FINALIZE(RS_RET_OK);
 	}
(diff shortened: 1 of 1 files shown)


KGuillemot:
Hello,
Since the upgrade of the package libmaxminddb on FreeBSD (1.3.2_2 -> 1.4.2), the module mmdblookup returns the first entry of the mmdb database even if the entry is not found.
After some debug, I found the solution in the official maxminddb repository : to check if the entry is in database, we must check the `found_entry` attribute, otherwise the function `MMDB_get_entry_data_list` will return the first entry of the database if the entry is not found in it.

rgerhards:
@KGuillemot  thx for the PR! Please note that I am currently working on CI - we right now have some known false positives. I'll let you know if something needs your attention!

KGuillemot:
Ok no problem. Nice to see that this PR will be in 8.2002 release ! :-)

Issue #4120, opened 2020-01-14
test-suite failes with 8.2001.0


mbiebl:
Version: 8.2001.0
Debian sid

Trying to build the latest version of rsyslog for Debian, I get

```
PASS: immark.sh
... [SNIPPED by admin for easier readability]
SKIP: imfile-logrotate-nocopytruncate.sh
FAIL: imfile-growing-file-id.sh
PASS: glbl-oversizeMsg-truncate-imfile.sh
PASS: config_enabled-on.sh
PASS: config_enabled-off.sh
PASS: imfile-endmsg.regex-with-example.sh
make[6]: Entering directory '/home/test/rsyslog-8.2001.0/tests'
make[6]: Nothing to be done for 'imfile-escapelf.replacement.log'.
make[6]: Nothing to be done for 'imfile-escapelf.replacement-empty.log'.
make[6]: Leaving directory '/home/test/rsyslog-8.2001.0/tests'
fatal: making test-suite.log: failed to create imfile-escapelf.replacement.trs
fatal: making test-suite.log: failed to create imfile-escapelf.replacement.log
fatal: making test-suite.log: failed to create imfile-escapelf.replacement-empty.trs
fatal: making test-suite.log: failed to create imfile-escapelf.replacement-empty.log
make[5]: *** [Makefile:3414: test-suite.log] Error 1
```

mbiebl:
Multiple issues here
- `make check` doesn't complete and doesn't produce a summary and test-suite.log
- imfile-growing-file-id.sh fails
- imfile-statefile-no-file_id-TO-file_id.sh

The problem is reproducible, so I'm happy to run further diagnostics

mbiebl:
Fortunately, individual log files have been written.
[imfile-growing-file-id.log](https://github.com/rsyslog/rsyslog/files/4061664/imfile-growing-file-id.log)
[imfile-statefile-no-file_id-TO-file_id.log](https://github.com/rsyslog/rsyslog/files/4061665/imfile-statefile-no-file_id-TO-file_id.log)

mbiebl:
I'm running `make check -j1` as unprivileged user.

mbiebl:
git bisect points at ac30968b7858d4ca3743d2b4d296eca543864fe2 as the first faulty commit
```
commit ac30968b7858d4ca3743d2b4d296eca543864fe2 (HEAD, refs/bisect/bad)
Author: Jiri Vymazal <jvymazal@redhat.com>
[...]

PR #4144, opened 2020-01-21
imfile bugfix: misadressing and potential segfault

rgerhards:
Commit 3f72e8c introduced an invalid memory allocation size. This lead to
too-short alloc and thus to overwrite of non-owned memory. That in turn
could lead to segfaults or other hard to find problems.

The issue was detected by our upgraded CI system. We did not receive
any problem reports in practice. Nevertheless, the problem is real and
people should update affected versions to patched ones.

see also: https://github.com/rsyslog/rsyslog/issues/4120


davidelang:
did this commit make it into any released versions? if so, can you update the commit message to indicate which versions?

rgerhards:
@davidelang amended it, was 8.1911.0

rgerhards:
CI failures are known-unrelated and being worked on elsewhere. So all go!

PR #4151, opened 2020-01-24
testbench: add script for improved "failed test" log output

rgerhards:
The main intent is to show only failed tests (not SKIPed ones)
to make it easier to interpret test failures.

This also limits the max output size of tests. This is especially
important because of a bug in buildbot that leads to DoS when
very large test logs are sent back to the build master (which
then aborts).

see also https://github.com/buildbot/buildbot/issues/4504


rgerhards:
CI failure is known problem being worked on.

PR #4159, opened 2020-01-30
testbench: create random test_id

rgerhards:
We use the test_id to generate unique files during parallel
runs. So far we used a combination of ms-based timestamp and
hash test name. Practice shows that under some circumstances
the generated ids are not unique and thus test get file conflicts
and false positives.

This is now changed so that the test_id we add 4 bytes of random
data to the test ID. This is hopefully sufficient to make the IDs
of test runing in parallel unique.

### DIFF
--- a/tests/diag.sh
+++ b/tests/diag.sh
@@ -2486,7 +2486,7 @@ case $1 in
 			echo "hint: was init accidentally called twice?"
 			exit 2
 		fi
-		export RSYSLOG_DYNNAME="rstb_$(./test_id $(basename $0))"
+		export RSYSLOG_DYNNAME="rstb_$(./test_id $(basename $0))$(tr -dc 'a-zA-Z0-9' < /dev/urandom | fold -w 4 | head -n 1)"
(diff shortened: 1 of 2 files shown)


davidelang:
remember that truely random numbers will generate collisions as well. can you 
use your pid as part of the id? That should be unique on the system.

David Lang

rgerhards:
> remember that truely random numbers will generate collisions as well. can you use your pid as part of the id? That should be unique on the system.

good point

as far as the pid goes, I would need the running shell's pid - is there any reliable way to get it.

Else I can again resort to including the hash of the test name. Or maybe add the random data to current hash-ms based id. I do not want to get the ID too long, as this makes file pathes also grow very long.

rgerhards:
@davidelang I changed the commit to add just 4 bytes of random data to the previous ID. The previous system worked in ~99% of the cases, so I hope this works out.

davidelang:
On Thu, 30 Jan 2020, Rainer Gerhards wrote:

> as far as the pid goes, I would need the running shell's pid - is there any reliable way to get it.

can you get parent pid from proc/self somehow?

David Lang

PR #4160, opened 2020-02-04
Add per minute rate limiting to imfile plugin

qu0zl:
Add MaxBytesPerMinute and MaxLinesPerMinute options.
These take integer values and, respectively, limit the number
of bytes or lines that may be sent in a minute.

This can be used to put a limit on the count or volume of logs
that may be sent for an imfile.

I have run into problems in production where developers ship changes that vastly increase the quantity of logs sent. I would like to be able to set rate limits on the number of messages or bytes sent per imfile.

This change solves this problem for me.

Initially I went per second, but that was too spiky and difficult to pick a sensible value for. Per minute works well for me.

I'm interested in feedback as to whether this could be merged, before I update the docs to match it.

### DIFF
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -33,6 +33,7 @@
 #include <pthread.h>
 #include <sys/types.h>
 #include <unistd.h>
+#include <time.h>
 #include <glob.h>
 #include <poll.h>
 #include <json.h>
@@ -108,6 +109,15 @@ static int bLegacyCnfModGlobalsPermitted;/* are legacy module-global config para
 	#define GLOB_BRACE 0
 #endif
 
+typedef struct per_minute_rate_limit_s per_minute_rate_limit_t;
+
+struct per_minute_rate_limit_s {
+	uint64_t maxBytesPerMinute;
+	uint32_t maxLinesPerMinute;
+	uint64_t bytesThisMinute; /* bytes sent so far this minute */
+	uint32_t linesThisMinute; /* lines sent to far this minute */
+	time_t rateLimitingMinute; /* minute we are currently rate limiting for */
+};
 
 static struct configSettings_s {
 	uchar *pszFileName;
@@ -120,6 +130,8 @@ static struct configSettings_s {
 	int iSeverity;  /* notice, as of rfc 3164 */
 	int readMode;  /* mode to use for ReadMultiLine call */
 	int64 maxLinesAtOnce;	/* how many lines to process in a row? */
+	uint64_t maxBytesPerMinute; /* maximum bytes per minute to send before rate limiting */
+	uint64_t maxLinesPerMinute; /* maximum lines per minute to send before rate limiting */
 	uint32_t trimLineOverBytes;  /* 0: never trim line, positive number: trim line if over bytes */
 } cs;
 
@@ -133,6 +145,7 @@ struct instanceConf_s {
 	uchar *pszStateFile;
 	uchar *pszBindRuleset;
 	int nMultiSub;
+	per_minute_rate_limit_t perMinuteRateLimits;
 	int iPersistStateInterval;
 	int iFacility;
 	int iSeverity;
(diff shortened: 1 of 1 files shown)


plugins/imfile/imfile.c
```diff
@@ -1295,6 +1310,39 @@ getStateFileName(const act_obj_t *const act,
 	return buf;
 }
 
+static rsRetVal
+checkPerMinuteRateLimits(per_minute_rate_limit_t *per_minute_rate_limits,
+			     const size_t msgLen)
+{
+	DEFiRet;
+	time_t current_minute = time(NULL)/60;
```
qu0zl, line 1318: minutes since the epoch. Safer than current minute of the day as this way each minute will only ever occur once.

rgerhards:
Thx for the PR. As a preliminary feedback; the feature looks appealing to me and a very quick and rough code review looks positive (bare a more in-depth review). Let's see what CI says. Note that we currently have some flakes in it - I'll keep an eye on that and let you know if something needs your attention.

plugins/imfile/imfile.c
```diff
@@ -1295,6 +1310,39 @@ getStateFileName(const act_obj_t *const act,
...
+	if(per_minute_rate_limits->maxBytesPerMinute) {
+		if (per_minute_rate_limits->rateLimitingMinute == current_minute) {
+			per_minute_rate_limits->bytesThisMinute += msgLen;
+			/* if we would breach our rate limit then do not send the message. */
```
qu0zl, line 1322: I prefer this to truncating based on the remaining bytes rate limit, and sending partial messages.

rgerhards:
milestone tentatively assigned

qu0zl:
Thank you, Rainer. I can write the docs, and fix anything the CI throws up.

GregFarrell:
@rgerhards the large majority of the buildbots passed, so I hope that those that failed are due to the flakey specs that you mentioned. They do not seem immediately linked to my code change.

I took a look at https://github.com/rsyslog/rsyslog-doc/blob/master/README.md , to extend the document, but figured that it was too early to write documentation for this.

Please let me know if you would like me to do anything on this.

Thanks again :)

rgerhards:
I am currently working on the flakes. My plan is to re-run them after I have merged the fix.

davidelang:
a thought on this, if imfile is going to fall behind in reading a file, there 
should be some way of tracking how far behind it is.

how expensive would it be to check the size of the file and make a stat (showing 
up in impstats) showing the difference between the current file position and the 
end of the file? (or something along those lines)

this stat would be 0 when it's not throttled.

David Lang

qu0zl:
That is an interesting thought. It would be ideal for me if I could easily raise an alarm to the team's responsible for the box when this rate limiting occurs. A monitorable stat would be very useful for that.

I don't know the Rsyslog code base very well, so no obvious lightweight approach for this came to mind. Light weight and negligible impact on CPU/log-throughput were my focus.

davidelang:
I'll wait for Rainer to weigh in on how to add a stat for pstats to output, but 
in term of overhead, a stats call to the filesystem would only happen when 
rsyslog is trottling the input, I don't know how frequently the stats call would 
need to be done (is there a housekeeping timer that could be used for this?). 
updating the stat each time you would normally read more data from the file but 
don't because of the throttling should work, and should still be far cheaper 
than reading/processing the data.

David Lang

rgerhards:
I would say let's merge this as-is and add the stats on top of it with another PR. That's even more important as I will be very limited available the next couple of days ;-)

I need to check how we can maintain it with low overhead, but I guess following @davidelang advice will lead us the right way.

qu0zl:
We've been using this in our production environment for 2 1/2 weeks now, with a couple of thousand EC2 instances. It protected us yesterday from a runaway host producing more than 6 million logs a minute :)

I'm happy with it as designed, and I hope we can merge it or something similar, so that we no longer need to maintain our own binary.

rgerhards:

qu0zl:
Docs added in https://github.com/rsyslog/rsyslog-doc/pull/889 , @rgerhards 

(Qu0zl is my personal account, and GregFarrell my work one).

rgerhards:
thx again, doing this as late merge for today's release :-)

GregFarrell:
Thank you, @rgerhards . That's brilliant :)

PR #4167, opened 2020-02-11
Fixed processing of 'cofig.enabled' directive

### LINKED ISSUE #4033: config.enabled directive does not work completely for all statements
### Expected behavior
config.enabled actually has the effect and does not produce errors in log

### Actual behavior
see below

see below

### Environment
- rsyslog version: rsyslog-8.1911.0-3.el8.x86_64
- platform: RHEL

```
# grep -n -v -e '^\s*#' -e '^\s*$' "/etc/rsyslog.conf"
12:module(load="imuxsock"         # provides support for local system logging (e.g. via logger command)
13:       SysSock.Use="off") # Turn off message reception via local log socket;
18:module(load="imjournal"          # provides access to the systemd journal
19:       StateFile="imjournal.state") # File to store the position in the journal
52:global(workDirectory="/var/lib/rsyslog")
57:module(load="builtin:omfile" Template="RSYSLOG_TraditionalFileFormat")
62:include(file="/etc/rsyslog.d/*.conf" mode="optional")
79:*.info;mail.none;authpriv.none;cron.none                /var/log/messages
82:authpriv.*                                              /var/log/secure
85:mail.*                                                  -/var/log/maillog
89:cron.*                                                  /var/log/cron
92:*.emerg                                                 :omusrmsg:*
95:uucp,news.crit                                          /var/log/spooler
98:local7.*                                                /var/log/boot.log
115:if $syslogfacility-text == 'local0' then {
116:  action(type="omfile" file="/var/log/mytest.log" config.enabled="off")
117:}

# rsyslogd -N 1
rsyslogd: version 8.1911.0-3.el8, config validation run (level 1), master config /etc/rsyslog.conf
rsyslogd: error during parsing file /etc/rsyslog.conf, on or before line 116: errors occured in file '/etc/rsyslog.conf' around line 116 [v8.1911.0-3.el8 try https://www.rsyslog.com/e/2207 ]

# grep -n -v -e '^\s*#' -e '^\s*$' "/etc/rsyslog.conf"
12:module(load="imuxsock"         # provides support for local system logging (e.g. via logger command)
13:       SysSock.Use="off") # Turn off message reception via local log socket;
18:module(load="imjournal"          # provides access to the systemd journal
19:       StateFile="imjournal.state") # File to store the position in the journal
52:global(workDirectory="/var/lib/rsyslog")
57:module(load="builtin:omfile" Template="RSYSLOG_TraditionalFileFormat")
62:include(file="/etc/rsyslog.d/*.conf" mode="optional")
79:*.info;mail.none;authpriv.none;cron.none                /var/log/messages
82:authpriv.*                                              /var/log/secure
85:mail.*                                                  -/var/log/maillog
89:cron.*                                                  /var/log/cron
92:*.emerg                                                 :omusrmsg:*
95:uucp,news.crit                                          /var/log/spooler
98:local7.*                                                /var/log/boot.log
115:global(localHostname="hostname1" config.enabled="on")
116:global(localHostname="hostname2" config.enabled="off")
117:local0.info action(type="omfile" file="/var/log/rsyslog-LocalHostName-test.log")
# rsyslogd -N 1
rsyslogd: version 8.1911.0-3.el8, config validation run (level 1), master config /etc/rsyslog.conf
rsyslogd: error during parsing file /etc/rsyslog.conf, on or before line 116: parameter 'localhostname' specified more than once - one instance is ignored. Fix config [v8.1911.0-3.el8 try https://www.rsyslog.com/e/2207 ]
rsyslogd: error processing global config parameters [global(...)] [v8.1911.0-3.el8 try https://www.rsyslog.com/e/2211 ]


# grep -n -v -e '^\s*#' -e '^\s*$' "/etc/rsyslog.conf"
12:module(load="imuxsock"         # provides support for local system logging (e.g. via logger command)
13:       SysSock.Use="off") # Turn off message reception via local log socket;
18:module(load="imjournal"          # provides access to the systemd journal
19:       StateFile="imjournal.state") # File to store the position in the journal
52:global(workDirectory="/var/lib/rsyslog")
57:module(load="builtin:omfile" Template="RSYSLOG_TraditionalFileFormat")
62:include(file="/etc/rsyslog.d/*.conf" mode="optional")
79:*.info;mail.none;authpriv.none;cron.none                /var/log/messages
82:authpriv.*                                              /var/log/secure
85:mail.*                                                  -/var/log/maillog
89:cron.*                                                  /var/log/cron
92:*.emerg                                                 :omusrmsg:*
95:uucp,news.crit                                          /var/log/spooler
98:local7.*                                                /var/log/boot.log
115:global(localHostname="hostname1" config.enabled="off")
116:global(localHostname="hostname2" config.enabled="on")
117:local0.info action(type="omfile" file="/var/log/rsyslog-LocalHostName-test.log")
# rsyslogd -N 1
rsyslogd: version 8.1911.0-3.el8, config validation run (level 1), master config /etc/rsyslog.conf
rsyslogd: error processing global config parameters [global(...)] [v8.1911.0-3.el8 try https://www.rsyslog.com/e/2211 ]


# grep -n -v -e '^\s*#' -e '^\s*$' "/etc/rsyslog.conf"
12:module(load="imuxsock"         # provides support for local system logging (e.g. via logger command)
13:       SysSock.Use="off") # Turn off message reception via local log socket;
18:module(load="imjournal"          # provides access to the systemd journal
19:       StateFile="imjournal.state") # File to store the position in the journal
52:global(workDirectory="/var/lib/rsyslog")
57:module(load="builtin:omfile" Template="RSYSLOG_TraditionalFileFormat")
62:include(file="/etc/rsyslog.d/*.conf" mode="optional")
79:*.info;mail.none;authpriv.none;cron.none                /var/log/messages
82:authpriv.*                                              /var/log/secure
85:mail.*                                                  -/var/log/maillog
89:cron.*                                                  /var/log/cron
92:*.emerg                                                 :omusrmsg:*
95:uucp,news.crit                                          /var/log/spooler
98:local7.*                                                /var/log/boot.log
115:parser(name="parser.rfc3164" type="rsyslog.rfc3164" config.enabled="on")
116:ruleset(name="myRuleset" parser="parser.rfc3164") {}
# rsyslogd -N 1
rsyslogd: version 8.1911.0-3.el8, config validation run (level 1), master config /etc/rsyslog.conf
rsyslogd: parser module name 'rsyslog.rfc3164' is unknown [v8.1911.0-3.el8 try https://www.rsyslog.com/e/2209 ]
rsyslogd: error: parser 'parser.rfc3164' unknown at this time (maybe defined too late in rsyslog.conf?) [v8.1911.0-3.el8 try https://www.rsyslog.com/e/2159 ]


# grep -n -v -e '^\s*#' -e '^\s*$' "/etc/rsyslog.conf"
12:module(load="imuxsock"         # provides support for local system logging (e.g. via logger command)
13:       SysSock.Use="off") # Turn off message reception via local log socket;
18:module(load="imjournal"          # provides access to the systemd journal
19:       StateFile="imjournal.state") # File to store the position in the journal
52:global(workDirectory="/var/lib/rsyslog")
57:module(load="builtin:omfile" Template="RSYSLOG_TraditionalFileFormat")
62:include(file="/etc/rsyslog.d/*.conf" mode="optional")
79:*.info;mail.none;authpriv.none;cron.none                /var/log/messages
82:authpriv.*                                              /var/log/secure
85:mail.*                                                  -/var/log/maillog
89:cron.*                                                  /var/log/cron
92:*.emerg                                                 :omusrmsg:*
95:uucp,news.crit                                          /var/log/spooler
98:local7.*                                                /var/log/boot.log
115:timezone(id="CET" offset="+01:00" config.enabled="on")
116:timezone(id="CEST" offset="+02:00" config.enabled="off")
# rsyslogd -N 1
rsyslogd: version 8.1911.0-3.el8, config validation run (level 1), master config /etc/rsyslog.conf
rsyslogd: error processing timezone config parameters [v8.1911.0-3.el8 try https://www.rsyslog.com/e/2211 ]
```

jvymazal:
Previously the directive was processed way too late which caused
false errors whenever it was set to 'off' and sometimes directive not
working at all.

closes #4033

### DIFF
--- a/grammar/rainerscript.c
+++ b/grammar/rainerscript.c
@@ -699,6 +699,23 @@ nvlstFindNameCStr(struct nvlst *lst, const char *const __restrict__ name)
 	return lst;
 }
 
+/* check if the nvlst is disabled, and mark config.enabled directive
+ * as used if it is not. Returns 1 if block is disabled, 0 otherwise.
+ */
+int nvlstChkDisabled(struct nvlst *lst)
+{
+	struct nvlst *valnode;
+
+	if((valnode = nvlstFindNameCStr(lst, "config.enabled")) != NULL) {
+		if(es_strbufcmp(valnode->val.d.estr, (unsigned char*) "on", 2)) {
+			return 1;
+		} else {
+			valnode->bUsed = 1;
+		}
+	}
+	return 0;
+}
+
 
 /* check if there are duplicate names inside a nvlst and emit
  * an error message, if so.
@@ -1204,23 +1221,6 @@ nvlstGetParams(struct nvlst *lst, struct cnfparamblk *params,
 		}
 		if(!nvlstGetParam(valnode, param, vals + i)) {
 			bInError = 1;
-		}
-	}
-
-	/* now config-system parameters (currently a bit hackish, as we
-	 * only have one...). -- rgerhards, 2018-01-24
-	 */
-	if((valnode = nvlstFindNameCStr(lst, "config.enabled")) != NULL) {
-		if(es_strbufcmp(valnode->val.d.estr, (unsigned char*) "on", 2)) {
-			dbgprintf("config object disabled by configuration\n");
-			/* flag all params as used to not emit error mssages */
-			bInError = 1;
-			struct nvlst *val;
-			for(val = lst; val != NULL ; val = val->next) {
-				val->bUsed = 1;
-			}
-		} else {
-			valnode->bUsed = 1;
 		}
 	}
 
@@ -4418,8 +4418,13 @@ cnfstmtNewAct(struct nvlst *lst)
 	struct cnfstmt* cnfstmt;
 	char namebuf[256];
 	rsRetVal localRet;
-	if((cnfstmt = cnfstmtNew(S_ACT)) == NULL)
+	if((cnfstmt = cnfstmtNew(S_ACT)) == NULL) {
 		goto done;
+	}
+	if (nvlstChkDisabled(lst)) {
+		dbgprintf("action disabled by configuration\n");
+		cnfstmt->nodetype = S_NOP;
+	}
 	localRet = actionNewInst(lst, &cnfstmt->d.act);
 	if(localRet == RS_RET_OK_WARN) {
 		parser_errmsg("warnings occured in file '%s' around line %d",
@@ -5281,6 +5286,11 @@ includeProcessCnf(struct nvlst *const lst)
 	if(lst == NULL) {
 		parser_errmsg("include() must have either 'file' or 'text' "
 			"parameter - ignored");
+		goto done;
+	}
+
+	if (nvlstChkDisabled(lst)) {
+		DBGPRINTF("include statement disabled\n");
 		goto done;
 	}
 
--- a/grammar/rainerscript.h
+++ b/grammar/rainerscript.h
@@ -340,6 +340,7 @@ void nvlstDestruct(struct nvlst *lst);
 void nvlstPrint(struct nvlst *lst);
 void nvlstChkUnused(struct nvlst *lst);
 struct nvlst* nvlstFindName(struct nvlst *lst, es_str_t *name);
+int nvlstChkDisabled(struct nvlst *lst);
 struct cnfobj* cnfobjNew(enum cnfobjType objType, struct nvlst *lst);
 void cnfobjDestruct(struct cnfobj *o);
 void cnfobjPrint(struct cnfobj *o);
--- a/runtime/rsconf.c
+++ b/runtime/rsconf.c
@@ -438,6 +438,16 @@ cnfDoObj(struct cnfobj *const o)
 
 	dbgprintf("cnf:global:obj: ");
 	cnfobjPrint(o);
+
+	/* We need to check for object disabling as early as here to cover most
+	 * of them at once and avoid needless initializations
+	 * - jvymazal 2020-02-12
+	 */
+	if (nvlstChkDisabled(o->nvlst)) {
+		dbgprintf("object disabled by configuration\n");
+		return;
+	}
+
 	switch(o->objType) {
 	case CNFOBJ_GLOBAL:
 		glblProcessCnf(o);


jvymazal:
@rgerhards this seems to fix most of the problems pointed in the issue, however the action is still complaining, from code it seems that `action()` statements are processed seperately from all others, but s far I did not manage to find where. Any idea?

jvymazal:
ok, I added a second evaluation of `config.enabled` directly for actions, however if you know of a way to have it in only one place feel free to edit the PR. Thanks

rgerhards:
> ok, I added a second evaluation of `config.enabled` directly for actions, however if you know of a way to have it in only one place feel free to edit the PR. Thanks

I think if we cannot go with the previous generic approach, that code is probably needed for every "object". It looks like "include" also needs it (if I look at CI).

I could not try it, but could we use the current code and return the vals pointer even in case the object is disabled?

jvymazal:
> I think if we cannot go with the previous generic approach, that code is probably needed for every "object". It looks like "include" also needs it (if I look at CI).
> 
> I could not try it, but could we use the current code and return the vals pointer even in case the object is disabled?

Unfortunately not, once we return non-null, in case of disabled object we would have to ensure in all the places it is used (from rough query 100+) that the code knows how to handle it and not continue with processing or doing whatever it does per the disabled block. Currently in few places where are looked, once it gets the vals object it assumes it is alright and immediately start instantiation or other initialization work. Also, when we not return NULL that means we have allocated the `vals` and we need to care about that the caller will free it (which will require additional changes as none of the code using it was designed to care about object it will not use).

rgerhards:
yeah, it's a bit bad ... maybe we need to choose between two solutions, each of which with disadvantages. I'll try to have a more in-depth look today, but cannot promise (and then will be off for a couple of days..). The CI fails are actually showing one more place where the current fix needs to be copied over to.

jvymazal:
> yeah, it's a bit bad ... maybe we need to choose between two solutions, each of which with disadvantages. I'll try to have a more in-depth look today, but cannot promise (and then will be off for a couple of days..). The CI fails are actually showing one more place where the current fix needs to be copied over to.

Yes, I see it now, that is 3 places already, I will try to refactor the patch and make a function out of it, hopefully that will simplify things a bit.

jvymazal:
@rgerhards, while I am at it, should we specifically check for "off" value on the directive and issue some error otherwise or treat anything other than "on" as legitimate off? (current state)

rgerhards:
> @rgerhards, while I am at it, should we specifically check for "off" value on the directive and issue some error otherwise or treat anything other than "on" as legitimate off? (current state)

The framework should already bark if its not on/off (or yes/no I think for historical reasons IIRC). If I am right, that would require further framework changes - best probably the introduction of a new param type.

jvymazal:
> > @rgerhards, while I am at it, should we specifically check for "off" value on the directive and issue some error otherwise or treat anything other than "on" as legitimate off? (current state)
> 
> The framework should already bark if its not on/off (or yes/no I think for historical reasons IIRC). If I am right, that would require further framework changes - best probably the introduction of a new param type.

Alright, was not aware of framework handling that, will leave it as it is then

jvymazal:
Just for the record, I have just tried putting some garbage in the `config.enabled` value and nothing seems to complain, even under debug it behaves the same as with "off", along with printing the garbage value into the debug output.

jvymazal:
@rgerhards unless you see some more problems with current state we should be good to merge, I tried the scenarios from the linked issue and results are looking alright.

rgerhards:
@jvymazal I checked the grammar definitions - I think you nailed it ;-)

rgerhards:
> Just for the record, I have just tried putting some garbage in the `config.enabled` value and nothing seems to complain, even under debug it behaves the same as with "off", along with printing the garbage value into the debug output.
> 
> EDIT: I just checked the docs and the above behavior is actually mentioned there, so it works as expected.

I also re-checked. The initial approach was rather hackish - The bottom line is that this functionality (and this only) does not and cannot use the regular config syntax processors. The reason is that it is implemented outside of the usual framework.

Do you think we need any more special treatment in the light of this?

jvymazal:
> Do you think we need any more special treatment in the light of this?

I think we should have the required ports covered, the object bit should care about most of config, only exceptions I found are the `action()` and `include()` which are handled separately. So at least per current documentation we are covering everything mentioned there, plus few extra things - with this patch I think there might be even possibility to have different sets of main queue parameters and template definitions which are switchable by environment variables (per the code it should work, however I did not try that explicitly).

Issue #4158, opened 2020-01-30
tcpflood erroneously splits big relp messages in two (or more) distinct messages


StrongestNumber9:
### Expected behavior
Expect relp message to either truncate at library defaults, or try to send it as a whole. Splitting makes rsyslog treat it as a new message.
```sh
Got hostname: hostname, app-name: tagname, inputname: imrelp, strlen: 250000
Got hostname: hostname, app-name: tagname, inputname: imptcp, strlen: 250000
```

### Actual behavior
Two (or more) distinct relp packets for one event: 
```sh
1138.806473344:imrelp.c       : imrelp.c: relp session read 32768 octets, buf '2 syslog 205824 <134>1 2020-01-01T00:00:00+02:00 hostname tagname - - ====<snip>

1138.818296922:imrelp.c       : imrelp.c: relp session read 32768 octets, buf '3 syslog 44231 ========<snip>
```

As seen from example config:
```sh
# rsyslogd -f /rsyslog.conf -n
Got hostname: hostname, app-name: tagname, inputname: imrelp, strlen: 205770
Got hostname: localhost, app-name: ===============<snip>================, inputname: imrelp, strlen: 43720
Got hostname: hostname, app-name: tagname, inputname: imptcp, strlen: 250000
```

### Steps to reproduce the behavior
rsyslog.conf
```sh
$LocalHostName test
$AbortOnUncleanConfig on
$PreserveFQDN on
main_queue(queue.type="Direct")
global(
    workDirectory="/tester"
    maxMessageSize="256k"
)

module(load="omstdout")
module(load="imrelp")
module(load="imptcp")
input(
    type="imrelp"
    name="imrelp"
    port="601"
    ruleset="print"
    MaxDataSize="256k"
)

input(
    type="imptcp"
    name="imptcp"
    port="514"
    ruleset="print"
)

template(name="print_message" type="list"){
    constant(value="Got hostname: ")
    property(name="hostname")
    constant(value=", app-name: ")
    property(name="app-name")
    constant(value=", inputname: ")
    property(name="inputname")
    constant(value=", strlen: ")
    property(name="$!strlen")
    constant(value="\n")
}
      
ruleset(name="print") {
    set $!strlen = strlen($msg);
    action(
        type="omstdout"
        template="print_message"
    )
}
```
Commands used:
```sh
[...]

PR #4168, opened 2020-02-11
testbench: increased max extra data length in tcpflood from 256 to 512KB

alorbach:
testbench: increased max extra data length in tcpflood from 256 to 512KB

Added a imrelp test for big messages (256KB).

closes: https://github.com/rsyslog/rsyslog/issues/4158


rgerhards:
not sure if this really addresses #4158 - waiting for response on that issue before merge

PR #4174, opened 2020-02-16
testbench: modify es_response_get_msgnum.py for python3

jwslater0823:
python3 does not like mixing spaces and tabs for indentation, so we get rid of the tab before print().


rgerhards:
thx for the PR! much apreciated.

PR #4176, opened 2020-02-17
Fix return for act_obj_add if open fails

tapakund:
act_obj_add should not return RS_RET_OK if open fails

### DIFF
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -711,10 +711,11 @@ act_obj_add(fs_edge_t *const edge, const char *const name, const int is_file,
 	if(fd < 0) {
 		if (is_file) {
 			LogError(errno, RS_RET_ERR, "imfile: error accessing file '%s'", name);
+			ABORT_FINALIZE(RS_RET_NO_FILE_ACCESS);
 		} else { /* reporting only in debug for dirs as higher lvl paths are likely blocked by selinux */
 			DBGPRINTF("imfile: error accessing directory '%s'", name);
+			ABORT_FINALIZE(RS_RET_FILE_OPEN_ERROR);
 		}
-		FINALIZE;
 	}
 	DBGPRINTF("add new active object '%s' in '%s'\n", name, edge->path);
 	CHKmalloc(act = calloc(sizeof(act_obj_t), 1));


tapakund:
ping. Please help to review the change.

rgerhards:
> ping. Please help to review the change.

keep in mind that this is volunteer work - I am pretty busy at the moment ;-)

rgerhards:
@tapakund please let me know why this patch is provided aka "what problem does ist fix?". I suggest to update the commit message accordingly.

tapakund:
> @tapakund please let me know why this patch is provided aka "what problem does ist fix?". I suggest to update the commit message accordingly.

@rgerhards  I have updated the commit message with the problem statement and the fix. Thanks for looking into it.

rgerhards:
@tapakund would you be willing to also submit a new testbench test to cover this scenario? You can have a look at the imfile tests in `./tests` - I can also help with any questions.

tapakund:
> @tapakund would you be willing to also submit a new testbench test to cover this scenario? You can have a look at the imfile tests in `./tests` - I can also help with any questions.

@rgerhards sure, i will take a look and update this PR.

tapakund:
@rgerhards have added a test for covering up this scenario.

tapakund:
Missed to push a change in Makefile. Will update.

tests/imfile-file-descriptor-exhasution.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+# This is part of the rsyslog testbench, licensed under GPLv3
```
rgerhards: Are you willing to change this to ASL 2.0? We are moving towards that license and I would prefer not having to remove things once (whenever this is...) we can do the switch.

tests/imfile-file-descriptor-exhasution.sh
```diff
@@ -0,0 +1,61 @@
...
+do
+	echo test >> $RSYSLOG_DYNNAME.input.$i.log;
+	if content_check --check-only "Too many open files" $RSYSLOG_DYNNAME.started; then
+		printf 'Failed to open file $RSYSLOG_DYNNAME.input.%d.log a FD exhasuted\n' "$i"
+		break
+	fi
+done
+
+#sleeping for 3 sec, as all FDs are exhasuted and no 
+#listner port available as well incomming connection
```
rgerhards: I admit I do not understand why the sleep **here at this point** does help. Can you elaborate?

tapakund: Sleep is not as such required here, i kept it to check even if all FDs exhausted, rsyslog process is not crashing. Let me know your comments as well.

tests/imfile-file-descriptor-exhasution.sh
tapakund: Sure, i will make the change and update. Thanks.

tests/imfile-file-descriptor-exhasution.sh
rgerhards: if we do not need `sleep` we should remove it. This has two advantages

1. testbench runs quicker (is already quite long running and, yes, single tests sum up)
2. `sleep` inside tests always cause doubt of raciness and thus make devs question why it is there ever and ever again. In sum, this wastes a lot of time.

tapakund: Thanks for your comment. I have removed sleep from the change.

tests/imfile-file-descriptor-exhasution.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+# This file is part of the rsyslog testbench, released  under ASL 2.0
```
rgerhards: I may be wrong, but I think the test can actually never fail. Did you forget the check?

tests/imfile-file-descriptor-exhasution.sh
```diff
@@ -0,0 +1,62 @@
...
+#keep a boolean to track if FD exhasuted
+
+# keep adding files and check wether the FD exhasuted
+num_of_files=0
+
+for (( i = 100; i < 1200 ; ++i ))
+do
+	echo test >> $RSYSLOG_DYNNAME.input.$i.log;
+	if content_check --check-only "Too many open files" $RSYSLOG_DYNNAME.started; then
+		printf 'Failed to open file $RSYSLOG_DYNNAME.input.%d.log a FD exhasuted\n' "$i"
```
rgerhards: now that we are at it again: typo "exhasuted" vs. "exhausted"

tests/imfile-file-descriptor-exhasution.sh
tapakund: My thought was, as this test is to open more files than the FD limit set for this process. 
So, if we cross that limit the process should not crash and should gracefully exit. So, in case the process crashed, then this test case will be marked failed.

I can add a test if the FD limit did not exhaust, but that won't be actually covering the scenario which we fixed.

Pls feel free to share your thoughts.

tests/imfile-file-descriptor-exhasution.sh
tapakund: i will modify this. Thanks for pointing out.

tests/imfile-file-descriptor-exhasution.sh
```diff
@@ -0,0 +1,62 @@
...
+do
+	echo test >> $RSYSLOG_DYNNAME.input.$i.log;
+done
+
+# Start rsyslog now
+startup
+
+#keep a boolean to track if FD exhasuted
+
+# keep adding files and check wether the FD exhasuted
```
tapakund: Need to even correct this typo, exhasuted vs exhausted.

tests/imfile-file-descriptor-exhasution.sh
rgerhards: > So, in case the process crashed, then this test case will be marked failed.

That's OK, but you do a content check for the error message in any case. So we should ensure that the message is present when we shut down rsyslog. Along the same lines, it probably makes sense to explicitly set the fd ulimit to 1024 so that we can test even if the OS by default provides more (or is this somehow a bad idea - not sure).

tapakund: Have added a check to make sure if the process crashed or still running, and depending on it test will Pass or Fail.

Also as you mentioned, i think its better to keep the fd limit 1024 for the specific PID when we are running the test. 

I executed the test with the failure and with changes and it looks fine.

Let me know your comments on the same.

tapakund: Also, as the crash was in process_symlink(), so i have modified the test case to exactly go through the same behaviour.

tests/imfile-file-descriptor-exhasution.sh
```diff
@@ -0,0 +1,69 @@
...
+
+# keep adding files and check wether the FD exhausted
+
+num_of_files=0
+for (( i = 100; i < 1200 ; ++i ))
+do
+	echo test >> $RSYSLOG_DYNNAME.input.$i.file;
+	ln -s $RSYSLOG_DYNNAME.input.$i.file $RSYSLOG_DYNNAME.input.$i.log
+	if  content_check --check-only "failed to processed incoming connection" $RSYSLOG_DYNNAME.started || \
+		content_check --check-only "Too many open files" $RSYSLOG_DYNNAME.started; then
```
rgerhards: shouldn't we do line 42 after shutdown as a mandatory check?

rgerhards: To be precise - line 42 I think should remain, just **adding** an additional check at the end.

tapakund: We can keep the check here as i don't want to keep creating files if FD is already exhausted.

In addition we can add one check at the end for confirming if we have hit this condition, before exiting.

Let me know if you had the same thoughts.

tapakund:
I see the change failing on contos6 as prlimit is not available. 

./imfile-file-descriptor-exhasution.sh: line 32: prlimit: command not found

Other failures are on other tests:
imfile-endregex-timeout-with-shutdown.sh
sndrcv_tls_ossl_certvalid_tlscommand.sh

I don't think these are relevant. Let me know you comments on the same.

Thanks.

tapakund:
> I see the change failing on centos6 as prlimit is not available.
> 
> ./imfile-file-descriptor-exhasution.sh: line 32: prlimit: command not found
> 
> Other failures are on other tests:
> imfile-endregex-timeout-with-shutdown.sh
> sndrcv_tls_ossl_certvalid_tlscommand.sh
> 
> Let me know your comments on the same.
> 
> Thanks.

@rgerhards The other way to fix this is to set LimitNOFILE=1024 in service file, which I see is set at LimitNOFILE=16384. Let me know your thoughts?


https://github.com/rsyslog/rsyslog/blob/ac0c4c0d84587f49bb0bae59d0e1734d3b8f8d81/rsyslog.service.in#L17

tapakund:
> @tapakund sry, I lost track on the PR :-( I think the best route probably is to skip the test if `prlimit` cannot be found. The service file does not help, because the CI does not run under systemd.
> 
> Are you still (after the long time) up for a change? Else I'll apply it :-)

@rgerhards Thanks for your response. I have updated the change as requested (skipped the test part).

Thanks.

tapakund:
I see https://github.com/rsyslog/rsyslog/commit/89ff6436b55cd81c54dcb076490b0c4de98d508d have fixed this issue. Thanks.

Issue #4170, opened 2020-02-11
NSD does not allow for intermediate / chain CA files


onet-git:
### Expected behavior

What is expected is the ability to specify additional certificate files to the openssl/gnutls chain. This is needed in the case where the client and server have different roots of trust / intermediate certs. This is a requirement for complying with RFC 5246 Section 7.4.2 certificate_list, which explicitly allows servers to send intermediate CA certs during the handshake process. 

A specific example for this need is where clients are using self-signed certs while the server is using a cert from an intermediate CA. The server needs the self-signed root to verify clients (e.g the cacert option) as well as the option to send the intermediate cert to clients.

nsd_ossl.c calls `SSL_CTX_use_certificate_file()` which by API specification reads only the first certificate in the file rather than the entire chain.

Per the openssl 1.1.1 documentation it is recommend that `SSL_CTX_use_certificate_chain_file()`, instead of `SSL_CTX_use_certificate_file()`. That call reads all certificates from the file instead of just the first.

An alternate solution is to explicitly allow add extra certs with the `SSL_CTX_add_extra_chain_cert()` call.

Note this is related to and I believe would fully fix https://github.com/rsyslog/rsyslog/issues/2547 in the openssl world. I believe gnutls has a similar calls at add additional certs to the chain.

There appears to be a work around introduced, but not documented, in https://github.com/rsyslog/rsyslog/pull/3611. Specifically the priority string can used to set SSL_CONF settings. `Certificate` setting calls `SSL_CTX_use_certificate_chain_file()` or the `ChainCAFile` setting if the chain is provided as a separate file. Perhaps a more holistic approach to supporting more advanced settings would be to encourage the use of this string and making the result codes visible outside of debug mode.

### Steps to reproduce the behavior

Have a server with a cert signed thru an intermediate CA that the client does not have in its certificate store. They will not be able to connect.

### Environment

rsyslog 8.2001.0

alorbach:
I took a look to the API description, I think it should be no problem for us to add an option to set the certificate chain file (SSL_CTX_use_certificate_chain_file). 

It would be helpful if you could provide us some help creating test certificates with a chain file. 

I will take care on this.

onet-git:
Thanks for fixing this already! You can create test cert using letsencrypt, which provides free certs signed using its intermediate signing key:

https://letsencrypt.org/getting-started/

PR #4182, opened 2020-02-19
openssl nsd driver: Switch to SSL_CTX_use_certificate_chain_file API

alorbach:
Switched from SSL_CTX_use_certificate_file to SSL_CTX_use_certificate_chain_file
API for loading the certificate into the certificate store.

According to the openssl doc, the SSL_CTX_use_certificate_chain_file API has the
advantage that it can load the main certificate (First in the PEM) and many chain
certificates after that automatically.

See notes section for more:
https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_use_certificate_chain_file.html

closes: https://github.com/rsyslog/rsyslog/issues/4170

### DIFF
--- a/runtime/nsd_ossl.c
+++ b/runtime/nsd_ossl.c
@@ -462,7 +462,7 @@ osslGlblInit(void)
 		osslLastSSLErrorMsg(0, NULL, LOG_ERR, "osslGlblInit");
 		ABORT_FINALIZE(RS_RET_TLS_CERT_ERR);
 	}
-	if(bHaveCert == 1 && SSL_CTX_use_certificate_file(ctx, certFile, SSL_FILETYPE_PEM) != 1) {
+	if(bHaveCert == 1 && SSL_CTX_use_certificate_chain_file(ctx, certFile) != 1) {
 		LogError(0, RS_RET_TLS_CERT_ERR, "Error: Certificate could not be accessed. "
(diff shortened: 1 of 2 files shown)


rgerhards:
do we have (or need) a similar patch for gnutls? Remeber: the functionality must be the same no matter which driver is used.

alorbach:
> do we have (or need) a similar patch for gnutls? Remeber: the functionality must be the same no matter which driver is used.

I will see what I can do for gnutls

alorbach:
gnutls driver is using "gnutls_x509_crt_list_import" to read the certificate file. 
As far as I understand the API description, this should import all certificates (chain) within the PEM file:
https://www.gnutls.org/reference/gnutls-x509.html#gnutls-x509-crt-list-import

We need to create a chained test certificate in order to test this.

rgerhards:
@alorbach This for me boils down to:

* the current openssl patch is actually a fix to get equal with gnutls
* we cannot test the new functionality as we do not have a test cert

In other words we are basically good to merge, but need to see that we can create a chained test cert in order to integrate a CI test?

rgerhards:
no doc needed - this just implements a missing feature that gnutls already had.

onet-git:
For visibility, you can create test certs using letsencrypt, which provides free certs signed using its intermediate signing key:

https://letsencrypt.org/getting-started/

Issue #3892, opened 2019-10-08
improve protection against duplicate instances


davidelang:
currently when rsyslog starts, it checks to see if a pidfile exists, and if it exists, rsyslog refuses to start.

However, if rsyslog crashes or is killed with a -9, it does not have a chance to remove the pidfile and so a replacement cannot be started

As an enhancement, rather than just depending only on the existance of a pid file, rsyslog should look in the pid file and check to see if there is a process running with that pid (ideally if it's an rsyslog process, but at least that _some_ process with that pid exists on the system), if there is no process with that pid, rsyslog should log a warning that there was a orphaned pid file and overwrite it.

smowafy:
I'm interested in this issue, if it's possible I can take it over.

**Disclaimer:** This would be my first meaningful contribution in a C project, so in case I took it please excuse me in case I was a little slow on this one.

davidelang:
go for it! The Adiscon team has such a long to-do list that they won't mind you 
taking this, and it's an issue that's been around forever, so it's not a 
pressing bug that needs to get fixed immediatly.

David Lang

rgerhards:
At least in theory, rsyslog should already do exactly that: https://github.com/rsyslog/rsyslog/blob/master/tools/rsyslogd.c#L326

Might it be that due to some SELinux or similar rsyslog has no permissions to send the signal to itself? Or Priv Drop? Not sure if we can come around this via a non-privileged interface.

davidelang:
My idea was not any attempt to communicate between instances, merely that if 
a pid file exists, look at the pid in it and if such a pid doesn't exist on the 
system, consider the pidfile orphaned and act like it didn't exist

if you do a normal kill, rsyslog has a chance to clean up after itself and 
remove the pid file, if you do a kill -9 it doesn't and the pid file remains

David Lang

rgerhards:
El mié., 9 oct. 2019 a las 9:32, David Lang (<notifications@github.com>)
escribió:

> My idea was not any attempt to communicate between instances, merely that
> if
> a pid file exists, look at the pid in it and if such a pid doesn't exist
> on the
> system, consider the pidfile orphaned and act like it didn't exist
>
>
This is exactly what the mentioned code does (or intends to do ;-)).

Rainer

skupkosk:
Some reading about new pidfd in Linux 5.1 kernel.
https://lwn.net/Articles/794707/
[...]

Issue #407, opened 2015-06-23
PID checking does not work properly.


ido:
If another process is running under the same PID as the PID currently in the `rsyslogd.pid` file, rsyslogd assumes it is still running when I try to start it up, and refuses to start. For example, after a forced reboot, `dbus-daemon` was running as PID 249, which was the same value as that in `rsyslogd.pid` retained from before the reboot when `rsyslogd` was shut down forcefully.

There are a number of ways to fix this, one of which is to consider rsyslogd as not running if the file is not currently locked (a la `flock`), even if a process currently exists with the same PID as the number in the contents of the PID file.

rgerhards:
This is no longer an issue in today's world.

PR #4191, opened 2020-02-22
improved pid checking and instance locking

jmarckel:
This is my first contribution, so I will introduce myself. My name is Jeff Marckel, and I have a considerable amount of experience as a C programmer on Linux systems that I would like to share back with the community.

This PR focuses on "PID checking and Instance Locking" that I feel should cover at least 2 open issues.

   https://github.com/rsyslog/rsyslog/issues/3892

   Here a user sees that a pidfile not tied to a running process can still
   block a new instance from starting. I believe that this is likely a case
   of #407, noted below, but some testing is required to exclude other issues.

   https://github.com/rsyslog/rsyslog/issues/407

   Here it is seen that pid re-use is an apparent vulnerability, where an
   unrelated process that just happens to get the pid that rsyslogd last used
   can block a new instance from starting.

In either case, the correct approach is to rely on using `fcntl()` to pass a `struct flock`, where any exclusive lock held by the dead process would be cleared by the system, meaning custom pid detection logic can be removed from the rsyslogd code entirely. From a brief examination of the code, it seems that there is also a race condition between checkStartupOK() and the creation of the actual pidfile in writePidFile(), which could pose a risk of multiple instances.

I implemented and have tested these changes, added a test to cover multiple instance startup of rsyslogd, and then used the new features of the lockfile to cross check the pidfile functionality in tests/diags.sh. I verified my testing on Travis. 

The pid validation function does not yet cause any tests to fail, but maybe that would be a good idea to turn on, but I did not know if it might cause problems for other tests. I have spot checked all tests I can and see good results. I am also thinking I will do more to take advantage of the lockfile to clean up diags.sh further, in terms of killing hung processes in a future PR. Feedback is welcome. 

Thank you,

Jeff

### DIFF
--- a/tests/diag.sh
+++ b/tests/diag.sh
@@ -515,6 +515,19 @@ startup() {
 	reassign_ports
 }
 
+# this function would only be valid after another good
+# startup() call and does nothing but try to start rsyslogd
+# just like the eval line in the function above, which
+# should fail to do more than cause some logging as the
+# forced process should just hit the pidfile's lock barrier and
+# exit after logging a message to stderr, which a test can
+# then detect in the log
+#
+# $1 is instance, some other vars might be missing, but ...
+startup_forced_second_instance() {
+	eval LD_PRELOAD=$RSYSLOG_PRELOAD $valgrind ../tools/rsyslogd -C $n_option -i$RSYSLOG_PIDBASE$1.pid -M../runtime/.libs:../.libs -f$CONF_FILE $RS_REDIR &
+}
+
 
 # assign TCPFLOOD_PORT from port file
 # $1 - port file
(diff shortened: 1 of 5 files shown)


jmarckel:
I see some failures popping up on older platforms that lack flock() support. I will have to find some work arounds for that, but will be working on it. Also, I may end up asking for advice on the new test, and the proper place for it to be looking for the stderr. I used the name I knew the output log would be using, which may be the issue on the centos6 fail.

rgerhards:
@jmarckel Thanks for contributing. We are happy to help. Just as a reminder, fewer and fewer distros are using the pid file functionality. It is most important especially for the older platforms.

Looking forward to the conversation.

jmarckel:
You are very welcome @rgerhards - I am happy for the chance to contribute to rsyslogd.

All the solaris build issues should be easy to fix. Using fcntl locking will work well, but I will need to move the lock taking code to take place after any call to fork(). I am going to look closer to make sure, but I don't really see any downside, as long as the child asserts the lock early, while the parent is still waiting for the OK. Honestly, after reading more about the implementation of flock() on some platforms, depending on the inheritance of that lock by a child process may not be reliable.

Regarding the other failing tests:

* I can't quite figure out the centos6 issue, but I suppose that the output file is just not sync'd to disk and so the test is stuck in a loop looking for it.
* the ubuntu16 thing is a failure from functionality I did not change, not that I am claiming innocence of course, but I am not sure how that happened...

Thank you for the context. I will push out a fix shortly.

rgerhards:
I admit I haven't looked at all the failures in detail. We still have some flakes, albeit I managed to get things much better in the past couple of weeks. Usually if there is a large number of failures I just assume there is something real in it. Failures in unrelated functionality can be considered as flakes until we see it is differently.

Note: if we have few failures, I usually check them individually and babysit where needed.

jmarckel:
Yes @rgerhards - I agree and will make those changes to `tests/diag.sh` that I had flagged as TODO where we will use `fuser -k` instead of trying to kill the pid in the pidfile and also to fail tests that enter `shutdown_when_empty()` without a running instance of `rsyslogd` holding the lock.

rgerhards:
@jmarckel great, thx!

jmarckel:
Hello @rgerhards - I had not really anticipated the `fuser` command being missing, and will need to examine that more closely. Also, on Solaris, the `fuser` command uses other signal indications, so will need to examine that too.

jmarckel:
I will submit a PR for changes to the appropriate Dockerfile for the various dev containers to include the `pstools` package so we can have the `fuser` command available to the test infrastructure.

rgerhards:
> I will submit a PR for changes to the appropriate Dockerfile for the various dev containers to include the `pstools` package so we can have the `fuser` command available to the test infrastructure.

I have some updates for the containers on my TODO list, I can just add the package while I do them.
Please also consider the `check_command_available` check inside the test. For a sample, see https://github.com/rsyslog/rsyslog/blob/master/tests/omkafka.sh#L6 . This is important as we never know in which environment the tests are running.

rgerhards:
@jmarckel FYI: at least on Ubuntu 18.04 it's `psmisc`, not `pstools`. Preping a container with it.

tests/diag.sh
```diff
@@ -880,6 +893,87 @@ wait_queueempty() {
...
+	else
+		pidfile_pid=$(cat "$PF")
+		if [ -z $pidfile_pid ]; then
+			echo "validate_pidfile: pidfile pid was not detected"
+			RETCODE=1
+		fi
+	fi
+
+	if [ -f ${PF}.lock ]; then
+		fuser_pid=$(fuser "${PF}.lock" 2> /dev/null)
```
rgerhards, line 944: I just noticed that you use `fuser` inside the testbench plumbing and thus in all tests. We cannot do this, basic tests need to be able to run on as plain vanilla systems as possible (we cannot force folks to install everything). So please check if `fuser` is available and downgrade gracefully when not (but otherwise keep the tests running).

For single tests, it is OK if they skip when the tool is not available.

jmarckel:
Hello @rgerhards thank you, sorry for the delay, I had some things come up. I will get this change finished up this week.

tests/diag.sh
jmarckel: Hello @rgerhards - I am finishing changes now that will fall back to old way of getting pid from pidfile when `fuser` command is not available. Would that be ok? My thinking here was that when `fuser` is available, we should use it, because it makes the test stronger.

rgerhards: sorry for the late reply - I was swamped with quite some things due to the Corona crisis. I am now re-reviewing.

Yes, if we fall back to the old way, that would be ok. The important thing is that everything continues as before on platforms where fuser is not present (else we will effectively reduce platform coverage).

tests/diag.sh
```diff
@@ -2567,7 +2692,13 @@ case $1 in
 		fi
 		;;
    'kill-immediate') # kill rsyslog unconditionally
-		kill -9 $(cat $RSYSLOG_PIDBASE.pid)
+		command -v fuser > /dev/null 2>&1
```
rgerhards: I think it would make sense to put this common code into a function.

tools/rsyslogd.c
```diff
@@ -1621,10 +1658,8 @@ initAll(int argc, char **argv)
 	hdlr_enable(SIGCHLD, hdlr_sigchld);
 	hdlr_enable(SIGHUP, hdlr_sighup);
 
-	if(rsconfNeedDropPriv(ourConf)) {
-		/* need to write pid file early as we may loose permissions */
-		CHKiRet(writePidFile());
-	}
+	CHKiRet(checkStartupOK());
```
rgerhards: in the case of non-priv drop, this now writes the pid file earlier than before. That can cause races in some exotic situations. I know the chance is slim, but I remember cases where we had this in the testbench, for example.

tests/diag.sh
jmarckel: Hi @rgerhards - I agree that a function makes sense here, and will push that change shortly.

tools/rsyslogd.c
jmarckel: Thank you @rgerhards - apologies for delays on my side as well, these are interesting times. I will refresh my memory on this code path and look at what should be done. It would be interesting to understand that race condition better too. Do you happen to recall which testbench test caused the race?

tests/diag.sh
jmarckel: I have pushed this change, with one more fix for the non-priv drop code path mentioned pending.

tools/rsyslogd.c
jmarckel: Hello @rgerhards - hope you are well. Sorry for the delay, but I have pushed a change that puts the non-priv drop parts of the code path back the way it was. Thank you.

tests/diag.sh
jmarckel: Yes, we fall back to the old way when needed.

jmarckel:
Hello @rgerhards - I wanted to reach out to see if there was anything else I could to to help get this PR merged. I believe that the issues we discussed have been resolved. The 2 tests that had failed on the last run do not appear to be related to the change set.

rgerhards:
sry for loosing track. Given the advancements in state until now, I will close unmerged. If you think it should still be merged, please let me know and I'll re-check.

My apologies for the very late reply and action.

Issue #2133, opened 2017-11-29
remove "strcat" from source


rgerhards:
```strcat()``` is prone to buffer overruns. We need to remove that function and replace it by a more secure solution.

rgerhards:
I close this issue because the unsafe `strcat()` use has been removed from current source.

On current `main`, `git grep strcat` only finds historical `ChangeLog` entries and no tracked source use. That satisfies the original cleanup request.

If a new unsafe string append pattern is found, it should be handled as a new, specific cleanup issue.

PR #4193, opened 2020-02-25
imjournal: remove strcat call

jmarckel:
Hello @rgerhards -

I noticed the #2133 issue and thought I would make a run at some of them.

This change removes usage of strcat call from the imjournal plugin. I also looked for sprintf and strcpy calls while I was in there. I do not like the code here too much even though it is safer from the buffer overrun perspective. The way it could truncate the last 4 bytes of a very long state file path with a short state file name might result in a hidden file - or worse  in some other directory completely. I thought it would be better to just mention what I saw though, since it is unrelated, and another can of worms entirely. Happy to write another bug if you wanted.

I also spent some time on the test, it is more stable for me now.

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -520,24 +520,32 @@ persistJournalState(void)
 	DEFiRet;
 	FILE *sf = NULL; /* state file */
 	char tmp_sf[MAXFNAME];
-	size_t n;
 
 	DBGPRINTF("Persisting journal position, cursor: %s, at head? %d\n",
 			  journalContext.cursor, journalContext.atHead);
 
 	/* first check that we have valid cursor */
 	if (!journalContext.cursor) {
+		DBGPRINTF("Journal cursor is not valid, ok...\n");
 		ABORT_FINALIZE(RS_RET_OK);
 	}
 
 	/* we create a temporary name by adding a ".tmp"
 	 * suffix to the end of our state file's name
+	 *
+	 * we use snprintf() to safely honor the boundaries
+	 * of the temporary state file name buffer by using
+	 * a precision specifier, which will limit the number
+	 * of bytes taken from cs.stateFile to what will fit
+	 *
+	 * TODO: figure out a better way to avoid the PATH_MAX
+	 * problem. The truncated cs.stateFile with .tmp at the
+	 * end is not optimal
 	 */
-	n = snprintf(tmp_sf, sizeof(tmp_sf), "%s.tmp", cs.stateFile);
-	if (n >= sizeof(tmp_sf)) { /* backup in case of too long path name */
-		strncpy(tmp_sf, cs.stateFile, sizeof(tmp_sf) - 5);
-		strcat(tmp_sf, ".tmp");
-	}
+#define IM_SF_TMP_SUFFIX ".tmp"
+	snprintf(tmp_sf, sizeof(tmp_sf), "%.*s%s",
+			(int)(sizeof(tmp_sf) - strlen(IM_SF_TMP_SUFFIX) -1),
+			cs.stateFile, IM_SF_TMP_SUFFIX);
 
 	sf = fopen(tmp_sf, "wb");
 	if (sf == NULL) {
@@ -572,6 +580,8 @@ persistJournalState(void)
 			ABORT_FINALIZE(RS_RET_IO_ERROR);
 		}
 	}
+
+	DBGPRINTF("Persisted journal to '%s'\n", cs.stateFile);
 
 finalize_it:
 	if (sf != NULL) {
--- a/tests/imjournal-statefile.sh
+++ b/tests/imjournal-statefile.sh
@@ -25,11 +25,22 @@ if [ $? -ne 0 ]; then
         echo "SKIP: failed to put test into journal."
         error_exit 77
 fi
-journalctl -an 200 | fgrep -qF "$TESTMSG"
+
+# sleep 1 to get this test to reliably detect the message
+sleep 1
+
+journalctl -an 200 > /dev/null 2>&1
 if [ $? -ne 0 ]; then
         echo "SKIP: cannot read journal."
         error_exit 77
 fi
+
+journalctl -an 200 | grep -qF "$TESTMSG"
+if [ $? -ne 0 ]; then
+        echo "SKIP: cannot find '$TESTMSG' in journal."
+        error_exit 77
+fi
+
 # do first run to process all the stuff already in journal db
 startup
 


tests/imjournal-statefile.sh
```diff
@@ -25,11 +25,22 @@ if [ $? -ne 0 ]; then
...
+
+journalctl -an 200 > /dev/null 2>&1
 if [ $? -ne 0 ]; then
         echo "SKIP: cannot read journal."
         error_exit 77
 fi
+
+journalctl -an 200 | grep -qF "$TESTMSG"
+if [ $? -ne 0 ]; then
+        echo "SKIP: cannot find '$TESTMSG' in journal."
```
jmarckel, line 40: Here I tried to split a failure to run journalctl from a failure to find the message.

I considered failing the test something other than 77, but I don't know all the details about why that might be a bad idea. :)

rgerhards: That's a guard if things go wrong for non-rsyslog related tests. The idea is if we cannot find the message ourself, rsyslog can for sure also not find it. So not checking would create a false positive (actually, it did - that's why the guard is now in place).

rgerhards:
@jvymazal could you pls have a look at this PR? Would be great!

tests/imjournal-statefile.sh
jmarckel: Hi @rgerhards - thank you. The `sleep 1` line was important here.  Additionally, I should note that I split the `journalctl` guard you mention into two pieces. Assuming we can run `journalctl` at all, the second test could fail, taking an opportunity to close a risk of a false negative. If we can get the sleep interval correct (and maybe on some platforms it is 5.001 minutes) then the failure of `journalctl` to produce the message to the test is a failure to examine.

plugins/imjournal/imjournal.c
```diff
@@ -520,24 +520,23 @@ persistJournalState(void)
...
 	FILE *sf = NULL; /* state file */
 	char tmp_sf[MAXFNAME];
-	size_t n;
 
 	DBGPRINTF("Persisting journal position, cursor: %s, at head? %d\n",
 			  journalContext.cursor, journalContext.atHead);
 
 	/* first check that we have valid cursor */
 	if (!journalContext.cursor) {
+		DBGPRINTF("Journal cursor is not valid, ok...\n");
```
jvymazal, line 529: More precisely we do not have any cursor at all. I, as journald API user, cannot in fact decide by myself whether cursor is valid (journald API call is needed for that)

plugins/imjournal/imjournal.c
```diff
@@ -520,24 +520,23 @@ persistJournalState(void)
...
 
 	/* we create a temporary name by adding a ".tmp"
 	 * suffix to the end of our state file's name
 	 */
-	n = snprintf(tmp_sf, sizeof(tmp_sf), "%s.tmp", cs.stateFile);
-	if (n >= sizeof(tmp_sf)) { /* backup in case of too long path name */
-		strncpy(tmp_sf, cs.stateFile, sizeof(tmp_sf) - 5);
-		strcat(tmp_sf, ".tmp");
-	}
+#define SUFFIX ".tmp"
```
jvymazal: maybe a nitpick, however `SUFFIX` seems quite generic, I would be afraid of it clashing with something.

plugins/imjournal/imjournal.c
```diff
@@ -520,24 +520,23 @@ persistJournalState(void)
...
+	snprintf(tmp_sf, sizeof(tmp_sf), "%.*s%*.s",
```
jvymazal: While this, after lot of digging trough docs seems to do the trick, IMHO it warrants some explaining comment, preferably longer one ;-) 
Also, what is the advantage of doing the argument-defined minimum width with statically defined constant?

jmarckel: I really smiled when it occurred to me that this would work well, so yes, I suppose a comment is in order. I will add that. As for the advantage of the minimum, I am not sure exactly why I left it there, it isn't necessary and I really should have minimized it in advance. I will remove it to save some typing in the comment. :)

plugins/imjournal/imjournal.c
jmarckel: :)

plugins/imjournal/imjournal.c
jmarckel: Agreed, I will fix it.

plugins/imjournal/imjournal.c
```diff
@@ -520,24 +520,32 @@ persistJournalState(void)
...
 		ABORT_FINALIZE(RS_RET_OK);
 	}
 
 	/* we create a temporary name by adding a ".tmp"
 	 * suffix to the end of our state file's name
+	 *
+	 * we use snprintf() to safely honor the boundaries
+	 * of the temporary state file name buffer by using
+	 * a precision specifier, which will limit the number
+	 * of bytes taken from cs.stateFile to what will fit
```
rgerhards, line 539: We do not need to take care of this - snprintf already does. If the buffer is too slow, snprintf returns a size larger than the buffer **without doing any writes**. This is what the previous code uses, which IMHO is perfectly OK (but I may be wrong, pls point out where exactly if so).

tests/imjournal-statefile.sh
rgerhards: practice was different: it was always either

1. bug in journal
2. journal was rotated away

I don't like to go back to the myriad of false positives we had with the journal tests around a year or so ago. If journalctl cannot find the message, let's assume it was rotated away.

jmarckel: Hello @rgerhards - this code will skip in either case. The difference is that the log will indicate if case 1 or case 2 was more likely to have happened. Are you asking that we not do that differentiation?

plugins/imjournal/imjournal.c
jmarckel: Hello @rgerhards - The new snprintf call writes the file name and extension all in one go. The comment is explaining the use of the precision specifier to truncate just the cs.stateFile part of the string, and avoid the need for the alternate strncpy() strncat() block. Are you asking for the comment to change?

tests/imjournal-statefile.sh
rgerhards: @jmarckel  got you - overlooked it's just two skip cases

plugins/imjournal/imjournal.c
```diff
@@ -520,24 +520,32 @@ persistJournalState(void)
...
+	 * end is not optimal
 	 */
-	n = snprintf(tmp_sf, sizeof(tmp_sf), "%s.tmp", cs.stateFile);
-	if (n >= sizeof(tmp_sf)) { /* backup in case of too long path name */
-		strncpy(tmp_sf, cs.stateFile, sizeof(tmp_sf) - 5);
-		strcat(tmp_sf, ".tmp");
-	}
+#define IM_SF_TMP_SUFFIX ".tmp"
+	snprintf(tmp_sf, sizeof(tmp_sf), "%.*s%s",
+			(int)(sizeof(tmp_sf) - strlen(IM_SF_TMP_SUFFIX) -1),
```
rgerhards, line 547: I suggest
```
(int)(sizeof(tmp_sf) - sizeof(IM_SF_TMP_SUFFIX) -1),
```
which I guess was the original intent (sizeof vs. strlen). With strlen I think the remainung buffer space may be one byte too few.

plugins/imjournal/imjournal.c
rgerhards: looks like I had a bad day during the review. Makes sense.

plugins/imjournal/imjournal.c
jmarckel: I believe it is clearer to leave it with `strlen()` since that is the intent - to reserve space for the length of the suffix and one for the null, and also because it would remain correct even if `IM_SF_TMP_SUFFIX` were changed to be declared as `char *`. Using -1 is correct to reserve space for the null byte in this case. I think `- sizeof() - 1` would reserve too much space.

rgerhards: OK. Just two words of warning

1. not all compilers optimize out `strlen("constant")`, so this can lead to bigger and slower code
1. the rest of the code uses `(sizeof("constant")-1)`, so this would be the only occasion where `strlen("constant")` is used - which may cause some confusion (like mine on the meaning of  `-1`)

I would still accept this, because it is imjournal, which is special in many ways and slow in any case. If it were some other component, I'd insist on `sizeof()` ;-)

PR #4196, opened 2020-02-26
imsolaris: remove unsafe strcat, sprintf calls

jmarckel:
Hi @rgerhards - this commit is very raw until it runs through some Solaris CI as I do not currently have development access to this OS platform.

### DIFF
--- a/plugins/imsolaris/sun_cddl.c
+++ b/plugins/imsolaris/sun_cddl.c
@@ -228,10 +228,12 @@ sun_open_door(void)
 				    info.di_target);
 
 				if (info.di_target > 0) {
-					(void) sprintf(line, "syslogd pid %ld"
-					    " already running. Cannot "
-					    "start another syslogd pid %ld",
-					    info.di_target, getpid());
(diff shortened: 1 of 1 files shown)


jmarckel:
Hi @rgerhards - I see that this code has built well and passed all tests on solaris platforms. I downloaded the stdio log from a long `make check` buildbot run and looked closely for signs of problems with this change. No problems stood out, however I do not see where either of these error messages were triggered by the tests. I do not see any test coverage for provoking door failure states, so not seeing the log messages is probably OK.

rgerhards:
very busy atm. I am hesitant to change solaris-only code as, like you have seen, we do not have test nor real know-how available for that platform. Do you think the change is worth the risk (I did **not** look at the code so far)?

jmarckel:
Yes, I understand you are busy, with this being a low priority change. I will pull the PR for now. If I get a chance I will set up a Solaris VM and test it, maybe I can come back with a better story. It should be very easy to add an option to the module to pretend the door is broken, so proving the code works will be easier. Thank you @rgerhards .

PR #4198, opened 2020-02-26
imfile: remove strcat, strcpy calls

jmarckel:
Hello @rgerhards - Related to issue #2133, in this PR I have removed uses of strcat and strcpy from the imfile plugin. I also added a new test which enables the imfile option addceetag, providing some coverage to the change. I ran the test with debug enabled, and examined logs to make sure that the changes were exercised.

### DIFF
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -710,7 +710,7 @@ act_obj_add(fs_edge_t *const edge, const char *const name, const int is_file,
 	char basename[MAXFNAME];
 	DEFiRet;
 	int fd = -1;
-	
+
 	DBGPRINTF("act_obj_add: edge %p, name '%s' (source '%s')\n", edge, name, source? source : "---");
 	for(act = edge->active ; act != NULL ; act = act->next) {
 		if(!strcmp(act->name, name)) {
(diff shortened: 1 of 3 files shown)


plugins/imfile/imfile.c
```diff
@@ -1374,8 +1374,9 @@ enqLine(act_obj_t *const act,
 		size_t ceeMsgSize = msgLen + CONST_LEN_CEE_COOKIE + 1;
 		char *ceeMsg;
 		CHKmalloc(ceeMsg = malloc(ceeMsgSize));
-		strcpy(ceeMsg, CONST_CEE_COOKIE);
-		strcat(ceeMsg, (char*)rsCStrGetSzStrNoNULL(cstrLine));
+		snprintf(ceeMsg, ceeMsgSize, "%s%s", CONST_CEE_COOKIE,
```
rgerhards, line 1377: I don't like snprintf here - it is so much slower. What is the issue with strcpy at this point inside the code? I guess it is more the bounding of the strcat(), right? If so, it probably makes sense to query cstrLine for its size - but this is done in line 1362 already, so we know the buffer is sufficiently large.  We could probably replace the strcpy() by a memcpy() if we fear that an attacker might change the buffer between the point where the size is taken and then used. What do you think?

plugins/imfile/imfile.c
```diff
@@ -1789,8 +1791,9 @@ checkInstance(instanceConf_t *const inst)
 					inst->pszFileName);
 				ABORT_FINALIZE(RS_RET_ERR);
 			}
-			curr_wd[len_curr_wd] = '/';
-			strcpy((char*)curr_wd+len_curr_wd+1, (char*)inst->pszFileName);
+			p_fname = (char *)curr_wd + len_curr_wd;
+			const size_t p_fname_max = sizeof(curr_wd) - len_curr_wd;
+			snprintf(p_fname, p_fname_max, "/%s", (char*)inst->pszFileName);
```
rgerhards, line 1796: We should really not use snprintf for such things ;-) - stopping review here

PR #4201, opened 2020-02-28
fix -o which produces missing space between call and rulename

### LINKED ISSUE #3634: add a capability to create consolidated config file
This goes as a side-line to https://github.com/rsyslog/rsyslog/issues/3584

We should have a capability to create an output config file that explodes all "includes" into a single file. This provides a much better overview of how exactly the configuration is crafted. That could often be a great troubleshooting aid.

### LINKED ISSUE #3761: 'rsyslogd -o' option produces config file with missing space between "call" and ruleset name
### Expected behavior

Generated config file with space between `call` and ruleset name

### Actual behavior

Config file with entries like this one (note the lack of a space):

```
callclient-record-local-properties
```

`rsyslogd -f /etc/rsyslog.conf -N3 -o rsyslog-combined-config.conf` when using configuration that includes multiple `call` statements

### Environment
- rsyslog version: `8.1908.0.c19b240adbcf` 
- platform: Ubuntu 16.04

leneron:
closes #3761

### DIFF
--- a/grammar/lexer.l
+++ b/grammar/lexer.l
@@ -312,7 +312,7 @@ int fileno(FILE *stream);
 <EXPR>.				{ cnfPrintToken(yytext); parser_errmsg("invalid character '%s' in expression "
 					        "- is there an invalid escape sequence somewhere?",
 						yytext); }
-<INCALL>[ \t\n]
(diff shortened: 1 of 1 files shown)


rgerhards:
Thx for the PR, much apreciated.

PR #4211, opened 2020-03-10
Hardcoded default hash algorithm KSI_HASHALG_SHA2_256 replaced with

taavi-valjaots:
KSI_getHashAlgorithmByName("default").

Function rsksiSetHashFunction and SetCnfParam updated.


rgerhards:
@taavi-valjaots thx for the PR. I would appreciate if you could start the commit comment with a standard header line like
```
ksi bugfix: xxxx
```
This makes sense better readable. Also pls remember keeping an empty line between title and commit message.

PR #4212, opened 2020-03-10
When KSI module in async. mode is suddenly closed, files are finalized

taavi-valjaots:
correctly. All pending signature requests are closed immediately and
unsigned block marker is attached with message about sudden closure.
Similar approach is used for blocks that already contain some records.
Empty blocks are just closed without any metadata.

### DIFF
--- a/runtime/lib_ksils12.c
+++ b/runtime/lib_ksils12.c
@@ -90,9 +90,14 @@ typedef enum QITEM_type_en {
 
 /* Worker queue item status identifier */
 typedef enum QITEM_status_en {
+	/* State assigned to any item added to queue (initial state). */
 	QITEM_WAITING = 0x00,
+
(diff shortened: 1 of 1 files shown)


rgerhards:
@taavi-valjaots thx for the PR. I would appreciate if you could start the commit comment with a standard header line like
```
ksi bugfix: xxxx
```
This makes sense better readable. Also pls remember keeping an empty line between title and commit message.

rgerhards:
@taavi-valjaots sry, looks like the merge sequence created a merge conflict. Could you do a rebase on this PR?

rgerhards:
@taavi-valjaots I notice it is just a quick quirk.. will see that I can fix it via manual merge...

rgerhards:
@taavi-valjaots  done!

Issue #4210, opened 2020-03-10
Fix LibreSSL support (does not support for @SECLEVEL and SSL_CONF_CTX_set_flags())


alorbach:
See https://github.com/rsyslog/librelp/pull/180 , simular fix needs to be done for rsyslog to enable libressl support. 

@rgerhards  Maybe we need a machine for libressl testing?!

rgerhards:
> @rgerhards Maybe we need a machine for libressl testing?!

That's a good point. Can you find out if we can do this e.g. with Debian? I could then craft a container.

alorbach:
According to this (https://www.libressl.org/), FreeBSD would be the right choice

rgerhards:
ok, than I think it is too much effort. BSD doesn't work in containers.

PR #4213, opened 2020-03-10
openssl: add check for LIBRESSL

alorbach:
Disable use of "@SECLEVEL" in default cipher string and
avoid SSL_CONF_CTX_set_flags() API when LIBRESSL is used.
This means tlscommands will not work.

closes: https://github.com/rsyslog/rsyslog/issues/4210

### DIFF
--- a/runtime/nsd_ossl.c
+++ b/runtime/nsd_ossl.c
@@ -633,7 +633,7 @@ osslInitSession(nsd_ossl_t *pThis) /* , nsd_ossl_t *pServer) */
 
 	if (bAnonInit == 1) { /* no mutex needed, read-only after init */
 		/* Allow ANON Ciphers */
-		#if OPENSSL_VERSION_NUMBER >= 0x10100000L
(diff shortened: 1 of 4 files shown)


alorbach:
There is already an error logged, see here https://github.com/rsyslog/rsyslog/blob/master/runtime/nsd_ossl.c#L1862

I can adjust the message like "OpenSSL Version to old or SSL_CONF_cmd not supported. "

rgerhards:
> I can adjust the message like "OpenSSL Version to old or SSL_CONF_cmd not supported. "

I think that message would be mis-leading. I'd suggest to go with something like
```
TLS library does not support SSL_CONF_cmd (maybe it is too old?) "
```

alorbach:
> OpenSSL Version to old

this needs to remain in the error message, because some test check for this string.

rgerhards:
> OpenSSL Version to old

Well... I would suggest to fix the tests -- especially is there is a typo "to" vs "too" ;-)

alorbach:
Should be ok now.

PR #4219, opened 2020-03-12
omhttp: Adding multiple http headers capability

callmegar:
Allows the inclusion of multiple http headers on the REST call, uses ":" as a delimiter for multiple header keys and values using the existing httpheaderkey and httpheadervalue parameters, you can specify up to 20 headers. Example:
 
    httpheaderkey="X-Insert-Key:X-Event-Source"
    httpheadervalue="key:logs"

### DIFF
--- a/contrib/omhttp/omhttp.c
+++ b/contrib/omhttp/omhttp.c
@@ -119,6 +119,8 @@ typedef struct instanceConf_s {
 	uchar *httpheaderkey;
 	uchar *httpheadervalue;
 	uchar *headerBuf;
+	uchar **httpHeaders;
+	int nHttpHeaders;
 	uchar *restPath;
 	uchar *checkPath;
 	uchar *tplName;
@@ -189,6 +191,7 @@ static struct cnfparamdescr actpdescr[] = {
 	{ "httpcontenttype", eCmdHdlrGetWord, 0 },
 	{ "httpheaderkey", eCmdHdlrGetWord, 0 },
 	{ "httpheadervalue", eCmdHdlrString, 0 },
+	{ "httpheaders", eCmdHdlrArray, 0 },
 	{ "uid", eCmdHdlrGetWord, 0 },
 	{ "pwd", eCmdHdlrGetWord, 0 },
 	{ "restpath", eCmdHdlrGetWord, 0 },
@@ -299,6 +302,11 @@ CODESTARTfreeInstance
 	free(pData->headerContentTypeBuf);
 	free(pData->httpheaderkey);
 	free(pData->httpheadervalue);
+	for(i = 0 ; i < pData->nHttpHeaders ; ++i) {
+		free((void*) pData->httpHeaders[i]);
+	}
+	free(pData->httpHeaders);
+	pData->nHttpHeaders = 0;
 	free(pData->pwd);
 	free(pData->authBuf);
 	free(pData->headerBuf);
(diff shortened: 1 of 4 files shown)


gintrator:
rsyslog does support an array data configuration type, which might lend itself more naturally to supporting this feature, instead of extending `httpheaderkey/httpheadervalue` with a delimiter.

Something like this..
```
action(
  type="omhttp"
  ...
  httpheaders=[
    "X-Insert-Key: key",
    "X-Event-Source: logs"
  ]
  ...
)
```

omkafka uses this to implement [confParam](https://www.rsyslog.com/doc/master/configuration/modules/omkafka.html#confparam), here's the source for that. https://github.com/rsyslog/rsyslog/blob/master/plugins/omkafka/omkafka.c#L1871-L1880.

Thoughts?

callmegar:
Will look into the code @gintrator using an array feels definitively better, could not find a good example on how to process that type of parameters.

callmegar:
Allows the inclusion of multiple http headers on the REST call, adds a new `httpheaders` array parameter containing the list of headers to add,  Example:

```
action(
  type="omhttp"
  ...
  httpheaders=[
    "X-Insert-Key: key",
    "X-Event-Source: logs"
  ]
  ...
)
```

contrib/omhttp/omhttp.c
```diff
@@ -351,6 +359,11 @@ CODESTARTdbgPrintInstInfo
 		(uchar*)"(not configured)" : pData->httpheaderkey);
 	dbgprintf("\thttpheadervalue='%s'\n", pData->httpheadervalue == NULL ?
 		(uchar*)"(not configured)" : pData->httpheadervalue);
+	dbgprintf("]\n");
```
gintrator: Extra `dbgprintf("]\n");` line here?

contrib/omhttp/omhttp.c
```diff
@@ -1772,6 +1816,14 @@ CODESTARTnewActInst
 			pData->httpheaderkey = (uchar*)es_str2cstr(pvals[i].val.d.estr, NULL);
 		} else if(!strcmp(actpblk.descr[i].name, "httpheadervalue")) {
 			pData->httpheadervalue = (uchar*)es_str2cstr(pvals[i].val.d.estr, NULL);
+		} else if(!strcmp(actpblk.descr[i].name, "httpheaders")) {
+			pData->nHttpHeaders = pvals[i].val.d.ar->nmemb;
+			CHKmalloc(pData->httpHeaders = malloc(sizeof(uchar *) * pvals[i].val.d.ar->nmemb ));
+			for(int j = 0 ; j <  pvals[i].val.d.ar->nmemb ; ++j) {
+				char *cstr = es_str2cstr(pvals[i].val.d.ar->arr[j], NULL);
+				CHKiRet(processHeaderParam(cstr, (const uchar **)&pData->httpHeaders[j]));
```
gintrator: Since you already own the memory to `cstr` here, is it necessary to do a `strdup` inside `processHeaderParam` and then free the string in the next line? Can this function just do validation?

contrib/omhttp/omhttp.c
callmegar: Ah, yes!

contrib/omhttp/omhttp.c
callmegar: Got it, doing some adjustments

gintrator:
Looks pretty good @callmegar ! Can you please open an associated PR in rsyslog-docs for this parameter? https://github.com/rsyslog/rsyslog-doc/blob/master/source/configuration/modules/omhttp.rst

We might also consider marking params `httpheaderkey` and `httpheadervalue` as deprecated in the docs as well, since `httpheaders=[ .. ]` is the general case. Additionally, `httpcontenttype` can be specified as `Content-Type: application/json` as well if a user wants, but both shouldn't be specified (might just be worth writing this down in the docs to avoid confusion).

callmegar:
Submitted PR #892 on rsyslog/rsyslog-doc

gintrator:
You linked to PR 829 on rsyslog/rsyslog, but I see your PR https://github.com/rsyslog/rsyslog-doc/pull/892, looks good.

rgerhards:
@callmegar sorry for the late response, I had unfortunately only limited time available due to the corona crisis. That said, I am ready to merge, but would appreciate if you  could update the commit text to contain a clear description (just like in the tracker here) what and why was made. I'd like to have this as a permanent record inside the git history.

callmegar:
Sure @rgerhards can you point me to an example. Would it be just updating the first comment here? or is it something different?

rgerhards:
@callmegar sry for the delay, still swamped. I guess this has all the info you need: https://rainer.gerhards.net/2019/03/howto-great-pull-request.html#descriptive_commit_message

PR #4228, opened 2020-03-24
stream: Fix for segfault on imfile read

jaankit:
- if cstrLen(pThis->prevMsgSegment) > maxMsgSize then len calculation
  become negative if cstrLen(thisLine) < cstrLen(pThis->prevMsgSegment)
  This causes illegal access to memory location and thus causing segfault.
- assigning len = 0 if cstrLen(pThis->prevMsgSegment) > maxMsgSize so that
  it access the correct memory location.


rgerhards:
Thx for the PR!

Issue #4233, opened 2020-03-30
omfile: invalid log message split when using $outchannel 


alorbach:
When using $outchannel to do logrotation based on size, syslog messages can be incorrectly split if the incoming batch of messages exceeds the default omfile IO Buffer size of 4096 bytes.

PR #4234, opened 2020-03-30
stream.c: Moved doSizeLimitProcessing check to strmWrite

alorbach:
The check was done in strmPhysWrite before which caused syslog
messages to split in the middle if the syslog message batch exceeded
the default IO Buffer size.

closes: https://github.com/rsyslog/rsyslog/issues/4233


rgerhards:
Travis is known issue which can be ignored in this context.

Issue #4245, opened 2020-04-14
Can't get ID for group with large number of users when using 'fileGroup' parameter


emilbart:
rsyslogd complains when in action we set to 'fileGroup' parameter the name of group with large number of users.

The problem is in function doGetGID where getgrnam_r is called. The buffer with not enough size is passed to getgrnam_r.

### Expected behavior
rsyslog should find ID for group with large number of users when using 'fileGroup' parameter in rainerscript configuration.
### Actual behavior
The rsyslog complains with the following messages:
parameter 'filegroup': error to obtaining group id for 'students': Numerical result out of range [v8.32.0]
error during parsing file /etc/rsyslog.d/test.conf, on or before line 1: parameter 'filegroup': ID for group students could not be found [v8.32.0 try http://www.rsyslog.
### Steps to reproduce the behavior
1. Create a group with large number of users. For example assign to group at least 200 user.
2. Create file /etc/rsyslog.d/test.conf with the following body:
\*.\* action(type="omfile" FileCreateMode="0644" fileOwner="student1" fileGroup="students" File="/tmp/logging")

### Environment
rsyslogd -v
rsyslogd 8.32.0, compiled with:
	PLATFORM:				x86_64-pc-linux-gnu
	PLATFORM (lsb_release -d):		
	FEATURE_REGEXP:				Yes
	GSSAPI Kerberos 5 support:		Yes
	FEATURE_DEBUG (debug build, slow code):	No
	32bit Atomic operations supported:	Yes
	64bit Atomic operations supported:	Yes
	memory allocator:			system default
	Runtime Instrumentation (slow code):	No
	uuid support:				Yes
	systemd support:			Yes
	Number of Bits in RainerScript integers: 64

See http://www.rsyslog.com for more information.

rgerhards:
Thanks for the report, and sorry it sat open for so long. rsyslog does not close old issues just because they are old; I checked the current code while working through the backlog.

This appears to be fixed in current `main`. Both group-name resolution paths now retry `getgrnam_r()` with a dynamically growing buffer when libc returns `ERANGE`:

- RainerScript/object parameters: `grammar/rainerscript.c::doGetGID()`
- legacy `$FileGroup`-style config lines: `runtime/cfsysline.c::doGetGID()`

That means a group entry with a large member list is no longer limited to the old fixed-size buffer that produced `Numerical result out of range` in the original report.

[...]

PR #4246, opened 2020-04-14
rainerscript: call getgrnam_r repeatedly to get all group members

emilbart:
This patch provides a fix for a problem reported in https://github.com/rsyslog/rsyslog/issues/4245


rgerhards:
Travis was a known unrelated issue which already is fixed. The fail can  be ignored here.

rgerhards:
Thx for the PR!

PR #4253, opened 2020-04-24
testbench tooling: update kafka components

rgerhards:
those used so far are outdated, bump to newest version


rgerhards:
Traivs is a false positive (and needs to be looked into separately).

PR #4255, opened 2020-04-25
omprog: fix bad fd errors in daemon mode

### LINKED ISSUE #4247: "bad file descriptor" errors using omprog with rsyslogd in daemon mode
### Expected behavior
rsyslogd starts in daemon mode without errors and passes log messages to my program via omprog plugin

### Actual behavior
When run in daemon mode, I see "bad file descriptor" errors and my program is not started successfully.

apr 14 23:07:38.968884 RSYSLO  omprog: error sending message to program: Bad file descriptor [v8.2002.0 try https://www.rsyslog.com/e/2119 ] 
apr 14 23:07:38.968978 RSYSLO  action 'action-1-omprog' suspended (module 'omprog'), retry 0. There should be messages before this one giving the reason for suspension. [v8.2002.0 try https://www.rsyslog.com/e/2007 ] 
apr 14 23:07:39.970158 RSYSLO  action 'action-1-omprog' resumed (module 'omprog') [v8.2002.0 try https://www.rsyslog.com/e/2359 ] 
apr 14 23:07:39.970283 RSYSLO  omprog: error sending message to program: Bad file descriptor [v8.2002.0 try https://www.rsyslog.com/e/2119 ] 
apr 14 23:07:39.970364 RSYSLO  action 'action-1-omprog' suspended (module 'omprog'), retry 0. There should be messages before this one giving the reason for suspension. [v8.2002.0 try https://www.rsyslog.com/e/2007 ] 

If I run rsyslogd in the foreground `rsyslogd -n`,  I don't get the errors and messages are passed to my program as expected.

These errors also go away if I set `forceSingleInstance="off"`

1 - configure omprog
module(load="omprog")
action(type="omprog"
           binary="/usr/bin/logoutput"
           forceSingleInstance="on"
           template="output")

2 - start rsyslogd as a daemon
`rsyslogd` on the command line, or
/etc/init.d/syslog start

### Environment
root@zcu102-zynqmp:~# rsyslogd -v
rsyslogd  8.2002.0 (aka 2020.02) compiled with:
	PLATFORM:				aarch64-oe-linux-gnu
	PLATFORM (lsb_release -d):		
	FEATURE_REGEXP:				Yes
	GSSAPI Kerberos 5 support:		No
	FEATURE_DEBUG (debug build, slow code):	No
	32bit Atomic operations supported:	Yes
	64bit Atomic operations supported:	No
	memory allocator:			system default
	Runtime Instrumentation (slow code):	No
	uuid support:				Yes
	systemd support:			No
	Config file:				/etc/rsyslog.conf
	PID file:				/var/run/rsyslogd.pid

[syslog_debug.log.txt](https://github.com/rsyslog/rsyslog/files/4479011/syslog_debug.log.txt)
[rsyslog.conf.txt](https://github.com/rsyslog/rsyslog/files/4479009/rsyslog.conf.txt)

jsiwrk:
When omprog was used with the 'forceSingleInstance=on' option, and/or
the 'output' setting, "bad file descriptor" errors occurred, which
prevented the external program to be executed and/or the program output
to be correctly captured. The bug could also manifest as "resource
temporarily unavailable" errors, or other errors related to the use of
invalid/reassigned file descriptors. These errors only happened when
rsyslog ran in daemon mode (i.e. they didn't happen if rsyslogd was
run with the '-n' option).

The cause of the bug was that omprog opened the pipe fds needed by
these flags during the configuration load phase (in the 'newActInst'
module entrypoint). This is a bad place since the fork of the daemon
occurs after this phase, and all fds are closed when the daemon process
is started (see 'initAll' in rsyslogd.c), hence invalidating the
previously opened fds.

To correct this, the single child process and the output capture thread
are now started later, when the first log message is received by the
first worker thread. (Note: the 'activateCnf' module entrypoint, despite
being invoked after the fork, cannot be used for this purpose, since it
is invoked per module, not per action instance.)

Currently no automated test exists for this use case since the testbench
always runs rsyslog in non-daemon mode.

Affected versions: v8.38 and later

Closes #4247

### DIFF
--- a/plugins/omprog/omprog.c
+++ b/plugins/omprog/omprog.c
@@ -67,7 +67,7 @@ DEF_OMOD_STATIC_DATA
 #define MAX_FD_TO_CLOSE 65535
 
 typedef struct childProcessCtx {
-	int bIsRunning;		/* is program currently running? 0-no, 1-yes */
+	int bIsRunning;		/* is the program running? (if 0, next fields are uninitialized) */
 	pid_t pid;			/* pid of currently running child process */
 	int fdPipeOut;		/* fd for sending messages to the program */
 	int fdPipeIn;		/* fd for receiving status messages from the program, or -1 */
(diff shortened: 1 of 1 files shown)


jsiwrk:
The CI failures seem unrelated.

I have not added any new test since the testbench seems not designed for testing rsyslog in daemon mode (?). Nevertheless, a potential future regression seems highly unlikely, since it would require moving the refactored code back to the `newActInst` endpoint (definitely not something you can easily do by mistake).

@rgerhards: It would be great if this PR could make it into the next release, since it is a rather critical bugfix for omprog.

rgerhards:
sorry for the silence, didn't manage to get earlier to it... The test fails look indeed like false positives. There is also something weired with Travis. I'll see that I get things through.

rgerhards:
Travis is a know issue which is irrelevant in regard to this PR.

PR #4258, opened 2020-04-27
Travis: something is wrong with Travis, this just makes it non-fail

rgerhards:
This is not a proper solution, but keeps rsyslog CI erroring out on
every run due to Travis issue. We didn't change anything, but somehow
Travis now runs only a subset of tests and fails on some others. This
needs to be investigated.


rgerhards:
Rest of CI runs was irrelevant, as this only changed Travis settings.

PR #4261, opened 2020-04-28
omprog: fix assert failed on HUP with output flag

jsiwrk:
If the 'output' setting of omprog was used and rsyslog received a HUP
signal just after starting (and before the omprog action received the
first log to process), an internal assertion could fail, causing
rsyslog to terminate. The failure message was "rsyslogd: omprog.c:660:
closeOutputFile: Assertion `pCtx->bIsRunning' failed."

The failure could also occur if rsyslog received a HUP signal during
the shutdown sequence.

This bug was introduced in v8.2004 by PR #4255.

Although a test already existed that checked the interaction of HUPs
with the 'output' setting, it didn't always fail in this particular case
due to timing conditions. The test has been improved to cover this case
more reliably.

### DIFF
--- a/plugins/omprog/omprog.c
+++ b/plugins/omprog/omprog.c
@@ -657,7 +657,6 @@ done:
 static void
 closeOutputFile(outputCaptureCtx_t *pCtx)
 {
-	assert(pCtx->bIsRunning);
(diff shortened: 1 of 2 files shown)


jsiwrk:
Well, I hoped v8.2004 would be free of (known) omprog bugs, but I've caught this one too late... Sorry @rgerhards.

Issue #4227, opened 2020-03-24
imtcp---server can not accept the first message，the second message is acceptable


wsp1991:
### Expected behavior
server can accept the first message
### Actual behavior
server can not accept the first message，The second message is acceptable
### Steps to reproduce the behavior
1、server： systemctl restart rsyslog
2、client：Print a log to the server（eg：logger -it local6 "111111111111111111117"）
3、Observe whether the server receives the message
### Environment
- rsyslog version: 
rsyslog-8.1907.0
- platform: 
- for configuration questions/issues, include rsyslog.conf and included config files

client：
$DefaultNetstreamDriverCAFile /root/certificate/CA.cert
$DefaultNetstreamDriverCertFile /root/certificate/client.cert
$DefaultNetstreamDriverKeyFile /root/certificate/client.key
#$ActionSendStreamDriverAuthMode x509/certvalid
#$ActionSendStreamDriverMode 1
if      ($programname == 'local6')\
then    action( type="omfwd" protocol="tcp" target="9.82.254.194" port="11514" StreamDriver="gtls" StreamDriverMode="1" StreamDriverAuthMode="x509/certvalid")
&       ~

server：
$ModLoad imtcp
$DefaultNetstreamDriver gtls
$DefaultNetstreamDriverCAFile /root/certificate/CA.cert
$DefaultNetstreamDriverCertFile /root/certificate/server.cert
$DefaultNetstreamDriverKeyFile /root/certificate/server.key
$InputTCPServerStreamDriverAuthMode x509/certvalid
$InputTCPServerKeepAlive on
$InputTCPServerStreamDriverMode 1
$InputTCPServerRun 11514
$template Remote, "/var/log/syslog/%fromhost-ip%/%$YEAR%-%$MONTH%-%$DAY%.log"
:fromhost-ip, !isequal, "127.0.0.1" ?Remote

wsp1991:
Failure Cause Analysis
In function CheckConnection,
……
   rc = recv(pThis->sock, msgbuf, 1, MSG_DONTWAIT | MSG_PEEK);
    if(rc == 0 && errno != EAGAIN) {
       dbgprintf("CheckConnection detected broken connection - closing it (rc %d, errno %d)\n", rc, errno);
      /* in this case, the remote peer had shut down the connection and we
       * need to close our side, too.
If the server restarted rsyslog, the connection with the client was lost. 
[...]

PR #4300, opened 2020-05-27
the return value judgment of recv is err in CheckConnecttion

wsp1991:

### DIFF
--- a/runtime/msg.c
+++ b/runtime/msg.c
@@ -2619,22 +2619,21 @@ MsgGetStructuredData(smsg_t * const pM, uchar **pBuf, rs_size_t *len)
 uchar * ATTR_NONNULL(1)
 getProgramName(smsg_t *const pM, const sbool bLockMutex)
 {
-	if(pM->iLenPROGNAME == -1) {
-		if(pM->iLenTAG == 0) {
-			uchar *pRes;
-			rs_size_t bufLen = -1;
-			getTAG(pM, &pRes, &bufLen, bLockMutex);
-		}
+	if(pM->iLenTAG == 0) {
+		uchar *pRes;
+		rs_size_t bufLen = -1;
+		getTAG(pM, &pRes, &bufLen, bLockMutex);
+	}
 
-		if(bLockMutex == LOCK_MUTEX) {
-			MsgLock(pM);
-			/* need to re-check, things may have change in between! */
-			if(pM->iLenPROGNAME == -1)
-				aquireProgramName(pM);
-			MsgUnlock(pM);
-		} else {
+	if(bLockMutex == LOCK_MUTEX) {
+		MsgLock(pM);
+		/* need to re-check, things may have change in between! */
+		if(pM->iLenPROGNAME == -1)
 			aquireProgramName(pM);
-		}
+		MsgUnlock(pM);
+	} else {
+		if(pM->iLenPROGNAME == -1)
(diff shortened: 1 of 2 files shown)


wsp1991:
close https://github.com/rsyslog/rsyslog/issues/4227

wsp1991:
> This change looks just plainly wrong. Can you describe why you made it and why it is correct?

the recv（）function RETURN VALUE
       These calls return the number of bytes received, or **-1 if an error occurred.  In the event of an error, errno is set to indicate the error.**

       **When a stream socket peer has performed an orderly shutdown, the return value will be 0 (the traditional "end-of-file" return).**

       Datagram sockets in various domains (e.g., the UNIX and Internet domains) permit zero-length datagrams.  When such a datagram is received, the return value is 0.

       The value 0 may also be returned if the requested number of bytes to receive from a stream socket was 0.

---> so when the return value is 0, the errno is not set. 
> This change looks just plainly wrong. Can you describe why you made it and why it is correct?

wsp1991:
> This change looks just plainly wrong. Can you describe why you made it and why it is correct?
recv() function RETURN VALUE
       These calls return the number of bytes received, or -1 if an error occurred.  In the event of an error, errno is set to indicate the error.

       When a stream socket peer has performed an orderly shutdown, the return value will be 0 (the traditional "end-of-file" return).

       Datagram sockets in various domains (e.g., the UNIX and Internet domains) permit zero-length datagrams.  When such a datagram is received, the return value is 0.

       The value 0 may also be returned if the requested number of bytes to receive from a stream socket was 0.
---> so when the return value is 0, the errno is not set. only the return value is -1,the errno is set.

wsp1991:
> This change looks just plainly wrong. Can you describe why you made it and why it is correct?


>recv()functon RETURN VALUE
       These calls return the number of bytes received, or -1 if an error occurred.  In the event of an error, errno is set to indicate the error.
       When a stream socket peer has performed an orderly shutdown, the return value will be 0 (the traditional "end-of-file" return).
       Datagram sockets in various domains (e.g., the UNIX and Internet domains) permit zero-length datagrams.  When such a datagram is received, the return value is 0.
       The value 0 may also be returned if the requested number of bytes to receive from a stream socket was 0.

---> so when the return value is 0, the errno is not set. only the return value is -1,the errno is set.

rgerhards:
Thx for the clarification on the errno, I had overlooked the part. Sorry for the noise.

I just now need to see how I handle this PR which actually addresses two things. I think we also need to fixup the commit descriptions so that it is clearer what the commit does and why. I will probably create two separate PRs out of this one with the updated commit text.

rgerhards:
@wsp1991 can you comment on why the second commit does what it does? In which situations does the segfault occur?

First commit is now handled by https://github.com/rsyslog/rsyslog/pull/4417

rgerhards:
@wsp1991 I assume you envision a race because iLenProgram name is not guarded by mutex. This could indeed be an issue if the update happens in an related thread. However, I think a better fix for that would be to protect the full sequence with mutex look.

I have replaced the second commit by https://github.com/rsyslog/rsyslog/pull/4418 -- feedback is appreciated.

rgerhards:
"merged" by other PRs mentioned in conversation.

@wsp1991 thx for the PR, as you can see it has been merged, albeit in an unusual way.

PR #4301, opened 2020-05-28
http input module

n2yen:
- uses http library to provide http input.
user would need to configure an 'endpoint' as input, along
with a ruleset, defining how the input should be routed in
rsyslog.

### DIFF
--- a/tests/imhttp-post-payload.sh
+++ b/tests/imhttp-post-payload.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+# This is part of the rsyslog testbench, licensed under ASL 2.0
+
+. ${srcdir:=.}/diag.sh init
+generate_conf
+IMHTTP_PORT="$(get_free_port)"
+add_conf '
+template(name="outfmt" type="string" string="%msg%\n")
+module(load="../contrib/imhttp/.libs/imhttp"
+       ports="'$IMHTTP_PORT'")
+input(type="imhttp" endpoint="/postrequest" ruleset="ruleset")
+ruleset(name="ruleset") {
+	action(type="omfile" file=`echo $RSYSLOG_OUT_LOG` template="outfmt")
+}
+'
+startup
+sleep 1
+NUMMESSAGES=50
+for (( i=1; i<=NUMMESSAGES; i++ ))
+do
+  curl -si -H Content-Type:application/json http://localhost:$IMHTTP_PORT/postrequest -d '[{"foo":"bar","bar":"foo"},{"one":"two","three":"four"}]' &
+  pids[${i}]=$!
+done
+
+# wait for all pids
+for pid in ${pids[*]};
+do
+  wait $pid
+done
+
+sleep 2
+wait_queueempty
+echo "doing shutdown"
+shutdown_when_empty
+wait_shutdown
+echo "file name: $RSYSLOG_OUT_LOG"
+content_count_check '[{"foo":"bar","bar":"foo"},{"one":"two","three":"four"}]' $NUMMESSAGES
+exit_test
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,7 +75,7 @@ matrix:
        services:
          - mysql
          - postgresql
-       env: RUN="run.sh",BUILD_FROM_TARBALL="YES", GROK="YES", CHECK="YES", CFLAGS="-g -O2", RS_TESTBENCH_VALGRIND_EXTRA_OPTS="--gen-suppressions=all", EXTRA_CONFIGURE="--disable-default-tests --disable-elasticsearch --disable-impstats --disable-imfile --disable-imptcp --disable-gnutls -disable-openssl --disable-relp --disable-pmsnare --disable-pmlastmsg"
+       env: RUN="run.sh",BUILD_FROM_TARBALL="YES", GROK="YES", IMHTTP="YES", CHECK="YES", CFLAGS="-g -O2", RS_TESTBENCH_VALGRIND_EXTRA_OPTS="--gen-suppressions=all", EXTRA_CONFIGURE="--disable-default-tests --disable-elasticsearch --disable-impstats --disable-imfile --disable-imptcp --disable-gnutls -disable-openssl --disable-relp --disable-pmsnare --disable-pmlastmsg"
 
 #     - os: linux
 #       compiler: "gcc"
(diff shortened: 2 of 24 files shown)


n2yen:
initial "working" implementation - welcome feedback, for improvements, thanks!

rgerhards:
Thx for this very interesting PR! I guess even if we receive no feedback, we should merge it - after all it's not bad if we get feedback after that.

please also add doc to https://github.com/rsyslog/rsyslog-doc/

rgerhards:
@n2yen I just tried to update the CI containers, but I cannot find any packages for civitweb at all. Can you help?

n2yen:
Hi @rgerhards ,
Yes that's true, i don't think there are any maintained packages available. I have been building/installing from source in my environment. Would something like this work: https://github.com/rsyslog/rsyslog/blob/05e7ae66da186e88c699a6b96dee91f253ba2d6c/tests/travis/install.sh#L48

rgerhards:
@n2yen yes, that works! :-) I create the containers, so I have non-packaged software under my control (at least if they don't have a large dependency tail, which usually is not the case). I'll give it a try this evening or tomorrow.

rgerhards:
@n2yen pls have a look at https://build.rsyslog.com/#/builders/136/builds/3170 -- this is the first container where I (finally) enabled imhttp and it shows a valid concern. Working on more environments.

rgerhards:
@n2yen this also looks valid to me, at least I think I do not see the parameters inside the module parameter block: https://build.rsyslog.com/#/builders/137/builds/5919/steps/4/logs/failed-tests_log

rgerhards:
I'll continue to work on additional containers after the fixes - makes it a bit easier for me to track what is missing :-)

n2yen:
@rgerhards it seems like this test failed because tests/testsuites/docroot/file.txt couldn't be located. The tests configures docroot=`testsuites/docroot`
and serves `file.txt` as a static page. Did i miss something else besides adding it here? https://github.com/rsyslog/rsyslog/blob/9e146f166c4725d4bbcd3e4f360c837252733db3/tests/Makefile.am#L2145
 https://build.rsyslog.com/#/builders/137/builds/5932/steps/4/logs/failed-tests_log

rgerhards:
I am pretty sure that your path names are not hardened for `make distcheck`. Unfortunately `make distcheck` sets up a different directory structure. Have a look at how e.g. the mmdblookup tests address their data files: https://github.com/rsyslog/rsyslog/blob/master/tests/mmdb.sh#L21

The key point is the use of `srcdir` variable.

rgerhards:
@n2yen could you re-push the PR so that we get a new CI run? That would be most helpful (`git commit --amend` without changes will do to force it). 

n2yen:
@rgerhards thanks for the heads up. will do - been bit swamped, will try to update the docs this weekend!

n2yen:
@rgerhards hard to be sure, but it looks like the targets with TSAN *might be failing due to this macro: https://github.com/civetweb/civetweb/blob/d637328500c3f4bf9451b63f5b81cfa32fa45603/src/civetweb.c#L3743-L3748
Are civetweb compile definitions in the container definitions? If so, wonder if it's worth enabling the macro: `REENTRANT_TIME` ?

n2yen:
hi @rgerhards,
Added a PR to docker images to help with the TSAN issues: https://github.com/rsyslog/rsyslog-docker/pull/39 tested it locally, and seems to help.

rgerhards:
> Added a PR to docker images to help with the TSAN issues: [rsyslog/rsyslog-docker#39](https://github.com/rsyslog/rsyslog-docker/pull/39) tested it locally, and seems to help.

sorry for the delay, I'll try to get to it either today or tomorrow .. again some unexpected tasks... :-(

rgerhards:
@n2yen I still see some failures after updating the container and civetweb - can you have a look? Any troubleshooting suggestion?

n2yen:
Hi @rgerhards ,
haven't had time to dig into this issue yet - i suspect this current issue may be an Ubuntu 18 issue as we no longer see errors from the sanitychecks... will need to dig into it a bit more, there is an error in the log `truncating vsnprintf buffer: [, ERR_free_stri]` which is coming from `civetweb` itself and `ERR_free_str` may be related to one of the crypt libraries when SSL feature is set. A few folks have also reported Ubuntu 18 quirks: https://github.com/civetweb/civetweb/issues/641

rgerhards:
@n2yen looks like you nailed it, except for one test...

n2yen:
i can reproduce the issue occasionally - appears to happen if the port is in use... i'm adding verbose to the curl command to confirm.

rgerhards:
> i can reproduce the issue occasionally - appears to happen if the port is in use... i'm adding verbose to the curl command to confirm.

oh wait, that sounds like a known issue for which (depending on the actual circumstances) a solution exists. How do you assign the port?

tests/imhttp-post-payload.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+# This is part of the rsyslog testbench, licensed under ASL 2.0
+
+. ${srcdir:=.}/diag.sh init
+generate_conf
+IMHTTP_PORT="$(get_free_port)"
```
rgerhards, line 6: that's the known issue - while this usually works OK, it is racy. There is no guarantee that the port will remain free.

A better solution, if possible, is:

* bind with port 0 
* make code ask OS which port it got
* persist that port to file
* make testbench pick port from there

The imtcp tests use that mode. But it depends on civitweb providing the necessary functionality.

n2yen: yes, I'll need to have a look at this. thanks for the tip!

rgerhards:
Not sure on the last pushes,,, Looking at the code, you could not use port 0. Anything you still work on? Asking because I wonder when we can merge.

n2yen:
yes, one thing changed which helped dramatically was moving the server initialization into `activateCnf`. There may have been a race condition with http server initialization and when the post request begin. 
There may still be a race condition with the os port, but looks like the above may have been the main source of the errors.

Passing the CI tests was the main thing I was working on, so I think its at a good merging point. I'm still working on docs though

rgerhards:
@n2yen moving to next schedule release milestone because of the docs, but ready to merge it early (which means it turns up in the daily stable release)

n2yen:
sure, thanks @rgerhards!  I'll be updating the docs this weekend.

rgerhards:
@n2yen did you have a chance to do the doc?

n2yen:
@rgerhards yes, will be finishing it up over the weekend!

rgerhards:
@n2yen thx again for all your work!

PR #4307, opened 2020-06-01
[WIP/RFC] ossl: initial OCSP support

gollub:
This implements certificate revocation support using OCSP (RFC 6960).
This does not introduce TLS OCSP stapling. CRL support is also not
included. If a certificate holds only a CRL distribution point reference
the revocation check will fail.

The OCSP revocation patch implements a "strict" revocation policy,
which means any OCSP error responses will let the verification fail.

Only the OCSP over HTTP transport protocol is implemented, not HTTPS.

Supported OCSP (RFC 6960) extensions:
 * Nonce

Signed-off-by: Daniel Gollub <dgollub@att.com>


gollub:
There might be more adjustments on default knobs or additional configuration knobs to control the OCSP client behavior.

One delicate item to decide on: should this be enabled by default in the NSD ossl?

gollub:
@rgerhards , @alorbach , is there any trick to make Codacy happy with respect to the "unknown" SO_BINDTODEVICE value?

I'm also puzzled about "Condition '!bio' is always false".

runtime/nsd_ossl.c
```diff
@@ -215,22 +219,595 @@ void osslLastSSLErrorMsg(int ret, SSL *ssl, int severity, const char* pszCallSou
...
+	bio = BIO_new_socket(fd, BIO_CLOSE);
+	if (bio == NULL) {
+		LogError(0, RS_RET_NO_ERRCODE, "OCSP: Could attach socket to BIO object\n");
+		goto err;
+	}
+err:
+
+	freeaddrinfo(rp);
+
+	if (!bio) /* don't close the socket in case of success */
```
alorbach: try to change it to 

if (bio == NULL)

to pass codacy.

alorbach: Long time no see Daniel ;)

> @rgerhards , @alorbach , is there any trick to make Codacy happy with respect to the "unknown" SO_BINDTODEVICE value?

I think this issue is a false positive and can be ignored. 
I have commented above on the other issue.

gollub: Indeed, pretty long time :)

Pushed a bio==NULL variant.

Let's see if that passes codacy.


Do you have any preference or suggestion on additional configuration knobs?

I guess a knob to turn of all certificate revocation checks off would probably be helpful, to mimic the previous behavior. Personally I would prefer if (online) revocation checks would be done by default, which might be expected by some TLS users.

I could contribute online CRL support as well, but that is rather complex, with delta-CRLs and such things. And not sure if online CRLs has really a future. Most or all public CAs do OCSP today, might be also a CAB Forum requirement (haven't checked). Letsencrypt does OCSP-only. So I wonder if online CRL is worth the burden.

gollub: No luck with codacy:

   `Condition 'bio==NULL' is always false`

alorbach: > No luck with codacy:
> 
> `Condition 'bio==NULL' is always false`

@rgerhards do you see the reason why this is detected as issue?

alorbach: > Indeed, pretty long time :)
> 
> Pushed a bio==NULL variant.
> 
> Let's see if that passes codacy.
> 
> Do you have any preference or suggestion on additional configuration knobs?

Yes indeed I would recommend to have configuration options in imtcp and omfwd enabling  / disabling OCSP checking with default OFF to maintain existing behavior. 
With a name like "StreamDriver.TlsVerifyOcsp" (See other TLS Options in imtcp/omfwd doc): 
https://www.rsyslog.com/doc/master/configuration/modules/imtcp.html
https://www.rsyslog.com/doc/master/configuration/modules/omfwd.html

You needs some additional coding to pass the configuration option from the module down to the nsd driver. And also "not implented" wrappers in nsd_gtls and nsd_ptcp drivers. Here is an older PR where I added "streamdriver.permitexpiredcerts" which you can use a template: 
https://github.com/rsyslog/rsyslog/pull/3366

> I could contribute online CRL support as well, but that is rather complex, with delta-CRLs and such things. And not sure if online CRLs has really a future. Most or all public CAs do OCSP today, might be also a CAB Forum requirement (haven't checked). Letsencrypt does OCSP-only. So I wonder if online CRL is worth the burden.

For now I would also recommend too leave it with OCSP for now. If CRL becomes more important in future, it can be discussed again.

Btw., it would be great if we could get a testcase for OCSP. Do you have an idea how to do make one that works with a free and reliable OCSP service?

gollub: > > Indeed, pretty long time :)
> > Pushed a bio==NULL variant.
> > Let's see if that passes codacy.
> > Do you have any preference or suggestion on additional configuration knobs?
> 
> Yes indeed I would recommend to have configuration options in imtcp and omfwd enabling / disabling OCSP checking with default OFF to maintain existing behavior.
> With a name like "StreamDriver.TlsVerifyOcsp" (See other TLS Options in imtcp/omfwd doc):
> https://www.rsyslog.com/doc/master/configuration/modules/imtcp.html
> https://www.rsyslog.com/doc/master/configuration/modules/omfwd.html
> 
> You needs some additional coding to pass the configuration option from the module down to the nsd driver. And also "not implented" wrappers in nsd_gtls and nsd_ptcp drivers. Here is an older PR where I added "streamdriver.permitexpiredcerts" which you can use a template:
> #3366


Ok, I'll look into introduce those knobs.

With respect to the default behavior: Wouldn't a user/operator expect that the TLS drivers perform such online revocation check by default, if the CA/PKI is including those? E.g. if someone hosts a TLS rsyslog server with a public CA vendor, the server gets compromised the  end-entity certificate of the server revoked by the CA. The OCSP server would know about the revoked certificate, the TLS rsyslog clients could then be "protected" talking to a compromised/revoked certificate. At least that would be my naive expectation that TLS client comply with: https://tools.ietf.org/html/rfc5280#section-6 /  https://tools.ietf.org/html/rfc5280#section-6.3
(which is referred by https://tools.ietf.org/html/rfc5425#section-4.2.1 )


I understand that this might break existing deployments under certain circumstances. But on the long run I would encourage to perform online revocation checks if OCSP (or CRL) URL are embedded in the  end-entity or in the certificate path:

```
$ openssl s_client -connect github.com:443 < /dev/null 2> /dev/null | openssl x509 -text -noout | grep -A5 -E '(Authority Information Access|CRL Distribution Points)'
            X509v3 CRL Distribution Points: 

                Full Name:
                  URI:http://crl3.digicert.com/sha2-ha-server-g6.crl

                Full Name:
--
            Authority Information Access: 
                OCSP - URI:http://ocsp.digicert.com
                CA Issuers - URI:http://cacerts.digicert.com/DigiCertSHA2HighAssuranceServerCA.crt

            X509v3 Basic Constraints: critical
                CA:FALSE
```

This would not affect existing deployment using private PKI environment without any CRL/OCSP URL in the issued certificate.

Not sure if "ignoring" OCSP/CRL URLs as part of certificate validation would be categorized as https://cwe.mitre.org/data/definitions/299.html

(Yes, the browsers don't do by default live online reovcation either IIUC, but they have some sort of regular caches updated transparently to the user. So there is at least some revocation checking going on)

On the long run OCSP stapling support would be obviously nice. The initial motivation for us implementing this was fulfilling some security requirements as syslog+TLS client, not as server. Not sure if there is on the market already a OCSP stapling aware syslog server. Maybe something worth to look into to add for rsyslog in future.

> 
> > I could contribute online CRL support as well, but that is rather complex, with delta-CRLs and such things. And not sure if online CRLs has really a future. Most or all public CAs do OCSP today, might be also a CAB Forum requirement (haven't checked). Letsencrypt does OCSP-only. So I wonder if online CRL is worth the burden.
> 
> For now I would also recommend too leave it with OCSP for now. If CRL becomes more important in future, it can be discussed again.
> 
> Btw., it would be great if we could get a testcase for OCSP. Do you have an idea how to do make one that works with a free and reliable OCSP service?

Sure, first you need to issue a set of certificates with a OCSP URI (Authority Information Access) included. To generate such testing purposes I build this: https://github.com/danos/vyatta-security-vpn/tree/master/docs/x509

I'll look into the existing rsyslog UTs if I can enhance/add something to existing ossl related UTs.

For development I used openssl as OCSP responder/server:
```
openssl ocsp -index index.txt -port 8888 -rsigner RSA_ocsp1.pem -rkey RSA_ocsp1.key -CA RSA_CA.pem -text -nmin 10
```

The index.txt is the file which openssl generates when you use OpenSSL to build a PKI. In this example the certificate for server13 got revoked. And `openssl ocsp -index indext.txt` basically uses that file as revocation database.

```
$ cat index.txt
V       191110011802Z           00      unknown /C=US/ST=CA/O=AT&T/OU=Labs/CN=client1/emailAddress=client1+rsa@vyatta.att-mail.com
V       191110011816Z           01      unknown /C=US/ST=CA/O=AT&T/OU=Labs/CN=serverA/emailAddress=serverA+rsa@vyatta.att-mail.com
R       191110011822Z   181110011831Z   02      unknown /C=US/ST=CA/O=AT&T/OU=Labs/CN=server13/emailAddress=server13+rsa@vyatta.att-mail.com
```

To make use of OCSP nonces add `-nonce`:

```
openssl ocsp -index index.txt -port 8888 -rsigner RSA_ocsp1.pem -rkey RSA_ocsp1.key -CA RSA_CA.pem -text -nmin 10 -nonce
```


I haven't checked other free/open implementations yet. FreeIPA/dogtag PKI might be worth a look. But for self-contained UTs `openssl ocsp` might be sufficient I guess.

The only other detailed behavior change between different production OCSP responder/server deployments, which are easy to overlook if you just use `openssl ocsp`: the HTTP requests need to hold a "Host" header entry. Most or many public CA vendor seem to use CDN providers to guarantee their OCSP availability and responsiveness.

alorbach: > With respect to the default behavior: Wouldn't a user/operator expect that the TLS drivers perform such online revocation check by default, if the CA/PKI is including those? E.g. if someone hosts a TLS rsyslog server with a public CA vendor, the server gets compromised the end-entity certificate of the server revoked by the CA. The OCSP server would know about the revoked certificate, the TLS rsyslog clients could then be "protected" talking to a compromised/revoked certificate. At least that would be my naive expectation that TLS client comply with: https://tools.ietf.org/html/rfc5280#section-6 / https://tools.ietf.org/html/rfc5280#section-6.3
> (which is referred by https://tools.ietf.org/html/rfc5425#section-4.2.1 )
> 
> I understand that this might break existing deployments under certain circumstances. But on the long run I would encourage to perform online revocation checks if OCSP (or CRL) URL are embedded in the end-entity or in the certificate path:

Thanks for the details, @rgerhards I think under these circumstances it is save and reasonable to have OCSP checking enabled by default with the possibility to disable it if needed.

> On the long run OCSP stapling support would be obviously nice. The initial motivation for us implementing this was fulfilling some security requirements as syslog+TLS client, not as server. Not sure if there is on the market already a OCSP stapling aware syslog server.

Probably not yet ;).

> Sure, first you need to issue a set of certificates with a OCSP URI (Authority Information Access) included. To generate such testing purposes I build this: https://github.com/danos/vyatta-security-vpn/tree/master/docs/x509
> 
> I'll look into the existing rsyslog UTs if I can enhance/add something to existing ossl related UTs.
> 
> For development I used openssl as OCSP responder/server:
> 
> ```
> openssl ocsp -index index.txt -port 8888 -rsigner RSA_ocsp1.pem -rkey RSA_ocsp1.key -CA RSA_CA.pem -text -nmin 10
> ```
> 
> The index.txt is the file which openssl generates when you use OpenSSL to build a PKI. In this example the certificate for server13 got revoked. And `openssl ocsp -index indext.txt` basically uses that file as revocation database.
> 
> ```
> $ cat index.txt
> V       191110011802Z           00      unknown /C=US/ST=CA/O=AT&T/OU=Labs/CN=client1/emailAddress=client1+rsa@vyatta.att-mail.com
> V       191110011816Z           01      unknown /C=US/ST=CA/O=AT&T/OU=Labs/CN=serverA/emailAddress=serverA+rsa@vyatta.att-mail.com
> R       191110011822Z   181110011831Z   02      unknown /C=US/ST=CA/O=AT&T/OU=Labs/CN=server13/emailAddress=server13+rsa@vyatta.att-mail.com
> ```
> 
> To make use of OCSP nonces add `-nonce`:
> 
> ```
> openssl ocsp -index index.txt -port 8888 -rsigner RSA_ocsp1.pem -rkey RSA_ocsp1.key -CA RSA_CA.pem -text -nmin 10 -nonce
> ```
That would be fully sufficient as ocsp server to build a testcase on. 
It just needs to be started in background and killed on shutdown. 
This needs some additional plumbing in diag.sh like done for this test here:
https://github.com/rsyslog/rsyslog/blob/master/tests/sndrcv_omsnmpv1_udp.sh
(See snmp_start_trapreceiver  and snmp_stop_trapreceiver )

If you create sample ca / certifcates based on your script, but make them last longer (Like 36500 days) which is ok for the testbench, add them to here for example (tests/testsuites/x.509), you can use these tests as template to build your own test on: 

https://github.com/rsyslog/rsyslog/blob/master/tests/sndrcv_tls_certvalid.sh
(Just replace '$RS_TLS_DRIVER' with ossl, as this will only run for openssl). 
And this tests: 
https://github.com/rsyslog/rsyslog/blob/master/tests/imtcp-tls-ossl-basic.sh

You can run single tests manually when you build rsyslog like this:
- go to rsyslog source dir
- export srcdir=./
- autoreconf -vfi 
- ./configure ... your options ...
- make && make check 
- ...wait until first tests runs then break with CTRL + C
- cd tests
- ./yourtestcase.sh

davidelang:
enabled by default? what happens if you are unable to do the validation? how 
much of a delay does this cause?

David Lang

rgerhards:
> @rgerhards , @alorbach , is there any trick to make Codacy happy with respect to the "unknown" SO_BINDTODEVICE value?

 adding some info: the root cause is that SO_BINDTODEVICE is not present Codacy's build environment. That can happen elsewhere as well. So the proper cure is to do an
```
#ifndef SO_BINDTODEVICE
```
and make to code react accordingly.

gollub:
> enabled by default? what happens if you are unable to do the validation?

If your PKI/CA is not using OCSP, there will be no online revocation validation at all. So in case you have a private PKI,  without OCSP URLs in the issued certificate chain, there will be no online revocation validation at all. Only few CPU cycles to check if there is a URL in the certificate as part of the certificate path validation

Only if there is an OCSP URL in the certificate and only if this URL is using http:// as transport an online revocation check will be done.

In case of a revocation "failure" (not a "cert revoked" response) there are two options:

1. strict revocation policy aka. hard fail: this is what this PR right now holds, hardcoded. The certificate validation will fail. 

2. relaxed revocation policy aka. soft fail: I'll introduce an update of the PR which allows this to be configurable, or even make this the (upstream) default. That relaxed revocation policy tolerates lookup OCSP "failures" and only fails if a OCSP response indicates "certificate revoked!".

The problem wit the relaxed/soft option: The reason is: OCSP error responses are not signed. Mallory, Marvin or Mallet could respond with unsigned bogus OCSP errors to hide the fact that the certificate got compromised and already revoked.

There is a nice article, which I highly recommend, which illustrates the problem quite nicely: https://arstechnica.com/information-technology/2017/07/https-certificate-revocation-is-broken-and-its-time-for-some-new-tools/

For the original case this patch got build there was demand for the strict/hard failure. For an upstream default usage I guess and agree that probably the best option would be: have OCSP checks enabled by default (so public PKI users get the revocation capabilities they pay their CA vendor for), but to with a relaxed/soft OCSP failure policy. To tolerate setup issues, if the PKI infrastructure is not reachable or mis-configured.

> how much of a delay does this cause?

If your PKI is not using OCSP there is pretty much no delay. If you build a private PKI and you don't want a revocation lookup delay, either don't put OCSP URL in the issue certificate or disable it in the ossl driver.

That depends on your PKI infrastructure, how long takes the HTTP request to the OCSP infrastructure. OCSP response lifetime can be also cached. The CAB Forum allows approx. 7 or 10 days of OCSP/CRL lifetime before renewal. OCSP/CRL clients  usually use that lifetime, controlled by the PKI, to cache the response, so they don't perform the lookup every time. For OCSP clients it's not unusual to cache that only in memory, so the cache is only valid for the lifetime of the process lifetime.

If public CAs are used: those CAs usually advertise how quick and reliable their OCSP/CRL infrastructure is, most of the time they utilize established CDN networks.

gollub:
> > @rgerhards , @alorbach , is there any trick to make Codacy happy with respect to the "unknown" SO_BINDTODEVICE value?
> 
> adding some info: the root cause is that SO_BINDTODEVICE is not present Codacy's build environment. That can happen elsewhere as well. So the proper cure is to do an
> 
> ```
> #ifndef SO_BINDTODEVICE
> ```
> 
> and make to code react accordingly.
> 
> EDIT:
> Sometimes it is as simple as doing
> 
> ```
> #ifndef SO_BINDTODEVICE
> #define SO_BINDTODEVICE 0
> #endif
> ```
> 
> but I am not sure this is a proper cure here.

I thought about that as well, but I wonder why the SO_BINDTODEVICE is working for runtime/nsd_ptcp.c?

Is Codacy only check diffs/commits? Or the full tree?

davidelang:
On Wed, 3 Jun 2020, Daniel Gollub wrote:

>> how much of a delay does this cause?
>
> If your PKI is not using OCSP there is pretty much no delay. If you build a private PKI and you don't want a revocation lookup delay, either don't put OCSP URL in the issue certificate or disable it in the ossl driver.
>
> That depends on your PKI infrastructure, how long takes the HTTP request to the OCSP infrastructure. OCSP response lifetime can be also cached. The CAB Forum allows approx. 7 or 10 days of OCSP/CRL lifetime before renewal. OCSP/CRL clients  usually use that lifetime, controlled by the PKI, to cache the response, so they don't perform the lookup every time. For OCSP clients it's not unusual to cache that only in memory, so the cache is only valid for the lifetime of the process lifetime.
>
> If public CAs are used: those CAs usually advertise how quick and reliable their OCSP/CRL infrastructure is, most of the time they utilize established CDN networks.

what about when you use a public CA, but your syslog server doesn't have access 
to the Internet?

what if the only access to the Internet is through a proxy, how would you 
configure that?

David Lang

gollub:
> what about when you use a public CA, but your syslog server doesn't have access to the Internet? 

In that case it's questionable why a public CA is used? How do you handle compromised end entity certificates?

If you are not running a public HTTPS server and webshop, one reason you should pay a public CA vendor is the service that the ensure that they  will revoke any compromised certificate within 24 hours or so.

Overall, with OCSP enabled by default and soft/relaxed policy by default, this will not make the certification path validation fail ever.

> what if the only access to the Internet is through a proxy, how would you configure that? 

Yes, I have not looked in proxy support yet. If there is demand, I'm sure someone is willing to implement that. The OpenSSL library is used as HTTP client implementation, might not be the best pick to add proxy support.

But the revocation lookup can be bind to certain source interface, in case of usage of a management VRF, if the default VRF has no internet cases, which is pretty common.

It's not going to break existing proxy setups. The OCSP capabilities for proxy setups will be the same as before.

gollub:
David, with the CT being common/mandatory for a public CAs, I wonder how common it would be to setup a syslog server with a public CA issued certificate.

OTOH:
https://crt.sh/?q=syslog%25
https://crt.sh/?q=rsyslog%25

... but I guess CT is still pretty unknown to some operators.

alorbach:
> > but I am not sure this is a proper cure here.
> 
> I thought about that as well, but I wonder why the SO_BINDTODEVICE is working for runtime/nsd_ptcp.c?

nsd_ptcp.c actually used this: 
#		if defined(SO_BINDTODEVICE)
		if(setsockopt(pThis->sock, SOL_SOCKET, SO_BINDTODEVICE, device, strlen(device) + 1) < 0)
#		endif

I didn't find that before either for some reason.

rgerhards:
> I tried the trick @rgerhards suggested:
> 
> ```
> #ifndef SO_BINDTODEVICE
> #define SO_BINDTODEVICE 0
> #endif
> ```
> 
> ... and that worked.

did you evaluate that it actually *works* vs. "does not bring up analyzer errors"? We do not want to trick the analyzer. The question is if the functionality can be provided if this functionality is not present. If it "just" causes some minor degradation, that's probably fine. If it breaks things, we need to err our and tell the user that the platform does not offer the necessary plumbing and so the feature cannot be used. I don't know how hard this hits here,...

gollub:
We need to error out, if SO_BINDTODEVICE is not available but a specific source-interface/device configured we should fail rather using a "random" other interface/device.

That would be also consistent with nsd_ptcp.c behavior.

With that "trick" it will not have the correct behavior:

```
#ifndef SO_BINDTODEVICE
#define SO_BINDTODEVICE 0
#endif
[...]
        if (device) {
#if defined(SO_BINDTODEVICE)
                if(setsockopt(fd, SOL_SOCKET, SO_BINDTODEVICE, device,
                              strlen(device) + 1) < 0)
#endif
                {
                        LogError(0, RS_RET_NO_ERRCODE, "Failed to bind to interface \"%s\" for "
                                        "OCSP request.\n", device);
                        goto err;
                }
        }
```

At least I guess #if defined(0) will be a true expression.

IMHO ```if defined(X)``` should be accepted by Codacy, even if X is not defined.

As it is the case for nsd_ptcp.c :

```
        if(device) {
#               if defined(SO_BINDTODEVICE)
                if(setsockopt(pThis->sock, SOL_SOCKET, SO_BINDTODEVICE, device, strlen(device) + 1) < 0)
#               endif
                {
                        dbgprintf("setsockopt(SO_BINDTODEVICE) failed\n");
                        ABORT_FINALIZE(RS_RET_IO_ERROR);
                }
        }
```

rgerhards:
late, but...
> Is Codacy only checking diffs/commits? Or the full tree?

both- the check is done on full tree, but than only diagnostics newly introduced are shown.

rgerhards:
> With that "trick" it will not have the correct behavior:

ok, so this is no cure

> At least I guess #if defined(0) will be a true expression.



> At least I guess #if defined(0) will be a true expression.

I would guess it is wrong, as `0` is not a preprocessor macro and as such no such macro is defined.

To the root problem: I don't see at the moment why cppcheck (which is run by Coday) diagnoses this. It's not as complex as e.g. this case: https://github.com/danmar/simplecpp/issues/73

@gollub Could you replace
```
#               if defined(SO_BINDTODEVICE)
                if(setsockopt(pThis->sock, SOL_SOCKET, SO_BINDTODEVICE, device, strlen(device) + 1) < 0)
#               endif
```
by
```
#               ifdef SO_BINDTODEVICE
                if(setsockopt(pThis->sock, SOL_SOCKET, SO_BINDTODEVICE, device, strlen(device) + 1) < 0)
#               endif
```
and see what happens?

gollub:
Ah ok, great. That might explain via nsd_ptcp.c probably passed that check. I'll try to come up with something to trick Codacy and keep the intended functionality.

gollub:
Pushed:

```
@@ -57,10 +57,6 @@
 /* things to move to some better place/functionality - TODO */
 // #define CRLFILE "crl.pem"
 
-#ifndef SO_BINDTODEVICE
-#define SO_BINDTODEVICE 0
-#endif
-
 #define  OCSP_TIMEOUT 5
 
 MODULE_TYPE_LIB
@@ -449,7 +445,7 @@ static BIO* ocsp_connect(const char *host, const char *port, const char *device)
        }
 
        if (device) {
-#if defined(SO_BINDTODEVICE)
+#ifdef SO_BINDTODEVICE
                if(setsockopt(fd, SOL_SOCKET, SO_BINDTODEVICE, device,
                              strlen(device) + 1) < 0)
 #endif
```

rgerhards:
> I'll try to come up with something to trick Codacy and keep the intended functionality.

Let's better try to get to the root cause - I can always merge even when codacy fails. But my experience is that those "dumb analyzers with their false positives" are often right, even if it does not look so ;-) Doesn't look so in the `#if defined()` case, but I am also concerend about the other one.

Maybe it's really a bug in cppcheck. It might be worth a try to remove the last (and redundant) `goto err`  more or less in front of the `err:` label. I don't say it's wrong, it's perfectly OK. But it could confuse cppcheck. Especially in cases like this second one trying hard to understand the root cause showed me more often than not a real bug.

gollub:
I'm fine with finding the root-cause.

The ```#ifdef``` attempt failed:
https://app.codacy.com/manual/rsyslog/rsyslog/pullRequest?prid=5619292

I'll push a dummy change in nsd_ptcp.c to see if that also barfs.

And try your advice for the err: label.

rgerhards:
> I'll push a dummy change in nsd_ptcp.c to see if that also barfs.

The problem is that you need to trigger a new diagnostic, what will possibly not happen.

You can also try cppcheck locally, e.g.
```
cppcheck --force runtime/nsd_ptcp.c
```
I'll see if I can find the exact calling parameters Codacy uses.

rgerhards:
@gollub Codacy shows the diagnostic is also present in nsd_ptcp:

https://app.codacy.com/manual/rsyslog/rsyslog/file/45262348346/issues/source?bid=8903817&fileBranchId=8903817

If you cannot access this page, send me a mail.

gollub:
Thanks that link worked.

I think I found the cppcheck flag to trigger this:

```
dgollub@vyatta-dev:~/projects/rsyslog$ dpkg -l cppcheck
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name           Version      Architecture Description
+++-==============-============-============-=========================================
ii  cppcheck       1.86-1       amd64        tool for static C/C++ code analysis (CLI)
dgollub@vyatta-dev:~/projects/rsyslog$ cppcheck --version
Cppcheck 1.86
dgollub@vyatta-dev:~/projects/rsyslog$ cppcheck --enable=all --force runtime/nsd_ossl.c
Checking runtime/nsd_ossl.c ...
[runtime/nsd_ossl.c:201]: (style) The scope of the variable 'iSSLErr' can be reduced.
[runtime/nsd_ossl.c:1617]: (style) The scope of the variable 'ret' can be reduced.
[runtime/nsd_ossl.c:1618]: (style) The scope of the variable 'err' can be reduced.
Checking runtime/nsd_ossl.c: DEBUG...
Checking runtime/nsd_ossl.c: DEBUGLESS...
Checking runtime/nsd_ossl.c: ENABLE_LIBLOGGING_STDLOG...
Checking runtime/nsd_ossl.c: HAVE_ATOMIC_BUILTINS...
Checking runtime/nsd_ossl.c: HAVE_ATOMIC_BUILTINS64...
Checking runtime/nsd_ossl.c: HAVE_BUILTIN_EXCEPT...
Checking runtime/nsd_ossl.c: HAVE_LSEEK64;__FreeBSD__...
Checking runtime/nsd_ossl.c: HAVE_PTHREAD_SETSCHEDPARAM...
Checking runtime/nsd_ossl.c: HAVE_STRUCT_SOCKADDR_SA_LEN...
Checking runtime/nsd_ossl.c: LIBRESSL_VERSION_NUMBER...
Checking runtime/nsd_ossl.c: LOG_CRON...
Checking runtime/nsd_ossl.c: NDEBUG...
Checking runtime/nsd_ossl.c: OS_BSD...
Checking runtime/nsd_ossl.c: OS_SOLARIS...
Checking runtime/nsd_ossl.c: O_CLOEXEC;_AIX...
[runtime/nsd_ossl.c:449]: (information) Skipping configuration 'SO_BINDTODEVICE' since the value of 'SO_BINDTODEVICE' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
Checking runtime/nsd_ossl.c: SO_BINDTODEVICE...
Checking runtime/nsd_ossl.c: _AIX...
Checking runtime/nsd_ossl.c: __GNUC__...
Checking runtime/nsd_ossl.c: __hpux...
[runtime/nsd_ossl.c:2647]: (style) The function 'modInit' is never used.
[runtime/nsd_ossl.c:173]: (style) The function 'opensslh_THREAD_cleanup' is never used.
(information) Cppcheck cannot find all the include files (use --check-config for details)
dgollub@vyatta-dev:~/projects/rsyslog$ 
```

https://github.com/codacy/codacy-cppcheck/blob/master/src/main/scala/codacy/cppcheck/CPPCheck.scala#L58-L72

gollub:
Btw. removing the redundant "goto err" made codacy happy w.t.r. the 2nd issue.

rgerhards:
I think `cppcheck-suppress` is (in this exceptional case) a good thing for the `err:` case (maybe with a bug report for them at https://trac.cppcheck.net/). Search in doc for suppression functionality:

http://cppcheck.sourceforge.net/manual.pdf

rgerhards:
aha! cppcheck natively reports:
```
[runtime/nsd_ptcp.c:890]: (information) Skipping configuration 'SO_BINDTODEVICE' since the value of 'SO_BINDTODEVICE' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
```
Look closely - it's an information. So Codacy IMHO is overdoing a bit in reporting this as a diagnostic. Not sure why the standard definition is not used - it should be in the system header. Maybe still a cppcheck bug?

Trying
```
cppcheck --force --enable=all -D SO_BINDTODEVICE 0 runtime/nsd_ptcp.c
````
brings up more, similiar, informational messages.

**Comment by rgerhards** [comment]
This "skipping..." informational message sounds strange to me. I have turned it off once and forever in codacy cppcheck config. I don't it provides any value at all. Please correct me if I am wrong.

**Comment by gollub** [comment]
Reading http://cppcheck.sourceforge.net/manual.pdf chapter 5.1 gives the impression that this informational message just saying it's skipping code check for in this particular situation:

```
                if(setsockopt(fd, SOL_SOCKET, SO_BINDTODEVICE, device,
                              strlen(device) + 1) < 0)
```

... because cppcheck needs some hint which value it should test for SO_BINDTODEVICE.

The information message might be helpful to catch other stuff, or highlight which code gets actually skipped by cppcheck. And then decide on case-by-case basis which checks to skip.

Codacy seems to run with ```--inline-suppr``` which allows fine-grain suppression inline. I just pushed a new set which suppresses the ID "ConfigurationNotChecked" for this case. I could apply that also for nsd_ptcp.c via separate PR, in case you want to go forward with informational cppchecks.

**Comment by gollub** [comment]
> Not sure why the standard definition is not used - it should be in the system header. Maybe still a cppcheck bug?

http://cppcheck.sourceforge.net/manual.pdf , Chapter 5.2:

```
[...]

[...] Actually, it is recommended to not give all
include paths. While it is useful for cppcheck to see the declaration of a class
when checking the implementation of its members, passing standard library
headers is highly discouraged because it will result in worse results and longer
checking time. 
```

So that sounds like it is done by design.

But option obviously would be to call cppcheck twice with ```-DSO_BINDTODEVICE=25``` and ```-USO_BINDTODEVICE``` to test both permutation. But on a quick glance I could not figure out how to get both permutations tested with one cppcheck call, so there is no need to call cppcheck actually twice via codacy.

Overall the the ```SO_BINDTODEVICE``` usage in nsd_ptcp and nsd_ossl is exactly one line of code which gets skipped by cppcheck. I think that's acceptable and needs to carefully manually reviewed when an inline cppcheck-suppression gets added.

I would not encourage to do this for larger code blocks.

**Comment by rgerhards** [comment]
> I would not encourage to do this for larger code blocks.

looking at the big picture, I think completely disabling this informational message poses no real risk. Number one, as you say, we cannot easily do the permutations within the constraints of Codacy. Granted, we could run cppcheck ourselves, but that's quite a bit of effort. Also and most importantly, cppcheck (via Codacy) is just one of three static analyzers we run routinely:

* cppcheck
* clang static analyzer
* LGTM (pretty powerful)

When it works (far from always) we also use coverity.

We also do quite intensive dynamic testing on multiple platforms during CI - including involving dynamic sanitizers (memory, thread, undefined behavior) [side-note: that causes some flakes which we continuously fight].

So all in all, I think turning off this informational message is not a real loss - now that we know exactly why and how it happens. Providing always the correct potential values is also a cumbersome duty.

Any concerns, pls let me know. Just giving my honest reasoning for evaluation by others. Working together and discussion helps prevent disasters :-)

**Comment by gollub** [comment]
I'm fine with that approach.


Do you want me to remove the inline cppcheck suppression of the SO_BINDTODEVICE / ConfigurationNotChecked case?

**Comment by gollub** [comment]
David, I understand that this is a delicate decision. I pointed out earlier that the decision the "default" OCSP behavior will be delicate. Especially because OCSP certificates by public CAs is already the norm.

There are two options:

 - security by default via behavior change -> potentially breaks some, not all, existing deployments
 - not breaking existing behavior -> might cause overlook of the fact that OCSP checks are not performed by default, which might not meet thee expectation of existing/new users/customers


And there are the RFCs:

syslog & TLS: https://tools.ietf.org/html/rfc5425#section-4.2.1

```
                                            [....] Certification path validation
      is performed as defined in [RFC5280].  This method is useful where
      there is a Public Key Infrastructure (PKI) deployment.
```


https://tools.ietf.org/html/rfc5280#section-6
https://tools.ietf.org/html/rfc5280#section-6.3

```
Conforming implementations that support CRLs
   are not required to implement this algorithm, but they MUST be
   functionally equivalent to the external behavior resulting from this
   procedure when processing CRLs that are issued in conformance with
   this profile. 
```
(I just assume here that OCSP is the natural successor of CRL. E.g. see letsencrypt.org which is doing OCSP only, no CRL)


If you have a farm syslog clients and servers and use certificates-based authentication, you must be prepared to revoke individual end-entity certificates. Many PKI setups will use OCSP (today, and in future) and some will still use CRL-only (but hopefully that comes to an end soon). And their expectation will be that revoked certificates will get rejected as part of the certificate path validation.

That's an essential PKI functionality and IMHO that justifies potential user-behavior change/breakage which should be release-noted in such a fashion it's hard to miss.

In an ideal world this OCSP, OCSP-Stapling and CRL support would have been added with the first release of the rsyslog TLS implementation. But that ship sailed already.

I would be fine if initially this functionality would be for some period disabled by default and later with some heads-up enabled by default (with soft/relaxed revocation-policy). E.g. by adding some informational logging on start-up if TLS is configured that OCSP is currently disabled, but will be disabled in future releases or so and it's encouraged to consider enabling it already today manually.

**Comment by alorbach** [comment]
Perhaps we can find a compromise with having OCSP enabled by default but do not break existing configurations with environments not connected to the internet using public certificates or issues with invalid OCSP URLs. 

How about adding an additional configuration option called "StreamDriver.TlsVerifyOcspFailOnTimeout" for example with default 0. 
0 = If timeout occurs while trying to reach OCSP URL, the check will be considered as success.
1 = If timeout occurs while trying to reach OCSP URL, the check will be considered as fail.

So in worst case, we would have a 5 seconds delay when a new TLS Connection is being established. 

Would do you think @rgerhards  ?

**Comment by rgerhards** [comment]
default: pls be a bit patient, I have an idea, but I need to think a little bit about it. Trying to avoid emitting dumb things ;-)

**Comment by gollub** [comment]
> How about adding an additional configuration option called "StreamDriver.TlsVerifyOcspFailOnTimeout" for example with default 0.
> 0 = If timeout occurs while trying to reach OCSP URL, the check will be considered as success.
> 1 = If timeout occurs while trying to reach OCSP URL, the check will be considered as fail.
> 


To me that sounds like "soft-failure"/relaxed revocation policy anyway.


My proposal was also to have "soft-failure" by default. That means following OCSP failure gets tolerated and will not fail the certificate path validation:

1. OCSP request timeout
2. DNS lookup timeout/failure
3. any HTTP failure
4. any OCSP error response (other then "the cert is revoked! Don't trust it!")
...

The only thing which would make in a "soft-failure" configuration make the validation fail would be a revoked certificate.

```
StreamDriver.TlsVerifyRevocationPolicy = soft (default)
StreamDriver.TlsVerifyRevocationPolicy = strict

StreamDriver.TlsVerifyRevocationOCSP = enable (default)
StreamDriver.TlsVerifyRevocationOCSP = disable
```

**Comment by gollub** [comment]
As non-Web-browser reference of an protocol which performs certificate-based authentication: IKE/IPsec: strongsawn/charon: https://wiki.strongswan.org/projects/strongswan/wiki/Swanctlconf#connectionsltconngtremoteltsuffixgt-section

```

onnections.<conn>.remote<suffix>.revocation | relaxed
-- | --
Certificate revocation policy for CRL or OCSP revocation.A strict revocation policy fails if no revocation information is available, i.e. the certificate is not known to be unrevoked.ifuri fails only if a CRL/OCSP URI is available, but certificate revocation checking fails, i.e. there should be revocation information available, but it could not be obtained.The default revocation policy relaxed fails only if a certificate is revoked, i.e. it is explicitly known that it is bad.
```

Their default is also soft/relaxed.

Keep in mind, certificates which don't have any CRL/OCSP URL embedded, will not see any behavior change. E.g. (some) private PKIs not using CRL/OCSP at all or self-signed end-entity certificates.

**Comment by rgerhards** [comment]
Note: my post here **only is for defaults** and this in respect for prior versions.

We have a very strong policy that an old rsyslog.conf must work, as expected, with newer versions of rsyslog. Only for extremely strong reasons we break this. I think we did do this only two or three times in the past 10+ years.

The idea behind this is that there never should exist a fear to try a newer version because it does no longer work as intended. I know that the current trend is to ignore backwards compatibility. I think this is a bad and damaging trend.

As such, I stay to our policy and say all new features affecting existing behavior must require an explicit opt-in.

So far, so good. The bad thing obviously is that the casual user who does not want to dig into the reference doc as such never gets a new, desirable, behavior. For this PR, I agree that from a security perspective the new behavior is desirable. From a backwards compatibility PoV it is not.

This is a general problem we face every now and then, so it was time to think against about it. What I can come up with is an indication of which version of rsyslog the config file was written for, via e.g. a new config construct, let's say
```
config(compatibility="8.2006")
```
That would need to be the first statement inside the config or rsyslog errors out.

What would it mean?
If given above, it would tell rsyslog and plugins that behavior of 8.2006 is expected. In the case at question, this would mean the new behavior should be the default. If the statement is absent or specifies an older version, than the former behavior should be used as default.

I think this is a solution to address the compatibility issue. But it also has it's quirks: the casual user may never set a compatibility level, and so we are still stuck with the same default.

Nevertheless, I tend to say it is the best solution we can currently provide.

What do you think? @gollub @davidelang @alorbach @Whissi @mbiebl @jvymazal

**Comment by gollub** [comment]
That proposal works for me.

**Comment by alorbach** [comment]
Sounds like a good compromise to me. 
Perhaps a warning / friendly reminder could be logged on startup if the compatibility level is not set / configured.

**Comment by jvymazal** [comment]
From a distro maintenance standpoint this sounds reasonable, the top level config file is usually under maintainer control so compatibility can be set a appropriate doc updated to reflect that with any potential side-effects.

**Comment by davidelang** [comment]
I don't think it will help, and is likely to confuse matters.

As you say, casual users won't use it, and I see a LOT of configs that have been 
carried forward for a _very_ long time that will never see it.

So I think it adds complexity without really solving the problem.

I don't think that the problem is nearly as significant as it's made out to be 
in the first place. Very few syslog servers are exposed to the Internet, and in 
my experience, very few syslog servers that handle sensitive data are able to 
reach the Internet.

When you have websites maintained by one entity that are accessed by lots of 
people who have no relationship with the entity that manages the website, haing 
a way to revoke certs in the unlikely event that someone steals a cert (and it's 
key) from the website, and sets up both a fake website and fake DNS to point 
people to the website makes some amount of sense (but note that sophistication 
and multiple layers of the attack, and if they can manipulate DNS to point you 
at a bad site, can't they also redirect your CRL check?)

In the case of syslog, in almost all cases, the same entity is managing both 
sides of the connection, and the only people you would defend against with a CRL 
are internal bad actors who can also defeat your CRL check.

So I see this as checking boxes for theoretical security more than providing any 
actual security. There ARE cases where it is useful, but I don't see them as the 
common case.

So I would either disable the check by default, or make it a soft-fail check at 
most (what is the timeout on a check? how painful is that timeout going to be)

having an update of rsyslog suddently start generating requests to the Internet 
is going to set of security concerns (and bad press as they publish how rsyslog 
has become spyware, without ever researching what it's doing and why), which 
would argue to strict opt-in

David Lang

**Comment by davidelang** [comment]
we just phased out the -c5 etc compatibility flags, do we really want to add 
something like that back?

I think a better thing to do would be to have someone who really cares about 
certificate and CRL support to write up a help page about how to configure 
things (for example, how to ensure TLS 1.2+ support, probably more significant a 
risk than CRL checking)

If the how-to examples all enable CRL checking, then the casual user is going to 
have CRL checking enabled due to the power of cut-n-paste (distros can also 
include an example of how to enable certs that would include the CRL checking if 
they want)

**Comment by gollub** [comment]
> [...] I don't think that the problem is nearly as significant as it's made out to be in the first place. [..] There ARE cases where it is useful, but I don't see them as the common case. 


https://tools.ietf.org/html/rfc5280#section-3.3

```
   When a certificate is issued, it is expected to be in use for its
   entire validity period.  However, various circumstances may cause a
   certificate to become invalid prior to the expiration of the validity
   period.  Such circumstances include change of name, change of
   association between subject and CA (e.g., an employee terminates
   employment with an organization), and compromise or suspected
   compromise of the corresponding private key.  Under such
   circumstances, the CA needs to revoke the certificate.
```

Compromise is just one case. The critical case in context of syslog & TLS is certificate-based authentication and **authorization**. See https://tools.ietf.org/html/rfc5425#section-5

```
   Different environments have different security requirements and
   therefore would deploy different security policies.  This section
   discusses some of the security policies that may be implemented by
   syslog transport receivers and syslog transport senders.  The
   security policies describe the requirements for authentication and
   authorization.  The list of policies in this section is not
   exhaustive and other policies MAY be implemented.

   If the peer does not meet the requirements of the security policy,
   the TLS handshake MUST be aborted with an appropriate TLS alert.
[...]
5.2.  Subject Name Authorization

   Implementations MUST support certification path validation [RFC5280].
   In addition, they MUST support specifying the authorized peers using
   locally configured host names and matching the name against the
   certificate as follows.
```


In other words: if you do certificate-based authentication **and authorization** your certificate revocation strategy is not only required to block compromised certificates, but also syslog peers/clients which are no longer allowed to access your central syslog infrastructure.

RFC 5425 is not just about providing confidentially of the syslog traffic.

And that's why I think the comparison to "simple" TLS HTTP server usecase is not a good/fitting comparison. Unless you also take HTTPS client authentication **and authorization** into account as well.


> can't they also redirect your CRL check?

Yes, same is true for OCSP URIs. For that reason OCSP status responses are signed and CRL files are signed by the CA or a delegated CRL issuer. Not checking the signature I would consider as a security vulnerability and violating the PKI/CRL and OCSP RFCs IMHO.
 


> [...] start generating requests to the Internet is going to set of security concerns (and bad press as they publish how rsyslog has become spyware, without ever researching what it's doing and why) [...]

Yes, online revocation checks done out-of-band as proposed by this PR for OCSP or in case of CRL  are posing privacy concerns. OCSP-stapling would avoid this out-of-band look up from the client/initiator side and I think that will be on the long run the future. But for interoperability reasons with non OCSP-stapling aware syslog implementations (including rsyslog) out-of-band online revocation checks are mandatory.


> I think a better thing to do would be to have someone who really cares about
> certificate and CRL support to write up a help page about how to configure
> things (for example, how to ensure TLS 1.2+ support, probably more significant a
> risk than CRL checking)


Your arguments were all referring to "talking to the internet", so I assume for now you were always talking about using some public CA vendor which has to comply with CA/B forum guidelines ( https://cabforum.org/ )

As mentioned earlier, CRLs are signed:

https://tools.ietf.org/html/rfc5280#page-94
```
      (f)  Obtain and validate the certification path for the issuer of
           the complete CRL.  The trust anchor for the certification
           path MUST be the same as the trust anchor used to validate
           the target certificate.  If a key usage extension is present
           in the CRL issuer's certificate, verify that the cRLSign bit
           is set.

      (g)  Validate the signature on the complete CRL using the public
           key validated in step (f).
```

And if they come from a public CA, which complies with CA/B, those CRLs will have a short lifetime attached.

https://github.com/cabforum/documents/blob/d9b360f15d6ce3079c735e69539666e09b0c85e7/docs/BR.md#497-crl-issuance-frequency-if-applicable

```
If the CA publishes a CRL, then the CA SHALL update and reissue CRLs at least once every seven days, and the value of the nextUpdate field MUST NOT be more than ten days beyond the value of the thisUpdate field.
```

The nextUpdate / thisUpdate basically allows the PKI also to control how often the PKI aware application need to fresh their CRL / OCSP response caches.

So the CRL documentation you are looking for would hold: "If you are using a  public CA vendor, and don't want online revocation checking due to privacy concerns or lack of internet access, you need to setup a custom logic or manually fetch every X days the signed CRL from your CA / PKI vendor."

(And that wouldn't work for CA vendors which do OCSP only - e.g. letsencrypt.org)

Due to the signing of the CRLs by the CA, you can't issue your own Sub-CRL for a separate without loosing the signature verification or violating RFC 5280.

So yes, TLS syslog setups using public CAs not having internet access will just not have a full compliant with to implemented Syslog & TLS: certificate base authentication ( https://tools.ietf.org/html/rfc5425#section-4.2.1 ). They will just have encrypted syslog traffic, if that's confidential it's up the state if one of the used keys got compromised. The only authentication/authorization you have is certificate fingerprint matching. Which will be in future also unfeasible in case the CA/B decides to reduce the maximum certificate lifetime to less then a year. (Last year their was a discussion/proposal to reduce it to 90 days. Overall the industry trend will probably reducing the lifetimes not increasing it.)

I fully agree that the combination of public CA and no Internet access is problematic and IMHO not fit for some production deployments.

If you are working in an environment with no internet access, and you don't want any central revocation management, just use your private PKI/CA or use end-entity certificate matching as described in https://tools.ietf.org/html/rfc5425#section-4.2.1

```
   o  End-entity certificate matching: The transport sender or receiver
      is configured with information necessary to identify the valid
      end-entity certificates of its authorized peers.  The end-entity
      certificates can be self-signed, and no certification path
      validation is needed.  Implementations MUST support certificate
      fingerprints in Section 4.2.2 and MAY allow other formats for
      end-entity certificates such as a DER-encoded certificate.  This
      method provides an alternative to a PKI that is simple to deploy
      and still maintains a reasonable level of security.

   Both transport receiver and transport sender implementations MUST
   provide means to generate a key pair and self-signed certificate in
   the case that a key pair and certificate are not available through
   another mechanism.
```

If you want central revocation management but you can't have internet access, just spin your own private PKI with PKI infrastructure accessible internally without any internet access. You have then full control of the CRL or OCSP URLs.

If you want to do offline CRL checks by configuring and manually updating files, you can, by building again a public PKI. In that case you could even control the lifetime of the CRL file. (I don't recommend offline CRLs, due to the risk of silent CRL expiration and overall update/syncing complexity). But most probably the End-entity certificate matching makes more sense in such environment and certificate fingerprint matching makes more sense in such environments.


Private CAs or OCSP-stapling would solve most of those discussed problems. But due to interoperability with non-OCSP-stapling aware syslog peers, out-of-band revocation checks are required to have Certificate-Based Authentication and authorization complying with RFC 5425.


Whatever you decide on, lots of certificate revocation/compromise-protection and all the criticism about revocation checking is a problem of public/internet PKIs.The trend in the security and/or browser community seems to be lower certificate lifetimes, which the CAs hesitate to do, because their customer will face problems with the high degree of certificate renewal automation.

With private PKIs you have full control of certificate lifetimes you can control where the out-of-band revocation checks are directed to and how often they are done. And you can control the certificate lifetimes to your need. If you have no Internet access, you should consider that.

**Comment by d1eanotherday** [comment]
Is there an update on this PR... will it be included into a future release?
We are in need for this functionality on a project and would be very happy if OCSP would be supported from rsyslog.

**Comment by gollub** [comment]
I'm currently busy with another project.
Once I have spare cycles (in few months or so) I plan to continue on this and implement what was proposed here: https://github.com/rsyslog/rsyslog/pull/4307#issuecomment-639412346

But obviously if anyone else has free cycles feel free to pick up the commits from this PR.

**Comment by rgerhards** [comment]
@cubic-dev-ai review

**Comment by rgerhards** [comment]
@cubic-dev-ai review

File: runtime/net_ossl.c [hunk 2]
```diff
@@ -912,6 +915,568 @@ rsRetVal net_ossl_chkpeercertvalidity(net_ossl_t __attribute__((unused)) * pThis
...
+
+    memset(&hints, 0, sizeof(hints));
+    hints.ai_family = AF_UNSPEC;
+    hints.ai_socktype = SOCK_STREAM;
+
+    /* TODO: DNS resolution is blocking with no timeout. A malicious certificate
+     * with OCSP responder URLs pointing to slow/unresponsive DNS names can cause
+     * the TLS handshake to block indefinitely. This is called from the TLS verify
+     * callback. Consider using async DNS resolution or making OCSP optional.
+     */
```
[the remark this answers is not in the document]
rgerhards: that's acceptable for the current PoC

runtime/net_ossl.c
```diff
@@ -912,6 +915,568 @@ rsRetVal net_ossl_chkpeercertvalidity(net_ossl_t __attribute__((unused)) * pThis
...
+    if (BIO_flush(bio) != 1) {
+        LogError(0, RS_RET_NO_ERRCODE, "OCSP: Failed to flush BIO\n");
+        goto err;
+    }
+
+    /* Read HTTP response - skip headers */
+    char buf[1024];
+    int in_headers = 1;
+    long content_length = -1;
+    const long MAX_OCSP_RESPONSE_SIZE = 1024 * 1024; /* 1MB limit */
```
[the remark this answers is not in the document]
rgerhards: follow-up work

rgerhards:
remaining AI review comments are ok for the merge of experimental functionality

PR #4314, opened 2020-06-04
build system: add --enable-libdbi=optional functionality

rgerhards:
If set, builds a dummy module which just emits a "module not supported
on this platform" error message when loaded.

Primary use case for this system is Debian-ish builds on SUSE OBS,
where we prefer to have a single package definition for all versions
(else things get much more complicated).


rgerhards:
Somehow notifications seem to have not been received. Anyhow, this looks irrelevant to the PR, so merging it.

PR #4322, opened 2020-06-09
Fixing imfile segfaulting on selinux denial

jvymazal:
If imfile is denied access to file watched trough symlink there is
unchecked condition resulting in access to not initialized memory.


rgerhards:
There is something fishy with the recently updated Ubuntu 16 container. I'll sort it out and will merge before the 8.2006 final.

PR #4327, opened 2020-06-24
build bugfix: version month value misinterpreted as octal

rgerhards:
This causes issue for x.08 releases, because 8 is not an
octal digit ;-) The leading zero was obviously useless and
can be removed.


rgerhards:
LGTM fails due to bug in current master (which this PR fixes).

PR #4330, opened 2020-06-25
MMDARWIN:: improve configuration flexibility, and successive use with the same message

frikilax:
# Improvements
- now able to get fields in local variables ('$.' variables)
- now able to configure a custom root container (module-wise, default is still $!mmdarwin) to put results in
- now able to put result in nested keys (example: "!key1!key2")

# Fix
- don't generate a new uuid when being called successively with the same message, reuse previously put uuid from message


rgerhards:
@frikilax I haven't checked the others, but this looks very valid and probably causes the remaining failures: https://build.rsyslog.com/#/builders/136/builds/3172

frikilax:
> @frikilax I haven't checked the others, but this looks very valid and probably causes the remaining failures: https://build.rsyslog.com/#/builders/136/builds/3172

I agree there is a problem here, I'll have a look

PR #4342, opened 2020-07-02
fixed ChkDisabled method to make config.enabled work

jvymazal:
There was wrong negation in the method so it returned 0/1 in reverse
and also it did not mark the node to not be reported as unknown at all
times which is needed after all.


jvymazal:
@rgerhards the 2 CI failures looks like some errors, could you please restart them?

rgerhards:
The centos 7 seems to be a CI problem, occurs everywhere. Will check, but may need a day or two.

PR #4349, opened 2020-07-08
CI: add github actions

rgerhards:

### DIFF
--- a/.github/workflows/ci.yml.template
+++ b/.github/workflows/ci.yml.template
@@ -1,0 +1,113 @@
+# Copyright 2020 Rainer Gerhards and Others
+#
+# https://github.com/rsyslog/rsyslog-pkg-ubuntu
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# References:
+#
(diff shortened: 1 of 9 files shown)


rgerhards:
@Whissi the "find group to drop privs to" logic seems to have an issue with the container/github action runner interaction. I'll implement a workaround so that we use `id -g`, but it would be great if you could have a look and check what causes the issue. Here in this env gid != uid.

Whissi:
Do we have have some error logs available showing the problem?

rgerhards:
I'll switch the workaround back so that we get the original error. This is from the early runs https://github.com/rsyslog/rsyslog/runs/853278746?check_suite_focus=true without the workaround.

It kind of looks like it is no problem on debian.  The "workaround" does not yet work. Possibly it's a container issue as well. At least on buildbot everything worked. Not sure if we can harden for any situation. I begin to think that gid 116 used in this case by the github runners is not present in /etc/groups inside the container and that causes the issue. Probably hard to get right in any case.

Whissi:
Mh, so the error is
> rsyslogd: could not set requested group id: Operation not permitted [v8.2008.0.master try https://www.rsyslog.com/e/2432 ]

But `privdrop_common.sh` will detect

> Will use user "user1001" (#&#x2060;1001) and group "grp1001" (#&#x2060;1001)

which means that
```bash
local testgid=$(id --group ${testuser} 2>/dev/null)
```
worked...

I have no clue about GitHub runners yet. From Dockers I know about FSETID capability. Do you know if you can set capabilities somewhere? At the moment it would assume that the error message is real and that process running in container isn't allowed to become group 1001...

rgerhards:
well, the group reported by `id -g` is 161, so something is fishy here....

rgerhards:
@Whissi I think the problem was related to the container environment, which did not properly reflect the execution environment user and group wise. I think I have found a workaround at the container workflow level.

rgerhards:
@Whissi sry for the trouble: it indeed turns out that this is a container-env issue. I have worked around it and this work-around is sufficiently clean (maybe even cleaner than the previous approach).

PR #4357, opened 2020-07-14
CI: run centos 6 checks via github action

rgerhards:


rgerhards:
The "hanging" worker is no longer active in Buildbot and was moved to github actions. So all well.

PR #4369, opened 2020-08-10
gnutls: Propagate CheckExtendedKeyPurpose when accepting connection

ueno:
Previously, when the server accepts a new connection, it doesn't properly set the dataTypeCheck field based on the listening socket. That results in skipping ExtendedKeyUsage (EKU) check on the client certificates.


rgerhards:

Issue #4127, opened 2020-01-16
omelasticsearch failure parsing elasticsearch reply in  8.2001.0 


iashwash:
### Expected behavior
omelasticsearch can parse an elasticsearch response from a bulkmode insert

### Actual behavior
After rsyslog was upgraded to 8.2001.0, omelasticsearch fails to parse valid reply that includes "items"
```
2020-01-15T20:54:10.461319+00:00 our-hostname rsyslogd: omelasticsearch: error in elasticsearch reply: bulkmode insert does not return array, reply is: 1579121650 20:54:10 elk-production green 8 4 550 275 0 0 0 0 - 100.0% {"took":10,"errors":false,"items":[{"index":{"_index":"rsyslog-2020.01.15","_type":"rsyslog","_id":"TpX8qm8BhsTOmI8LK9EE","_version":1,"result":"created","_shards":{"total":2,"successful":2,"failed":0},"_seq_no":43071185,"_primary_term":1,"status":201}}]} [v8.200
1.0 try https://www.rsyslog.com/e/2218 ]
```
### Steps to reproduce the behavior

I am not too sure. I think it has to do with logic in this PR: https://github.com/rsyslog/rsyslog/pull/4073

But basically increasing the buffer and then trying to send messages smaller than that increase. The problem persists after restart of rsyslogd, but maybe that's just due to the volume of messages in our buffer.

### Environment
- rsyslog version: 8.2001.0 (aka 2020.01) 
- platform: Ubuntu 18.04

Note: we are running Ubuntu 18.04 hosts that still have rsyslogd  `8.1911.0 (aka 2019.11)` which aren't experiencing issues to the same elasticsearch cluster.

On hosts experiencing this error, I do not see any logged allocation errors `"realloc failed in curlResult"` in the log history.

rgerhards:
Please create a debug log. Instructions: https://www.rsyslog.com/doc/v8-stable/troubleshooting/howtodebug.html

Be sure that the problem in question occurs at least once during the debug run. Once this is done, please post the debug log (e.g. via pastbin or github). After that, we can review the contents and check how exactly rsyslog processing worked.

iashwash:
Sorry for the delay on this, wanted to set-up a simpler environment to grab the debug log from. You can grab it from this gist, just remove REMOVETHISPARTOFTHELINK from it (a lazy attempt against scraping):
[...]

PR #4375, opened 2020-08-13
omelasticsearch: Fix reply buffer reset after health check

julthomas:
This is a proposal to fix github issue #4127 "omelasticsearch
failure parsing elasticsearch reply in 8.2001.0".

The issue happens when more than one server is defined on the
action. On that condition a health check is made through
checkConn() before sending the POST. The replyLen should be
set back to 0 after the health check, otherwise the response
data received from the POST gets appended to the end of the
last health check.


rgerhards:

Issue #4282, opened 2020-05-14
DiskQueue: segfault if queue file cannot be created


alorbach:
When using Disk Queue and a queue.filename that can not be created by rsyslog, the service does not switch to another queue type as supposed to and crashes at a later step: 

> ==111610== Process terminating with default action of signal 11 (SIGSEGV): dumping core
> ==111610==  Access not within mapped region at address 0x0
> ==111610==    at 0x17E832: strmWriteChar (stream.c:2087)
> ==111610==    by 0x16DA88: objSerializeHeader (obj.c:229)
> ==111610==    by 0x16DDE9: BeginSerialize (obj.c:261)
> ==111610==    by 0x158C24: MsgSerialize (msg.c:1222)
> ==111610==    by 0x18A5D0: qAddDisk (queue.c:1031)
> ==111610==    by 0x18A9D3: qqueueAdd (queue.c:1166)
> ==111610==    by 0x1912CC: doEnqSingleObj (queue.c:3098)
> ==111610==    by 0x1915B7: qqueueMultiEnqObjNonDirect (queue.c:3144)
> ==111610==    by 0x120472: multiSubmitMsg2 (rsyslogd.c:1114)
> ==111610==    by 0x19811B: ratelimitAddMsg (ratelimit.c:295)
> ==111610==    by 0x7ED7211: defaultDoSubmitMessage (tcps_sess.c:263)
> ==111610==    by 0x7ED7C6B: processDataRcvd (tcps_sess.c:451)

davidelang:
rsyslog does not try to open anything until it needs to, so the first time that 
rsyslog discovers that the file cannot be created is when it tries to create it.

David Lang

rgerhards:
bug confirmed thx to test by @alorbach

PR #4393, opened 2020-08-26
Queue: add test for segfault issue 4282

rgerhards:
Also added some debug output to trace the error.

see also: https://github.com/rsyslog/rsyslog/issues/4282


rgerhards:
moving to my own branch naming

PR #4422, opened 2020-09-15
rsyslog: add anon to omrelp/imrelp

### LINKED ISSUE #3838: tls.authMode="anon" is not supported for imrelp and results in error ""
### Related

Multiple fixes closed most of the items raised in this issue except the handling of 'anon'. https://github.com/rsyslog/rsyslog/issues/435#issuecomment-429672053 provides more complete info and attempts to find source code handling config that doesn't pass this mode through.

Also: https://github.com/rsyslog/rsyslog/issues/1738.

### Expected behavior

Use of `authMode='anon'` with imrelp input allows clients to connect without client certs.

Consistency with properties used to configure TLS for other modules, e.g. `imtcp`.

### Actual behavior

Rsyslog warns that `authMode='anon'` is invalid. 

```
rsyslogd: imrelp: invalid auth mode 'anon' [v8.1907.0 try https://www.rsyslog.com/e/2291 ]
```

E.g. config

```
input(
   type="imrelp"
   port="7514"
   tls="on"
   tls.caCert="/etc/pki/tls/certs/ca-bundle.crt"
   tls.myCert="/usr/local/etc/pki/test/test_syslog_server.cert.pem"
   tls.myPrivKey="/usr/local/etc/pki/test/test_syslog_server.key.pem"
   tls.authMode="anon"
)
```

After startup, the error message is shown.

### Environment
- rsyslog version: 8.1907.0
- platform: CentOS 7

outscale-mgo:
It seems that librelp consider an anonymous user, an user that have no own Cert.
omtcp/imtcp use authmode="anon" for the same purpose,
this PR uniformize plugins behaviours by adding tls.authMode="anon" to imrelp/omrelp.

To do so, I simply skip the call to relpSrvSetAuthMode and to relpSrvSetOwnCert if authmode is set as "anon".

fix #3838

Signed-off-by: Matthias Gatto <matthias.gatto@outscale.com>

### DIFF
--- a/plugins/imrelp/imrelp.c
+++ b/plugins/imrelp/imrelp.c
@@ -369,6 +369,7 @@ addListner(modConfData_t __attribute__((unused)) *modConf, instanceConf_t *inst)
 	int relpRet;
 	uchar statname[64];
 	int i;
+	int isAnon;
 	DEFiRet;
 
 	if(!inst->bEnableLstn) {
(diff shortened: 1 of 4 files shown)


rgerhards:
@outscale-mgo it looks like this fails miserably ;-) -- see CI

PR #4424, opened 2020-09-18
Replace GNUTLS_SHUT_RDWR by GNUTLS_SHUT_WR when ending TLS connections

rmetrich:
Some TLS servers don't reply to graceful shutdown requests "for
optimization". This results in rsyslog's omfwd+gtls client to wait
forever for a reply of the TLS server which never comes, due to shutting
down the connection with gnutls_bye(GNUTLS_SHUT_RDWR).

On systemd systems, commands such as "systemctl restart rsyslog" just
hang for 1m30 and rsyslogd gets killed upon timeout by systemd.

This patch replaces call to gnutls_bye(GNUTLS_SHUT_RDWR) by calls to
gnutls_bye(GNUTLS_SHUT_WR) which is sufficient and doesn't wait for a
server reply.

A Red Hat customer reproduces the hang reliably when sending the logs to
his Kiwi Syslog server, which apparently doesn't send the TLS reply upon
connection termination request.

### DIFF
--- a/runtime/nsd_gtls.c
+++ b/runtime/nsd_gtls.c
@@ -1321,9 +1321,9 @@ gtlsEndSess(nsd_gtls_t *pThis)
 
 	if(pThis->bHaveSess) {
 		if(pThis->bIsInitiator) {
-			gnuRet = gnutls_bye(pThis->sess, GNUTLS_SHUT_RDWR);
+			gnuRet = gnutls_bye(pThis->sess, GNUTLS_SHUT_WR);
 			while(gnuRet == GNUTLS_E_INTERRUPTED || gnuRet == GNUTLS_E_AGAIN) {
-				gnuRet = gnutls_bye(pThis->sess, GNUTLS_SHUT_RDWR);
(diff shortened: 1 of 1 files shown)


rmetrich:
@rgerhards 
Hi,
Could you please have a look at this PR. This fixes a hang of rsyslog when acting as a log forwarder using TLS and a Kiwi backend.
Thank you.

rgerhards:
@rmetrich thx for the PR; I am doing rounds for merging thus it often takes a couple of days

rgerhards:
@alorbach is this something to consider also for librelp?

rgerhards (approved):
Verified the changed setting, agree to analysis - thx!

Just for info, gnutls doc: https://linux.die.net/man/3/gnutls_bye

rgerhards:
@rmetrich please be aware the the same can happen to librelp (used in im/omrelp). PR for patch based on yours is here: https://github.com/rsyslog/librelp/pull/220

rmetrich:
@rgerhards Thank you!

PR #4426, opened 2020-09-21
core/network: obey net.enableDNS=off when querying local hostname

carbn:
Local hostname resolution uses DNS queries even if the enableDNS is set to off, and this can cause unexpected delays in the HUP signal handling if the DNS server is not responsive. This change makes sure that the resolver is not used if it has been explicitly disabled.


rgerhards (approved):

rgerhards:
thx for the PR!

PR #4433, opened 2020-10-01
Improving nsd ossl logs for better debugging

Dinesh-Ramakrishnan:
Currently the nsd_ossl module doesn't contains logs for few failure scenarios.  This commit adds those additional logs.
Also we have added the nsd_ossl tag to the logs for better identification

### DIFF
--- a/runtime/nsd_ossl.c
+++ b/runtime/nsd_ossl.c
@@ -51,7 +51,6 @@
 #include "nsdsel_ossl.h"
 #include "nsd_ossl.h"
 #include "unicode-helper.h"
-
 /* things to move to some better place/functionality - TODO */
 // #define CRLFILE "crl.pem"
 
@@ -621,7 +620,7 @@ osslInitSession(nsd_ossl_t *pThis) /* , nsd_ossl_t *pServer) */
 		osslLastSSLErrorMsg(0, pThis->ssl, LOG_ERR, "osslInitSession");
 		ABORT_FINALIZE(RS_RET_NO_ERRCODE);
 	}
-	
+
 	// Set SSL_MODE_AUTO_RETRY to SSL obj
 	SSL_set_mode(pThis->ssl, SSL_MODE_AUTO_RETRY);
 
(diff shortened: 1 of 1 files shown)


alorbach (approved):
This looks fine to me, very useful addition to openssl error / info output.

rgerhards:
@Dinesh-Ramakrishnan thx for the PR. Regarding the CI fail, I suspect you need to check openssl version number and downgrade reporting if needed

alorbach:
@Dinesh-Ramakrishnan 
For older OpenSSL Versions, there needs to be a check because SSL_get_shared_curve  is available since OpenSSL 1.0.2 and the build is failing now on Centos6. 

You can use this: 
#if OPENSSL_VERSION_NUMBER >= 0x10002000L
#endif

rgerhards:
@Dinesh-Ramakrishnan pls also use Tabs for indention, that's the style we use inside rsyslog

Dinesh-Ramakrishnan:
Thank you very much for the comments.  I will address the same and update this pr

Anusha-Pai95:
@rgerhards Rainer Gerhards , Thank you so much for your comments. I have addressed all your comments regarding indentation and version compatibility of the api. But I see there is commit squash issue. we may need to merge all the 6 commits as a single commit. I see that code inspect failure. If possible can that be taken care by the infra or it should be explicitly taker care?

rgerhards:
@Dinesh-Ramakrishnan thx for your persistence! I would appreciate if you could squash, but I can also do this for you. In any case, it would make sense to re-word the commit text to what the commit actually does. Much like description here in github issue. Remember: github may go away in the future, but your commit log won't. So the only real thing is the commit text.

Anusha-Pai95:
@rgerhards , We have squashed all the commits and all the checks are passed now. Are we good to merge this pull request to the upstream?

rgerhards (approved):
LGTM

rgerhards:
> We have squashed all the commits and all the checks are passed now. Are we good to merge this pull request to the upstream?

I think so, but let's give @alorbach a final chance to have a look

alorbach:
Looks good (Y) can go for merge ;)

Issue #4385, opened 2020-08-22
Add functionality to check whether the variable was set (isset function)


cybersecuritypoet:
Add "isset" function to the functions manipulating variables (set,unset,rest) that would return true if the variable is set or false if not. Said function should work on all types of variables.
Currently the behaviour can be somehow replicated by comparing a json variable to empty string but it doesn't allow to check a variable that actually contains an empty string (false positive). Also, using a dedicated function/check would be more readable in general.

This is how it works now (variable not set or empty string):
`if ( $!mytree!myvar == "" ) 

It should work this way (variable not set):
`if( not isset( $!mytree!myvar ) )`

This would be especially useful when manipulating logs that have json property tree generated by mmnormalized or mmjsonparse in advanced log processing scenarios.

PR #4434, opened 2020-10-02
Do not create empty objects when accessing non-existent keys

### LINKED ISSUE #4430: accessing a non-existing key creates an empty parent object
Hello,

Here is a behavior that is a bit annoying.
I tried to have a look at it a while ago but seems to be a bit difficult for me...

### Expected behavior

I would expect that rainerscript statement  `if ($!foo!bar != "") then ...` do not create a `$!foo` empty object if `$!foo` did not exist before the statement is executed.

### Actual behavior

Performing a condition (eg: check for an empty string) on a subtree key that do not exists (depth > 1 from the root container), creates an empty "parent" object. Depending on your context, you may end up with (kind of...) annoying garbage when producing object documents (for instance to index in ES).

I think I could live with it if I was able to check for object type on the parent before performing the condition, or maybe just having a sort of isset() test function. However that really feels like a bug, so I felt it was a good idea to start sharing on this topic, because Rainer or someone else could have an idea for a quick fix!

```
unset $!; 

if ($!source!ip == "") then
    set $!source_ip_was_empty_or_not_set = 1;
else
    set $!source_ip_was_empty_or_not_set = 0;

if ($!a!b!c!d == "") then
    set $!a_b_c_d_was_empty_or_not_set = 1;
else
    set $!a_b_c_d_was_empty_or_not_set = 0;  
```
Produces:
```
{ "$!": { "source_ip_was_empty_or_not_set": 1, "a": { "b": { "c": { } } }, "a_b_c_d_was_empty_or_not_set": 1 } }
```

### Environment
- rsyslog version: 8.2006.0
- platform: CentOS 8

### LINKED ISSUE #4436: hang condition in some plugins due to param-passing mode
A plugin can instruct rsyslog to provide template content as JSON. This is used very seldomly, currently only seen in omjournal and ommongodb, Even more, this use does not look valid (to be left for other issues). So the practical consequences of this bug look like nil.

Rsyslog core hangs when
* OMSR_TPL_AS_JSON passing mode is requested by plugin
* at least one message or local variable is present in message object
* the template uses subtree mode
* the subtree path contains at least one non-existing component

Under these conditions, rsyslog enters an infinite loop.

Severity: none, as the OMSR_TPL_AS_JSON option is not used by any important plugin and seems not to work in the two plugins that seem to support it.

julthomas:
This is a proposal for Github issue rsyslog/rsyslog#4430:
accessing a non-existing key creates an empty parent object
https://github.com/rsyslog/rsyslog/issues/4430

When looking up an object property, the tree of intermediate
object containers was ceated by get and del functions. The
patch is an attempt to fix that behavior by passing 0 to the
bCreate argument of jsonPathFindParent().

There is also one case where the return value of
jsonPathFindParent() was not checked, in the recurssive call
of jsonPathFindParent() itself. This was leading to infinite
loops if bCreate was 0.

closes #4430 
closes #4436

### DIFF
--- a/runtime/msg.c
+++ b/runtime/msg.c
@@ -3142,7 +3142,7 @@ getJSONPropVal(smsg_t * const pMsg, msgPropDescr_t *pProp, uchar **pRes, rs_size
 		field = *jroot;
 	} else {
 		leaf = jsonPathGetLeaf(pProp->name, pProp->nameLen);
-		CHKiRet(jsonPathFindParent(*jroot, pProp->name, leaf, &parent, 1));
+		CHKiRet(jsonPathFindParent(*jroot, pProp->name, leaf, &parent, 0));
 		if(jsonVarExtract(parent, (char*)leaf, &field) == FALSE)
 			field = NULL;
 	}
@@ -3197,7 +3197,7 @@ msgGetJSONPropJSONorString(smsg_t * const pMsg, msgPropDescr_t *pProp, struct js
 		ABORT_FINALIZE(RS_RET_NOT_FOUND);
 	}
 	leaf = jsonPathGetLeaf(pProp->name, pProp->nameLen);
-	CHKiRet(jsonPathFindParent(*jroot, pProp->name, leaf, &parent, 1));
+	CHKiRet(jsonPathFindParent(*jroot, pProp->name, leaf, &parent, 0));
 	if(jsonVarExtract(parent, (char*)leaf, pjson) == FALSE) {
 		ABORT_FINALIZE(RS_RET_NOT_FOUND);
 	}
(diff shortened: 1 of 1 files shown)


rgerhards:
Thx, for the PR. I'll evaluate. I would suspect that there are some side-effects, but CI looks fine. Co-incidentally I am working on the `exist()` (#4385) function right now.

julthomas:
Thank you Rainer.
Indeed the exist() function would be great to have as well!

rgerhards:
@julthomas if things turn out well, I'll probably merge your PR first (neat and lean) and rebase mine. But I guess we need some more check to ensure we do not accidently break things. Will see that I create and merge them first, than we can carry on. I'll possibly focus on #4435 first so that I do not lose content by jumping to other things. So pls bear with me maybe a couple of days.

rgerhards:
As a side-note, I think the endless loop you discovered could affect existing functionality as well. It's one of the first things on my tests-to-create list.

julthomas:
OK Rainer no worries. So you now I've been running the patch for a while 
now on a setup with lots of liblognorm and $! manipulations (an 
Elasticsearch front); so far I did not see problems.

rgerhards:
I found the bug: #4436 -- I guess this PR also fixes it, but will verify.

rgerhards:
@julthomas yes, it fixes #4436 ; it will not add a test to CI as the whole code path looks pretty unused, and hard to test due to no valid way to use it ;-)

rgerhards (approved):
Looks like you have exactly found and fixed the cause -- thx!

PR #4435, opened 2020-10-02
add 'exists()' script function to check if variable exists

rgerhards:
This implements a way to check if rsyslog variables (e.g. '$!path!var') is
currently set of not.

closes https://github.com/rsyslog/rsyslog/issues/4385

### DIFF
--- a/grammar/grammar.y
+++ b/grammar/grammar.y
@@ -8,7 +8,7 @@
  * PRI filter) are very hard to beat in ease of use, at least for simpler
  * cases.
  *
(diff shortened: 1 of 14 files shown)


rgerhards:
see also #4434 which is related and in parts conflicting; we need some additional tests for both this PR and #4434 to ensure we do not break backwards compatibility.

rgerhards:
there was only one issue in regard to backwards compatibility, and this is solved: #4436

rgerhards:
The PR itself is now ready to be merged, except for missing doc. Tests for new functionality added and passing. @davidelang @julthomas

PR #4443, opened 2020-10-09
msg: Fix segfault and memory leak on set operation

julthomas:
A segfault may occur in jsonPathFindNext() when bCreate is 1 and when the <root> container where to insert the namebuf key is not an object.

Here is simple reproducible test case:
```
// ensure we start fresh
// unnecessary if there was no previous set
unset $!;

set $! = "";
set $!event!created = 123;
```
There is also a memory leak in msgAddJSON() if jsonPathFindParent() failed because of a missing call to json_object_put(json).


julthomas:
Push forced to address code style failed tests.

julthomas:
This PR is for issue rsyslog/rsyslog#4444

rgerhards (approved):

PR #4445, opened 2020-10-19
do_inotify block only once for events

kortemik:
do_inotify block only once for events with timeouts. there is possible scenario where no data is available and read() enters blocking while poll() is released.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,11 @@
 ----------------------------------------------------------------------------------------
 Scheduled Release 8.2204.0 (aka 2022.04) 2021-04-??
+- 2022-04-04: bugfix: cosmetic data races
+  there was a more or less cosmetic data race which could happen when children
+  processes died in quick sequence. Even then, no real harm happened, as all
+  children were reaped eventually.
+  A similar data race exists for HUP processing.
+  However, these races polluted TSAN test runs, and so we fixed them
 - 2022-04-01: add property options to support ISO week/year number
   Thanks to Mattia Barbon for the patch.
 - 2022-04-01: core bugfix: "action suspended" message was emitted even when turned off


kortemik:
the centos_6 cancelled after timeout may be a bug:
"rsyslogd debug: input thread term: timeout expired waiting on thread imfile termination - canceling"
though it was there all the time as EINTR is being used as poll restart condition without checking if shutdown is the cause of EINTR invocation. should be fixed now.

rgerhards:
note: inotify is racy, thus it can happen that you get a notification but no data is left for reading

kortemik:
i think this is now ready for merge

rgerhards:
I cannot merge because this would introduce a bug. pls see https://github.com/rsyslog/rsyslog/pull/4445#issuecomment-712235951

kortemik:
I am sorry for not understanding that the note was for me. Would it be possible to explain in very brief the bug introduced as I am unable to undestand how the functionality changes?

kortemik:
I think the change introduced here changes the racy reading to be same for the triggered file as for the rest of the files:

current:
1) Inotify is triggered
2) Triggering file is read for new data
3) Rest of the files are read for timer expiration
4) more inotifys are processed *these may be at the end of the file already as well then*

new:
1) Inotify is triggered
3) All files are read for timer expiration, which includes the triggering file
2) Triggering files is read for data, *this may be at the end of the file already then*
4) more inotifys are processed *these may be at the end of the file already as well then*

therefore I find that the change introduced harmonises the way triggering and the future triggers are handled, by not making the triggering one as an exceptional one, so it should be easier to test the changes as well.

If I still have it wrong then please close.

rgerhards:
@kortemik Please add relevant info to the commit message itself. Among others makes reviewing easier (also for distro packagers who think about back-porting). I am going through the code soon,

rgerhards (approved):
code looks good to me, but we need a more descriptive commit message

rgerhards:
I don't know why this was closed when I fixed the commit message, but I have re-opened it at https://github.com/rsyslog/rsyslog/pull/4844

PR #4456, opened 2020-11-04
add new system property $now-unixtimestamp

rgerhards:
Among others, this may be used as a monotonic counter
for doing load-balancing and other things.

Thanks to Nicholas Brown for suggesting this feature.

### DIFF
--- a/runtime/msg.c
+++ b/runtime/msg.c
@@ -7,7 +7,7 @@
  * of the "old" message code without any modifications. However, it
  * helps to have things at the right place one we go to the meat of it.
  *
- * Copyright 2007-2019 Rainer Gerhards and Adiscon GmbH.
+ * Copyright 2007-2020 Rainer Gerhards and Adiscon GmbH.
  *
  * This file is part of the rsyslog runtime library.
  *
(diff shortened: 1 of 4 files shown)


nickbroon:
This is fantastic, thanks @rgerhards !

Just to link to the original email discussion: http://rsyslog-users.1305293.n2.nabble.com/timegenerated-utc-or-high-precision-now-utc-td7596138.html

nickbroon:
Just to check; will this property be available outside of templates as well? So that it can used as part of arithmetic expression, in the same way the that 'parse_time($timegenerated)', as I see in the code change that it's returned as a string not an integer.

That is; could $now-unixtimestamp replace parse_time($timegenerated) in the below?

```
# Get Unix timestamp of current message
set $.tnow = parse_time($timegenerated);

# Rate limit info to 5 every 60 seconds
if ($!severity == 6 and $!facility == 17) then {
  if (($.tnow - $/trate) > 60) then {
    # 5 seconds window expired, allow more messages
    set $/trate = $.tnow;
    set $/ratecount = 0;
  }
  if ($/ratecount > 5) then {
    # discard message
    stop
  } else {
    set $/ratecount = $/ratecount + 1;
  }
}
```

rgerhards:
@nickbroon yes -- I have just amended the test case to cover this scenario as well. Thx for pointing this out. Is it OK if I use your sample inside the doc as a use case?

nickbroon:
Thanks for confirming @rgerhards 
You are welcome to use the above snippet as an example in the docs.

nickbroon:
Fab, thanks @rgerhards :-)

PR #4470, opened 2020-11-22
imhttp updates - query parameter ingestion & basic auth support

n2yen:
- Basic Authentication support & tests
  * configured via imhttp option "basicAuthFile". This option should be configured
    to point to your htpasswd file generated via a standard htpasswd tool.
  tests:
  * imhttp-post-payload-basic-auth.sh
  * imhttp-post-payload-basic-auth-vg.sh

- Query parameter ingestion capability & tests
  use `addmetadata` option to inject query parameters into
  metadata for imhttp input.

### DIFF
--- a/tests/imhttp-post-payload-basic-auth.sh
+++ b/tests/imhttp-post-payload-basic-auth.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+# This is part of the rsyslog testbench, licensed under ASL 2.0
+
+. ${srcdir:=.}/diag.sh init
+generate_conf
+IMHTTP_PORT="$(get_free_port)"
+add_conf '
+template(name="outfmt" type="list") {
+  property(name="msg")
+  constant(value=" - ")
+  constant(value="{ baz: ")
+  property(name="$!metadata!httpheaders!baz")
+  constant(value=", daddle: ")
+  property(name="$!metadata!httpheaders!daddle")
+  constant(value=", fiddle: ")
+  property(name="$!metadata!httpheaders!fiddle")
+  constant(value=" }")
+  constant(value=" - ")
+  constant(value=" !metadata: ")
+  property(name="$!metadata")
+  constant(value="\n")
+}
+
+#
+# note htpasswd file contains entry:
+# user1:1234
+#
+module(load="../contrib/imhttp/.libs/imhttp"
+       ports="'$IMHTTP_PORT'")
+input(type="imhttp" endpoint="/postrequest"
+      ruleset="ruleset"
+      addmetadata="on"
+      basicauthfile="'$srcdir'/testsuites/htpasswd")
+ruleset(name="ruleset") {
+	action(type="omfile" file=`echo $RSYSLOG_OUT_LOG` template="outfmt")
+}
+'
+startup
+# validate invalid passwd
+ret=$(curl -v -s -H Content-Type:application/json -H "BAZ: skat" -H "daddle: doodle" -H "fiddle: faddle" \
+  --user user1:badpass --write-out '%{http_code}' \
+  http://localhost:$IMHTTP_PORT/postrequest -d '[{"foo":"bar","bar":"foo"},{"one":"two","three":"four"}]')
+
+echo "response: $ret"
+if ! ( echo "$ret" | grep "Error 401: Unauthorized" )
+then
+  echo "ERROR: should not be unauthorized"
+  error_exit 1
+fi
+
+# validate correct password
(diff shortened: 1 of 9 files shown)


n2yen:
may require docker image update to include:
- libaprutil1-dev (ubuntu)
- apr-util-devel (red-hat)

rgerhards:
@n2yen sry for being late to support the merge ;-)

I have one question, though: would it make sense to have fallback capability to run with the old feature set if the newly required libs are not available? I ask because we had situations in the past where a module could not longer be provided or built for some platforms due to missing dependencies.

rgerhards:
@n2yen I have experimentally begun to add the lib to some containers (looks like it's not a new thing ;-)).

Can you please do a simple `git commit --amend -S ; git push -f` to trigger a new CI run. Unfortunately github prevents me from manually initiating it after a certain while. If you currently have no time, I can do this, but it looks nicer when you do.

n2yen:
> @n2yen sry for being late to support the merge ;-)
> 
> I have one question, though: would it make sense to have fallback capability to run with the old feature set if the newly required libs are not available? I ask because we had situations in the past where a module could not longer be provided or built for some platforms due to missing dependencies.

yes, I'll have a look at adding a configuration switch for this. Looks like we can do this in another PR or would it be better to add to this PR?

rgerhards:
@n2yen 

> yes, I'll have a look at adding a configuration switch for this. Looks like we can do this in another PR or would it be better to add to this PR?

Thx, but hold on for a moment -- it looks like the dependency seems to be present everywhere... I'll see.

Any thoughts in that (anyone)?

rgerhards:
@n2yen fighting with my company VM environment, this currently blocks me from editing the containers. Just so that you know. Once fixed, I'll go ahead.

rgerhards:
@n2yen going forward :-)

We now also have some real failure results. Pls have a look at the "Ubuntu 20" fails.

n2yen:
Hi @rgerhards , for some reason the  imhttp-post-payload-basic-auth.sh unit test is failing only in `ubuntu 20 distcheck`. Appears to pass in the other enabled platforms, like [ubuntu 20 SAN, ubuntu 18, ubuntu 16].  
It seems to be getting an `empty response` when the tests executes a curl command with basic auth headers: https://github.com/rsyslog/rsyslog/pull/4470/checks?check_run_id=3559544137#step:4:81

Any ideas why this particular check is different from others? What's the best way to run this check locally?

davidelang:
differences in the version of curl between distos??

tests/imhttp-post-payload-basic-auth.sh
```diff
@@ -0,0 +1,62 @@
...
+#
+# note htpasswd file contains entry:
+# user1:1234
+#
+module(load="../contrib/imhttp/.libs/imhttp"
+       ports="'$IMHTTP_PORT'")
+input(type="imhttp" endpoint="/postrequest"
+      ruleset="ruleset"
+      addmetadata="on"
+      basicauthfile="testsuites/htpasswd")
```
rgerhards: `make distcheck` environment is differently, you will not find the files inside the subdir of the current working directory (because the build is moved out of if). you need to use the `$srcdir` variable to get to the tarball extract point. I think this will work (but may be wrong):
```
basicauthfile="'$srcdir'/testsuites/htpasswd"
```

Note that you can run `make distcheck` also locally, what will get you the same issue. If you place a bash call inside the test, you see the directory structure first hand.

rgerhards (changes requested):

rgerhards:
> differences in the version of curl between distos??

It's a `make distcheck` "issue" (actually a test script bug that because it doesn't follow the autotools check details - which are a bit strange) -- see my source comment.

rgerhards:
It also looks like there is no need to support cases with missing apr library - it seems to be universally available on relevant builds. If I run into problems in later stages, we can always talk about it ;-)

tests/imhttp-post-payload-basic-auth.sh
n2yen: thanks @rgerhards, making the changes!

n2yen: think that was it! one of the tests failed, but looks unrelated to the change.

rgerhards: Looks like flakes - we sometimes have timing issues. I restart and guess that will solve the issue.

rgerhards (approved):

rgerhards:
@n2yen That did it! Thx again for the PR, will merge soon.

PR #4473, opened 2020-11-26
testbench: fix the wrong message injection object of instance 1

VincentZhu007:
In some client-server test cases, messages are supposed to be injected into the instance 2(client), but they are actually injected into instance 1(server), which may lead to false negative results. This patch fixed it by replacing 'injectmsg' with 'injectmsg2', and dealt with some minor issues.


rgerhards:
@VincentZhu007 sorry, I am a bit puzzled with what github did. I addressed some minor issues and force-pushed back to your branch (which was named "master" if I am not totally mistaken). Then github closed the issue while it still is not merge.

I will now continue with https://github.com/rsyslog/rsyslog/pull/5032, which is your exact same PR + the minor additions.

PR #4482, opened 2020-12-07
CI: temporarily disable Ubuntu 18 distcheck

rgerhards:
We have problems with third-party dependencies after container
update and need to get things going for the upcoming scheduled
stable release. So we disable this and fix it after the release.


rgerhards:
This work-around is no longer needed.

PR #4497, opened 2021-01-14
testbench: fix kafkacat issues

rgerhards:
The kafkacat tool has an upper limit of how many messages it can send
at once. Going over that limit causes messages loss. The exact limit
seems to depend on the environment. This cases testbench false positives.

This commit fixes two related issues:
- errors during kafkacat run were not detected - this has been added
- we now have a "max messages at once" setting, after which kafkacat
  is restarted for the next batch of messages. It currently is set
  to 25,000 msgs per incarnation. All tests loop now to send the
  required number of messages. This has been fixed at the testbench
  framework level, so no need to adjust individual tests.

### DIFF
--- a/tests/diag.sh
+++ b/tests/diag.sh
@@ -442,10 +442,24 @@ injectmsg_kafkacat() {
 		printf 'TESTBENCH ERROR: TESTMESSAGES env var not set!\n'
 		error_exit 1
 	fi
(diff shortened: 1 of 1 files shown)


rgerhards:
CI failure as expected due to other outstanding PR, so good to merge.

PR #4498, opened 2021-01-14
testbench: fix year-dependendt clickhouse test

rgerhards:
A test had the year value hardcoded and as such failed whenever the
year changed. This patch corrects that.


rgerhards:
CI fails are related to issues which have separate PR. So good to merge.

Issue #4494, opened 2021-01-08
rsyslog leaves imfile state files behind (these get never deleted)


rmetrich:
### Expected behavior

`imfile-state:INODE:HASH` get deleted once rsyslog finished processing the file corresponding to that inode after a log rotation occurred.

### Actual behavior

No deletion happening at all

### Steps to reproduce the behavior

1. Compile rsyslog with imfile support

   ~~~
   $ ./configure --enable-imfile
   $ make
   $ sudo make install
   ~~~

2. Create a dedicated mount point for the test (formatted with XFS preferably, so that inodes get reused)

   ~~~
   # mkdir -p /disk2/rsyslog
   ~~~

3. Create rsyslog.conf configuration file

   ~~~
   global(workDirectory="/disk2/rsyslog")
   module(load="imfile" mode="inotify")
   input(type="imfile"
      File="/disk2/input.log"
      Tag="imfile:"
      Severity="info"
      Facility="local6"
   )
   local6.info			/disk2/output.log
   ~~~

4. Create the following `logger.sh` script

~~~
#!/bin/sh

file="/disk2/input.log"
maxfilesize=$((8 * 1024 * 1024))

maxlogs=20

rotatelogs() {
	if [ -e "$file.$maxlogs" ]; then
		rm "$file.$maxlogs"
	fi

	for i in $(seq $(($maxlogs - 1)) -1 1); do
		[ -e "$file.$i" ] || continue
		mv "$file.$i" "$file.$(($i + 1))"
	done
	mv "$file" "$file.1"
	touch "$file"
}

statlogs() {
	echo
	date
	echo "$file $(stat -c "%i" $file)"
	for i in $(seq 1 $maxlogs); do
		[ -e "$file.$i" ] || break
		echo "$file.$i $(stat -c "%i" $file.$i)"
	done
}

seqnum=1
filler=$(printf "X%.0s" $(seq 1 250))
msgsize=$(printf "%09d $filler\n" $seqnum | wc -c)

writelog() {
	printf "%09d $filler\n" $seqnum >> "$file"
	let seqnum++
}

iters=$(($maxfilesize / $msgsize))
while :; do
	i=0; while [ $i -le 32140 ]; do
		if [ $(($seqnum % 1000)) -eq 0 ]; then
			echo $seqnum
		fi
		writelog
		let i++
	done
	rotatelogs
	statlogs
done
~~~

5. Execute rsyslogd with this configuration

   ~~~
[...]

PR #4500, opened 2021-01-14
imfile: fix state files leaking (Issue #4494)

rmetrich:
There were two issues with the current code:
1. the state file being deleted upon imfile vanishing was not the state
   file with 'file_id'
2. when a imfile was renamed (e.g. after a rotation), the file was
   flagged with 'in_move' which resulted in the state file to never be
   destroyed

The state file was not destroyed because of the following sequence:
1. imfile is renamed, a IN_MOVED_FROM event is fired and 'in_move' flag set
2. polltree executes and finds the imfile is gone, hence deletes it
3. due to having 'in_move' flag, the state file isn't destroyed

Finally, not related to the issue, but close enough, the inotify 'wd'
descriptor was never deleted, causing an internal error to occur if a
rotated file was being modified.

The new code gets rid off 'in_move' flag completely, because it is
useless: upon file renaming, polltree will detect that file vanished and
will delete the entry after flushing.
The old code was anyway unreliable in the case more than 1 file was
being renamed in a given directory simultaneously: since polltree
executed as soon as first notification IN_MOVED_FROM was received, it
deleted the entry for the other files being renamed as well before the
'in_move' flag could have been set.

Signed-off-by: Renaud Métrich <rmetrich@redhat.com>


rgerhards:
Thx for the PR - and sorry I had to be unresponsive for quite a while.

I need to review this in depth. The root cause is that imfile is not "leaking" state files, but they are rather intentionally not deleted. The reason is a number of folks who require that when a log file is moved elsewhere and moved back later to the same location, it is not treated as new but processing continues from where it previously ended.

I had plans to implement a kind of background scheduler to delete such files after a configurable interval (e.g. 2 days). Not sure yet if I managed to do or not (Corona volunteer work hit me hard the past couple of months). It also makes sense to have an **option** to remove state files immediately.

Also, just FYI: rsyslog does not rely on inotify. A prime reason is that it is not available on all platforms. The more subtle reason is that inotify does not come with guarantees even where it is available. This is why take inotify just as a hint and then poll the file system ourselves.

Just wanted to let you know. Still struggling with catching up after I could reduce part of my volunteer work.

rmetrich:
Hi @rgerhards , no problem, take your time.

My scenario is a very basic one: log gets rotated (through logrotate) or is an audit log (so auditd rotates it itself) and I do see state files remaining forever.

I understand some people would like to move away a file and bring it back and it should continue, but it's not a common use case to me, whereas mine is.

davidelang:
both modes have their place, so I think this needs to be configurable (and there 
is also an enhancement request in to make imfile delete the file once it's 
finished reading it, in that mode it should also remove the state file)

David Lang

rmetrich:
@davidelang I understand, this needs to be made configurable.
However I don't see how the current design (i.e. imstate-file based on inode) could work in that case either:
if the log file is moved away and moved back again later, it has not much chance to get the same inode, hence having this kind of functionality requires to keep track of the filename, not the inode.

rgerhards:
The inode is not that important. It's primarily a a fallback mechanism when we do not have sufficient data for the file hash. And after we reworked imfile to do processing as it does now, we did not get any complaints from the user base I mentioned.

davidelang:
a mv just changes the name of the file, it doesn't change the inode. as long as 
the move away and back are on the same filesystem, the inode will remain the 
same.

rmetrich:
Then this is already handled fine: the inode being "sticky" the file being renamed will be processed.
I was more thinking of moving a file out of file system and copying it back later.

plugins/imfile/imfile.c
```diff
@@ -2388,15 +2408,23 @@ in_processEvent(struct inotify_event *ev)
...
 		ev->mask, etry->act->edge->is_file, etry->act->name);
 
-	if((ev->mask & IN_MOVED_FROM)) {
-		flag_in_move(etry->act->edge->node->edges, ev->name);
-	}
-	if(ev->mask & (IN_MOVED_FROM | IN_MOVED_TO))  {
-		fs_node_walk(etry->act->edge->node, poll_tree);
-	} else if(etry->act->edge->is_file && !(etry->act->is_symlink)) {
-		in_handleFileEvent(ev, etry); // esentially poll_file()!
+	if(!etry->act->edge->inotifyNewBehaviour) {
```
Cropi, line 2411:
```suggestion
	if(!etry->act->edge->instarr[0]->inotifyNewBehaviour) {
```

plugins/imfile/imfile.c
```diff
@@ -977,9 +984,13 @@ act_obj_destroy(act_obj_t *const act, const int is_deleted)
 	if(act == NULL)
 		return;
 
-	DBGPRINTF("act_obj_destroy: act %p '%s' (source '%s'), wd %d, pStrm %p, is_deleted %d, in_move %d\n",
-		act, act->name, act->source_name? act->source_name : "---", act->wd, act->pStrm, is_deleted,
-		act->in_move);
+	if(act->edge->instarr[0]->inotifyNewBehaviour) // TODO: same file, multiple instances?
```
Cropi, line 987: You will need to guard both  **if** and **else** blocks with curly brackets, because of DBGPRINTF expansion.

plugins/imfile/imfile.c
```diff
@@ -996,13 +1007,16 @@ act_obj_destroy(act_obj_t *const act, const int is_deleted)
 		pollFile(act); /* get any left-over data */
 		if(inst->bRMStateOnDel) {
 			statefn = getStateFileName(act, statefile, sizeof(statefile));
-			getFullStateFileName(statefn, "", toDel, sizeof(toDel)); // TODO: check!
+			if(!inst->inotifyNewBehaviour)
+				getFullStateFileName(statefn, "", toDel, sizeof(toDel)); // TODO: check!
+			else
+				getFullStateFileName(statefn, act->file_id, toDel, sizeof(toDel)); // TODO: check!
```
Cropi, line 1013: I think we don't need to distinguish between old and new style anymore, due to https://github.com/rsyslog/rsyslog/pull/4633/commits/fca8e14d65019ee1b814c4ed16ec7ff04de30809#diff-cfff74f7d8ebca9eeb47097882fc54089b1973aeb56c4b2d8240f40f77b2ca38R999 but I might be wrong in this case.

Cropi (commented):
Is the _inotifyNewBehaviour_  supposed to be an imfile module parameter?

Cropi:
Hello @rmetrich,

I am interested in the proposed changes and would like to use it. Would it be possible to do a rebase here? There were some changes in the imfile module since the beginning of this PR. Let me know if you don't have time to deal with it. Thanks.

rgerhards:
@Cropi please note (my first comment) I still have concerns that this break valid use cases. I need to check if the new code is conditional, and the default is "off", but wanted to leave this info for you as well.

Cropi:
> @Cropi please note (my first comment) I still have concerns that this break valid use cases. I need to check if the new code is conditional, and the default is "off", but wanted to leave this info for you as well.

Sure, I don't want to change the default behavior either. I just wonder if it should be a module parameter or instance parameter(or both?).

sopos:
Actually, I agree with @rmetrich that his usecase is more common so I would not hesitate to even change the default. I think it is much nicer situation that rsyslog does not leave a garbage after itself.
If there is someone mangling files under rsyslog hands they might just enable a support for that..

I could imagine an option `StateFilePurge=immediate|delayed|never` where the delayed would be implemented later.

CossyCosmas:
Quick newbie question:

Why is a new feature required if the 'deleteStateOnFileDelete' parameter already exists to delete orphaned state files?

My use case is fairly simple in that my audit logs have unique names and are very transient. The exist for a few days and are then deleted with no likelihood of them ever reappearing. 

Is this not already covered by the 'deleteStateOnFileDelete’ parameter?  Or, do I need to wait for the additional functionality described in this discussion to be delivered first? 

Thank you. 
Cossy

twogee:
There is a downstream patch for `deleteStateOnFileMove` at [RedHat](https://bugzilla.redhat.com/show_bug.cgi?id=1909639), would you please integrate it instead?

rgerhards:
This has functionality-wise been fixed by a couple of other (also long-standing) PRs that I applied in the past days. As such, I close this one.

Issue #4490, opened 2020-12-18
Possible bug in  build_SparseArrayTable


mabobr:
Function _build_SparseArrayTable()_ has this code:

`pThis->table.sprsArr->entries[i].key = (uint32_t) json_object_get_int(jindex);`

when value of jindex is >  2147483647 then **value 2147483647 is always returned/used**  which breaks searches for values above this limit.
Also function _build_ArrayTable()_ uses _json_object_get_int()_ and must be checked also.
Perhaps _int64_t json_object_get_int64()_ may help somehow.

Related to issue: lookup on sparsearray - wrong result #3394

rgerhards:
@janmejay do you happen to have a couple of minutes to comment on this?

janmejay:
@mabobr Yep, that certainly seems to be a problem. Since we use libfastjson (our own fork), we should probably create another function `json_object_get_uint32` and use that here and in other such call-sites.

janmejay:
@mabobr @rgerhards does [this](https://github.com/rsyslog/libfastjson/pull/162) look ok?

janmejay:
@mabobr @rgerhards actual [bugfix](https://github.com/rsyslog/rsyslog/pull/4502), uses the function the former PR adds.

mabobr:
Hi, this should work for you example:

# how to test:
# install current rsyslog
# put this file into /etc/rsyslog.d
# /usr/sbin/rsyslogd -N1 && sudo systemctl restart rsyslog
# in terminal one run: nc -u -l 51514
# in terminal two run (beware on quotes, ticks):  echo
'{"time":"2020-02-02T18:00:00", "mode":"imuxsock", "host":"venus", "msg":"
{\"name\":\"bill\", \"address\":\"123 Anywhere\", \"City\":\"Cleveland\",
\"Stat
e\":\"Ohio\", \"zip\":\"12345\"}"}' >/dev/udp/127.0.0.1/2514
# check result in termnal 1

module(load="imudp")
template (name="t_parsed_to_json" type="string" string="%$!all-json%\n")
ruleset(
          name="nestedjson"
          queue.type="LinkedList"
){
 unset $!;
 set $.ret = parse_json( $rawmsg, "\$!");
 if $.ret != 0 then { set $!error = "1 - bad json on input"; }
 if strlen($!msg) == 0 then { set $!error = $!error & "2 - msg not found";
}
 set $.ret = parse_json( ltrim($!msg), "\$.lev2");
[...]

PR #4502, opened 2021-01-17
Fix incorrect coercing of json-object int to uint32 in sparse-array-lookup-table implementation

janmejay:
Fix incorrect coercing of json-object int to uint32 
 in sparse-array-lookup-table implementation.

This aims to fix broken int-max case described
 https://github.com/rsyslog/rsyslog/issues/4490 and
 https://github.com/rsyslog/rsyslog/issues/4490.


janmejay:
@rgerhards I haven't bumped libfastjson version. Do I need to? or is it ok to defer it to the release process?

rgerhards:
> @rgerhards I haven't bumped libfastjson version. Do I need to? or is it ok to defer it to the release process?

Requiring a new libfastjson is always a bit hard, as some distros may not carry it for many month. I would prefer if we could do a configure check on the version and emit a warning/error message if and only if the function in question is actually needed by current config. Does that make sense to you?

janmejay:
> > @rgerhards I haven't bumped libfastjson version. Do I need to? or is it ok to defer it to the release process?
> 
> Requiring a new libfastjson is always a bit hard, as some distros may not carry it for many month. I would prefer if we could do a configure check on the version and emit a warning/error message if and only if the function in question is actually needed by current config. Does that make sense to you?

There was a line with incorrect formatting, fixed that. 

Since lookup-table is always enabled, its part of the core, dependency isn't really optional.

rgerhards:
> Since lookup-table is always enabled, its part of the core, dependency isn't really optional.

yes, but if the config does not use any lookup table at all - then we do not need it, right? I think that's a very common case.

rgerhards:
side-note: I have updated the Ubuntu 20.04 containers and re-initiated check runs for them - so we should get a result now

janmejay:
> yes, but if the config does not use any lookup table at all - then we do not need it, right? I think that's a very common case.

Compilation wouldn't succeed without the new function `json_object_get_uint`. So runtime check wouldn't be useful.

rgerhards:
> side-note: I have updated the Ubuntu 20.04 containers and re-initiated check runs for them - so we should get a result now

mistake on my side... will look into it

rgerhards:
@janmejay FYI: I am working on updgrading the CI infrastructure to the new (yet-unreleased) librelp version. I'll let you know when I am done.

rgerhards:
@janmejay FYI: I'll add e commit to your branch to force newer libfastjson and see how this works out

rgerhards:
OK, I cannot simply add. Will create a new PR and then let's see what we can mege.

rgerhards:
@janmejay please have a look at this result: https://build.rsyslog.com/#/builders/142/builds/5175/steps/4/logs/test-suite_log - there seems to be a problem on ARM platforms.

This is from #4507, which is basically your PR, plus an extra commit from my side. Feel free to cherry-pick it. If possible, I would appreciate if you could squash the end result to a single commit.

Note: some of the CI Platforms still fail due to too-old libfastjson (like Solaris), but that's a different matter. The majority should now work, including arm.

rgerhards:
removing milestone as this PR looks stalled

rgerhards:
merged via #4507

PR #4507, opened 2021-01-25
Ammend PR4502 with some cleanup/extra work

rgerhards:


rgerhards:
@janmejay OK... I admit this took veeeery long and my apologies for this. But you might be interested the #4502 is finally merged via this follow-up PR :-) Thanks again for all your work in the past!

PR #4512, opened 2021-01-29
lookup table bugfix: data race on lookup table reload

rgerhards:
A data race could happen when a lookup table was reloaded. We found
this while moving to newer version of TSAN, but have no matching
report from practice. However, there is a potential for this to cause
a segfault under "bad circumstances".

@janmejay  FYI

### DIFF
--- a/grammar/rainerscript.c
+++ b/grammar/rainerscript.c
@@ -2377,6 +2377,8 @@ doFunct_Lookup(struct cnffunc *__restrict__ const func,
 		return;
 	}
 	cnfexprEval(func->expr[1], &srcVal, usrptr, pWti);
+	pthread_rwlock_rdlock(&((lookup_ref_t*)func->funcdata)->rwlock);
+	pthread_rwlock_unlock(&((lookup_ref_t*)func->funcdata)->rwlock);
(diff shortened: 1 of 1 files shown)


grammar/rainerscript.c
```diff
@@ -2377,6 +2377,8 @@ doFunct_Lookup(struct cnffunc *__restrict__ const func,
 		return;
 	}
 	cnfexprEval(func->expr[1], &srcVal, usrptr, pWti);
+	pthread_rwlock_rdlock(&((lookup_ref_t*)func->funcdata)->rwlock);
+	pthread_rwlock_unlock(&((lookup_ref_t*)func->funcdata)->rwlock);
```
janmejay, line 2381:
```suggestion
```

janmejay (commented):
@rgerhards This is insufficient, race still exists. Will create another PR.

janmejay:
@rgerhards PR: https://github.com/rsyslog/rsyslog/pull/4513

rgerhards:
> @rgerhards This is insufficient, race still exists. Will create another PR.

umm ... you caught me, copy&paste error. Unfortunately TSAN was happy after that :-( THX!

PR #4518, opened 2021-02-01
Implement resume counting

thinkst-marco:
The impstats module reports a bunch of counters for actions, including the resume count. However, the resume counter is never actually used and the value stays at 0 in the stats output, regardless of whether an action has been resumed or not.

This tiny change adds a counter increment in the resume path.


rgerhards:
Thx for the PR, much appreciated!

Issue #4534, opened 2021-02-16
Fix librelp tlscommand tests


alorbach:
Adapt tlscommand test changes from: 
https://github.com/rsyslog/librelp/pull/225

They work on all openssl versions platforms of the testbench

rgerhards:
you should do a check if the requried librelp version is available on the platform and, if not, skip. We cannout guarantee that all platforms always have the most recent dependencies

alorbach:
The commit changes the tests with multiple tlscommands that should work on all librelp versions that support tlscommands.

PR #4535, opened 2021-02-16
testbench: changed tlscommands for librelp tls tests.

alorbach:
tlscommands for the following tests have been changed:
- imrelp-tls-cfgcmd.sh
- sndrcv_relp_tls-cfgcmd.sh

This will make sure the librelp tlscommand tests run well on all librelp versions and with all OpenSSL Version. 
Newer OpenSSL Versions would run TLSv1.3 on those tests which is valid for the old tlscommand. However we can NOT disable TLSv1.3 with one tlscommand as this would break the tlsconfcmd for older OpenSSL Versions.

closes: https://github.com/rsyslog/rsyslog/issues/4534

### DIFF
--- a/tests/sndrcv_relp_tls-cfgcmd.sh
+++ b/tests/sndrcv_relp_tls-cfgcmd.sh
@@ -11,7 +11,11 @@ module(	load="../plugins/imrelp/.libs/imrelp"
 	tls.tlslib="openssl")
 # then SENDER sends to this port (not tcpflood!)
 input(	type="imrelp" port="'$PORT_RCVR'" tls="on"
-	tls.tlscfgcmd="Protocol=ALL,-SSLv2,-SSLv3,-TLSv1,-TLSv1.2" 
+	tls.tlscfgcmd="Protocol=ALL,-SSLv2,-SSLv3,-TLSv1,-TLSv1.2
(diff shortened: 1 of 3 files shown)


tests/sndrcv_relp_tls-cfgcmd.sh
```diff
@@ -55,7 +63,7 @@ if [ $ret == 0 ]; then
 	skip_test
 else
 	# Kindly check for a failed session
-	content_check "librelp error 10031" $RSYSLOG_DEBUGLOG
+	content_check "librelp: generic error: ecode 10031" $RSYSLOG_DEBUGLOG
```
mbiebl, line 66: typo? ecode -> code ?

alorbach: its part of the debug output: 
https://github.com/rsyslog/librelp/blob/91aef5c53b92f93e47a36beee9c5937c0ec8365b/src/relp.c#L65

rgerhards:
A better explanation for folks who come across this PR:

test bugfix: some tests did not work with newer TLS library versions

Newer versions provide TLS versions that cannot be disabled in older versions as they are unkonwn there. This is solved by setting restrictions in multiple steps. For older library versions, the final step will error out, but the other one be applied. This permits to achieve proper test results.

PR #4542, opened 2021-02-23
MMDBLOOKUP::Improvement:: allow mmdb to be reloaded on HUP signal

frikilax:
Current version of mmdblookup doesn't handle the mmdb file(s) safely, this improvement allows safe reloading of mmdb files after changing them, without having to restart Rsyslog

### DIFF
--- a/plugins/mmdblookup/mmdblookup.c
+++ b/plugins/mmdblookup/mmdblookup.c
@@ -61,11 +61,13 @@ typedef struct _instanceData {
 		char **name;
 		char **varname;
 	} fieldList;
+	sbool reloadOnHup;
 } instanceData;
 
 typedef struct wrkrInstanceData {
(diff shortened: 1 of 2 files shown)


rgerhards:
@friedl gentle reminder: I think the CI fail has some valid points

rgerhards:
@frikilax gentle reminder: I think the CI fail has some valid points

frikilax:
> @frikilax gentle reminder: I think the CI fail has some valid points

You're right, I didn't pay attention to those tests, will try to look into it shortly

rgerhards:
looks like we (still?) have issues inside the code - memory leaks?

frikilax:
Looks like a known issue with libmaxminddb, the leak is 0 bytes and was already suppressed in tests, but is popping again in another form

PR #4543, opened 2021-02-23
NEW CONTRIB::IMHIREDIS:: input redis module

frikilax:
Redis Input Plugin using hiredis library

REQUIREMENTS:

* hiredis ( https://github.com/redis/hiredis.git )
* libevent ( https://github.com/libevent/libevent.git )

USAGE:

This plugin has two supported "modes":

# **queue** mode
The queue mode will LPOP or RPOP your message from a redis list.
Following parameters are required:
 - mode: Set mode to "queue" to enable the queue mode
 - key: The key to xPOP on
 - server: The name or IP address of the redis server
 - port: The redis listening port

Following parameters are optional:
 - password: If set, the plugin will issue an "AUTH" command before calling xPOP
 - uselpop: If set to "1", LPOP will be used instead of default RPOP

Imhiredis will query Redis every second to see if entries are in the list, if that's the case they will be dequeued
continuously by batches until none remains.
Redis pipelining is used inside the workerthread, with a hardcoded batch size of 10 entries.

Due to its balance between polling interval and pipelining and its use of lists, this mode is quite performant and reliable.
However, due to the 1 second polling frequency, one may consider using the `subscribe` mode instead if very low latency is required and/or bursts of messages every second (approx.) must be avoided.

```
module(load="imhiredis")

input(
    type="imhiredis"
    mode="queue"
    key="vulture"
    server="127.0.0.1"
    port="6379"
    uselpop="1"
    password="foobar"
)
```

# **subscribe** mode
The subscribe mode will SUBSCRIBE to a redis channel. The "key"
parameter is required and will be used for the subscribe channel.

Following parameters are required:
 - mode: Set mode to "subscribe" to enable the subscribe mode
 - key: The key to subscribe to (aka the "channel")
 - server: The name or IP address of the redis server
 - port: The redis listening port

Following parameters are optional:
 - password: If set, the plugin will issue an "AUTH" command before listening to a channel
 - uselpop: If set to "1", LPOP will be used instead of default RPOP

```
module(load="imhiredis")

input(
    type="imhiredis"
    mode="subscribe"
    key="vulture"
    server="127.0.0.1"
    port="6379"
    password="foobar"
)
```

# Master/Replica and fail-safing
This module handles Redis Master/replica(s) environments, and can adapt to a failing master with good resilience:
- at startup, the provided is node is queried to get its role
  - if it's a master, the module recovers the list of current replicas as fail-safes
  - if it's a replica, it tries to follow to the information provided by the replica to get to the master, then keeping the list of current replicas as fail-safes
- when a master becomes unavailable or crashes, imhiredis will try to connect to kept known replicas and discover the new master node
- the list of saved nodes depends on discoveries, this list is created every time the module starts and the master is disconnected
  - the node provided through configuration is always kept in the list, even when unavailable
  - during discoveries, some nodes may be tried several times, this is not a bug but a current simplicity in the code
- imhiredis will only attach to a master Redis node, connecting to a replica is currently impossible

# TODO
* TLS support


rgerhards:
@frikilax it would be great if you could fix the missing LF inside the source code. Otherwise I'll do it after the merge :-)

rgerhards:

rgerhards:
@frikilax Do you think you find time to update the doc? Else I will merge, but that probably means folks will not know about the new feature.

frikilax:
I already started, but couldn't finish it yet.
I can try and finish it by the end of the week.

rgerhards:
> I already started, but couldn't finish it yet.
> I can try and finish it by the end of the week.

perfect! Thx a lot!

PR #4551, opened 2021-03-09
fmunflatten: rscript function to unflatten keys in a JSON object

julthomas:
This commit adds a new rainerscript function to unflatten keys in a JSON tree. It provides a way to expand dot separated fields.

`<result> = unflatten(<source-tree>, <key-separator-character>);`

It allows for instance to produce this: `{ "source": { "ip": "1.2.3.4", "port": 443 } }`
from this source data: `{ "source.ip": "1.2.3.4", "source.port": 443 }`

Here is a sample use case:
```
# Say you have the following tree, obtained for instance
# with mmnormalize.
set $!source.ip = "1.2.3.4";
set $!source.bytes = 3258;
set $!source.geo.country_iso_code = "FR";
set $!destination.ip = "4.3.2.1";

# Now unflatten the keys in the $! tree.
set $.unflatten = unflatten($!, ".");

# You may do this to set back safely the result in $! because
# the function could return a default dummy value of 0 (rsyslog
# number) if $! was not touched (it would evaluates to an empty
# rsyslog string, which is not a JSON datatype).
if (script_error() == 0) then {
    unset $!;
    set $! = $.unflatten;
    unset $.unflatten;
}
```
An output of $! would give this, in pretty-print
```
{
  "source": {
    "ip": "1.2.3.4",
    "bytes": 3258,
    "geo": {
      "country_iso_code": "FR"
    }
  },
  "destination": {
    "ip": "4.3.2.1"
  }
}
```

This patch should address (as a post-processing workaround) the needs expressed in rsyslog/liblognorm#105: _permit specifying a multi-level name_. I believe this is somehow what @rgerhards suggested in a comment on the issue.

The patch was originally written to address a lack of functionality in Elasticsearch ingest pipeline processors. Some processors would not accept dotted fields as input (or output, can't remember exactly, at least in 7.8.1). The dot expander processor can be used to unflatten fields but this is not generic (and thus, difficult to maintain) as you have to apply the processor on a known list of fields.

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -1054,6 +1054,30 @@ TESTS +=  \
 	rscript_hash64-vg.sh
 endif # ENABLE_FMHASH
 endif # HAVE_VALGRIND
+if ENABLE_FMUNFLATTEN
+TESTS +=  \
+	rscript_unflatten_arg1_unsuitable.sh \
+	rscript_unflatten_arg2_invalid.sh \
+	rscript_unflatten_conflict1.sh \
+	rscript_unflatten_conflict2.sh \
+	rscript_unflatten_conflict3.sh \
+	rscript_unflatten_key_truncated.sh \
+	rscript_unflatten_non_object.sh \
+	rscript_unflatten_object.sh \
+	rscript_unflatten_object_exclamation.sh
+if HAVE_VALGRIND
+TESTS +=  \
+	rscript_unflatten_arg1_unsuitable-vg.sh \
+	rscript_unflatten_arg2_invalid-vg.sh \
+	rscript_unflatten_conflict1-vg.sh \
+	rscript_unflatten_conflict2-vg.sh \
+	rscript_unflatten_conflict3-vg.sh \
+	rscript_unflatten_key_truncated-vg.sh \
+	rscript_unflatten_non_object-vg.sh \
+	rscript_unflatten_object-vg.sh \
+	rscript_unflatten_object_exclamation-vg.sh
+endif # HAVE_VALGRIND
+endif # ENABLE_FMUNFLATTEN
 endif
 
 if ENABLE_MMPSTRUCDATA
@@ -1739,6 +1763,24 @@ EXTRA_DIST= \
 	rscript_exists-not2.sh \
 	rscript_exists-not3.sh \
 	rscript_exists-not4.sh \
+	rscript_unflatten_arg1_unsuitable.sh \
+	rscript_unflatten_arg2_invalid.sh \
+	rscript_unflatten_conflict1.sh \
+	rscript_unflatten_conflict2.sh \
+	rscript_unflatten_conflict3.sh \
+	rscript_unflatten_key_truncated.sh \
+	rscript_unflatten_non_object.sh \
+	rscript_unflatten_object_exclamation.sh \
+	rscript_unflatten_object.sh \
+	rscript_unflatten_arg1_unsuitable-vg.sh \
+	rscript_unflatten_arg2_invalid-vg.sh \
+	rscript_unflatten_conflict1-vg.sh \
+	rscript_unflatten_conflict2-vg.sh \
+	rscript_unflatten_conflict3-vg.sh \
+	rscript_unflatten_key_truncated-vg.sh \
+	rscript_unflatten_non_object-vg.sh \
+	rscript_unflatten_object_exclamation-vg.sh \
+	rscript_unflatten_object-vg.sh \
 	rs-cnum.sh \
 	rs-int2hex.sh \
 	rs-substring.sh \
--- a/Makefile.am
+++ b/Makefile.am
@@ -198,6 +198,10 @@ if ENABLE_FMHASH
 SUBDIRS += contrib/fmhash
 endif
 
+if ENABLE_FMUNFLATTEN
+SUBDIRS += contrib/fmunflatten
+endif
+
 if ENABLE_OMKAFKA
 SUBDIRS += plugins/omkafka
 endif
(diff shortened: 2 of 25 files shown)


davidelang:
could you please check (and possibly create a test) if this works using "!" as the separator rather than just '.'

there are a few coding style fixes to make. @rgerhards this is complaining about it being three commits, but this looks like a fairly reasonable split of commits to me. Also, are the Solaris build issues a problem with these changes or the solaris enviornment?

I also wonder if there is value in allowing multi-character separators (partly wondering about unicode character points) but even without that, this is a useful addition.

tests/Makefile.am
```diff
@@ -1054,6 +1054,28 @@ TESTS +=  \
 	rscript_hash64-vg.sh
 endif # ENABLE_FMHASH
 endif # HAVE_VALGRIND
+if ENABLE_FMUNFLATTEN
```
rgerhards, line 1057: The tests need also be given in EXTRA_DIST, that is why `make distcheck` fails.

rgerhards (changes requested):

rgerhards:
> this is complaining about it being three commits, but this looks like a fairly reasonable split of commits to me. 

This is a generic check which also serves as a warning to me. The current commit structure is fine with me, so this is not a show blocker (but in most other cases, it guards against very bad commit structure, that's why we do the check).

> Also, are the Solaris build issues a problem with these changes or the solaris enviornment?

It's a problem with the changes, as they do not take care of Solaris specialties. So this needs to be fixed.

julthomas:
Hello,
I have pushed an extra test using a "!" as separator, added the test scripts to EXTRA_DIST and swaped the config.h to fix the solaris build. Looks like there is still an issue with the distcheck test, @rgerhards do you have an idea? make check passes the tests okay on my system. Also I wonder if is possible to install jq in the CI container.
These are new commits added on top of the firsts, to ease review. We can do fix-up once everything is okay or at merge time.
Julien

davidelang:
the fact that the log shows a jq syntax error message says that jq is installed, it could potentially be a version of jq issue.

note that there are also a handful of overly long lines being complained about with the codestyle check.

davidelang:
ubuntu 20 has jq 1.6 it looks like most of the other distros have 1.5 (fedora 33 should be 1.6)

julthomas:
@davidelang I don't think jq is in the container where the tests used to failed (the tests were skiped because jq was not in $PATH). I think I've seen the message you mentioned but it was related to CI stuff in yaml. Actually the failure was due to the valgrind tests I missed (again...) to add in EXTRA_DIST.

Regarding jq, we would need to rebuild the docker image used in tests to include jq. Also adding apt-get install -y jq in the script ran from the container just felt wrong. I ended up removing the jq stuff and adding a notes "this test may need changes to produce a more deterministic output by sorting keys" (I am not sure if libfastjson guarantees anything about the order of keys when producing json output, but I found out the order of the "sets" makes a difference).

I could shorten the long lines (warnings) but it I believe it would produce something horrible to read. I knew that check would fail ;) Of course I will make the changes if required though!

@rgerhards I've added rsyslog/rsyslog-doc#922 for the documentation.

rgerhards:
@julthomas The line length is non-negotiable - it's terrible to read this on a smaller terminal, which is sometimes necessary. It's very easy to make it short and readable, e.g. 
```
DBGPRINTF("warning: flat key \"%.20s...\" truncated at depth #%d, buffer too small (max %zd)\n", key, depth, ctx->kbuf_len);
```
just becomes
```
DBGPRINTF("warning: flat key \"%.20s...\" truncated at depth #%d, buffer too "
     "small (max %zd)\n", key, depth, ctx->kbuf_len);
```
Also, if I would merge, all PRs from now on would fail ;-)

julthomas:
@rgerhards OK ;-) will be back

julthomas:
@rgerhards I've pushed the changes ;)

davidelang:
looking better. Ubuntu 20 is having trouble, but it doesn't look to be related 
to this PR. I just kicked off the tests again in the hope that it runs cleaner 
this time.

it still complains about there being multiple commits, but we've discussed that 
and it won't block the merge.

just need to understand what's going wrong with that one build.

David Lang

davidelang:
re-running the ubuntu 20 build worked, the solaris build is still showing pending and looks like it never started.  I don't have the ability to re-run that

julthomas:
OK thank you @davidelang. How do you retrigger the tests without pushing on the branch attached to the PR?

rgerhards:
> How do you retrigger the tests without pushing on the branch attached to the PR?

 you need admin permissions :-(
 
 On Solaris: this build is provided by openCSW. Looks like an issue there. Will contact them.

julthomas:
Hello @rgerhards, thanks for the merge. I think you may close rsyslog/liblognorm#105 so that it makes one less issue in that repository ;) Best regards, Julien.

OrangeDog:
Can I ask for usage clarification here?
How (assuming you can) would you use this in an output template?
For example, unflattening the `host.name` key in this one:
```
template(name="json-template" type="list" option.jsonf="on") {
    constant(outname="@version" value="1" format="jsonf")
    property(outname="@timestamp" name="timereported" dateFormat="rfc3339" format="jsonf")
    constant(outname="type" value="rsyslog" format="jsonf")
    property(outname="message" name="msg" position.from="2" format="jsonf")
    property(outname="host.name" name="hostname" format="jsonf")
    property(outname="severity" name="syslogseverity-text" format="jsonf")
    property(outname="facility" name="syslogfacility-text" format="jsonf")
    property(outname="programname" name="programname" format="jsonf")
    property(outname="procid" name="procid" format="jsonf")
}
```

PR #4553, opened 2021-03-10
testbench: Remove python shebang in tests/testsuites

yifan19:
py files should be run as ${PYTHON} *.py

reference issue: #3853

Signed-off-by: Yi Fan Yu <yifan.yu@windriver.com>


rgerhards:
Thx for the PR!

PR #4554, opened 2021-03-11
testbench: Skip tests that need logger -d for busybox

yifan19:
Busybox logger doesn't have the -d flag.

Skip instead of failing the tests that need -d.
Also fix typo s/-p/-d.

busybox reference:
https://busybox.net/downloads/BusyBox.html#logger

Signed-off-by: Yi Fan Yu <yifan.yu@windriver.com>


yifan19:
seems like it is skipping tests without busybox... checking...

rgerhards:
thx for the PR!

PR #4555, opened 2021-03-12
Update plugin.py to append to a file

prslnx:
Make skeleton plugin append to a file instead of truncating it.

I lost 2 days looking why omprog not work, almost give up, because i was running a tail in /tmp/logfile and didn´t was seen any messages at all.
After running rsyslog in debug mode i saw omprog calling plugin.py, so i double checked it and after change to append all worked fine.


rgerhards:
Thx for the PR

Issue #4271, opened 2020-05-06
inconsistent include behavior


davidelang:
### Expected behavior
include files in alpha sort order
### Actual behavior
it works differently on centos 6 vs centos 7 (also a ubuntu of some unknown version)
in my examples, the result ends up correct, but an inconsistancy like this (processing file by file vs recursive-like processing) worries me

### Environment
8.2001

on centos 6

# cat include1
# include1

# cat include2
#include2

# cat test.conf
$IncludeConfig ./include*

# rsyslogd -o test.result -f test.conf
rsyslogd: there are no active actions configured. Inputs would run, but no output whatsoever were created. [v8.2001.0 try https://www.rsyslog.com/e/2103 ]
rsyslogd: run failed with error -2103 (see rsyslog.h or try https://www.rsyslog.com/e/2103 to learn what that number means)

# cat test.result
##### BEGIN CONFIG: test.conf
$IncludeConfig ./include*./include*
##### BEGIN CONFIG: ./include1
# include1

##### END   CONFIG: ./include1
#include2

##### END   CONFIG: test.conf

on centos 7

# cat include1
#1

# cat include2
#2

# cat test.conf
$IncludeConfig include*

# rsyslogd -f test.conf -o test.results
rsyslogd: there are no active actions configured. Inputs would run, but no output whatsoever were created. [v8.2002.0 try https://www.rsyslog.com/e/2103 ]
rsyslogd: run failed with error -2103 (see rsyslog.h or try https://www.rsyslog.com/e/2103 to learn what that number means)
[root@s8392cplv ~]# cat test.results
##### BEGIN CONFIG: test.conf
$IncludeConfig include*include*
##### BEGIN CONFIG: include1
#1

##### END   CONFIG: include1
#2

rgerhards:
I guess this is a problem with the `-o'  option. It possibly takes the input buffer too early. The actual order is correct (99.99999% sure). Will verify.

davidelang:
I initially thought I had a sample that included files in the wrong order 
(resulting in a stop happening at the wrong place) but was not able to reproduce 
it

I started looking at this because things weren't behaving properly on one 
machine, and it looked like the include order was wrong (20*.conf after 
35*.conf) but the real problem ended up being elsewhere.

David Lang

alorbach:
I think the reverse read order of the configuration files is done intentional after reading the comments in the code: 
[...]

PR #4557, opened 2021-03-17
clarify some debug output in regard to config include load order

rgerhards:
The previous text seemed to cause the wrong impression that includes
were processed in the wrong order. Right the contrary is the case, as
config files are being put on a stack before they are processed.

closes https://github.com/rsyslog/rsyslog/issues/4271


rgerhards:
mmmhhh... looking deeper that the test and creating more instrumentation, I wonder why the impression of a wrong order was created. To me it looks obvious (in -o output) that the order is right. Am I overlooking something? @davidelang

PR #4566, opened 2021-04-06
omelastisearch: allow omitting deprecated _type field

### LINKED ISSUE #4387: Deprecation warning when using omelasticsearch with Elasticsearch 7.x (will break with ES 8.x)
### Expected behavior
I expect no warning in ``/var/log/elasticsearch/cluster-es_deprecation.log``

### Actual behavior
```
[2020-08-24T10:28:51,375][WARN ][o.e.d.a.b.BulkRequestParser] [node-es] [types removal] Specifying types in bulk requests is deprecated.
```
Note: the warning only appears once after Elasticsearch is restarted. However, the real issue is that starting with Elasticsearch 8.x, ``types`` will not be allowed at all, meaning that this will trigger fatal errors. See https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html#_schedule_for_removal_of_mapping_types for more information.

* Install Elasticsearch 7.x
* Use the following rsyslog configuration:
```
template(name="logTemplate" type="list"){
  constant(value="{" )
  constant(value="\"timereported\":\"")    property(name="timereported" dateFormat="rfc3339"  mandatory="on")
  constant(value="\",\"hostname\":\"")      property(name="hostname" format="json")
  constant(value="\",\"message\":\"")   property(name="msg" format="json" )
  constant(value="\"}")
}

   action(type="omelasticsearch"
          server="127.0.0.1"
          serverport="9200"
          usehttps="off"
          template="logTemplate"
          searchIndex="logs"
          # Note: removing the following line still triggers the warning because the value then defaults to "events"
          searchType="_doc" # <-- This line triggers the deprecation warning
          dynSearchIndex="off"
          bulkmode="on"
   )
```

### Environment
- rsyslog version: 8.2006.0
- elasticsearch version: 7.8.1
- platform: CentOS 7.8.2003 (64 bits)

See https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html for more information about types removal and how it impacts the various Elasticsearch APIs.

oranenj:
Fixes #4387

This patch allows you to explicitly set searchType="", which causes it to be omitted from the JSON request as well as the bulk API url.

I didn't yet test with an actual Elasticsearch server, but testing with just dumping POST requests onto stdout with a custom template, I can't see immediate problems with the output

in bulk mode:
```
{"index":{"_index": "rsyslog-write-default"}}
{"message":"test","fromhost":"localhost","facility":"user","priority":"notice","timereported":"2021-04-06T21:47:38.739697+03:00","timegenerated":"2021-04-06T21:47:38.739815+03:00"}
```
vector tells me the path is `/_bulk`

and in non-bulk mode the path is `/rsyslog-write-default` and the message does not change

Omitting the parameter restores the default behaviour:
```
{"index":{"_index": "rsyslog-write-default","_type":"events"}}
```
and path `/rsyslog-write-default/events` for non-bulk.

### DIFF
--- a/plugins/omelasticsearch/omelasticsearch.c
+++ b/plugins/omelasticsearch/omelasticsearch.c
@@ -623,7 +623,7 @@ setPostURL(wrkrInstanceData_t *const pWrkrData, uchar **const tpls)
 	/* since 7.0, the API always requires /idx/_doc, so use that if searchType is not explicitly set */
 	uchar* actualSearchType = (uchar*)"_doc";
 	es_str_t *url;
-	int r;
+	int r = 0;
 	DEFiRet;
 	instanceData *const pData = pWrkrData->pData;
 	char separator;
@@ -646,13 +646,12 @@ setPostURL(wrkrInstanceData_t *const pWrkrData, uchar **const tpls)
 		getIndexTypeAndParent(pData, tpls, &searchIndex, &searchType, &parent, &bulkId, &pipelineName);
 		if(searchIndex != NULL) {
 			r = es_addBuf(&url, (char*)searchIndex, ustrlen(searchIndex));
+			if(searchType != NULL && searchType[0] != '\0') {
+				actualSearchType = searchType;
+			}
 			if(r == 0) r = es_addChar(&url, '/');
-
-		if(searchType != NULL) {
-			actualSearchType = searchType;
+			if(r == 0) r = es_addBuf(&url, (char*)actualSearchType, ustrlen(actualSearchType));
 		}
-		if(r == 0) r = es_addChar(&url, '/');
-		if(r == 0) r = es_addBuf(&url, (char*)actualSearchType, ustrlen(actualSearchType));
 		if(pipelineName != NULL && (!pData->skipPipelineIfEmpty || pipelineName[0] != '\0')) {
 			if(r == 0) r = es_addChar(&url, separator);
 			if(r == 0) r = es_addBuf(&url, "pipeline=", sizeof("pipeline=")-1);
@@ -714,7 +713,11 @@ computeMessageSize(const wrkrInstanceData_t *const pWrkrData,
 		r += ustrlen(searchIndex);
 	}
 	if(searchType != NULL) {
-		r += ustrlen(searchType);
+		if(searchType[0] == '\0') {
+			r += 4; // "_doc"
+		} else {
+			r += ustrlen(searchType);
+		}
 	}
 	if(parent != NULL) {
 		r += sizeof(META_PARENT)-1 + ustrlen(parent);
@@ -759,7 +762,7 @@ buildBatch(wrkrInstanceData_t *pWrkrData, uchar *message, uchar **tpls)
 			if(r == 0) r = es_addBuf(&pWrkrData->batch.data, META_IX, sizeof(META_IX)-1);
 		if(r == 0) r = es_addBuf(&pWrkrData->batch.data, (char*)searchIndex,
 				 ustrlen(searchIndex));
-		if(searchType != NULL) {
+		if(searchType != NULL && searchType[0] != '\0') {
 			if(r == 0) r = es_addBuf(&pWrkrData->batch.data, META_TYPE, sizeof(META_TYPE)-1);
 			if(r == 0) r = es_addBuf(&pWrkrData->batch.data, (char*)searchType,
 				 ustrlen(searchType));
--- a/plugins/omelasticsearch/omelasticsearch.c
+++ b/plugins/omelasticsearch/omelasticsearch.c
@@ -620,6 +620,8 @@ setPostURL(wrkrInstanceData_t *const pWrkrData, uchar **const tpls)
 	uchar *parent;
 	uchar *bulkId;
 	char* baseUrl;
+	/* since 7.0, the API always requires /idx/_doc, so use that if searchType is not explicitly set */
+	uchar* actualSearchType = (uchar*)"_doc";
 	es_str_t *url;
 	int r;
 	DEFiRet;
@@ -645,11 +647,12 @@ setPostURL(wrkrInstanceData_t *const pWrkrData, uchar **const tpls)
 		if(searchIndex != NULL) {
 			r = es_addBuf(&url, (char*)searchIndex, ustrlen(searchIndex));
 			if(r == 0) r = es_addChar(&url, '/');
-			if(searchType != NULL) {
-				if(r == 0) r = es_addBuf(&url, (char*)searchType, ustrlen(searchType));
-			}
-		} else
-			r = 0;
+
+		if(searchType != NULL) {
+			actualSearchType = searchType;
+		}
+		if(r == 0) r = es_addChar(&url, '/');
+		if(r == 0) r = es_addBuf(&url, (char*)actualSearchType, ustrlen(actualSearchType));
 		if(pipelineName != NULL && (!pData->skipPipelineIfEmpty || pipelineName[0] != '\0')) {
 			if(r == 0) r = es_addChar(&url, separator);
 			if(r == 0) r = es_addBuf(&url, "pipeline=", sizeof("pipeline=")-1);
@@ -693,7 +696,7 @@ computeMessageSize(const wrkrInstanceData_t *const pWrkrData,
 	const uchar *const message,
 	uchar **const tpls)
 {
-	size_t r = sizeof(META_TYPE)-1 + sizeof(META_END)-1 + sizeof("\n")-1;
+	size_t r = sizeof(META_END)-1 + sizeof("\n")-1;
 	if (pWrkrData->pData->writeOperation == ES_WRITE_CREATE)
 		r += sizeof(META_STRT_CREATE)-1;
 	else
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -104,6 +104,7 @@ TESTS +=  \
 	es-basic-errfile-popul.sh \
 	es-bulk-errfile-empty.sh \
 	es-bulk-errfile-popul.sh \
+	es-searchType-empty.sh \
 	diskqueue-multithread-es.sh \
 	es-writeoperation.sh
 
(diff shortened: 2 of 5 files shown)


plugins/omelasticsearch/omelasticsearch.c
```diff
@@ -2092,6 +2099,12 @@ CODESTARTnewActInst
 	if(pData->searchType == NULL)
 		pData->searchType = (uchar*) strdup("events");
 
+	/* if the search type is explicitly set to an empty string, omit it entirely */
+	if(!strcmp((char*)pData->searchType, "")) {
+		free(pData->searchType);
```
oranenj: self-reviewing this bit, is freeing the dummy value here correct? I'm not a 100% sure on the ownership semantics here.

rgerhards: looks good to me

rgerhards (changes requested):
We should have at least one test for this in the testbench. I am happy to help with this. Please let me know the minimum ES version we need to tests it. A working config sample would be helpful.

If you like to dig yourself a bit into it, I am looking at least for a basic test like here: https://github.com/rsyslog/rsyslog/blob/master/tests/es-basic-es6.0.sh

rgerhards:
FYI: milestone tentatively assigned - may be for the June release actually (depending on how fast we can move forward).

rgerhards:

rgerhards:
Note to myself: code looks good except for the missing test

oranenj:
@rgerhards the deprecation is described here:
https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html

Reading that, this code isn't actually completely correct; After 7.0, it looks like you need to use index/_doc when POSTing to an index outside of bulk mode, and to avoid the deprecation you need to omit the _type field when using bulk mode.

The "cleanest" option would be to remove the type altogether, but that's not a backwards compatible change and is likely to break people's setups by accident :/

I'll fix non-bulk indexing and take a look at documenting this.

oranenj:
The code should now be correct for non-bulk indexing as well, though I still haven't had time for proper testing with an actual Elasticsearch instance.

oranenj:
I'm not sure what happened with that one build failure. Looks like it just randomly failed for an unrelated reason.

I'll squash the commits after review.

oranenj:
I added a dumb test for any deprecation messages with 7.10.2 (the last OSS elasticsearch version). It's not actually enabled, but it allows me to locally verify that things are working.

Running it requires https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-oss-7.10.2-linux-x86_64.tar.gz

Output looks like this:
```
$ ./es-basic-es7.0.sh 
testbench: TZ env var not set, setting it to UTC
------------------------------------------------------------
12:44:52[0]  Test: ./es-basic-es7.0.sh
------------------------------------------------------------
cat: es.pid: No such file or directory
cmp: elasticsearch.running: No such file or directory
cat: elasticsearch.running: No such file or directory
rm: cannot remove 'elasticsearch.running': No such file or directory
rm: cannot remove 'elasticsearch.running': No such file or directory
TEST USES ELASTICSEARCH BINARY /home/oranenj/git/rsyslog/tests/.dep_cache/elasticsearch-7.10.2.tar.gz
Setting transport tcp option to transport.port
Creating elastic search directories
ElasticSearch prepared for use in test.
Starting ElasticSearch
elasticsearch pid is 1898707
curl: (7) Failed to connect to localhost port 19200: Connection refused
--- waiting for ES startup: 0 seconds
curl: (7) Failed to connect to localhost port 19200: Connection refused
--- waiting for ES startup: 1 seconds
curl: (7) Failed to connect to localhost port 19200: Connection refused
--- waiting for ES startup: 2 seconds
curl: (7) Failed to connect to localhost port 19200: Connection refused
--- waiting for ES startup: 3 seconds
curl: (7) Failed to connect to localhost port 19200: Connection refused
--- waiting for ES startup: 4 seconds
ES startup succeeded
running elasticsearch instance: elasticsearch-7.10.2.tar.gz:19200:1898707
{"error":{"root_cause":[{"type":"index_not_found_exception","reason":"no such index [rsyslog_testbench]","resource.type":"index_or_alias","resource.id":"rsyslog_testbench","index_uuid":"_na_","index":"rsyslog_testbench"}],"type":"index_not_found_exception","reason":"no such index [rsyslog_testbench]","resource.type":"index_or_alias","resource.id":"rsyslog_testbench","index_uuid":"_na_","index":"rsyslog_testbench"},"status":404}config rstb_072725_d76c1a8bTMCf_.conf is:
     1	module(load="../plugins/imdiag/.libs/imdiag")
     2	global(inputs.timeout.shutdown="60000"
     3	       default.action.queue.timeoutshutdown="20000"
     4	       default.action.queue.timeoutEnqueue="20000")
     5	# use legacy-style for the following settings so that we can override if needed
     6	$MainmsgQueueTimeoutEnqueue 20000
     7	$MainmsgQueueTimeoutShutdown 10000
     8	$IMDiagListenPortFileName rstb_072725_d76c1a8bTMCf.imdiag.port
     9	$IMDiagServerRun 0
    10	$IMDiagAbortTimeout 580
    11	
    12	:syslogtag, contains, "rsyslogd"  ./rstb_072725_d76c1a8bTMCf.started
    13	###### end of testbench instrumentation part, test conf follows:
    14	
    15	template(name="tpl" type="string"
    16		 string="{\"msgnum\":\"%msg:F,58:2%\"}")
    17	
    18	module(load="../plugins/omelasticsearch/.libs/omelasticsearch")
    19	
    20	if $msg contains "msgnum:" then
    21		action(type="omelasticsearch"
    22		       server="127.0.0.1"
    23		       serverport="19200"
    24		       searchType=""
    25		       template="tpl"
    26		       searchIndex="rsyslog_testbench")
rsyslogd: NOTE: RSYSLOG_DEBUG_TIMEOUTS_TO_STDERR activated
main Q:Reg: worker start requested, num workers currently 0
main Q:Reg: wrkr start initiated with state 0, num workers now 1
rsyslog debug: main Q:Reg: worker 0x23a9780 started
rsyslog debug: main Q:Reg: started with state 3, num workers now 1
12:45:04[12]  rstb_072725_d76c1a8bTMCf:.pid found, pid 1898832
12:45:04[12]  rsyslogd startup msg seen, pid  1898832
waiting for file rstb_072725_d76c1a8bTMCf.imdiag.port
imdiag port: 40123
injecting 2000 messages
imdiag[40123]: 2000 messages injected
Shutting down instance 1
imdiag: wait q_empty: qsize 2000 nempty 0
imdiag: wait q_empty: qsize 1982 nempty 0
imdiag: wait q_empty: qsize 1726 nempty 0
imdiag: wait q_empty: qsize 1470 nempty 0
imdiag: wait q_empty: qsize 1214 nempty 0
imdiag: wait q_empty: qsize 958 nempty 0
imdiag: wait q_empty: qsize 702 nempty 0
imdiag: wait q_empty: qsize 446 nempty 0
imdiag: wait q_empty: qsize 190 nempty 0
imdiag: wait q_empty: qsize 0 nempty 1
imdiag[40123]: mainqueue empty
{"acknowledged":true}12:45:07[15]  es_shutdown_empty_check: success, have 2000 lines
rsyslogd debug: info: trying to cooperatively stop input ../plugins/imdiag/.libs/imdiag, timeout 60000 ms
rsyslog debug: main Q:Reg/w0: enter WrkrExecCleanup
rsyslog debug: 0x23a6bf0: worker exiting
rsyslog debug: main Q:Reg/w0: thread joined
12:45:08[16]  wait on shutdown of 1898832
{"acknowledged":true}12:45:08[16]  Test ./es-basic-es7.0.sh SUCCESSFUL (took 16 seconds)
-------------------------------------------------------------------------------
```
If the searchType parameter is omitted, the deprecation log check fails

oranenj:
@rgerhards what's needed to get the appropriate test integrated into the test system? I don't have access to rsyslog.com to upload another version of elasticsearch for the tests to use. I'm also not sure what the preferred way of structuring new tests is.

rgerhards:
sorry... I somehow lost track (strange times ATM...)

Let me see that I get the ES versions inside the testbench bumped, than we can look further. If I am kind of too late for you, I'll see how I can pursue this on my own. Assistance, of course, would be appreciated.

rgerhards:
@oranenj FYI: CI environment is now ready for the new test. I'll wait a day or two and then pursue finishing the PR :-)

oranenj:
@rgerhards I can completely understand issues getting "lost" when you're busy. Happens to me constantly too. Fortunately for us this issue merely causes a lot of log spam that we can at least work around, so it hasn't been a showstopper.

I don't think I'll have much time to put into this right now, so go ahead if you're not too busy.

rgerhards:
@oranenj OK, no problem, thx for the reply :-) I'll see that I get it integrated. I let you know if I run into a question. Else you'll see a merge :-)

rgerhards:
@oranenj if you have a minute, I would appreciate if you could have a look at e.g. the "check elasticsearch" run. It looks like the code connects to port 9300, while 9200 is configured. I admit right now I have no idea of why this happens. Will also try to find cause, but would appreciate comments.

oranenj:
@rgerhards The  "check elasticsearch" test seems to fail because the test is not configuring the "xpack.security.enabled=false" option explicitly and that causes a deprecation warning, erroring out because the test just checks for *any* deprecation.

rgerhards:
> @rgerhards The "check elasticsearch" test seems to fail because the test is not configuring the "xpack.security.enabled=false" option explicitly and that causes a deprecation warning, erroring out because the test just checks for _any_ deprecation.

Thx for the tip. I think I have now disabled it, let's see how it works out...

rgerhards:
@oranenj that setting seems not to be the problem. Can you check if it is correct? Also, I would appreciate if you could re-push so that CI starts anew. I am a bit out of ideas here.

rgerhards:
@oranenj sry, I screwd up once more (if bad things happen, they continue to do...).

I merge https://github.com/rsyslog/rsyslog/issues/4387 which brings, at it's core, most of the same change. I would still like to get the new test plus ES 7 functionality, so I try to fix my mistake in that PR here.

rgerhards (approved):

PR #4471, opened 2020-11-22
Add built-in function get_property() to access property vars

n2yen:
Provides ability to evaluate a rsyslog variable using dynamically
evaluated parameters.
  1st param is the rsyslog param, 2nd param is a key, can be an array
  index or key string.

Useful for accessing json sub-objects, where a key
needs to be evaluated at runtime. Can be used to access arrays as well.

see tests for examples

PR #4571, opened 2021-04-15
Add built-in function get_property() to access property vars

rgerhards:
Provides ability to evaluate a rsyslog variable using dynamically
evaluated parameters.
  1st param is the rsyslog param, 2nd param is a key, can be an array
  index or key string.

Useful for accessing json sub-objects, where a key
needs to be evaluated at runtime. Can be used to access arrays as well.

see tests for examples

remove unnecessary escape char


rgerhards:
sry for the long silence - have rebased to current master (also to get rid of CI issue), see #4571

rgerhards:
merged via original PR #4471

PR #4572, opened 2021-04-15
imjournal: flush buffer before fsync

gerd-rausch:
Flush the FILE* buffer before rename & fsync in order
to not end up syncing an empty file.

Also, close WorkDir on fsync in order to prevent
file descriptor leakage.

Signed-off-by: Gerd Rausch <gerd.rausch@oracle.com>
Signed-off-by: Venu Busireddy <venu.busireddy@oracle.com>


rgerhards (approved):

rgerhards:
Thx for the PR!

Issue #4567, opened 2021-04-12
omhttp dynrestpath fails when batching is enabled


orion84:
### Expected behavior
I have configured an omhttp action to send events to Splunk. This is working fine until I started testing the use of a templated restpath using dynrestpath="on" (See code below). 

### Actual behavior
When setting dynrestpath="off" and setting the restpath directly using the exact same string from the template it works. When using the templated approach, I get below error:

omhttp: checkResult error http status code: 404 reply: {"text":"The requested URL was not found on this server.","code":404} [v8.2102.0 try https://www.rsyslog.com/e/2218 ]

Temporarily disabling ssl, I see the following in tcpdump:
` POST /restpathhecraw HTTP/1.1`

Which indicates that it is ignoring the `dynrestpath = "on"` and just takes the template name as the restpath.

### Steps to reproduce the behavior
relevant parts of the config:
```
template(name="restpathhecraw" type="string" string="services/collector/raw?sourcetype=syslog&source=centos7_omhttp&host=centos7")

ruleset(name="tcp9001-omhttptest-HEC"){
	action(
		name="tcp9001-omhttptest-HEC"
		type="omhttp"
		...
		dynrestpath = "on"
		restpath="restpathhecraw"
		...
```

### Environment
- rsyslog version: 8.2102.0
- platform: Centos 7

alorbach:
I think you need to specify the restpathhecraw as template parameter: 

> template(name="restpathhecraw" type="string" string="services/collector/raw?sourcetype=syslog&source=centos7_omhttp&host=centos7")
> 
> ruleset(name="tcp9001-omhttptest-HEC"){
> 	action(
> 		name="tcp9001-omhttptest-HEC"
> 		type="omhttp"
> 		...
> 		dynrestpath = "on"
> 		template="restpathhecraw"
> 		...

Thats where omhttp is expected the template when dynrestpath is set to on.

orion84:
@alorbach the template parameter is for the message formatting, not for the dynamic rest path, right? It also says in the documentation:

> dynrestpath
> When this parameter is set to “on” you can specify a template name in the parameter **restpath** instead of the actual path.
(emphasis mine)

You can also see from the `POST /restpathhecraw HTTP/1.1` that I captured of the network that it tries to connect to an endpoint as per the value of the restpath parameter. It just fails to interpret that as a reference to a template rather than a literal value.

If it would be using whatever is specified in the template parameter (which I do also use for formatting the message) I would get very different results.

[...]

PR #4575, opened 2021-04-20
omhttp: Fix dynrestpath param in batch mode

alorbach:
Modified omhttp: Fix dynrestpath param in batch mode

When batchmode was used, the templates could not be used to    
expand dynrestpath. We are now storing the restpath param 
within the batch data if we are in batch mode.

- testbench: Added tests for omhttp dynrestpath param
  Testing setting the restpath by template - one with batch mode
  and one without batch mode

closes: https://github.com/rsyslog/rsyslog/issues/4567

### DIFF
--- a/contrib/omhttp/omhttp.c
+++ b/contrib/omhttp/omhttp.c
@@ -171,9 +171,11 @@ typedef struct wrkrInstanceData {
 	sbool bzInitDone;
 	z_stream zstrm; /* zip stream to use for gzip http compression */
 	struct {
-		uchar **data; /* array of strings, this will be batched up lazily */
-		size_t sizeBytes; /* total length of this batch in bytes */
-		size_t nmemb;	/* number of messages in batch (for statistics counting) */
+		uchar **data;		/* array of strings, this will be batched up lazily */
+		uchar *restPath;	/* Helper for restpath in batch mode */
+		size_t sizeBytes;	/* total length of this batch in bytes */
+		size_t nmemb;		/* number of messages in batch (for statistics counting) */
+
 	} batch;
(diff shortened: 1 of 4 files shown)


gintrator:
@alorbach Note that will lead to unexpected results if the dynamic path is based on a message attribute that is inconsistent among messages in the same batch.

To correctly do dynamic paths with batching, we'll need to "key" batches internally by the dynamic path and only post the messages that map to the same key to that path. 

Otherwise, everything in a batch will get posted to whatever the last message's dynamic path evaluated to, which is pretty similar to a static path configured for the whole action. In theory, every message in a batch could resolve to a different dynamic path.

rgerhards:
At least for a quick fix, we could (and probably should for a first try) check if the dynpath changes and, if so, terminate the previous batch at this point, commit, start a new one and continue as described.

alorbach:
I see what you mean. At the moment how the action is build, it is simple not possible to have unique dynrestpathes for each message in the batch. It would require some more rework on that action in order to split the batch and submit the messages based on the dynrestpath as key.

I think the approach from @rgerhards can be done within this PR and would be possible to implement quickly.

rgerhards:
I would strongly opt to have it fixed quickly, even though this can cause performance issues (that's the same as in other spots). Of course, we should add a warning to the doc. If we see complaints, we can look at a better method.

alorbach:
@gintrator I changed the PR as recommended by Rainer - the batch will be finalized and restarted once the dynpath changes. Can you confirm that this solution is working for you?

gintrator:
Looks like it produces correct behavior, good by me!

PR #4578, opened 2021-04-23
imtcp refactoring

rgerhards:
another try at supporting inputs with different stream drivers as well as getting code cleaner

see also https://github.com/rsyslog/rsyslog/issues/3727#issuecomment-525705318


rgerhards:
will use https://github.com/rsyslog/rsyslog/pull/4582 instead of this PR - better method

PR #4592, opened 2021-05-15
Include programname in a new JSON template

artbristol:
The default JSON template does not include the programname.
This makes it difficult to know which program wrote which log entry.

### DIFF
--- a/runtime/rsconf.c
+++ b/runtime/rsconf.c
@@ -115,6 +115,15 @@ static uchar template_StdJSONFmt[] = "\"{\\\"message\\\":\\\"%msg:::json%\\\",\\
 "%HOSTNAME:::json%\\\",\\\"facility\\\":\\\"%syslogfacility-text%\\\",\\\"priority\\\":\\\""
 "%syslogpriority-text%\\\",\\\"timereported\\\":\\\"%timereported:::date-rfc3339%\\\",\\\"timegenerated\\\":\\\""
 "%timegenerated:::date-rfc3339%\\\"}\"";
+static uchar template_FullJSONFmt[] = "\"{\\\"message\\\":\\\"%msg:::json%\\\","
+"\\\"fromhost\\\":\\\"%HOSTNAME:::json%\\\","
+"\\\"programname\\\":\\\"%programname%\\\","
+"\\\"procid\\\":\\\"%PROCID%\\\","
+"\\\"msgid\\\":\\\"%MSGID%\\\","
+"\\\"facility\\\":\\\"%syslogfacility-text%\\\","
+"\\\"priority\\\":\\\"%syslogpriority-text%\\\","
+"\\\"timereported\\\":\\\"%timereported:::date-rfc3339%\\\","
+"\\\"timegenerated\\\":\\\"%timegenerated:::date-rfc3339%\\\"}\"";
 static uchar template_StdClickHouseFmt[] = "\"INSERT INTO rsyslog.SystemEvents (severity, facility, "
 "timestamp, hostname, tag, message) VALUES (%syslogseverity%, %syslogfacility%, "
 "'%timereported:::date-unixtimestamp%', '%hostname%', '%syslogtag%', '%msg%')\",STDSQL";
@@ -1405,6 +1414,8 @@ initLegacyConf(void)
 	tplAddLine(ourConf, " StdPgSQLFmt", &pTmp);
 	pTmp = template_StdJSONFmt;
 	tplAddLine(ourConf, " StdJSONFmt", &pTmp);
+	pTmp = template_FullJSONFmt;
+	tplAddLine(ourConf, " FullJSONFmt", &pTmp);
 	pTmp = template_StdClickHouseFmt;
 	tplAddLine(ourConf, " StdClickHouseFmt", &pTmp);
 	pTmp = template_spoofadr;


davidelang:
the programname should be part of the syslogtag (or in the rfc5424 headers), is 
that not the case for your logs?

could you show us a sample log message?

David Lang

artbristol:
I'm using omelasticsearch. With the default configuration (no template configured so it uses the default JSON one) a log message looks like this in Elasticsearch

```
{
  "_index": "system",
  "_type": "events",
  "_id": "jdr_bXkBNi6M-_cW0TEk",
  "_version": 1,
  "_score": null,
  "fields": {
    "facility.keyword": [
      "auth"
    ],
    "timegenerated": [
      "2021-05-15T03:09:12.342Z"
    ],
    "message.keyword": [
      " Received disconnect from 103.136.42.230 port 51208:11: Bye Bye [preauth]"
    ],
    "fromhost.keyword": [
      "storage"
    ],
    "fromhost": [
      "storage"
    ],
    "message": [
      " Received disconnect from 103.136.42.230 port 51208:11: Bye Bye [preauth]"
    ],
    "priority": [
      "info"
    ],
    "timereported": [
      "2021-05-15T03:09:12.342Z"
    ],
    "facility": [
      "auth"
    ],
    "priority.keyword": [
      "info"
    ]
  },
  "highlight": {
    "fromhost": [
      "@kibana-highlighted-field@storage@/kibana-highlighted-field@"
    ]
  },
  "sort": [
    1621048152342
  ]
}
```
This message is from `sshd` but you can't tell that.

I used a custom template and added the programname to get messages that look like this in Elasticsearch

```
{
  "_index": "system",
  "_type": "events",
  "_id": "39rzbXkBNi6M-_cWSSLk",
  "_version": 1,
  "_score": null,
  "fields": {
    "facility.keyword": [
      "auth"
    ],
    "timegenerated": [
      "2021-05-15T02:55:31.298Z"
    ],
    "programname.keyword": [
      "sshd"
    ],
    "message.keyword": [
      " Received disconnect from 178.128.125.217 port 54098:11: Bye Bye [preauth]"
    ],
    "programname": [
      "sshd"
    ],
    "fromhost.keyword": [
      "storage"
    ],
    "fromhost": [
      "storage"
    ],
    "message": [
      " Received disconnect from 178.128.125.217 port 54098:11: Bye Bye [preauth]"
    ],
    "priority": [
      "info"
    ],
    "timereported": [
      "2021-05-15T02:55:31.298Z"
    ],
    "facility": [
      "auth"
    ],
    "priority.keyword": [
      "info"
    ]
  },
  "highlight": {
    "fromhost": [
      "@kibana-highlighted-field@storage@/kibana-highlighted-field@"
    ]
  },
  "sort": [
    1621047331298
  ]
}
```

davidelang:
all you would need to do with the default template is add the line

set $!programname.keyword = $programname;

to set the variable in the $! namespace, and then it will get picked up by the 
default template.

David Lang

artbristol:
> all you would need to do with the default template is add the line set $!programname.keyword = $programname; to set the variable in the $! namespace, and then it will get picked up by the default template. David Lang

Thanks for the reply. (I've already solved the issue as I mentioned)

The reason I opened this PR is to make it easier for people new to rsyslog to be able to hook it up to Elasticsearch. Adding the line you mention - do you think this should be documented somewhere? (Or is it there already and I missed it?)

davidelang:
set is well documented, the details of the line will depend on what you want it 
to look like (programename.keyword is what you used in your example, so that's 
what I used in mine)

There are other properties that are not in the default $! set that someone may 
want added (hostname, timestamp, fromhost-ip, actually just about any of them 
_could_ be wanted), so there isn't any documentations specifying any of them in 
particular.

David Lang

artbristol:
I appreciate your taking the time to reply and sharing your expertise. Could you step back and look at this issue as a newcomer to rsyslog might?

Someone following the documentation here https://www.rsyslog.com/doc/v8-stable/configuration/modules/omelasticsearch.html, in particular the Example 1 section, will end up with logs that do not include the programname. In my opinion, these logs are much less useful without it, to the point of being unusable.

Compare the file log output that Rsyslog, also in the default configuration, writes to /var/log/auth.log for the same event:

```
2021-04-28T22:33:57.069302+01:00 tilca sshd[9432] Received disconnect from 103.136.42.230 port 60906:11: Bye Bye [preauth]
```

This includes the timestamp, hostname, programname, and message.

I think sane defaults matter a lot and either the documentation I link to should be updated (I'm happy to make the change) or preferably the default JSON logging output should include the programname.

rgerhards (changes requested):
breaks backward-compatiblility (see main discussion thread).

artbristol:
> The problem with this PR is that it breaks backwards-compatibility. We would need to define a new canned template for this and update the doc. @artbristol still any interest in doing that? Else I would consider taking your PR as basis and doing it myself.

Hi, yes thanks for the reply. I'm happy to define a new template and update the docs.

davidelang:
On Mon, 7 Feb 2022, Art O Cathain wrote:

>> The problem with this PR is that it breaks backwards-compatibility. We would need to define a new canned template for this and update the doc. @artbristol still any interest in doing that? Else I would consider taking your PR as basis and doing it myself.
>
> Hi, yes thanks for the reply. I'm happy to define a new template and update the docs.

While you are doing this, look to see what other properties are not in the 
existing template and let's see if we can include anything significant.

David Lang

rgerhards:
@artbristol interest in finishing this PR? Else, I'll close it unmerged (doing some cleanup right now). Thx!

artbristol:
@rgerhards I've moved the changes to the default template into a new template and updated the docs in the linked PR

rgerhards (approved):

PR #4601, opened 2021-06-10
bugfix: _sender_stats reports integer counter as string

rgerhards:


jay7x:
Any chance to have "origin" field there as well? I really hope nobody is parsing this JSON manually..

rgerhards:
I am looking at `origin` - I admit I must first re-check how this fits into the overall plumbing (this stat is indeed a bit different).

rgerhards:
I also added `origin` -- actually, that part of the code is not tied at all into the general framework, and so it ultimately was a simple change at the output line itself.

PR #4626, opened 2021-07-06
imfile add `ignoreolderthanoption`

n2yen:
Submit on behalf of the primary author @yanjunli76

instructs imfile not to ingest a file that has not been modified in the
specified number of seconds.

### DIFF
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -65,6 +65,7 @@
 #include "ratelimit.h"
 #include "srUtils.h"
 #include "parserif.h"
+#include "datetime.h"
 
 #include <regex.h>
 
@@ -90,6 +91,7 @@ DEFobjCurrIf(glbl)
 DEFobjCurrIf(strm)
 DEFobjCurrIf(prop)
 DEFobjCurrIf(ruleset)
+DEFobjCurrIf(datetime)
 
 extern int rs_siphash(const uint8_t *in, const size_t inlen, const uint8_t *k,
 	uint8_t *out, const size_t outlen); /* see siphash.c */
@@ -169,6 +171,7 @@ struct instanceConf_s {
 	sbool fileNotFoundError;
 	int maxLinesAtOnce;
 	uint32_t trimLineOverBytes;
+	uint32_t ignoreOlderThan;
 	int msgFlag;
 	uchar *escapeLFString;
 	ruleset_t *pBindRuleset;	/* ruleset to bind listener to (use system default if unspecified) */
@@ -343,6 +346,7 @@ static struct cnfparamdescr inppdescr[] = {
 	{ "freshstarttail", eCmdHdlrBinary, 0},
 	{ "filenotfounderror", eCmdHdlrBinary, 0},
 	{ "needparse", eCmdHdlrBinary, 0},
+	{ "ignoreolderthan", eCmdHdlrInt, 0},
 	{ "maxbytesperminute", eCmdHdlrInt, 0},
 	{ "maxlinesperminute", eCmdHdlrInt, 0}
 };
@@ -702,6 +706,22 @@ fs_node_print(const fs_node_t *const node, const int level)
 	}
 }
 
+static sbool
+isIgnoreOlderFile(const instanceConf_t *const inst, const char *const name)
+{
+	if (inst->ignoreOlderThan)
+	{
+		struct stat stat_buf;
+		time_t tt;
+		/* skip old files */
+		datetime.GetTime(&tt);
+		if (stat((char *)name, &stat_buf) == 0 && difftime(tt, stat_buf.st_mtime) > inst->ignoreOlderThan) {
+			return 1;
+		}
+	}
+	return 0;
+}
+
 /* add a new file system object if it not yet exists, ignore call
  * if it already does.
  */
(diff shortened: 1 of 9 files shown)


tests/Makefile.am
```diff
@@ -1464,6 +1464,14 @@ TESTS += \
...
 	imfile-logrotate-nocopytruncate.sh \
 	imfile-growing-file-id.sh \
+	imfile-ignore-old-file-0.sh \
+	imfile-ignore-old-file-1.sh \
+	imfile-ignore-old-file-2.sh \
+	imfile-ignore-old-file-3.sh \
+	imfile-ignore-old-file-4.sh \
+	imfile-ignore-old-file-5.sh \
+	imfile-ignore-old-file-6.sh \
+	imfile-ignore-old-file-7.sh
```
rgerhards: There is a '\\' missing at the end of the line. This causes the CI failure.

rgerhards (changes requested):

tests/Makefile.am
n2yen: ahh! thanks for the tip @rgerhards .

rgerhards:
sorry I didn't mention this earlier: please add doc to https://github.com/rsyslog/rsyslog-doc/

plugins/imfile/imfile.c
```diff
@@ -1855,6 +1883,7 @@ addInstance(void __attribute__((unused)) *pVal, uchar *pNewVal)
 		}
 	}
 	inst->trimLineOverBytes = cs.trimLineOverBytes;
+	inst->ignoreOlderThan = cs.ignoreOlderThan;
```
rgerhards: please remove support for the old config system - we do NOT provide new features via the old system.

plugins/imfile/imfile.c
```diff
@@ -2042,6 +2073,7 @@ CODESTARTbeginCnfLoad
 	cs.readMode = 0;
 	cs.maxLinesAtOnce = 10240;
 	cs.trimLineOverBytes = 0;
+	cs.ignoreOlderThan = 0;
```
rgerhards: please remove support for the old config system - we do NOT provide new features via the old system.

plugins/imfile/imfile.c
```diff
@@ -2846,6 +2879,7 @@ resetConfigVariables(uchar __attribute__((unused)) *pp, void __attribute__((unus
...
```
rgerhards: please remove support for the old config system - we do NOT provide new features via the old system.

plugins/imfile/imfile.c
```diff
@@ -2892,6 +2927,8 @@ CODEmodInit_QueryRegCFSLineHdlr
 	  	NULL, &cs.maxLinesAtOnce, STD_LOADABLE_MODULE_ID));
 	CHKiRet(omsdRegCFSLineHdlr((uchar *)"inputfiletrimlineoverbytes", 0, eCmdHdlrSize,
 	  	NULL, &cs.trimLineOverBytes, STD_LOADABLE_MODULE_ID));
+	CHKiRet(omsdRegCFSLineHdlr((uchar *)"inputfileignoreolderthan", 0, eCmdHdlrSize,
```
rgerhards: please remove support for the old config system - we do NOT provide new features via the old system.

tests/imfile-ignore-old-file-0.sh
```diff
@@ -0,0 +1,32 @@
+#! /bin/bash
```
rgerhards: I think this test is not needed, as it only covers the case where the parameter is not used. That is already being tested by all existing tests. CI runtime is a limited resource, so we try to avoid redundant tests.

tests/imfile-ignore-old-file-1.sh
```diff
@@ -0,0 +1,40 @@
...
+'
+
+# create a log for testing
+./inputfilegen -m 1000 > ${RSYSLOG_DYNNAME}.input
+
+touch -m -t 201806010000.00 ./${RSYSLOG_DYNNAME}.input
+
+startup
+
+# sleep a little to give rsyslog a chance to begin processing
```
rgerhards: this sleep should no longer be necessary with the current testbench plumbing; I understand it is carried over from an older test. If you like, I can update one of the tests to the newest standards (I could also patch them after merge, so this is not a show stopper).

rgerhards (changes requested):
More or less small issues. I would appreciate a fix. I can still merge as-is. If so, I'll fix it with a follow-up patch.

In any case, we need the doc (but please NOT for the legacy config).

rgerhards:
@n2yen I have begun to do a cleanup commit, so that you have a chance to review (not yet 100% finished). I will squash both before merge.

n2yen:
> @n2yen I have begun to do a cleanup commit, so that you have a chance to review (not yet 100% finished). I will squash both before merge.

Hi @rgerhards , apologies for slow reply - yes that sounds good! i'll work on the doc for this as well.

n2yen:
> > apologies for slow reply - yes that sounds good
> 
> all is well! I just thought it's pretty straight-forward, so I wanted not to hold merge (I plan the next scheduled release for next Tuesday).
> 
> I have made one more change. If all goes well (I let you know), you may want to squash yourself. That way I do not interfere with your "git credits".

sure sounds good. I'll keep an eye out for your signal!

rgerhards:
> I'll keep an eye out for your signal!

and here it is ;-)
CI looks happy, and I think I have all cleaned-up where it decently makes sense (due to policies etc as mentioned).

rgerhards (approved):

PR #4629, opened 2021-07-08
fix imklog crash in modExit

wxiaoguang:
Commit 94c4a871d0bcea445c8ee6d94e3f52b099730146 introduced a crash bug in `imklog`.

There is a `free(runModConf->pszBindRuleset)` in `BEGINmodExit`, but the `runModConf` seems to have been freed before.

This patch fixes the crash.

### DIFF
--- a/plugins/imklog/imklog.c
+++ b/plugins/imklog/imklog.c
@@ -453,6 +453,10 @@ ENDactivateCnf
 
 BEGINfreeCnf
 CODESTARTfreeCnf
+	if (runModConf != NULL) {
+		free(runModConf->pszBindRuleset);
+		runModConf = NULL;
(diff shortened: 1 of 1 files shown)


rgerhards:
Thx for the PR; however I need to have an in-depth look, as I thinks this points to a more general problem.

rgerhards:
@wxiaoguang  Once again thanks for the PR. I have done the review. The PR definitely points into the right direction, but the proposed change has some slight problems as it prevents the free of the config object. Instead of trying to amend it, I have created a new bugfixing PR ar https://github.com/rsyslog/rsyslog/pull/4659 and given proper credits. Hope this is fine with you.

PR #4633, opened 2021-07-13
Fixing the deleteStateOnFileDelete option

### LINKED ISSUE #4494: rsyslog leaves imfile state files behind (these get never deleted)

pearseimperva:
Fixes #4494

Fixes the `deleteStateOnFileDelete` option for imfile.

When the log file is deleted, imfile would attempt to delete the statefile but it was missing the `file_id` part of the statefile name. This means the statefiles were only removed in the log file is less than 512 characters, because for very small files the hash is not created. 

Added tests that checks that the statefile is deleted if `deleteStateOnFileDelete` is not specified, and that the statefile is not deleted if `deleteStateOnFileDelete = "off"`

### DIFF
--- a/tests/imfile-statefile-delete.sh
+++ b/tests/imfile-statefile-delete.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+# added 2019-02-28
+# This is part of the rsyslog testbench, licensed under ASL 2.0
+. ${srcdir:=.}/diag.sh init
+export TESTMESSAGES=1000
+export TESTMESSAGESFULL=999 
+export RETRIES=50
+generate_conf
+add_conf '
+global(workDirectory="'${RSYSLOG_DYNNAME}'.spool")
+module(load="../plugins/imfile/.libs/imfile" mode="inotify" PollingInterval="1")
+input(type="imfile" tag="file:" file="./'$RSYSLOG_DYNNAME'.input")
+
+template(name="outfmt" type="string" string="%msg:F,58:2%\n")
+if $msg contains "msgnum:" then
+	action(type="omfile" file=`echo $RSYSLOG_OUT_LOG` template="outfmt")
+'
+./inputfilegen -m $TESTMESSAGES > $RSYSLOG_DYNNAME.input
+inode=$(get_inode "$RSYSLOG_DYNNAME.input")
+startup
+wait_file_lines $RSYSLOG_OUT_LOG $TESTMESSAGES $RETRIES
+rm $RSYSLOG_DYNNAME.input
+sleep_time_ms=0
+while ls $RSYSLOG_DYNNAME.spool/imfile-state:$inode:* 1> /dev/null 2>&1; do
+	./msleep 10
+	((sleep_time_ms+=10))
+	if [ $sleep_time_ms -ge 30000 ]; then
+	        printf 'FAIL: state file still exists when it should have been deleted\nspool dir is:\n'
+	        ls -l $RSYSLOG_DYNNAME.spool
+	        error_exit 1
+	fi
+done
+shutdown_when_empty
+wait_shutdown
+seq_check 0 $TESTMESSAGESFULL # check we got the  message correctly
+exit_test
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -996,7 +996,7 @@ act_obj_destroy(act_obj_t *const act, const int is_deleted)
 		pollFile(act); /* get any left-over data */
 		if(inst->bRMStateOnDel) {
 			statefn = getStateFileName(act, statefile, sizeof(statefile));
-			getFullStateFileName(statefn, "", toDel, sizeof(toDel)); // TODO: check!
+			getFullStateFileName(statefn, act->file_id, toDel, sizeof(toDel)); // TODO: check!
 			statefn = toDel;
 		}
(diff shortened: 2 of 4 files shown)


derekjsonar:
Hi @rgerhards, hope all is well.
Do you know how we can get past the test failure in the `solaris10x64` build?
```
FAIL: parsertest-snare_ccoff_udp2.sh
``` 
I don't see how it could be related to @pearseimperva's change.

Thanks.

rgerhards:
@derekjsonar sorry for the late reply. The CI failures look like issues in CI itself. I have re-initiated the runs. Will now also look at the patch.

tests/imfile-statefile-delete.sh
```diff
@@ -0,0 +1,35 @@
...
+template(name="outfmt" type="string" string="%msg:F,58:2%\n")
+if $msg contains "msgnum:" then
+	action(type="omfile" file=`echo $RSYSLOG_OUT_LOG` template="outfmt")
+'
+./inputfilegen -m $TESTMESSAGES > $RSYSLOG_DYNNAME.input
+inode=$(get_inode "$RSYSLOG_DYNNAME.input")
+startup
+wait_file_lines $RSYSLOG_OUT_LOG $TESTMESSAGES $RETRIES
+rm $RSYSLOG_DYNNAME.input
+# sleep a little to give rsyslog a chance to notice the deleted file
```
rgerhards: timing via sleep is racy and results in flakes. Do we have a chance to synchronize the wait in a different way?

rgerhards (changes requested):
more a "discussion request" on the test sync comment,.

tests/imfile-statefile-delete.sh
pearseimperva: Hi @rgerhards, thanks for looking into this PR. I remove the 2s sleep from `imfile-statefile-delete.sh`. It now waits until the state file is deleted, and if it waits for longer than 30s the test fails.

  `imfile-state-file-no-delete.sh` needs to wait long enough for rsyslog to notice the file was deleted and decide not to delete the state file because `deleteStateOnFileDelete="off”`. I haven't been able to think of a better way of doing this yet, but if you are able to point me in the right direction I'd be happy to change that test too. I decided to use sleeps in the tests after looking at `imfile-wildcards-dirs-multi.sh` and `imfile-endregex.sh` that do a similar thing.

rgerhards: it's indeed tricky, I agree. Sounds like a good solution. Let's see what CI now says. I am willing to merge and see if this test becomes racy in the long term. If so, we (I ;-)) need to think of a different way to handle it.

rgerhards (approved):

PR #4639, opened 2021-07-22
[omelasticsearch] Improve errorFile mutex handling

fpoirotte:
Lock the file during SIGHUPs to avoid issues with concurrent accesses by writeDataError().
Check the return value of pthread_mutex_init() in CODESTARTcreateInstance to make sure the mutex is actually usable.

### DIFF
--- a/plugins/omelasticsearch/omelasticsearch.c
+++ b/plugins/omelasticsearch/omelasticsearch.c
@@ -232,7 +232,7 @@ static rsRetVal curlSetup(wrkrInstanceData_t *pWrkrData);
 BEGINcreateInstance
 CODESTARTcreateInstance
 	pData->fdErrFile = -1;
-	pthread_mutex_init(&pData->mutErrFile, NULL);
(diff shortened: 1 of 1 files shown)


plugins/omelasticsearch/omelasticsearch.c
```diff
@@ -232,14 +232,15 @@ static rsRetVal curlSetup(wrkrInstanceData_t *pWrkrData);
 BEGINcreateInstance
 CODESTARTcreateInstance
 	pData->fdErrFile = -1;
-	pthread_mutex_init(&pData->mutErrFile, NULL);
+	CHKiRet(pthread_mutex_init(&pData->mutErrFile, NULL));
```
rgerhards, line 235: pthread_mutex_init() does not return a rsyslog error code. So this should be mapped. It's not a merge stopper, but I'll leave it open a bit more so that you can fix it. Otherwise I'll do after merge.

rgerhards (approved):
approved with given optional suggestion for improvement

PR #4640, opened 2021-07-22
Close file descriptor when freshStartTail is turned on

Cropi:
Close file descriptor when **freshStartTail** is turned on. Introduced in https://github.com/rsyslog/rsyslog/commit/b57a65287d0c17e4b28a99cbe5d25bdb5b58f02d.


rgerhards:
Thx for the PR, much appreciated!

PR #4651, opened 2021-08-04
debug support: add indication of "being HUPed" to debug log

rgerhards:
so far, we cannot clearly see when HUP processing begins


rgerhards:
Note: CodeFactor reports a false positive warning

Issue #4625, opened 2021-07-02
When log file is rotated, rsyslog may meet memory corruption and leads to incorrect logic or crash


wxiaoguang:
### Steps to reproduce the behavior

Possible steps:

```
nginx: write log to access.web.log (inode 2226240, wd=112)
logrotate: rename to access.web.log (inode 2226240) to access.web.log.1 (inode 2226240)
rsyslog: update wdmap, remove wd=112
nginx: write log to access.web.log.1 by opened fd (inode 2226240, wd=112)  # if I read document right, the watch id is related to file fd
rsyslog: receive inotify with wd=112
```

And I believe the crash is related to the writing of rotated file. But I can not get a core dump and haven't catch the crash in gdb.

```
[Fri Jul  2 13:08:28 2021] traps: rsyslogd[23639] general protection ip:7f9fdb25e9bd sp:7fffb5aaf310 error:0 in libc-2.28.so[7f9fdb1fc000+148000]
[Fri Jul  2 13:13:48 2021] traps: rsyslogd[26012] general protection ip:7f7e1ac039bd sp:7fff61bd2510 error:0 in libc-2.28.so[7f7e1aba1000+148000]
[Fri Jul  2 13:14:11 2021] traps: rsyslogd[7017] general protection ip:7fb05ab729bd sp:7ffcde2d5130 error:0 in libc-2.28.so[7fb05ab10000+148000]
[Fri Jul  2 13:14:11 2021] traps: rsyslogd[7943] general protection ip:7f4de54d49bd sp:7ffedae8f7a0 error:0 in libc-2.28.so[7f4de5472000+148000]
[Fri Jul  2 13:14:18 2021] traps: rsyslogd[8045] general protection ip:7f9df174b9bd sp:7ffeb00968b0 error:0 in libc-2.28.so[7f9df16e9000+148000]
[Fri Jul  2 13:14:22 2021] traps: rsyslogd[8287] general protection ip:7f3223f959bd sp:7ffdff791770 error:0 in libc-2.28.so[7f3223f33000+148000]
[Fri Jul  2 19:01:13 2021] traps: rsyslogd[19139] general protection ip:7fcc138729bd sp:7ffd306a6110 error:0 in libc-2.28.so[7fcc13810000+148000]
[...]

PR #4659, opened 2021-08-10
imklog: fix invalid memory adressing, could cause abort

rgerhards:
This is a regeression from commit 94c4a87. It introduced a free() call
using an object that was no longer valid (the main pointer to the
to-be-freed object) was already freed at time of use. This could
cause various issues, including a segfault.

Note: this bug was triggerred only during late phase of rsyslog
shutdown, so it did not affect regular operation.

Special thanks to github user wxiaoguang for analyzing the issue
and providing a draft fix proposal, on which this patch builds.

see also https://github.com/rsyslog/rsyslog/pull/4629
closes https://github.com/rsyslog/rsyslog/issues/4625

### DIFF
--- a/plugins/imklog/imklog.c
+++ b/plugins/imklog/imklog.c
@@ -453,6 +453,7 @@ ENDactivateCnf
 
 BEGINfreeCnf
 CODESTARTfreeCnf
+	free(pModConf->pszBindRuleset);
 ENDfreeCnf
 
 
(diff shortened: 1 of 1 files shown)


wxiaoguang:
Of course it's fine and very glad to help. `rsyslog` really helps me a lot. 

ps: I realized that I made a mistake in the patch, which is I meant to write `runModConf->pszBindRuleset = NULL`. But I misunderstood that `a more general problem` in the comment, I thought there was something complicated with rsyslog's memory model 😂

rgerhards:
> But I misunderstood that `a more general problem` in the comment, I thought there was something complicated with rsyslog's memory model

sorry for being imprecise. I did actually not know at this time what the real issue was. But it looked wrong compared to the other code. I wrote the generic mappers many years ago, and *my* memory vanished. When I refreshed it today, things became clear and simple.

PR #4666, opened 2021-08-24
imuxsock: fix rate limiting not working in default configuration

gajdusek:
This commit fixes the imuxsock's rate limiting, which does not work in the
default configuration (useSpecialParser="true" and ratelimit.severity="1")

When useSpecialParser was set to "true" all messages were treated with
severity 0 and therefore were not subjects to the rate limiting that by
default ignores messages with severity 0.

The test suite requires
./configure --enable-liblogging-stdlog

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -363,6 +363,7 @@ TESTS +=  \
 	dircreate_off.sh \
 	imuxsock_legacy.sh \
 	imuxsock_logger.sh \
+	imuxsock_logger_ratelimit.sh \
 	imuxsock_logger_ruleset.sh \
 	imuxsock_logger_ruleset_ratelimit.sh \
 	imuxsock_logger_err.sh \
@@ -2479,6 +2480,7 @@ EXTRA_DIST= \
 	imuxsock_legacy.sh \
 	imuxsock_logger_parserchain.sh \
 	imuxsock_logger.sh \
+	imuxsock_logger_ratelimit.sh \
 	imuxsock_logger_ruleset.sh \
 	imuxsock_logger_ruleset_ratelimit.sh \
 	imuxsock_logger_err.sh \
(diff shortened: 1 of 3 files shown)


rgerhards:
I think the analysis, and this cure, is slightly off: The problem is not that severity is not that the severity assignment per se is wrong. The current cure will use invalid severity values when no parser is called. I think the real cure is the move the second assignment of "severity" right below the if checking if parsing is needed (IAW: one line below). That will pick the severity from the message object.

Note that when `ratelimit->severity` (the configured ratelimiting severity) is 0, the actual message severity does not matter, because all messages are ratelimited. So we do NOT need to parse the message in early stage.

gajdusek:
> The current cure will use invalid severity values when no parser is called. 

You are right! I missed that the initial severity of msg object is LOG_DEBUG, not 0 as I falsely assumed.

> I think the real cure is the move the second assignment of "severity" right below the if checking if parsing is needed (IAW: one line below). That will pick the severity from the message object.

This was also my first attempt to fix it... but then I found the conditions logic too cluttered and I thought it can be simplified if the initial severity of each msg object is 0.

Thanks for review! I pushed another version.

rgerhards (approved):

rgerhards:
@gajdusek thx a lot for the follow-up and the patch in general - very helpful!

tests/Makefile.am
```diff
@@ -363,6 +363,7 @@ TESTS +=  \
 	dircreate_off.sh \
 	imuxsock_legacy.sh \
 	imuxsock_logger.sh \
+	imuxsock_logger_ratelimit.sh \
```
rgerhards, line 366: you need to add the new file to EXTRA_DIST as well (I know..., but that's historical and was not changed til now as it's quite some effort)

rgerhards (changes requested):

tests/Makefile.am
gajdusek: Done. I also changed the commit message to reflect that this is not imuxsock specific.

rgerhards (approved):

rgerhards:
Thx again for the PR and your support in getting it through!

PR #4668, opened 2021-08-25
WiP: Add support for wolfSSL for TLS.

haydenroche5:
Hello @rgerhards. I'm a software engineer with wolfSSL, and we recently had a customer ask us to port rsyslog from OpenSSL to wolfSSL. wolfSSL provides an OpenSSL compatibility layer that implements the majority of the OpenSSL API. As such, I was able to integrate wolfSSL support into the existing OpenSSL code in rsyslog. I'm curious if you would be interested in upstreaming this work, and if so, I'm also interested in any feedback you have on the patch. Thanks!

### DIFF
--- a/.github/workflows/run_checks.yml
+++ b/.github/workflows/run_checks.yml
@@ -1,4 +1,4 @@
-# Copyright 2024-2025 Rainer Gerhards and Others
+# Copyright 2024-2026 Rainer Gerhards and Others
 #
 # https://github.com/rsyslog/rsyslog-pkg-ubuntu
 #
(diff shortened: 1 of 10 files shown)


haydenroche5:
Hi @rgerhards and @alorbach, just circling back on this to see if you are still interested in having wolfSSL support upstream.

rgerhards:
@haydenroche5 sorry for the extremely late reply - we screwed up on that, unfortunately. Nevertheless, I thought I let you know that I will now see that I get your contribution into mergable state again.

PR #4671, opened 2021-09-02
gnutls: Propagate PrioritizeSAN when accepting a new connection

Cropi:
Until now, when the server accepts new connection via ```AcceptConnReq()```, it does not initialize the ```bSANpriority``` field.
While the option does work on the client side (meaning the server cert is checked), it does not on the server side.


rgerhards (approved):

rgerhards:

Issue #4684, opened 2021-09-13
Fix ElasticSearch Test brocken by ES incompatibility


rgerhards:
Elasticsearch has removed server settings. These queue settings do no longer exits:
https://github.com/rsyslog/rsyslog/blob/e5b62c7886fe09951ed5e32fa5420ef9e961830a/tests/es-bulk-retry.sh#L12

This leads to test failure under ES 7. We need to fix this, but I least I do not know how. Replacing `bulk` by queue did not work.

Ref: https://discuss.elastic.co/t/unknown-setting-thread-pool-bulk-queue-size/180120

@richm as original author of this test, can you help?

rgerhards:
fixed by #4685

PR #4685, opened 2021-09-13
Fix ElasticSearch Test broken by ES incompatibility

richm:
https://github.com/rsyslog/rsyslog/issues/4684
According to https://discuss.elastic.co/t/unknown-setting-thread-pool-bulk-queue-size/180120
The setting `thread_pool.bulk` has been renamed to `thread_pool.write`
"The bulk threadpool was renamed to write in 6.3.0, and support for the legacy bulk name was removed in 7.0.0 133."

### DIFF
--- a/tests/diag.sh
+++ b/tests/diag.sh
@@ -2213,6 +2213,8 @@ start_elasticsearch() {
 # $2 - ES port
 es_getdata() {
 	curl --silent -XPUT --show-error -H 'Content-Type: application/json' "http://localhost:${2:-$ES_PORT}/rsyslog_testbench/_settings" -d '{ "index" : { "max_result_window" : '${1:-$NUMMESSAGES}' } }'
+	# refresh to ensure we get the latest data
+	curl --silent localhost:${2:-$ES_PORT}/rsyslog_testbench/_refresh
 	curl --silent localhost:${2:-$ES_PORT}/rsyslog_testbench/_search?size=${1:-$NUMMESSAGES} > $RSYSLOG_DYNNAME.work
 	$PYTHON $srcdir/es_response_get_msgnum.py > ${RSYSLOG_OUT_LOG}
 }
--- a/tests/es-bulk-retry.sh
+++ b/tests/es-bulk-retry.sh
@@ -1,11 +1,5 @@
 #!/bin/bash
 # This file is part of the rsyslog project, released under ASL 2.0
-
-	echo The es-bulk-retry.sh test is temporarily suspended, because ElasticSearch 7 broke semantics
-	echo that the test depends on. See for more info and progress:
-	echo https://github.com/rsyslog/rsyslog/pull/4685
-	exit 77
-
 . ${srcdir:=.}/diag.sh init
 export ES_PORT=19200
 export NUMMESSAGES=100
@@ -14,9 +8,17 @@ override_test_timeout 120
 ensure_elasticsearch_ready --no-start
 
 # change settings to cause bulk rejection errors
+case "$ES_DOWNLOAD" in
+    elasticsearch-5.*) es_option="thread_pool.bulk"
+                       es_mapping_uses_type=true
+                       es_search_type="test-type" ;;
+    *) es_option="thread_pool.write"
+       es_mapping_uses_type=false
+       es_search_type="_doc" ;;
+esac
 cat >> $dep_work_dir/es/config/elasticsearch.yml <<EOF
-thread_pool.write.queue_size: 1
-thread_pool.write.size: 1
+${es_option}.queue_size: 1
+${es_option}.size: 1
 EOF
 start_elasticsearch
 
@@ -95,7 +97,7 @@ ruleset(name="try_es") {
 		       bulkmode="on"
 		       retryfailures="on"
 		       retryruleset="try_es"
-		       searchType="test-type"
+		       searchType="'${es_search_type}'"
 		       searchIndex="rsyslog_testbench")
 	}
 }
@@ -107,7 +109,8 @@ if $msg contains "msgnum:" then {
 action(type="omfile" file="'$RSYSLOG_OUT_LOG'")
 '
 
-curl -s -H 'Content-Type: application/json' -XPUT localhost:${ES_PORT:-19200}/rsyslog_testbench/ -d '{
+if [ "$es_mapping_uses_type" = true ]; then
+    curl -s -H 'Content-Type: application/json' -XPUT localhost:${ES_PORT:-19200}/rsyslog_testbench/ -d '{
   "mappings": {
     "test-type": {
       "properties": {
@@ -119,6 +122,23 @@ curl -s -H 'Content-Type: application/json' -XPUT localhost:${ES_PORT:-19200}/rs
   }
 }
 ' | $PYTHON -mjson.tool
+else
+  # we add 10 shards so we're more likely to get queue rejections
+    curl -s -H 'Content-Type: application/json' -XPUT localhost:${ES_PORT:-19200}/rsyslog_testbench/ -d '{
+  "settings": {
+    "index.number_of_shards": 10
+  },
+  "mappings": {
+    "properties": {
+      "msgnum": {
+        "type": "integer"
+      }
+    }
+  }
+}
+' | $PYTHON -mjson.tool
+fi
+
 #export RSYSLOG_DEBUG="debug nostdout noprintmutexaction"
 #export RSYSLOG_DEBUGLOG="debug.log"
 startup
@@ -128,6 +148,7 @@ if [ -n "${USE_GDB:-}" ] ; then
 fi
 success=50
 badarg=50
+./msleep 5000
 injectmsg 0 $NUMMESSAGES
 ./msleep 1500; cat $RSYSLOG_OUT_LOG # debuging - we sometimes miss 1 message
 wait_content '"response.success": 50' $RSYSLOG_DYNNAME.spool/es-stats.log
--- a/tests/es-bulk-retry.sh
+++ b/tests/es-bulk-retry.sh
@@ -15,8 +15,8 @@ ensure_elasticsearch_ready --no-start
 
 # change settings to cause bulk rejection errors
 cat >> $dep_work_dir/es/config/elasticsearch.yml <<EOF
-thread_pool.bulk.queue_size: 1
-thread_pool.bulk.size: 1
+thread_pool.write.queue_size: 1
+thread_pool.write.size: 1
 EOF
 start_elasticsearch
 


richm:
@rgerhards This PR is incomplete, but this is as far as I can go.  It seems that ES7 has changed the bulk api in some way that causes the response not to register any bulk rejections.  As far as I can tell, the good records are being correctly stored, and the bad records are being correctly rejected.  So one option is to change the test to skip looking for bulk index rejections (alternately, skip that check on ES6 or later).  I'll see if I can find someone - @praiskup @Cropi @portante - any interest in fixing this?

portante:
> @rgerhards This PR is incomplete, but this is as far as I can go. It seems that ES7 has changed the bulk api in some way that causes the response not to register any bulk rejections.

So if one overwhelms an ES V7 instance, you no longer see bulk rejections (status code in each response action payload of 429)?

> As far as I can tell, the good records are being correctly stored, and the bad records are being correctly rejected.

For "bulk rejection" (429) to happen, the Elasticsearch instance has to have exhausted its `thread_pool.write` count of threads.  Then a new request which can't be queued should see that response.  This test appears to be designed to make that happen, but perhaps the queue size setting is being ignored for some reason?  Has anybody stood up an Elasticsearch V7 instance to verify that setting `thread_pool.write.queue_size` to one actually does what we expect?

>  So one option is to change the test to skip looking for bulk index rejections (alternately, skip that check on ES6 or later). I'll see if I can find someone - @praiskup @Cropi @portante - any interest in fixing this?

If the goal of this test is to verify that the retry ruleset works, then just dropping the check for bulk index rejections from the test is fine.

But it appears that the code is retrying lots of records, which it shouldn't (400s, duplicates on create, etc.) where the re-try should not be once, but continuously until they get through.

rgerhards:
> It seems that ES7 has changed the bulk api in some way that causes the response not to register any bulk rejections.

Yeah - I tried the change myself, but saw that the test failed. It looks like (already discussed) ES changed not only the name but also semantics (I really hate braking changes, don't know why folks always do that...). I am good with anything that solves the situation. At worst, we need to remove the test if with newer ES it is impossible to actually test the functionality. So any work-around is appreciated. To go forward, I'll possibly disable the test temporarily in my PR, so that we can pursue integrating the proposed new functionality (which addresses another breaking change...).

rgerhards:
I have now updated the current development state to ES 7 (via https://github.com/rsyslog/rsyslog/pull/4683). I did so by temporarily disabling the test. You now need to rebase and remove the "disabling lines" at the top (https://github.com/rsyslog/rsyslog/blob/master/tests/es-bulk-retry.sh#L4). Then we see how any new test works out.

richm:
> > It seems that ES7 has changed the bulk api in some way that causes the response not to register any bulk rejections.
> 
> Yeah - I tried the change myself, but saw that the test failed. It looks like (already discussed) ES changed not only the name but also semantics (I really hate braking changes, don't know why folks always do that...).

The good news is that Elasticsearch only does breaking changes for major releases (e.g. semantic versioning).  The bad news is that ES has had several major releases in the past few years . . .

> I am good with anything that solves the situation. At worst, we need to remove the test if with newer ES it is impossible to actually test the functionality.

It will require someone doing some intensive investigation + trial-and-error to figure out how to test the bulk rejection handling with ES7.  I just don't have time to do this, which is why I was hoping @Cropi @praiskup could do this, or maybe @portante knows someone interested in rsyslog + elasticsearch.

> So any work-around is appreciated. To go forward, I'll possibly disable the test temporarily in my PR, so that we can pursue integrating the proposed new functionality (which addresses another breaking change...).

I think it is best for now to disable the test.

rgerhards:
@richm ack, sounds good.

If you could do me nevertheless the favor to rebase and remove the skip lines inside the test. That would leave this tracker here open for anyone potentially following up. I can also do it myself (assuming you permitted admin pushes), but this always sounds a bit "unclean" to me ;-) But don't feel pressed, I'll do in a few days if needed.

rgerhards:
@radu-gheorghe can you help with the new semantics in ES 7?

radu-gheorghe:
Certainly! I had a quick look and this line looks wrong: https://github.com/rsyslog/rsyslog/blob/master/tests/es-bulk-retry.sh#L112

We define a type, but types are not supported anymore, so the `test-type` level can go out entirely. I don't know if this is enough to fix the test, feel free to ping me if it doesn't. It's been a long time since I fiddled with rsyslog tests :( So if you want me to look deeper, can you give me a quick list of reproduction steps? Or a link to where they are?

rgerhards:
@radu-gheorghe thx for the quick reply, much appreciated. I'll update this PR now so that everyone can see the test failure. I'll also point you to the code in question and will check if your tipp already helps solve the situation.

radu-gheorghe:
Great! You're welcome! Just @ ping me if I can help with anything.

rgerhards:
@radu-gheorghe I checked the suggestion on line 112 ("type") and this is actually only executed when ES 5 is used.
I have also rebased and modified the test to be executed now with ES 7.14, Please have a look e.g. at https://github.com/rsyslog/rsyslog/pull/4685/checks?check_run_id=3621086247 (go to the "error log" step, not the one github opens - "error log" is right below)

In a nutshell, the issue is that ES is configured in a way that should trigger errors, which are then retried. @portante has described this much better in https://github.com/rsyslog/rsyslog/pull/4685/checks?check_run_id=3621086247 than I can ;-) -- actually, I have no idea (due to lack of ES know-how) of how exactly the test is supposed to work.

The problem is that seemingly not only the syntax of ES config parameter `thread_pool.bulk` has changed but also the semantic. This seems to break the test fail/success detection.

I hope I could clarify. The current cure is to disable the test (and remove it in the long term if we cannot re-enable it). It would be better to know what actually happens and how to fix the test. At worst, the test may indeed bring up an issue in rsyslog code which we see only with ES 7, but which we misunderstand as an ES 7 semantic change. So I don't like to disable the test as a cure without knowing what's happening ;-)

radu-gheorghe:
Let me start with the complaints :)
- I still don't know how to reproduce this issue in isolation so I can poke around until I can find a fix
- maybe the link to @portante's comment was a bad copy-paste? But I read this thread carefully, so I guess I didn't miss it

Now let me explain my understanding of the test as it stands:
- we're sending 100 messages
- we're assuming some will fail with 429 and we'll retry them
- in the end, all should get in successfully, but we should see the rejected counts
- the rejected counts are actually [kept by rsyslog](https://github.com/rsyslog/rsyslog/blob/6eaec67cf97f57b6e076091dd3b1e781cedbe92c/plugins/omelasticsearch/omelasticsearch.c#L2251) when it gets 429s

Finally, here are my findings so far:
- I checked the script that generates the config. Looks OK to me, with the exception of `>>` which looks like it appends to something, I'm not sure if that file is already there. If it's not, we should be good. The config is OK
- the thread pool settings are applied, I could check that via `curl localhost:9200/_cat/thread_pool?h=name,size,queue_size`
- my feeling is that we're betting on a race. We're sending 100 messages (in batches of 16? that's the default, I think) and we're hoping that one thread on the Elasticsearch side with a queue of 1 would be slow enough to reject. But we don't have such guarantees
- unfortunately, we can't force ES to reject by setting any thread values to 0 :(
- with values set to 1, if I do a simple `while true; do curl -X POST -H 'Content-Type: application/json' 'http://localhost:9200/test/_doc' -d '{"foo": "bar"}'; done`, I get no rejections. Elasticsearch outperforms bash+curl :)
- one change to ES 7 was moving to a default of 1 shard (from 5). This might make the race harder for the test, because writes would spawn one thread per bulk per shard. One bulk would spawn 6 threads before (one for "coordination" and 5 per shard) and it would spawn just one now. So maybe a simple fix could be to force the number of shards, by adding it to the index creation process, like:

```
curl -s -H 'Content-Type: application/json' -XPUT localhost:${ES_PORT:-19200}/rsyslog_testbench/ -d '{
  "settings": {
    "index.number_of_shards": 10
  },
  "mappings": {
    "properties": {
      "msgnum": {
        "type": "integer"
      }
    }
  }
}
'
```

That said, with the above, I still couldn't get rejections with curl :( But maybe with rsyslog we can, especially if we use multiple worker threads (do we, by default?) and/or a lower batch size (e.g. 2 or even 1 message). It would still be a race and a hack, but maybe if we consistently "win" it, we'd be good :)

I'd love to get to the bottom of this, but my main blocker is my inability to reproduce the issue in isolation. Worst case scenario, I'd reproduce bits (e.g. set up rsyslog with that config on a local ES), but I'm not sure if that would fix the test - especially since I would suggest changes that I won't be able to test (e.g. maybe a syntax error would sneak in...).

Still, maybe the above helps!

rgerhards:
@radu-gheorghe Thanks for the thoughtful reply.

The link to @portante's reply was correct, but maybe I understood it not sufficiently to evaluate it ;-)

Re local repo, sorry for the slim info. There are various ways to do it. The best possibility, IMHO, is to use a docke container. We use containers inside CI, and so you could use the very same environment that CI uses. The container also has the advantage that you do not need to get all the build dependencies.

Is docker OK with you? If so, I will provide detail instructions - it's not much, but you need the right pieces.

rgerhards:
Well, just in case: ;-)
```
$ cd some-workdir
$ git clone https://github.com/rsyslog/rsyslog.git
$ cd rsyslog
$ export RSYSLOG_DEV_CONTAINER='rsyslog/rsyslog_dev_base_ubuntu:20.04'
$ devtools/devcontainer.sh --rm -ti
# Now inside container
$ devtools/run-configure.sh
$ make -j
# rsyslog built
$ cd tests
$ ./es-bulk-retry.sh
# Test runs, and can be edited :-)
```

richm:
> Let me start with the complaints :)
> 
>     * I still don't know how to reproduce this issue in isolation so I can poke around until I can find a fix
> 
>     * maybe the link to @portante's comment was a bad copy-paste? But I read this thread carefully, so I guess I didn't miss it
> 
> 
> Now let me explain my understanding of the test as it stands:
> 
>     * we're sending 100 messages

We're sending a single HTTP message - a bulk api request - which consists of 100 operations, all of which are to create a new record in ES.

> 
>     * we're assuming some will fail with 429 and we'll retry them

yes

> 
>     * in the end, all should get in successfully, but we should see the rejected counts

Actually, half of the messages should get accepted successfully, and half should be rejected due to syntax error (response.badargument)

> 
>     * the rejected counts are actually [kept by rsyslog](https://github.com/rsyslog/rsyslog/blob/6eaec67cf97f57b6e076091dd3b1e781cedbe92c/plugins/omelasticsearch/omelasticsearch.c#L2251) when it gets 429s

yes

> 
> 
> Finally, here are my findings so far:
> 
>     * I checked the script that generates the config. Looks OK to me, with the exception of `>>` which looks like it appends to something, I'm not sure if that file is already there. If it's not, we should be good. The config is OK

note that in bash, if you use `>>` and the target file does not exist, it will be created

> 
>     * the thread pool settings are applied, I could check that via `curl localhost:9200/_cat/thread_pool?h=name,size,queue_size`
> 
>     * my feeling is that we're betting on a race. We're sending 100 messages (in batches of 16? that's the default, I think) and we're hoping that one thread on the Elasticsearch side with a queue of 1 would be slow enough to reject. But we don't have such guarantees

yes, that's essentially it - and this seems to work with ES 5 and earlier

> 
>     * unfortunately, we can't force ES to reject by setting any thread values to 0 :(

1 is the minimum

> 
>     * with values set to 1, if I do a simple `while true; do curl -X POST -H 'Content-Type: application/json' 'http://localhost:9200/test/_doc' -d '{"foo": "bar"}'; done`, I get no rejections. Elasticsearch outperforms bash+curl :)

That isn't doing the same operation - that's a "simple" record addition, not a bulk record index operation.  You would need to create a rather large JSON blob to simulate a bulk index operation.  Something like https://github.com/openshift/origin-aggregated-logging/blob/master/test/zzzz-bulk_rejection.sh, especially around https://github.com/openshift/origin-aggregated-logging/blob/master/test/zzzz-bulk_rejection.sh#L138

> 
>     * one change to ES 7 was moving to a default of 1 shard (from 5). This might make the race harder for the test, because writes would spawn one thread per bulk per shard. One bulk would spawn 6 threads before (one for "coordination" and 5 per shard) and it would spawn just one now. So maybe a simple fix could be to force the number of shards, by adding it to the index creation process, like:
> 
> 
> ```
> curl -s -H 'Content-Type: application/json' -XPUT localhost:${ES_PORT:-19200}/rsyslog_testbench/ -d '{
>   "settings": {
>     "index.number_of_shards": 10
>   },
>   "mappings": {
>     "properties": {
>       "msgnum": {
>         "type": "integer"
>       }
>     }
>   }
> }
> '
> ```
> 
> That said, with the above, I still couldn't get rejections with curl :( But maybe with rsyslog we can, especially if we use multiple worker threads (do we, by default?) and/or a lower batch size (e.g. 2 or even 1 message). It would still be a race and a hack, but maybe if we consistently "win" it, we'd be good :)

ok, sounds good

> 
> I'd love to get to the bottom of this, but my main blocker is my inability to reproduce the issue in isolation. Worst case scenario, I'd reproduce bits (e.g. set up rsyslog with that config on a local ES), but I'm not sure if that would fix the test - especially since I would suggest changes that I won't be able to test (e.g. maybe a syntax error would sneak in...).
> 
> Still, maybe the above helps!

radu-gheorghe:
Thanks, @rgerhards and @richm ! Yes, Docker is perfectly fine, let me try it out and come back.

rgerhards:
@radu-gheorghe I made a small mistake in the docker instructions: it's not `make -j` but `make -j check TESTS=`. The latter builds the testbench tools (but does not run the tests). Tools are needed to run the test manually.

radu-gheorghe:
Thanks, @rgerhards! Unfortunately, `make -j check TESTS=.` fails with:

```
miniamqpsrvr.c: In function 'amqp_srvr':
miniamqpsrvr.c:456:9: error: 'frame.data' may be used uninitialized in this function [-Werror=maybe-uninitialized]
  456 |     p = amqpFieldFprintf("Exchange:", frame.data + 2);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
miniamqpsrvr.c:507:19: error: 'frame.datalen' may be used uninitialized in this function [-Werror=maybe-uninitialized]
  507 |     if (frame.data[frame.datalen-1] != '\n')
      |                   ^
miniamqpsrvr.c:294:20: error: 'frame.method' may be used uninitialized in this function [-Werror=maybe-uninitialize]
  294 |  amqp_frame_type_t frame;
      |                    ^~~~~
make[2]: *** [Makefile:3538: miniamqpsrvr-miniamqpsrvr.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[2]: Leaving directory '/rsyslog/tests'
make[1]: *** [Makefile:3827: check-am] Error 2
make[1]: Leaving directory '/rsyslog/tests'
make: *** [Makefile:924: check-recursive] Error 1
```

But then I did `make -j` and that seemed to work. I'll keep going, I seem to be able to reproduce the issue.

rgerhards:
> Unfortunately, `make -j check TESTS=.` fails with:

that's interesting, because it is the same environment... I'll check.

radu-gheorghe:
I think I found a fix. I'm not sure how good it is, because there are things I'm surely missing in how the test framework runs. Maybe others can comment.

Changes:
- `thread_pool.bulk` -> `thread_pool.write`. I'm not sure if we even need the conditional logic there, maybe we'll just support ES 7+?
- `searchType="test-type` -> `searchType="_doc". Similarly, already in the PR
- added 10 shards, as written above. At this point I get bulk rejections, but not all messages are there
- I think the missing messages are also due to a race: Elasticsearch is [near-realtime](https://www.elastic.co/guide/en/elasticsearch/reference/current/near-real-time.html), your view on an index might be outdated until the next refresh
- I added `curl --silent localhost:${2:-$ES_PORT}/rsyslog_testbench/_refresh` to `diag.sh`'s `es_getdata()`, right before the `_search` call. This should ensure we get an updated view on the index. And now the test works (on my machine)

We're still betting on a bit of a race with size=1, but I don't see a better option. If it helps, I can prepare a commit (to this PR) with my changes to the index settings and diag.sh. But maybe others see better options, other things to improve?

rgerhards:
Thx - I am just supporting the plumbing here, guess it is a call for @richm. Maybe a PR might be useful for him, and I guess it's no real work to do (neither is drop it later on ;-))

richm:
> I think I found a fix. I'm not sure how good it is, because there are things I'm surely missing in how the test framework runs. Maybe others can comment.
> 
> Changes:
> 
>     * `thread_pool.bulk` -> `thread_pool.write`. I'm not sure if we even need the conditional logic there, maybe we'll just support ES 7+?

doesn't matter to me

> 
>     * `searchType="test-type` -> `searchType="_doc". Similarly, already in the PR
> 
>     * added 10 shards, as written above. At this point I get bulk rejections, but not all messages are there
> 
>     * I think the missing messages are also due to a race: Elasticsearch is [near-realtime](https://www.elastic.co/guide/en/elasticsearch/reference/current/near-real-time.html), your view on an index might be outdated until the next refresh
> 
>     * I added `curl --silent localhost:${2:-$ES_PORT}/rsyslog_testbench/_refresh` to `diag.sh`'s `es_getdata()`, right before the `_search` call. This should ensure we get an updated view on the index. And now the test works (on my machine)

mine too - tested on fedora 34 - works fine with the changes you suggested

> 
> 
> We're still betting on a bit of a race with size=1, but I don't see a better option.

Right.

A radical option would be to use a "mock" elasticsearch - a fake http server that returns specific responses/error codes.  We took a similar approach for rsyslog mmkubernetes testing since we didn't want to install an entire Kubernetes cluster for the test - instead, we create a mock k8s api server - https://github.com/rsyslog/rsyslog/blob/master/tests/mmkubernetes_test_server.py - the drawback for this is that Elasticsearch seems to change its api pretty often, so you would be continually trying to keep the mock http server in sync with the actual Elasticsearch behavior . . .

> If it helps, I can prepare a commit (to this PR) with my changes to the index settings and diag.sh. But maybe others see better options, other things to improve?

I think your changes are fine.

radu-gheorghe:
Thanks, @richm. Excuse my git noobness, but I can't push another commit to this PR, because I think I don't have permissions to your clone:

```
$ git push richm richm-fix-es-bulk-setting:fix-es-bulk-setting
remote: Permission to richm/rsyslog.git denied to radu-gheorghe.
fatal: unable to access 'https://github.com/richm/rsyslog.git/': The requested URL returned error: 403
```

So I'm attaching a patch [here](https://github.com/rsyslog/rsyslog/files/7209044/rsyslog.patch.txt), maybe you can apply it? Or if you give me permissions, I can push the change? Or maybe I'm missing something?

richm:
> Thanks, @richm. Excuse my git noobness, but I can't push another commit to this PR, because I think I don't have permissions to your clone:
> 
> ```
> $ git push richm richm-fix-es-bulk-setting:fix-es-bulk-setting
> remote: Permission to richm/rsyslog.git denied to radu-gheorghe.
> fatal: unable to access 'https://github.com/richm/rsyslog.git/': The requested URL returned error: 403
> ```
> 
> So I'm attaching a patch [here](https://github.com/rsyslog/rsyslog/files/7209044/rsyslog.patch.txt), maybe you can apply it? Or if you give me permissions, I can push the change? Or maybe I'm missing something?

I would suggest that you take over this work - close this PR, and open your own PR.

radu-gheorghe:
OK, I'll try to do that next week (sorry, I'm swamped right now).

rgerhards:
@radu-gheorghe thx again for the effort - and be relieved, I'll handle the point from here on (except when I run into ES issues again). I guess I can update the PR due to my project admin rights, else I'll create a new one. I let you know the outcome.

radu-gheorghe:
Thanks Rainer! And thanks, Richard! 🙏

radu-gheorghe:
Sweet! You're welcome and thanks for the update!

Issue #4678, opened 2021-09-08
An improper locking bug(e.g., deadlock) on ctx->module_lock


ryancaicse:
Hi, developers, thank you for your checking. It seems the lock `ctx->module_lock` is not released correctly when ret != `RSGTE_SUCCESS` in the function `rsksiCtxOpenFile`?

https://github.com/rsyslog/rsyslog/blob/91885676001c9df1c2c91514d144cf71755d5d14/runtime/lib_ksils12.c#L946-L957

Best,

taavi-valjaots:
Hi! Has it caused any trouble? It seems that if signer is not started correctly module is disabled and not used. But I agree that `ctx->module_lock` should be released before returning the function.

ryancaicse:
@taavi-valjaots No, it is not triggered but found by code scan. Thank you for your reply and confirmation.

rgerhards:
@taavi-valjaots let me know if you intend to fix this. We have roughly 3 to 4 weeks to the next scheduled release, just FYI.

rgerhards:
Sorry for the very late reply. This may no longer be directly relevant to the original static-analysis report, but I am adding the answer for people who find this issue through search while we work through the backlog. We do not lightly close old issues and we do not close them just because of age.

The specific lock path reported here is fixed in current `main`. In `runtime/lib_ksils12.c`, `rsksiCtxOpenFile()` now unlocks `ctx->module_lock` before returning on the `ret != RSGTE_SUCCESS` path:

```c
if (ret != RSGTE_SUCCESS) {
    report(ctx, "Unable to init. KSI module, signing service disabled");
    ctx->disabled = true;
    pthread_mutex_unlock(&ctx->module_lock);
    return NULL;
}
```

That addresses the deadlock scenario described in this issue. I am closing this as fixed. If a current version still has a KSI locking problem, please open a new issue with the current code path and reproduction details.

PR #4689, opened 2021-09-20
ksi bugfix: locking bug fixed in rsksiCtxOpenFile

taavi-valjaots:
Fix for issue #4678.

Thread lock in `rsksiCtxOpenFile` is released when KSI module initialization fails and this function is called.


rgerhards:
@taavi-valjaots thx a lot!

rgerhards:
@taavi-valjaots CI failures seem to be unrelated issues. I'll look at them.

PR #4691, opened 2021-09-20
testbench bugfix: no pid file wait on elasticsearch startup

rgerhards:
The testbench framework does not properly wait until ES has created
its pid file, which probably means it did basic initializiation.
This can cause test synchronization issues and ultimately failures.


rgerhards:
Ubuntu 18 CI failures are false positives (and being addressed in other PRs)

PR #4701, opened 2021-09-29
build issue: handle undefined MAXPATHLEN

rgerhards:
While we handled missing PATH_MAX, we did not handle missing MAXPATHLEN.
This happens under GNU/Hurd, because there is no official limit. However,
extremely long pathes are extremely uncommon, so we do not want to
use slow dynamic alloc each time we need to build pathes. So we
impose a limit of 4KiB, which should be fairly enough. Note that
this obviously increases stack requirements in GNU/Hurd.


mbiebl:
A few comments. The following places seem to contain references to PATH_MAX
```
runtime/modules.c:#ifndef PATH_MAX
runtime/modules.c:#	define PATH_MAX MAXPATHLEN
runtime/modules.c:#	ifdef PATH_MAX
runtime/modules.c:	uchar pathBuf[PATH_MAX+1];
plugins/imfile/imfile.c:	uchar ourPath[PATH_MAX];
plugins/imfile/imfile.c:	uchar name[PATH_MAX];
plugins/imjournal/imjournal.c:	 * TODO: figure out a better way to avoid the PATH_MAX
ChangeLog:- enhanced module loader to not rely on PATH_MAX
ChangeLog:- enhanced module loader to not rely on PATH_MAX
contrib/omtcl/omtcl.c:	char fileName[PATH_MAX+1];
contrib/omtcl/omtcl.c:	if(getSubString(&p, fileName, PATH_MAX+1, ',') || getSubString(&p, buffer, 4096, ';') || !strlen(buffer)) {
```

So we have runtime/modules.c, plugins/imfile/imfile.c and contrib/omtcl/omtcl.c. 
Enough places to maybe move this DEFINE into config.h so it only needs to be set once?

Also, if you always set PATH_MAX and MAXPATHLEN, why not simplify this code in modules.c
```
#	ifdef PATH_MAX
	uchar pathBuf[PATH_MAX+1];
#	else
	uchar pathBuf[4096];
#	endif
```
and just use `uchar pathBuf[PATH_MAX+1];`

mbiebl:
Related here: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=900718

rgerhards:
thx @mbiebl for taking note and the suggestion. I now have implemented it.

PR #4705, opened 2021-10-07
tcpsrv: fix compilation without exceptions

orgads:
```
tcpsrv.c:992:1: error: label at end of compound statement
 finalize_it:
 ^~~~~~~~~~~
```

Quoting from pthread.h:
>   pthread_cleanup_push and pthread_cleanup_pop are macros and must always
>   be used in matching pairs at the same nesting level of braces.

Amends commit bcdd220142ec9eb106550195ba331fd114adb0bd.


orgads:
This is the previous run: https://build.rsyslog.com/#/builders/175/builds/3799

PR #4708, opened 2021-10-11
Fix Github workflow file

orgads:


rgerhards:
@orgads just to be sure: the "fix" is the additional comment, so nothing really changed? Just asking as it looks a bit like the patch is partial...

orgads:
It's not an additional comment. I just fixed the indentation. The YAML file was broken.

orgads:
`daqueue-invld-qi.sh` fails, but obviously it's unrelated to this change. The same test failed in #4705.

alorbach:
bGlblSrvrInitDone is a leftover from the refactoring and the variable can be fully removed.

@orgads could you rebase your branch on the current master? I tried checking out your branch but it did not have the "sndrcv_tls_ossl_anon_ciphers.sh" test merged (Was added recently).

rgerhards:
@orgads thx

@alorbach I think removing the variable doesn't look like the right cure. We need to check why the race is triggered, this sounds pretty odd.

alorbach:
I will look into the the anon init issue in a seperated pr

orgads:
I moved the removal of bGlblSrvrInitDone to #4715, and set `continue-on-error` for this job, so you can merge it and investigate the failure afterwards.

When you solve it, just remove the `continue-on-error`.

rgerhards:
Thx again for the PR and your persistence!

PR #4728, opened 2021-11-05
NEW CONTRIB::FFAUP:: URL parser module function using libfaup

frikilax:
## Description
New module function(s) (RainerScript) to parse and extract parts of an URL.

These new functions allow anybody to parse any variable containing URLs, hostnames, DNS queries and such to extract:
- the scheme
- the credentials (if present)
- the tld (with support for second-level TLDs)
- the domain (with and without the tld)
- the subdomain
- the full hostname
- the port (if present)
- the resource path (if present)
- the query string parameters (if present)
- the fragment (if present)

## Requirements
This module relies on the [faup](https://github.com/stricaud/faup) library.
The library should be compiled (see link for instructions on how to compile) and installed on build and runtime machines.

| :exclamation:  Warning   |
|--------------------------|
Even if faup is statically compiled to rsyslog, the library still needs an additional file to work properly: the mozilla.tlds stored by the libfaup library in /usr/local/share/faup. It permits to properly match second-level TLDs and allow URLs such as www.rsyslog.co.uk to be correctly parsed into \<rsyslog:domain\>.\<co.uk:tld\> and not \<rsyslog:subdomain\>.\<co:domain\>.\<uk:tld\>

## HowTo
The module functions are fairly simple to use, and are divided in 2 classes:
- `faup()` allows to parse the entire URL and return all parts in a complete JSON
- `faup_<field>()` allows to parse the entire URL, but only returns the (potential) value of the requested field as string

#### using the `faup()` function
The `faup()` is the simplest function to use, simply provide a value or variable (any type) as the only parameter, and the function returns a json object containing every element of the URL.

*example code:*
```
set $!url = "https://user:pass@www.rsyslog.com:443/doc/v8-stable/rainerscript/functions/mo-faup.html?param=value#faup";

set $.faup = faup($!url);
```

*$.faup will contain:*
```json
{
	"scheme": "https",
	"credential": "user:pass",
	"subdomain": "www",
	"domain": "rsyslog.com",
	"domain_without_tld": "rsyslog",
	"host": "www.rsyslog.com",
	"tld": "com",
	"port": "443",
	"resource_path": "\/doc\/v8-stable\/rainerscript\/functions\/mo-faup.html",
	"query_string": "?param=value",
	"fragment": "#faup"
}
```

| :memo:     Note      |
|---------------------|
This is a classic rsyslog variable, and you can access every sub-key with `$.faup!domain`, `$.faup!resource_path`, etc...

#### using the `faup_<field>()` functions
Using the field functions is even simpler: for each field returned by the `faup()` function, there exists a corresponding function to get only that one field.
For example, if the goal is to recover the domain without the tld, the example above could be modified as follows:
*example code:*
```
set $!url = "https://user:pass@www.rsyslog.com:443/doc/v8-stable/rainerscript/functions/mo-faup.html?param=value#faup";

set $.faup = faup_domain_without_tld($!url);
```

*$.faup will contain:*
```
rsyslog
```

| :memo:     Note      |
|---------------------|
The returned value is no longer a json object, but a simple string

## Motivations
Those functions are the answer to a growing problem encountered in Rsyslog when using modules to enrich logs : some mechanics (like lookup tables or external module calls) require "strict" URL/hostname formats that are often not formatted correctly, resulting in lookup failures/misses.
This ensures getting stable inputs to provide to lookups/modules to enrich logs.

## Documentation
rsyslog-doc : https://github.com/rsyslog/rsyslog-doc/pull/952

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -402,6 +402,28 @@ if test "$enable_fmhash_xxhash" = "yes"; then
 fi
 
 
+#faup header checks
+AC_ARG_ENABLE(ffaup,
+	[AS_HELP_STRING([--enable-ffaup],[Enable ffaup support @<:@default=no@:>@])],
+        [case "${enableval}" in
+         yes) enable_ffaup="yes" ;;
+          no) enable_ffaup="no" ;;
+           *) AC_MSG_ERROR(bad value ${enableval} for --enable-ffaup) ;;
+         esac],
+	[enable_ffaup=no]
+)
+
+AM_CONDITIONAL(ENABLE_FFAUP, test x$enable_ffaup = xyes)
+if test "$enable_ffaup" = "yes"; then
+	AC_CHECK_LIB([faupl], [faup_init], [
+		AC_CHECK_HEADER([faup/faup.h], [
+			FAUP_LIBS="-lfaupl"
+			AC_SUBST(FAUP_LIBS)],
+			[AC_MSG_ERROR([Unable to add faup support for URL parsing.])])
+	])
+fi
+
+
 # rscript function unflatten
 AC_ARG_ENABLE(fmunflatten,
 	[AS_HELP_STRING([--enable-fmunflatten],[Enable fmunflatten support @<:@default=no@:>@])],
@@ -2705,6 +2727,7 @@ AC_CONFIG_FILES([Makefile \
 		contrib/omhttp/Makefile \
 		contrib/fmhash/Makefile \
 		contrib/fmunflatten/Makefile \
+		contrib/ffaup/Makefile \
 		contrib/pmsnare/Makefile \
 		contrib/pmpanngfw/Makefile \
 		contrib/pmaixforwardedfrom/Makefile \
@@ -2853,6 +2876,7 @@ echo "    fmhttp enabled:                           $enable_fmhttp"
 echo "    fmhash enabled:                           $enable_fmhash"
 echo "    fmhash with xxhash enabled:               $enable_fmhash_xxhash"
 echo "    fmunflatten enabled:                      $enable_fmunflatten"
+echo "    ffaup enabled:                            $enable_ffaup"
 echo
 echo "---{ debugging support }---"
 echo "    distcheck workaround enabled:             $enable_distcheck_workaround"
--- a/Makefile.am
+++ b/Makefile.am
@@ -202,6 +202,10 @@ if ENABLE_FMUNFLATTEN
 SUBDIRS += contrib/fmunflatten
 endif
 
+if ENABLE_FFAUP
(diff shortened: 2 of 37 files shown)


frikilax:
Note to maintainers : the library used does not have packages for many distributions (I could only find a FreeBSD package), so testing will require compilation of the project.
The library doesn't need anything other than building tools to compile, and is fairly fast to do so, so there shouldn't be any problems on that part.
Feel free to contact me if things needs to be changed/for questions.

frikilax:
Also : hello @stricaud ! 👋

rgerhards:
Thx! A first quick comment: I think the module should not be enabled by default.

I will also look into the requirements to run the tests.

stricaud:
Thanks for the ping @frikilax this looks very cool. Happy to help with anything if I can make the integration smoother.

configure.ac
```diff
@@ -402,6 +402,32 @@ if test "$enable_fmhash_xxhash" = "yes"; then
...
+        [case "${enableval}" in
+         yes) enable_ffaup="yes" ;;
+          no) enable_ffaup="no" ;;
+           *) AC_MSG_ERROR(bad value ${enableval} for --enable-ffaup) ;;
+         esac],
+	[enable_ffaup=yes]
+)
+
+AM_CONDITIONAL(ENABLE_FFAUP, [test "x$enable_ffaup" = "xyes"])
+
```
rgerhards: looks duplicated ;-)

rgerhards (changes requested):
I don't see exactly what is wrong, but the faup module is obviously also compiled if not selected. Even if selected, and libfaup is present, compilation fails (probably due to invalid USE_FAUP definition).

rgerhards:
Ubuntu 20 CI system now has libfaup.

rgerhards:
I fixed part of the configure.ac issue, but maintainer pushes to the repro are not permitted.  For starters, in line 413 of configure.ac you need to change to `[enable_ffaup=no]`.

rgerhards:
@frikilax Please have a look at the Ubuntu 20 and Fedora 34 CI failures. There are compile issues, at quick look it looks just like a little bit cleanup needs to be done on the code. Ubuntu 20 and F 34 are currently the only environments where libfaup is present. I'll enable others most probably when CI passes on these environments (maybe early, depending on workload).

configure.ac
frikilax: Solved !

rgerhards:
@frikilax The Fedora 34 fail looks like something is not really clean in regard to the tests (or some other platform foo...)

rgerhards:
@frikilax hold on, this may be a problem inside the container - looking at it

rgerhards (approved):

Issue #4725, opened 2021-11-01
mmanon: IPv6 addresses followed by a port (in not recommended from) are not rewritten properly 


xshadow:
Dear rsyslog team, thanks for your nice software. I found a bug in mmanons IPv6 log anonymization. The IPv6 log anonymization is not recognizing IPv6 addresses properly when the IPv6 address is followed by a port. 

If a 4 digit port is append to the IPv6 address, the first 4 digits are not rewritten, instead the port gets obfuscated.

```
1a00:c820:1180:c84c:ad3f:4024:d991:ec2e:4922 becomes 1a00:0:0:0:0:0:0:0:0
``` 

If a 5 digit port is append to the IPv6 address, the complete IPv6 address is written to the logfile.

```
1a00:c820:1180:c84c:ad3f:4024:d991:ec2e:49225 stays 1a00:c820:1180:c84c:ad3f:4024:d991:ec2e:49225
```

### Expected behavior

I would expect that the IPv6 address is properly removed from the syslog file.

### Actual behavior

An IPv6 address followed by a five digit port is not recognized as IPv6 address and the log filter is not applied.
Example log message: 

```
Nov  1 16:35:44 webserver apache: [Mon Nov 01 16:35:44.705418 2021] [authz_core:error] [pid 3195941:tid 139935243814656] [client ca00:6920:e180:c12c:ed3f:3024:d991:ec2e:64833] AH01630: client denied by server configuration: /var/www/www.mysite.to/www/xmlrpc.php`
```

### Steps to reproduce the behavior

```
ansible@host:~# logger "Catched IPv6 without port 1a00:c820:1180:c84c:ad3f:4024:d991:ec2e"
ansible@host:~# logger "Catched IPv6 with 4digit port 1a00:c820:1180:c84c:ad3f:4024:d991:ec2e:4922"
ansible@host:~# logger "Uncatched IPv6 with 5digit port 1a00:c820:1180:c84c:ad3f:4024:d991:ec2e:49225"

# File: /var/log/syslog

Nov  1 16:52:01 host ansible: Catched IPv6 without port 0:0:0:0:0:0:0:0
Nov  1 16:52:13 host ansible: Catched IPv6 with 4digit port 1a00:0:0:0:0:0:0:0:0
Nov  1 16:52:27 host ansible: Uncatched IPv6 with 5digit port 1a00:c820:1180:c84c:ad3f:4024:d991:ec2e:49225
```

### Environment
```
[...]

PR #4735, opened 2021-11-17
mmanon: relax IPv6 detection - improve anonymization

rgerhards:
We so far tried to ensure a value is really an IPv6 address, in order
to avoid to mangle with just similar-looking information elements.
However, this lead to misdetection for unusual formats, e.g. when a
port is appended to a numerical IPv6 adress given without braces [].
This has been changed now. In a sense, we now prefer to err on the
side of privacy.

BEHAVIOR CHANGE:
Previously, a suspect value was not anonymized, and thus some other
elements (like some MAC addresses) preserved. Now the opposite is
true, and we anonymize anything that looks close enough to be an
IPv6 address. This improves anonymization.

closes https://github.com/rsyslog/rsyslog/issues/4725

### DIFF
--- a/plugins/mmanon/mmanon.c
+++ b/plugins/mmanon/mmanon.c
@@ -577,6 +577,10 @@ syntax_ipv6(const uchar *const __restrict__ buf,
 				goto done;
 			}
 		}
+		if(ipParts == 8 && !hadAbbrev) {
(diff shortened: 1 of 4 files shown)


xshadow:
Thank you!

Issue #4646, opened 2021-07-28
Possible memory leak or false positive: tellModulesActivateConfigPrePrivDrop 


alorbach:
Librelp testbench fails with test rscript_http_request-vg.sh which is totally unrelated to librelp. 

> ==32374== 
> ==32374== HEAP SUMMARY:
> ==32374==     in use at exit: 97,573 bytes in 718 blocks
> ==32374==   total heap usage: 4,725 allocs, 4,007 frees, 2,488,703 bytes allocated
> ==32374== 
> ==32374== 24 bytes in 1 blocks are definitely lost in loss record 11 of 54
> ==32374==    at 0x483DD99: calloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
> ==32374==    by 0x61409F5: ???
> ==32374==    by 0x1475B2: tellModulesActivateConfigPrePrivDrop (rsconf.c:718)
> ==32374==    by 0x1475B2: activate (rsconf.c:886)
> ==32374==    by 0x122974: initAll (rsyslogd.c:1634)
> ==32374==    by 0x120020: main (rsyslogd.c:2143)
> ==32374== 
> {
>    <insert_a_suppression_name_here>
>    Memcheck:Leak
>    match-leak-kinds: definite
>    fun:calloc
>    obj:*
>    fun:tellModulesActivateConfigPrePrivDrop
>    fun:activate
>    fun:initAll
>    fun:main
> }
> ==32374== LEAK SUMMARY:
> ==32374==    definitely lost: 24 bytes in 1 blocks
> ==32374==    indirectly lost: 0 bytes in 0 blocks
> ==32374==      possibly lost: 0 bytes in 0 blocks
> ==32374==    still reachable: 2,684 bytes in 22 blocks
> ==32374==         suppressed: 94,865 bytes in 695 blocks
> ==32374== Reachable blocks (those to which a pointer was found) are not shown.
> ==32374== To see them, rerun with: --leak-check=full --show-leak-kinds=all
> ==32374== 
> ==32374== For lists of detected and suppressed errors, rerun with: -s
> ==32374== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 7 from 7)
> rsyslogd run exited with 10
> valgrind run FAILED with exceptions - terminating
> not reporting failure as RSYSLOG_STATSURL is not set
[...]

PR #4738, opened 2021-11-22
imtcp: Fix copy&paste error for pszLstnPortFileName init

alorbach:
When no listener portfile is configure, the wrong cnf_param
was initialized to NULL.

closes: https://github.com/rsyslog/rsyslog/issues/4646


tests/rscript_http_request-vg.sh
```diff
@@ -1,3 +1,5 @@
 #!/bin/bash
 export USE_VALGRIND="YES"
+#export RS_TEST_VALGRIND_EXTRA_OPTS="--leak-resolution=high --num-callers=100 --track-origins=yes --leak-check=full"
```
rgerhards: does this need to be present permanently?

alorbach: It was helpful for extending valgrind output, so it could be useful for future valgrind tests. But it can be removed.

Issue #4378, opened 2020-08-18
oversized message output is lacking detail


davidelang:
I'm using the oversizedmessage capability for the first time and was disappointed in the level of detail put into the oversized.file 
it has the rawmsg and the input and nothing else.

I would like to either be able to define a template to use for writing to the oversized file or have it write all-json so that it includes things like fromhost-ip, timegenerated, timereported, etc.

My first testing was oversized data from imfile, no timestamp, not tag, no filename metadata, just the raw message and input: imfile

it's better than nothing, but far less useful than it could be.

PR #4743, opened 2021-11-27
Avoid use of loadConf in favor of runConf in ratelimiting during runtime

Cropi:
The ```ratelimitNew()``` function used  the ```bReduceRepeatMsgs``` global option to acquire information from the configuration that is going to be loaded. While this function was created to be used(I believe) during "configuration loading time", multiple input plugins use it during runtime. It's true that it has no impact on the current system right now, because both ```loadConf``` and ```runConf``` point to the same config. However, in the future a dynamic configuration reload option might be added.

### DIFF
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -773,6 +773,7 @@ act_obj_add(fs_edge_t *const edge, const char *const name, const int is_file,
 	act->file_id[0] = '\0';
 	act->file_id_prev[0] = '\0';
 	act->is_symlink = is_symlink;
+	act->ratelimiter = NULL;
 	if (source) { /* we are target of symlink */
 		CHKmalloc(act->source_name = strdup(source));
 	} else {
@@ -800,6 +801,8 @@ act_obj_add(fs_edge_t *const edge, const char *const name, const int is_file,
 finalize_it:
 	if(iRet != RS_RET_OK) {
 		if(act != NULL) {
+			if (act->ratelimiter != NULL)
+				ratelimitDestruct(act->ratelimiter);
 			free(act->name);
 			free(act);
 		}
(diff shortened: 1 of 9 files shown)


Cropi:
I believe that https://github.com/rsyslog/rsyslog/pull/4738 will fix the valgrind related issues.

rgerhards:
I am a bit puzzled by the two platforms with timeout / many failures. Doesn't make sense to me, even in regard to #4378. There is also an issue with F33 on github, it often times out without any visible reason (that's the case on #4378, let's give it one more try). I'll merge #4378 later and then let's see what happens.

rgerhards:
I have merge #4378 and re-initiated failed CI runs

Cropi:
I've ran out of ideas on why those two specific platforms keep getting timeout failures.
- buildbot/rsyslog gen kafka distcheck 
- buildbot/rsyslog docker-ubuntu18 GnuTLS only

rgerhards:
sorry for the late reply, I was totally busy with some community help to fight covid.

I'll look at CI (hint: github workers also seem to have some issues) and will also go through the rest of this PR!

rgerhards:
@Cropi I am still fighting with CI, have initiated some related tests in separate PR.

If you need this one merged, just let me know and I'll do. Else I see to fix the CI issue first, so that it looks clean.

rgerhards:
Now that I wrote this, and had another look, I think it occurred to me what the problem is: https://github.com/rsyslog/rsyslog/pull/4751 -- so hope this is fixed a bit later today!

Cropi:
That's good news. In that case I will wait for #4751 to be merged. Thank you.

rgerhards:
@Cropi CI is currently really running wild in unexpected ways :-( Anyhow, I have merged my support PR, so let's no see how it works out. If I still see issues here, I'll merge anyhow. Takes roughly ~2 hours to go full cycle.

rgerhards:
we got it! :-)

Cropi:
Thanks a lot:)

PR #4745, opened 2021-11-30
Deallocate outchannel resources in rsconf destructor

Cropi:
The outchannel resources were not freed when rsyslog was stoped. The PR intends to fix it by invoking ```ochDeleteAll()``` in the rsconf destructor.


rgerhards:
now working on this one

PR #4751, opened 2021-12-13
testbench: do not use valgrind --keep-debuginfo=yes by default

rgerhards:
This option is not known by older valgrind versions. As such, it causes
false positives on older platforms. As the option is not necessary
(only helpful for some border cases), we remove it and add a comment
so that it may be enabled when necessary.


rgerhards:
Merging PR, as CI Failure looks highly unrelated and we cannot go forward without a merge.

PR #4754, opened 2021-12-15
Clarify the meaning of loadConf and RunConf

Cropi:
The PR intends to clarify the meaning of loadConf and runConf. These were originally introduced to facilitate a later move to dynamic config loads.  Over the years, the use of these pointers has become unclear. 

- loadConf was intended for candidate config (the new one)
- runConf for the currently running config

### DIFF
--- a/action.c
+++ b/action.c
@@ -393,7 +393,7 @@ rsRetVal actionConstruct(action_t **ppThis)
 	action_t *pThis;
 
 	assert(ppThis != NULL);
-	
+
 	CHKmalloc(pThis = (action_t*) calloc(1, sizeof(action_t)));
 	pThis->iResumeInterval = 30;
 	pThis->iResumeIntervalMax = 1800; /* max interval default is half an hour */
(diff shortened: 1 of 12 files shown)


rgerhards:
The systemd check is a CI issue. Some images seem to have changed, I am investigating this. Not a merge blocker.

Note: I'll probably move this to the Feb '22 release as I am preparing Dec today (to keep a bit away from xmas).

Cropi:
Sure, take your time. There is no need to rush, especially before the holidays:)

Cropi:
Regarding the "buildbot/rsyslog ElasticSearch codecov" failing test, is there a way to find out which test did not pass?

rgerhards:
> Regarding the "buildbot/rsyslog ElasticSearch codecov" failing test, is there a way to find out which test did not pass?

Maybe the CI Worker was overwhelmed (unlikely, but). To check, I reinitiated a new run.

rgerhards:
Note: Journal CI probably requires CI worker update. Ignoring issue for now.

PR #4760, opened 2021-12-28
Move rsyslog global variables to rsconf_t structure

Cropi:
The list of global parameters should be part of the ```rsconf_t``` configuration structure, so we can later distinguish between the actually running config(runConf) and the config that's being loaded(loadConf). Note that some of them are already part of it.
As soon as this way of handling global parameters is accepted, I will continue with other parameters as well.
The idea behind commit efbd2908e209772b27597a6eb5bd17754143df6e is to extend getters with an additional parameter, an rsconf_t structure specifying the config. With this approach, the caller will be able to query either ```loadConf``` or ```runConf``` . The setter always uses the ```loadConf``` structure, because right now I don't see a reason to modify it during "runtime". The other two commits demonstrate how I intend to apply changes on ```workDirectory``` and ```dropMsgsWithMaliciousDNSPtrRecords```.

Note that extending existing functions with additional parameter changes the interface structure. I would like to know whether it's feasible to do it, or it's a bad approach that should be not considered at all.

List of global parameters for reference: https://www.rsyslog.com/doc/master/rainerscript/global.html

### DIFF
--- a/plugins/imgssapi/imgssapi.c
+++ b/plugins/imgssapi/imgssapi.c
@@ -59,7 +59,7 @@
 #include "glbl.h"
 #include "debug.h"
 #include "unlimited_select.h"
-
+#include "rsconf.h"
 
 MODULE_TYPE_INPUT
 MODULE_TYPE_NOKEEP
@@ -203,7 +203,7 @@ onSessAccept(tcpsrv_t *pThis, tcps_sess_t *pSess)
 {
 	DEFiRet;
 	gsssrv_t *pGSrv;
-	
+
 	pGSrv = (gsssrv_t*) pThis->pUsr;
 
 	if(pGSrv->allowedMethods & ALLOWEDMETHOD_GSS) {
@@ -438,7 +438,7 @@ OnSessAcceptGSS(tcpsrv_t *pThis, tcps_sess_t *pSess)
 	allowedMethods = pGSrv->allowedMethods;
 	if(allowedMethods & ALLOWEDMETHOD_GSS) {
 		int ret = 0;
-		const size_t bufsize = glbl.GetMaxLine();
+		const size_t bufsize = glbl.GetMaxLine(runConf);
 		CHKmalloc(buf = (char*) malloc(bufsize + 1));
 
 		prop.GetString(pSess->fromHostIP, &pszPeer, &lenPeer);
@@ -586,7 +586,7 @@ OnSessAcceptGSS(tcpsrv_t *pThis, tcps_sess_t *pSess)
 		gssutil.display_ctx_flags(*sess_flags);
 		pGSess->allowedMethods = ALLOWEDMETHOD_GSS;
 	}
-	
+
 finalize_it:
 	free(buf);
 	RETiRet;
--- a/tools/rsyslogd.c
+++ b/tools/rsyslogd.c
@@ -320,7 +320,7 @@ checkStartupOK(void)
 		fprintf(stderr, "rsyslogd: error reading pid file, cannot start up\n");
 		ABORT_FINALIZE(RS_RET_ERR);
 	}
-	
+
 	/* ok, we got a pid, let's check if the process is running */
 	const pid_t pid = (pid_t) pf_pid;
 	if(kill(pid, 0) == 0 || errno != ESRCH) {
@@ -559,7 +559,8 @@ rsyslogd_InitStdRatelimiters(void)
 	CHKiRet(ratelimitNew(&dflt_ratelimiter, "rsyslogd", "dflt"));
 	CHKiRet(ratelimitNew(&internalMsg_ratelimiter, "rsyslogd", "internal_messages"));
 	ratelimitSetThreadSafe(internalMsg_ratelimiter);
-	ratelimitSetLinuxLike(internalMsg_ratelimiter, glblIntMsgRateLimitItv, glblIntMsgRateLimitBurst);
+	ratelimitSetLinuxLike(internalMsg_ratelimiter,
+		loadConf->globals.intMsgRateLimitItv, loadConf->globals.intMsgRateLimitBurst);
 	/* TODO: make internalMsg ratelimit settings configurable */
 finalize_it:
 	RETiRet;
@@ -851,8 +852,8 @@ static void
 logmsgInternal_doWrite(smsg_t *pMsg)
 {
 	const int pri = getPRIi(pMsg);
-	if(pri % 8 <= glblIntMsgsSeverityFilter) {
-		if(bProcessInternalMessages) {
+	if(pri % 8 <= runConf->globals.intMsgsSeverityFilter) {
+		if(runConf->globals.bProcessInternalMessages) {
 			submitMsg2(pMsg);
 			pMsg = NULL; /* msg obj handed over; do not destruct */
 		} else {
@@ -865,13 +866,13 @@ logmsgInternal_doWrite(smsg_t *pMsg)
 			 */
 			static warnmsg_emitted = 0;
 			if(warnmsg_emitted == 0) {
-				stdlog_log(stdlog_hdl, LOG_WARNING, "%s",
+				stdlog_log(runConf->globals.stdlog_hdl, LOG_WARNING, "%s",
 					"RSYSLOG WARNING: liblogging-stdlog "
 					"functionality will go away soon. For details see "
 					"https://github.com/rsyslog/rsyslog/issues/2706");
 				warnmsg_emitted = 1;
 			}
-			stdlog_log(stdlog_hdl, pri2sev(pri), "%s", (char*)msg);
+			stdlog_log(runConf->globals.stdlog_hdl, pri2sev(pri), "%s", (char*)msg);
 			#else
 			syslog(pri, "%s", msg);
 			#endif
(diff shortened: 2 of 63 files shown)


rgerhards:
> Note that extending existing functions with additional parameter changes the interface structure. I would like to know whether it's feasible to do it, or it's a bad approach that should be not considered at all.

It's OK - we learned that if we adapt all plugins inside the rsyslog tree, this works decently well. There are no public plugins that we know about and only one or two private extensions for some companies, as far as the rumor says ;-)

rgerhards:
Note that the systemd check is still defunct, so ignore bad outcome from it.

Cropi:
That's good to hear :) thank you for the quick feedback.

Cropi:
Proposed implementation change is not (yet?) done for the following list of global parameters:
- debug.logFile
- debug.unloadModules
- debug.files
- stdlog.channelspec
- localHostname
- preserveFQDN
- maxMessageSize

The main reason behind this is because most of these parameters are needed even before our configuration object exists(loadConf) or after it has been destructed(runConf).

### Is there some specific date after CLI options which have equivalent global parameters will be removed?
During debugging the implemented changes, I discovered that multiple global parameters can be activated via program arguments, e.g. ```net.ipprotocol="ipv4-only"``` can be achieved by the "-4" command line option. Some comments state that similar CLI options will go away in the future, for reference see https://github.com/rsyslog/rsyslog/blob/master/tools/rsyslogd.c#L1406 . Is there some specific date after this will take place? The reason why I am asking this is because I would need to make additional changes to make it work with current design(Right now it's disabled).

Cropi:
> Proposed implementation change is not (yet?) done for the following list of global parameters:
> 
>     * maxMessageSize
>     * stdlog.channelspec
Done
>     * debug.logFile
>     * debug.unloadModules
>     * debug.files
At this point, I don't think it's necessary to implement this for debug related global parameters. When one of these parameters change during a dynamic config reload, no harm is done, new values will be used instead of the old ones even in situations where config contains errors and the rest of the new config is not applied.
>     * localHostname
>     * preserveFQDN
Instead of making these two part of the ```rsconf_t``` structure, it's better to leave them in their original form. We need them earlier than our config exists because we may emit error log messages, which need the correct hostname.
With this in mind, I think that the implementation is ready for review.
> 
> The main reason behind this is because most of these parameters are needed even before our configuration object exists(loadConf) or after it has been destructed(runConf).
> ### Is there some specific date after CLI options which have equivalent global parameters will be removed?
> 
> During debugging the implemented changes, I discovered that multiple global parameters can be activated via program arguments, e.g. `net.ipprotocol="ipv4-only"` can be achieved by the "-4" command line option. Some comments state that similar CLI options will go away in the future, for reference see https://github.com/rsyslog/rsyslog/blob/master/tools/rsyslogd.c#L1406 . Is there some specific date after this will take place? The reason why I am asking this is because I would need to make additional changes to make it work with current design(Right now it's disabled).

rgerhards:
> Is there some specific date after CLI options which have equivalent global parameters will be removed?

not really, but the warning is inside the doc for quite a while (couple of years). So I think we can conclude that there is now time and real reason to disregard the command line params (with a meaningful error message).

rgerhards:
I had a couple of days off during the holiday break (extensive in southern Germany ;-)). Now back and reviewing. Will see to get this merged today or tomorrow. Some points:

`stdlog.channelspec`, I think, is a left-over that we do no longer need (as liblogging-stdlog was removed)

> At this point, I don't think it's necessary to implement this for debug related global parameters. When one of these parameters change during a dynamic config reload, no harm is done, new values will be used instead of the old ones even in situations where config contains errors and the rest of the new config is not applied.

full ack, that's fine

> Instead of making these two part of the `rsconf_t` structure, it's better to leave them in their original form. We need them earlier than our config exists because we may emit error log messages, which need the correct hostname.

I tend to agree here as well. But I would not outrule that reports from practice will tell us differently over time.

rgerhards:
@Cropi you should remove the commented-out code that is shown in the LGTM report. I think it's clear we do not need it any longer - and we also could revert via git history in any case.

side-note: with a new push, systemd journal check should also succeed. I fixed the definitions file, but github does not seem to take the new one from master for CI runs.

plugins/imgssapi/imgssapi.c
```diff
@@ -59,7 +59,7 @@
 #include "glbl.h"
 #include "debug.h"
 #include "unlimited_select.h"
-
+#include "rsconf.h" /* TODO remove once load/runModConf is introduced */
```
rgerhards: This is just one occasion of many: is the comment right? I think we are currently introducing them - or do you think about shuffling to a different include?

runtime/strms_sess.c
```diff
@@ -288,7 +288,7 @@ BEGINObjClassInit(strms_sess, 1, OBJ_IS_CORE_MODULE) /* class, version - CHANGE
 	CHKiRet(objUse(prop, CORE_COMPONENT));
 
 	CHKiRet(objUse(glbl, CORE_COMPONENT));
-	iMaxLine = glbl.GetMaxLine(); /* get maximum size we currently support */
+	iMaxLine = 8096; /* get maximum size we currently support */
```
rgerhards: This does not look right, given the previous code. Is it that at this stage we do not have the config object fully populated? In any case, if I am not mistaken, we should have one. So better put the constant in it's initialization. Or am I overlooking something (might very well be, haven't looked at this code area for quite a while).

tools/rsyslogd.c
```diff
@@ -1524,13 +1525,13 @@ initAll(int argc, char **argv)
 			fprintf (stderr, "rsyslogd: the -w command line option will go away "
 				 "soon.\nPlease use the global(net.permitWarning=\"off\") "
 				 "configuration parameter instead.\n");
-			glbl.SetOption_DisallowWarning(0);
+			// glbl.SetOption_DisallowWarning(0);
```
rgerhards: as mentioned in previous PR-wide comment, the commented out code should go away. Error message should be reworded to "Option *has* gone away".

rgerhards (changes requested):
overall, looks good to me - also some nice cleanup.

Please have a look at the mentioned comments (mostly cosmetic).

plugins/imgssapi/imgssapi.c
Cropi: I am not entirely sure but left it there for some later discussion.
Some(just a few) plugins do not explicitly require to include this header, especially those which have a pointer to the ```rsconf_t``` structure in their ```modConfData_s``` struct. Sometimes a plugin needs to query the value of a global parameter, in that case I can see 2 ways on how to get that:
- use loadConf/runConf global variables from rsconf.c
- use the plugin's pointer from loadModConf/RunModConf()
Which approach do you prefer?

I will remove the comments then :)

runtime/strms_sess.c
Cropi: Yes, you are right. I remember that I set it to a constant value when I was debugging something and then forgot about it. I am not familiar with this part of code, is it even used? I could not find any references where the ```strms_sess```  initializer is even invoked(although I might overlooked something). The last call I could found has been removed via commit https://github.com/rsyslog/rsyslog/commit/ae8d9853be406a2b5dd8b7353ab0fc6f99e41b3c.

rgerhards: let me dig into it, will take a short while

rgerhards:
I forgot to mention. Once we are ready to merge, squashing into a single commit would be preferable to me over doing the squash via github.

runtime/strms_sess.c
rgerhards: I overlooked it's the class initializer (yeah, you said it... ;-). In this case: yes it is called, it's done by the module loader.

But: I don't see that `iMaxline` is actually being used! I'll craft a small PR to remove it and see if something really strange comes up. Should not be the case as it is a static variable and nowhere else referenced.

rgerhards: Indeed, `iMaxLine` was no longer used. It's gone now. If you rebase to master, you have one problem less ;-)

Cropi: I updated the branch with the commits from master and also applied changes according to your comments :)

plugins/imgssapi/imgssapi.c
rgerhards: > use the plugin's pointer from loadModConf/RunModConf()

1. I would tend to say that this is cleaner. IIRC this also was the original intent. I don't know, however, if this has some drawbacks with special modules when going forward with later steps. But that will probably happen anyway if it does at all ;-) I still hope the answer is helpful.

rgerhards:
@Cropi I think we are more or less ready to merge - do you agree or have some more work in this PR?

FYI: My comment on which config to use can also go into later PRs, as far as I am concerned.

Cropi:
> @Cropi I think we are more or less ready to merge - do you agree or have some more work in this PR?
> 
> FYI: My comment on which config to use can also go into later PRs, as far as I am concerned.

I think we are ready with this PR. In my opinion that should be addressed in a separate pull request, it does not affect the functionality, it's more like some cosmetic changes:)

rgerhards:
> I think we are ready with this PR. In my opinion that should be addressed in a separate pull request, it does not affect the functionality, it's more like some cosmetic changes:)

full ack - I'll wait, solely for cosmetic reasons, on the Ubuntu 20 check run and will then merge. Thx again for all your hard work!

rgerhards (approved):

Issue #4741, opened 2021-11-25
plugin ompsql ssl connection


fdefilippo:
Hi, i did not find anywhere how to configure the ompsql plugin for ssl connections, is it possible?

$ rsyslogd -v
rsyslogd  8.2102.0-5.el8 (aka 2021.02) compiled with:
centos8

thank you.

alorbach:
You mean the "ompgsql" plugin?
There is no SSL support implemented yet as far as I can see

ericlenio:
Yes SSL is supported, as of [this commit](https://github.com/rsyslog/rsyslog/commit/e98e691ad82cde0e0ed68138cfe1f83bfeb638f6)

Side note: on my system, I use `$PrivDropToUser` and this created a conflict: during the postgresql SSL handshake, it always tries to access `$HOME/.postgresql/root.crt`, which results in `permission denied` (because `HOME=/root`, and the unprivileged user cannot read this directory), which makes the handshake fail. I worked around this in my config file by setting a fake value for HOME: `global(environment=["HOME=/no-home"])`.

rgerhards:
Apologies for the very late follow-up. This may no longer be relevant for the original reporter, but I am adding the answer for people who arrive here through search while we work through the backlog. We do not like to close issues just because of age; this is now documented/supported through the PostgreSQL connection string.

Current `ompgsql` supports `conninfo`, and PostgreSQL TLS settings can be passed through the libpq connection string/URI, for example with `sslmode=verify-full` and `sslrootcert=...`. The current documentation includes examples for `conninfo`-based configuration.

So the practical answer is: configure TLS through PostgreSQL/libpq connection parameters, not through a separate rsyslog-specific TLS block for `ompgsql`.

Closing this as answered/implemented. If a current `ompgsql` `conninfo` TLS configuration does not work, please open a new issue with the exact `conninfo` value with secrets redacted, rsyslog version, libpq/PostgreSQL version, and the server-side TLS error.

Issue #4698, opened 2021-09-23
ompgsl plugin should support DNS Names that exceed 64 characters in length


androiddrew:
### Expected behavior

ompgsl accepts a DNS Name greater than 64 characters. This is critical because AWS Endpoints and K8s DNS names can routinely exceed this limit. 

### Actual behavior

```
rsyslogd: error during parsing file /development//local/ip-10-94-162-10.ec2.internal/rsyslog/config/rsyslog.conf, on or before line 48: ompgsql: srv parameter longer than supported maximum of 64 characters [v8.40.0 try https://www.rsyslog.com/e/2207 ]
rsyslogd: error during parsing file /development//local/ip-10-94-162-10.ec2.internal/rsyslog/config/rsyslog.conf, on or before line 48: errors occured in file '/development/local/ip-10-94-162-10.ec2.internal/rsyslog/config/rsyslog.conf' around line 48 [v8.40.0 try https://www.rsyslog.com/e/2207 ]
rsyslogd: there are no active actions configured. Inputs would run, but no output whatsoever were created. [v8.40.0 try https://www.rsyslog.com/e/2103 ]
rsyslogd: run failed with error -2103 (see rsyslog.h or try https://www.rsyslog.com/e/2103 to learn what that number means)
```

### Steps to reproduce the behavior

Configure ompgsql plugin with a target Postgres server DNS name exceeding 64 characters.

skupkosk:
Hi Drew,
make sure K8S really exceeds the limit.

K8S documentation mention the 63 character limit for hostname according the
RFC and does not break it.
https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

Peter

PR #4762, opened 2021-12-31
ompgsql: allow connection params via connection string

inahga:
Happy new years! Add another action parameter `conninfo` that allows specifying
a postgres connection string. This enables us to use any of the postgres connection
parameters, particularly `sslmode` and `sslrootcert`.

Per the postgres docs, this connection string can be a URI or several key-value
pairs. https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING

Here's an example:
```
module(load="ompgsql")

action(
	type="ompgsql"
	conninfo="postgresql://postgres:password@localhost/Syslog?sslmode=require"
)
```

This addresses https://github.com/rsyslog/rsyslog/issues/4741 by allowing the
user to specify ssl options as part of the connection string. `libpq` will take
care of the rest.

This also addresses https://github.com/rsyslog/rsyslog/issues/4698 because
`libpq` is not constrained by MAXHOSTNAMELEN. Long hostnames will work.

`conninfo` can be specified in lieu of the other parameters. `ompgsql` will
prioritize using `conninfo` to connect over the other parameters.

If this is accepted, I can make updates to the docs https://github.com/rsyslog/rsyslog-doc

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -891,6 +891,11 @@ if test "x$enable_pgsql" = "xyes"; then
 	    [-L`$PG_CONFIG --libdir`]
 	  )
 	])
+	AC_CHECK_LIB(
+	  [pq],
+	  [PGsslInUse],
+	  [AC_DEFINE([HAVE_PGSSLINUSE], [1], [PGsslInUse function available])]
+	)
 fi
 AM_CONDITIONAL(ENABLE_PGSQL, test x$enable_pgsql = xyes)
 AC_SUBST(PGSQL_CFLAGS)
--- a/plugins/ompgsql/ompgsql.c
+++ b/plugins/ompgsql/ompgsql.c
@@ -229,7 +229,11 @@ static rsRetVal initPgSQL(wrkrInstanceData_t *pWrkrData, int bSilent)
 		iRet = RS_RET_SUSPENDED;
 	}
 
+#ifdef HAVE_PGSSLINUSE
 	sslStatus = PQsslInUse(pWrkrData->f_hpgsql);
+#else
+	sslStatus = PQgetssl(pWrkrData->f_hpgsql) == NULL ? 0 : 1;
+#endif
 	dbgprintf("initPgSQL: ssl status: %d\n", sslStatus);
 
 	RETiRet;
--- a/plugins/ompgsql/ompgsql.c
+++ b/plugins/ompgsql/ompgsql.c
@@ -59,10 +59,11 @@ MODULE_CNFNAME("ompgsql")
 DEF_OMOD_STATIC_DATA
 
 typedef struct _instanceData {
-	char            srv[MAXHOSTNAMELEN+1];   /* IP or hostname of DB server*/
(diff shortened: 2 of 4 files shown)


inahga:
Also note that the valgrind test I added fails, but that one's not my fault :). There's a pre-existing memory leak, and it looks like postgres tests aren't in CI.

rgerhards:
Thanks for the PR, much appreciated!

> Also note that the valgrind test I added fails, but that one's not my fault :). There's a pre-existing memory leak that fails all postgres tests, and it looks like postgres tests aren't in CI.

would you be willing to have a look at the root cause of the memleak? If not, could you put the test into a separate PR, so that we can merge this one here without totally breaking CI. Our postgres know-how is limited, so it may take some extra time to get hold of the memleak.

rgerhards:
@inahga on quick look, the PR looks fine, and is definitely something we would like to merge. Work on doc is appreciated.

rgerhards:
I guess the CentOS 7 CI fail means that the function in question is not always present and it must be checked, during configure time, if it is available. I can lend a hand doing this. Just let me know.

rgerhards:
@inahga any interest in fixing the issue?

inahga:
Yes! Just as free time allows :)

rgerhards:
> Yes! Just as free time allows :)

thx, just wanted to be sure. Should I look into the configure check or is this something you would like to do? I am not pushy, just a honest question (these kind of things gave me a really hard time when I started ;-)).

inahga:
I'll give it a solid try--good opportunity to learn autotools (I'm more of a Go developer :laughing:).

WRT the memory leak--interestingly I'm unable to reproduce it today. I had suspected it was a bug in `libpq` or `libssl`--the stack trace showed the leak originated there. I've since updated those packages on my system (F35). I tried with the `rsyslog_dev_base_fedora:34`, `rsyslog_dev_base_ubuntu:16.04`, `rsyslog_dev_base_centos:8` containers and no memory leak there either.

False alarm I think--I probably should have been using the containers in the first place :)

inahga:
OK @rgerhards I think I was successful. It was a _suspiciously_ simple change, let me know if I missed anything.

inahga:
Hi @rgerhards, have you had a change to look at this PR?

No rush, just checking in.

rgerhards (approved):

rgerhards:
@inahga sry for being late. Looks great now, but I overlooked that we need to update the doc (else nobody will know about that feature.

Would you like to make a doc PR? If not, I will just add the two sentences from the initial posting here, but maybe you can provide better value to users. The doc project lives here:  https://github.com/rsyslog/rsyslog-doc/

Just let me know, I would like to merge the PR within this week (for next week's scheduled stable release).

inahga:
Will do!

inahga:
Docs PR submitted. https://github.com/rsyslog/rsyslog-doc/pull/957

rgerhards:
Thx again for the PR, your persistence and the collaboration! Much appreciated :-)

PR #4763, opened 2022-01-01
Fix an use-after-free bug in omlibdbi.c

secretnonempty:
In the original code, there is a use-after-free bug, which renders the sqlite3 output module unusable. The bug lies in the function initConn, the buggy code is as follows:
```
293             char *const dn_org = strdup((char*)pData->dbName);
294             char *const dn = dirname(dn_org);
295             dbi_conn_set_option(pData->conn, is_sqlite3 ? "sqlite3_dbdir" : "sqlite_dbdir",dn);
296             free(dn_org); /* Free original buffer - dirname may return different pointer */
297 
298             char *tmp = strdup((char*)pData->dbName);
299             char *bn = basename(tmp);
300             free(tmp);
301             dbi_conn_set_option(pData->conn, "dbname", bn); 
```
In line 299, the bn points to the middle of the buffer referenced by tmp, which is allocated in line 298. And then line 300 frees the tmp buffer. After that, line 301 re-uses the pointer bn. In case we are lucky and the buffer content remains the same after being freed, then things stay civil. Otherwise, we will run into trouble.

### DIFF
--- a/plugins/omlibdbi/omlibdbi.c
+++ b/plugins/omlibdbi/omlibdbi.c
@@ -290,15 +290,9 @@ static rsRetVal initConn(instanceData *pData, int bSilent)
 		int is_sqlite2 = !strcmp((const char *)pData->drvrName, "sqlite");
 		int is_sqlite3 = !strcmp((const char *)pData->drvrName, "sqlite3");
 		if(is_sqlite2 || is_sqlite3) {
-			char *const dn_org = strdup((char*)pData->dbName);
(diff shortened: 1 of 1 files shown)


rgerhards (approved):

rgerhards:
CI journal failure is a CI System issue, which already has been fixed but does automatically apply to this PR (as github seems to apply CI changes only on new push) -- so we can safely ignore it.

rgerhards:
@secretnonempty thx for the PR, appreciated!

Issue #4766, opened 2022-01-05
omhttp: Fix memory leak in lokirest batch format


alorbach:
There is a memory leak in When using lokirest batch format, 

> ==176885== 
> ==176885== HEAP SUMMARY:
> ==176885==     in use at exit: 13,115,812 bytes in 100,053 blocks
> ==176885==   total heap usage: 743,972 allocs, 643,919 frees, 142,056,383 bytes allocated
> ==176885== 
> ==176885== 6 bytes in 1 blocks are possibly lost in loss record 4 of 37
> ==176885==    at 0x483B7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
> ==176885==    by 0x4B6350E: strdup (strdup.c:42)
> ==176885==    by 0x49FF217: fjson_object_object_add_ex (in /usr/lib/libfastjson.so.4.3.0)
> ==176885==    by 0x5A817BA: serializeBatchLokiRest (omhttp.c:1277)
> ==176885==    by 0x5A817BA: submitBatch (omhttp.c:1467)
> ==176885==    by 0x5A81B9F: doAction (omhttp.c:1549)
> ==176885==    by 0x175BED: actionCallDoAction (action.c:1265)
> ==176885==    by 0x175BED: actionProcessMessage (action.c:1307)
> ==176885==    by 0x175F53: doTransaction (action.c:1335)
> ==176885==    by 0x175F53: actionTryCommit (action.c:1369)
> ==176885==    by 0x17626B: actionCommit (action.c:1543)
> ==176885==    by 0x177A91: actionCommitAllDirect (action.c:1632)
> ==176885==    by 0x16F8F1: processBatch (ruleset.c:675)
> ==176885==    by 0x1206FB: msgConsumer (rsyslogd.c:694)
> ==176885==    by 0x16A803: ConsumerReg (queue.c:2164)
> ==176885== 
> 
> ==176885== 12,662,680 (8,711,424 direct, 3,951,256 indirect) bytes in 34,029 blocks are definitely lost in loss record 37 of 37
> ==176885==    at 0x483DD99: calloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
> ==176885==    by 0x49FF086: fjson_object_new_object (in /usr/lib/libfastjson.so.4.3.0)
[...]

PR #4767, opened 2022-01-05
omhttp: Fix memory leak in lokirest batchmode

alorbach:
A JSON object was created (valueObj) but not used and also not
released causing a memory leak. This was properly caused by the
initial copy&paste from serializeBatchKafkaRest.

- Also added a valgrind test for lokirest batchmode
  Uncomment "--keep-debuginfo" in testcaseif stack traces
  are incomplete

closes: https://github.com/rsyslog/rsyslog/issues/4766

### DIFF
--- a/contrib/omhttp/omhttp.c
+++ b/contrib/omhttp/omhttp.c
@@ -1244,7 +1244,6 @@ serializeBatchLokiRest(wrkrInstanceData_t *pWrkrData, char **batchBuf)
 {
 	fjson_object *batchArray = NULL;
 	fjson_object *recordObj = NULL;
-	fjson_object *valueObj = NULL;
 	fjson_object *msgObj = NULL;
 
 	size_t numMessages = pWrkrData->batch.nmemb;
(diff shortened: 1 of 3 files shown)


contrib/omhttp/omhttp.c
```diff
@@ -1260,12 +1259,6 @@ serializeBatchLokiRest(wrkrInstanceData_t *pWrkrData, char **batchBuf)
 	}
 
 	for (size_t i = 0; i < numMessages; i++) {
-		valueObj = fjson_object_new_object();
```
rgerhards, line 1263: Do I correctly understand that this code was simply useless?

alorbach: > Do I correctly understand that this code was simply useless?

Yes, valueObj  was not used at all - and I am pretty sure just a leftover from copy&paste of serializeBatchKafkaRest function.

rgerhards: thx!

rgerhards (approved):

rgerhards:
The failed systemd check is an issue in CI itself. Not a blocker here.

PR #4772, opened 2022-01-10
DO NOT MERGE: try to find issue in Fedora 33 testrun

rgerhards:
this often hangs for an unknown reason.


rgerhards:
F33 is being phased out anyways, so no need to try to fix this any longer.

Issue #4395, opened 2020-08-27
imkmsg (probably also imklog): If filedescriptor becomes invalid, logspam is caused.


incase:
### Expected behavior

If the filedescriptor for /dev/kmsg becomes invalid for whatever reason, rsyslog should try to reopen the file or shut down

### Actual behavior

Using a lot of CPU, the log gets spammed with:
rsyslogd: imkmsg: error reading kernel log - shutting down: Bad file descriptor

The code if readkmsg says:
```
			/* something went wrong - error or zero length message */
			if (i < 0 && errno != EINTR && errno != EAGAIN) {
				/* error occured */
				imkmsgLogIntMsg(LOG_ERR,
				       "imkmsg: error reading kernel log - shutting down: %s",
					rs_strerror_r(errno, errmsg, sizeof(errmsg)));
				fklog = -1;
			}
			break;
```
Equivalent code is in the imklog plugin. However, nothing seems to check fklog value and shut anything down, it appears that readkmsg is just called over and over again.
Similarly, I couldn't find anything in imklog either which would actually result in a shutdown of the plugin or rsyslog as the message would indicate, but I might have missed something and the module is actually shut down and restarted, but if so, it fails to re-open the file in the process.

### Steps to reproduce the behavior

Unclear, we are seeing this on some machines, but not all of them, sometimes this re-occurs shortly after a restart of rsyslog, sometimes it doesn't.

### Environment
- rsyslog version: 8.2004.0
- platform: Debian bullseye

hsaliak:
There seems to be a bug in the code below - the read's return value is checked for  -EPIPE, but read always returns -1 on error, and ERRNO is set. So they should be checking if errno == EPIPE instead.
When that happens fklog FD is set to -1 and wrongly concludes that it shuts down and spams.

I saw this happen - and the errno at the start of the spam was an EPIPE (Broken Pipe)

``` i = read(fklog, pRcv, 8192);

		if (i > 0) {
			/* successful read of message of nonzero length */
			pRcv[i] = '\0';
		} else if (i == -EPIPE) {
			imkmsgLogIntMsg(LOG_WARNING,
					"imkmsg: some messages in circular buffer got overwritten");
			continue;
		} else {
			/* something went wrong - error or zero length message */
			if (i < 0 && errno != EINTR && errno != EAGAIN) {
				/* error occured */
				imkmsgLogIntMsg(LOG_ERR,
				       "imkmsg: error reading kernel log - shutting down: %s",
[...]

PR #4775, opened 2022-01-13
ikmsg: fix EPIPE error check in the /dev/kmsg reader loop

hsaliak:
This is a patch for the issue reported in  #4395. 

kmsg is a unique device, which can recover from EPIPE errors. The code in the module checks for this, but the return value from the libc read call always returns -1, and sets the appropriate errno.

Before this patch, the code checks the return value of the read() call against -EPIPE, which never happens. 

This meant that when an EPIPE error actually happens, the fd  is set to -1 and indefinitely retried. rsyslog does not recover from this. Even though the `for` loop in `readikmsg` is broken out of, this function is repeatedly invoked in a `while` loop.

Note: there is an additional improvement possible - `readikmsg` needs better error checking on the fd. I suspect that this was rarely an issue because /dev/kmsg goes truly invalid when the system is actually shutting down.

The fix here is to check the return value as well as the errno. This will allow the intended logic in the code to happen, and for the reads to recover after a `read` on `/dev/kmsg` returns an `EPIPE`.

closes https://github.com/rsyslog/rsyslog/issues/4395

### DIFF
--- a/contrib/imkmsg/kmsg.c
+++ b/contrib/imkmsg/kmsg.c
@@ -214,7 +214,7 @@ readkmsg(void)
 		if (i > 0) {
 			/* successful read of message of nonzero length */
 			pRcv[i] = '\0';
-		} else if (i == -EPIPE) {
+		} else if (i < 0 && errno == EPIPE) {
 			imkmsgLogIntMsg(LOG_WARNING,
 					"imkmsg: some messages in circular buffer got overwritten");
 			continue;


hsaliak:
@rgerhards it seems that as a first time contributer, I would need a maintainer to approve running the workflows. Would you be able to help?

rgerhards:
> it seems that as a first time contributor, I would need a maintainer to approve running the workflows.

Done! Thx for the PR!

rgerhards (approved):

apolarus:
@hsaliak we saw this issue again yesterday on [DTP-18609](https://jira.int.aurora.tech/browse/DTP-18609) running  [AVDeployThud-2022.0119.52](https://shipyard.common.cloud.aurora.tech/release/AVDeployThud-2022.0119.52). Would this still be expected? I'm not clear on how the flow of _rsyslog:master_ should look.

rgerhards:
Well, that depends on when you build from master (or however you access rsyslog code). After merge, it's in the main branch (still called "master" because of some automation dependencies which require quite a bit of work).

PR #4783, opened 2022-01-25
Introduce dynamic configuartion reload global option

Cropi:
The pull request encapsulates multiple changes, let me break it down for smaller parts:

### Input threads

Name of a new input thread now consists of its name(e.g. imtcp) and its serial(e.g. 1 - in which dynamic conf reload has the thread been created). The purpose of this is to better distinguish between input threads that have been created during runtime.

### Globals

New global directive is introduced - hup.reload.config. If set to "on", rsyslog will dynamically reload its configuration(right now only imtcp is implemented to do so). Otherwise, old behavior is used.

## Modules

New function, ```reloadCnf()``` has been added to the interface. If a certain module implements it, the module will be signaled when rsyslog is HUPed, requesting the module to dynamically reload its config.

### The imtcp module - conf reload

1. If there is just a slight difference between the old and new ```imtcp``` module configs we can not use old instances -> we need to reload all of them, note that not all of the module parameters force us to reload all instances, but for simplicity I think it makes sense to do it this way and later improve it
2. Check if new instance is part of old config. If indeed is, remove newly allocated resources and reconfigure pointers(relocate it from the running config into the config being loaded)
3. Remove old instance when not part of the new config.
It would be a lot easier if I just closed all instances and restarted them later but in that case existing TCP connections would have been disturbed.
The idea behind this is to not touch existing TCP connections, just relocate them from the old config to the new one.

### Comparing two imtcp module/instance configs

I am open to any ideas for some change here. I could not come up with a better way when comparing them, also I am not sure where to place macros used for comparing two instances/configs -- other modules will need them as well.

### DIFF
--- a/plugins/imtcp/imtcp.c
+++ b/plugins/imtcp/imtcp.c
@@ -84,10 +84,7 @@ static rsRetVal resetConfigVariables(uchar __attribute__((unused)) *pp, void __a
 typedef struct tcpsrv_etry_s {
 	tcpsrv_t *tcpsrv;
 	pthread_t tid;	/* the worker's thread ID */
-	struct tcpsrv_etry_s *next;
 } tcpsrv_etry_t;
-static tcpsrv_etry_t *tcpsrv_root = NULL;
-static int n_tcpsrv = 0;
 
 static permittedPeers_t *pPermPeersRoot = NULL;
 
@@ -153,13 +150,14 @@ struct instanceConf_s {
 	int iKeepAliveIntvl;
 	int iKeepAliveProbes;
 	int iKeepAliveTime;
-	struct instanceConf_s *next;
+	instanceConf_t *next, *prev;
+	tcpsrv_etry_t *tcpsrv_etry;
 };
 
 
 struct modConfData_s {
 	rsconf_t *pConf;		/* our overall config object */
-	instanceConf_t *root, *tail;
+	instanceConf_t *root, *tail, *act;
 	int iTCPSessMax; /* max number of sessions */
 	int iTCPLstnMax; /* max number of sessions */
 	int iStrmDrvrMode; /* mode for stream driver, driver-dependent (0 mostly means plain tcp) */
@@ -191,6 +189,7 @@ struct modConfData_s {
 
 static modConfData_t *loadModConf = NULL;/* modConf ptr to use for the current load process */
 static modConfData_t *runModConf = NULL;/* modConf ptr to use for the current load process */
+static pthread_mutex_t mutex;
 
 /* module-global parameters */
 static struct cnfparamdescr modpdescr[] = {
@@ -353,6 +352,7 @@ createInstance(instanceConf_t **pinst)
 	CHKmalloc(inst = (instanceConf_t*) calloc(1, sizeof(instanceConf_t)));
 	CHKmalloc(inst->cnf_params = (tcpLstnParams_t*) calloc(1, sizeof(tcpLstnParams_t)));
 	inst->next = NULL;
+	inst->prev = NULL;
 	inst->pszBindRuleset = NULL;
 	inst->pszInputName = NULL;
 	inst->dfltTZ = NULL;
@@ -388,12 +388,14 @@ createInstance(instanceConf_t **pinst)
 	inst->iTCPSessMax = loadModConf->iTCPSessMax;
 
 	inst->cnf_params->pszLstnPortFileName = NULL;
+	inst->tcpsrv_etry = NULL;
 
 	/* node created, let's add to config */
 	if(loadModConf->tail == NULL) {
 		loadModConf->tail = loadModConf->root = inst;
 	} else {
 		loadModConf->tail->next = inst;
+		inst->prev = loadModConf->tail;
 		loadModConf->tail = inst;
 	}
 
@@ -540,9 +542,7 @@ addListner(modConfData_t *modConf, instanceConf_t *inst)
 	tcpsrv_etry_t *etry;
 	CHKmalloc(etry = (tcpsrv_etry_t*) calloc(1, sizeof(tcpsrv_etry_t)));
 	etry->tcpsrv = pOurTcpsrv;
-	etry->next = tcpsrv_root;
-	tcpsrv_root = etry;
-	++n_tcpsrv;
+	inst->tcpsrv_etry = etry;
 
 finalize_it:
 	if(iRet != RS_RET_OK) {
@@ -850,7 +850,7 @@ CODESTARTendCnfLoad
 	free(cs.pszStrmDrvrAuthMode);
 	cs.pszStrmDrvrAuthMode = NULL;
 
-	loadModConf = NULL; /* done loading */
+	// loadModConf = NULL; /* done loading */
 ENDendCnfLoad
 
 
@@ -876,6 +876,10 @@ CODESTARTcheckCnf
 				"no listeners defined - no input will be gathered");
 		iRet = RS_RET_NO_LISTNERS;
 	}
+	/* whenever a new configuration reload is invoked, new instances will be appended to the end of
+	 * the linked list, thus this information will tell us which instances have not been started yet.
+	 */
+	loadModConf->act = loadModConf->root;
 ENDcheckCnf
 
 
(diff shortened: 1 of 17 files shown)


rgerhards:
@davidelang your feedback on the topic is highly appreciated!

rgerhards:
@Cropi I think ubuntu 20 [T]SAN have valid complaints ;-)

rgerhards:
> New global directive is introduced - hup.reload.config. If set to "on", rsyslog will dynamically reload its configuration(right now only imtcp is implemented to do so). Otherwise, old behavior is used.

would it make sense to split this off to a separate PR or at least commit? I think this is laying ground for many changes to come and so IMHO it deserves to stand out in git history.

Cropi:
> @Cropi I think ubuntu 20 [T]SAN have valid complaints ;-)

Yes, you are right:D

> would it make sense to split this off to a separate PR or at least commit? I think this is laying ground for many changes to come and so IMHO it deserves to stand out in git history.

Of course, I will divide it into smaller chunks. There are some parts that will require changes but I wanted to execute the CI.

davidelang:
reloading configs is both an easy thing and a horribly hard thing, depending on 
how much the configs are changing. Can you go into a little detail about what 
sorts of things you are expecting to change?

for example, adding additional inputs feeding into existing queue is relatively 
easy (and even for such a simple thing, it's an extremely useful feature to 
have)

but making major changes to the flow of logs through multiple queues, changing 
queue names, and outputting to multiple places is an extremely complex problem 
that can leave logs orphaned in queues very easily. When you have the new config 
using different modules, it gets even messier as the allowed syntax changes 
depending on what modules are loaded.

In between you have a 'normal' config where you don't have a complex flow, and 
are just working through the main queue outputting to different destinations 
(where any additional queues that are in use are only tied to a particular 
action, so if that queue doesn't appear in the new config, you could leave it in 
place with it's existing action to finish processing pending messages)

I'll look through this PR later today, but if you could give us an idea of the 
scope that you are looking at (and a rough roadmap of the steps that you are 
thinking of) it would be useful.

David Lang

davidelang:
>> New global directive is introduced - hup.reload.config. If set to "on", rsyslog will dynamically reload its configuration(right now only imtcp is implemented to do so). Otherwise, old behavior is used.

I think we want this directive to be more nuanced than 'on' 'off' (see my other 
comment about the possible complexity)

David Lang

rgerhards:
> I think we want this directive to be more nuanced than 'on' 'off' (see my other comment about the possible complexity)

let me chime in. The goal here is actually to do the hard and complex thing, so we probably do not need the nuances. As code and PR develops, of course, I think @Cropi will go from least to more complex things.

Cropi:
Thank you @davidelang for the constructive feedback. I did not specify in the description what I really want to accomplish with this PR, sorry for that. My long term goal would be to dynamically reload the entire config file via sending a HUP signal to rsyslog. I know that this task is extremely large and will certainly require lot of changes in the project.

All I wanted to accomplish in this PR, is to provide a way to dynamically reload the ```imtcp``` input module in the following manner: every single ```input(type="imtcp")``` instance, which
- is both present in old and new config is not disturbed at all
- is present in the old config but not in the new one is gracefully shut down
- is present in the new config but not in the old one will begin to run
Please note that at this point, I do not take into consideration that the ruleset can change for an imtcp input. I know it's a bit silly but I think that will require a separate PR due to its complexity.

I totally agree that the hup.reload.config directive will require some changes and it won't be as simple as it is now. 

I will divide the PR into smaller parts later today and address issues discovered by CI.

Cropi:
Commit1 e84efad5ac42cb354c1a1289fada2905e433f328 - introduce reloadCnf() for modules, each module capable of reloading its config should implement it.
Commit2 c6cf04dcbc2ec5efbd24dba18059787535f5f508 - make input thread names consist of name+ID in which dynamic config reload the thread has been started
Commit 3 ce9080480ec0a6b0cae4a33361ce4f845b134df2 - extend ```tcpsrv``` interface to be able to terminate a single ```tcpsrv``` object
Commit 4 32974b17ae87bf97747697341703b7a11d835f71 - add functionality to compare two peers if they equal, will be used when comparing two imtcp instances
Commit 5 32974b17ae87bf97747697341703b7a11d835f71 - ```tcpsrv_etry_t``` instances were kept in a separate linked list. We need to connect them to the ```instanceConf``` struct, otherwise we can not dynamically reload the imtcp config because we are not able to determine which instance is connected to which ```tcpsrv_etry```.
Commit 6 13ba8610c19969cd86413072b94f5044f18282c9 - implement reloadCnf() for imtcp module(support for ruleset excluded)

The CI failed because of the following reasons: PR contains more than 1 commit, commented out code and elasticsearch issue connection issue.

Let me know if there is something I forgot to answer.

davidelang:
On Wed, 26 Jan 2022, Attila Lakatos wrote:

> All I wanted to accomplish in this PR, is to provide a way to dynamically reload the ```imtcp``` input module in the following manner: every single ```input(type="imtcp")``` instance, which
> - is both present in old and new config is not disturbed at all
> - is present in the old config but not in the new one is gracefully shut down
> - is present in the new config but not in the old one will begin to run
> Please note that at this point, I do not take into consideration that the ruleset can change for an imtcp input. I know it's a bit silly but I think that will require a separate PR due to its complexity.

This is a good starting point. Another very useful place to do early would be 
imfile.

will your initial version work with inputs going to rulesets? existing inputs 
with different rulesets? crypto changes? etc? or are you just looking at 
additions/removals?

David Lang

Cropi:
> On Wed, 26 Jan 2022, Attila Lakatos wrote: All I wanted to accomplish in this PR, is to provide a way to dynamically reload the ```imtcp``` input module in the following manner: every single ```input(type="imtcp")``` instance, which - is both present in old and new config is not disturbed at all - is present in the old config but not in the new one is gracefully shut down - is present in the new config but not in the old one will begin to run Please note that at this point, I do not take into consideration that the ruleset can change for an imtcp input. I know it's a bit silly but I think that will require a separate PR due to its complexity.
> This is a good starting point. Another very useful place to do early would be imfile. will your initial version work with inputs going to rulesets? existing inputs with different rulesets? crypto changes? etc? or are you just looking at additions/removals? David Lang

The current version works with inputs going to rulesets. When config is reloaded, **new** inputs will deliver logs to whatever the ruleset is configured to. It has only one drawback: when a certain input is present in both old and new configs, then the ruleset is not updated(its content might have changed). In order to make it work, some changes in the  ```tcpsrv``` interface might be needed. I can provide a fix for that but I do not really want to make this PR more complicated as it is now.

There are some globals which may affect ```imtcp``` instances. I've included some checks(thanks for notificing it) if these globals change, e.g. default netstream driver.
Some plans for next steps: as you mentioned, I would like to provide the ability to reload more than just one input module, maybe the next one can be imfile. Next, I plan to focus on other rsyslog directives as well starting from easier(dynstats, perctile stats) to more complex(rulesets). I wonder if we should track these steps somewhere else, to not get lost under a pull request. Separate github issue, mailing list or somewhere else?

rgerhards:
> Some plans for next steps: as you mentioned, I would like to provide the ability to reload more than just one input module, maybe the next one can be imfile. Next, I plan to focus on other rsyslog directives as well starting from easier(dynstats, perctile stats) to more complex(rulesets). I wonder if we should track these steps somewhere else, to not get lost under a pull request. Separate github issue, mailing list or somewhere else?

sorry, I overlooked this. This sounds like we should/could use github projects. Will look into it - there is some new beta feature, but not sure if this suits us well. The "old" one is clumsy but worked so far sufficiently well.

rgerhards:
lol... I began to create the project this morning and was than dragged away. It's there, and I have assigned this PR to it. Pls try if you can assign your other work (else we need to look into permission settings).

rgerhards:
@Cropi I plan the next scheduled stable release for the 15th. do you think we should include this PR then? If so, we may need to do a little of finish-up work. Pls let me know your thoughts.

Cropi:
> lol... I began to create the project this morning and was than dragged away. It's there, and I have assigned this PR to it. Pls try if you can assign your other work (else we need to look into permission settings).

Unfortunately, I am not able to assign a PR to the project.

rgerhards:
> Unfortunately, I am not able to assign a PR to the project.

thx, I'll have a look into permission requirements

Cropi:
> @Cropi I plan the next scheduled stable release for the 15th. do you think we should include this PR then? If so, we may need to do a little of finish-up work. Pls let me know your thoughts.

I think this PR will need additional changes, so it would be a good idea to postpone it to a later release.

rgerhards:
@Cropi thinks have evolved quite a bit in the interim. Do you remember the PR and think it would be useful to try to complete it? I am undecided. If you think it is a solid base, I could also look at it from the project's PoV and try to drive it (slowly) forward. Quite honestly, I try to close or merge the open PRs.

Cropi:
The idea is nice but only in theory. Because technically, this would require extremely big code changes and resolution of many corner cases. Some time ago, I spent some time on trying to achieve this. But the more I investigated, the harder it was to implement. Although this was before coding with help of AI was a case. I think it would introduce way more problems than it would actually solve. So I will close this for now.

rgerhards:
@Cropi thx - that was also my impression. As you say, with AI some things might get easier, but I think it is not a top prio at the moment. I have begun to make more helper entities re-loadable with a HUP.

Issue #4770, opened 2022-01-08
issues with numeric comparision (rainerscript)


mabobr:
Hi,

with this code:
```
  set $!lower_nr  = 1111;
  set $!higher_nr = 2222;

  if $!higher_nr < $!lower_nr
  then {  set $!result = "never reached branch"; }
  else {  set $!result = "always reached branch"; }
```
i got strange results:
```
5291.997345043:main Q:Reg/w0  : rainerscript.c:     SET !lower_nr =
5291.997353190:main Q:Reg/w0  : rainerscript.c:       1111
5291.997362233:main Q:Reg/w0  : rainerscript.c:     END SET
5291.997369818:main Q:Reg/w0  : rainerscript.c: eval expr 0x558574715c90, type 'N'
5291.997373129:main Q:Reg/w0  : rainerscript.c: eval expr 0x558574715c90, return datatype 'N':1111
5291.997378955:main Q:Reg/w0  : rainerscript.c:     SET !higher_nr =
5291.997387523:main Q:Reg/w0  : rainerscript.c:       2222
5291.997406663:main Q:Reg/w0  : rainerscript.c:     END SET
5291.997426385:main Q:Reg/w0  : rainerscript.c: eval expr 0x558574715d50, type 'N'
5291.997430688:main Q:Reg/w0  : rainerscript.c: eval expr 0x558574715d50, return datatype 'N':2222
5291.997438048:main Q:Reg/w0  : rainerscript.c:     IF
5291.997448566:main Q:Reg/w0  : rainerscript.c:         var '!higher_nr'
5291.997459747:main Q:Reg/w0  : rainerscript.c:       <
5291.997467930:main Q:Reg/w0  : rainerscript.c:         var '!lower_nr'
5291.997479442:main Q:Reg/w0  : rainerscript.c: eval expr 0x5585747172c0, type 'CMP_LT'
5291.997483867:main Q:Reg/w0  : rainerscript.c: eval expr 0x558574715cb0, type 'V'
5291.997499875:main Q:Reg/w0  : rainerscript.c: rainerscript: (json) var 200:!higher_nr: '2222'
5291.997504985:main Q:Reg/w0  : rainerscript.c: eval expr 0x558574715cb0, return datatype 'J':0
5291.997510621:main Q:Reg/w0  : rainerscript.c: eval expr 0x558574715d70, type 'V'
[...]

PR #4785, opened 2022-01-26
rscript: literal numbers were not compared correctly

rgerhards:
This problem occurred when numbers were used in rsyslog.conf in
the set statement, e.g.

set $nbr = 1234;

In this case, during comparisons, the number was actually interpreted
as a string with digits. Thus numerical comparisons lead to unexpected
results. Even more so, as in other places of the code they were
treated as native numbers.

This is now fixed. We cannot outrule that this causes, in border cases,
change of behavior to existing configs. But it is unlikely and the
previous behaviour was a clear bug and very unintuitive. This in our
opinion it is justified to risk a breaking change for an expected
very minor subset of installations, if any such exists at all.

closes https://github.com/rsyslog/rsyslog/issues/4770

### DIFF
--- a/grammar/rainerscript.c
+++ b/grammar/rainerscript.c
@@ -2,7 +2,7 @@
  *
  * Module begun 2011-07-01 by Rainer Gerhards
  *
- * Copyright 2011-2019 Rainer Gerhards and Others.
+ * Copyright 2011-2022 Rainer Gerhards and Others.
  *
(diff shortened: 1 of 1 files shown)


rgerhards:
PR https://github.com/rsyslog/rsyslog/pull/4789 had a better approach and replaces this one here.

PR #4791, opened 2022-01-31
Make the main message queue part of the config

Cropi:
The intent of this PR is to make the main message queue part of the main config. It will help us to proceed towards dynamic configuration reload.
Besides that I did some refactoring. I divided ```activateMainQueue()``` into two separate functions. The first one, ```loadMainQueue()``` is responsible for interpreting main queue parameters, the second one ```activateMainQueue()``` will start the queue.
I had to move some settings from ```qqueueStart()``` to ```qqueueConstruct()```, because a queue should be fully initialized by the end of construction phase. This is needed in order to compare an already constructed(not yet started) queue with previous queues(queue from previous config that might be the same) during dynamic configuration reload.

### DIFF
--- a/action.c
+++ b/action.c
@@ -560,6 +560,7 @@ actionConstructFinalize(action_t *__restrict__ const pThis, struct nvlst *lst)
 		qqueueSetDefaultsActionQueue(pThis->pQueue);
 		qqueueApplyCnfParam(pThis->pQueue, lst);
 	}
+	qqueueCorrectParams(pThis->pQueue);
 
 #	undef setQPROP
 #	undef setQPROPstr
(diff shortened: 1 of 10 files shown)


Cropi:
Let me rephrase/correct the description. The intent of this PR is to make the main message queue part of the main config. It will help us to proceed towards dynamic configuration reload.

I divided the activateMainQueue() into two separate functions:

1. ```loadMainQueue()``` is responsible for interpreting main queue parameters. Moreover, if there is an already running main queue, which has the exact same parameters as the new one, it is used instead. 
2. ```activateMainQueue()``` will actually start the queue. I had to move some settings from ```qqueueStart()``` to a separate function ```qqueueCorrectParams()```, because a queue should be fully initialized by the end of construction phase. This is needed in order to compare an already constructed(not yet started) queue with previous queues(queue from previous config that might be the same) during dynamic configuration reload.

I've also added a new field ```isRunning``` to the queue struct, which holds information whether the queue has already been started or not. This will come in handy during dynamic config reload, more precisely when an already running queue is relocated from one config to another(it's present in both old and new conf) and all the queues are instructed to be started(so we won't start it twice).

@rgerhards , the PR is ready for review. The centos7.5 buildbot got into a strange compilation problem, out of memory?

PR #4808, opened 2022-02-25
Elasticsearch 8.x support (Data Stream API)

EHerzog76:
# Elasticsearch 8.x support
#### musl-libc (Notice)
I dosen´t spent time to compile it with the musl-libc.
So I added the libc to the rsyslog-docker  Alpine-Appliance  (see the rsyslog-docker  pull-request)


rgerhards:
This is not compatible with ES < 8 - I will apply changes to make user able to add version info.

rgerhards:
do we need additional doc? If so, would you be willing to do a PR on https://github.com/rsyslog/rsyslog-doc/ -- otherwise, please let me know what I should mention.

rgerhards:
some very basic doc in https://github.com/rsyslog/rsyslog-doc/pull/968 -- it would be good to have more.

PR #4811, opened 2022-03-01
Add option to limit error file to configured size

### LINKED ISSUE #4733: action.errorfile size management option
Do you think it would be appropriate to implement an ```action.errorfile.maxsize``` option rsyslog?

One of our clients is having issues with the huge amount of logs generated by the ```action.errorfile``` option. When forwarding logs to a remote server that is "too often" offline, the number of error messages might grow really quickly.
My understanding is that  ```logrotate``` is the primary tool used for such purpose. However, according to them it is not enough because it has to be executed too many times.
This is more like a question than a request but if it's accepted I can help with the implementation.

sarroutbi:
action.errorfile.maxsize has been added to enable
option to limit the amount of bytes dumped to
configured errorfile

fixes #4733

Signed-off-by: Sergio Arroutbi <sarroutb@redhat.com>

### DIFF
--- a/action.c
+++ b/action.c
@@ -189,6 +189,7 @@ static struct cnfparamdescr cnfparamdescr[] = {
 	{ "name", eCmdHdlrGetWord, 0 }, /* legacy: actionname */
 	{ "type", eCmdHdlrString, CNFPARAM_REQUIRED }, /* legacy: actionname */
 	{ "action.errorfile", eCmdHdlrString, 0 },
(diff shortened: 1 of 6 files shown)


rgerhards:

rgerhards (approved):

rgerhards:
Thx for https://github.com/rsyslog/rsyslog-doc/pull/962

PR #4815, opened 2022-03-03
Don't refer to errno when the pthread library fails

miztake:
When the pthread library fails, errno is referenced even though errno is not set.
Fix to refer to the return code of the pthread library instead of errno.


rgerhards:
thx for catching this!

PR #4822, opened 2022-03-08
Ensure error file does not increase appropriately when action.errorfile.maxsize option is enabled

### LINKED ISSUE #4821: Max size for error file is not working appropriately if error file already has contents
### Expected behavior
When including a limitation of the size of the error file with option action.errorfile.maxsize, if error file already is created and has some content, minor or equal to the maximum size specified, error file should not increase to a bigger size than that.

### Actual behavior
When including a limitation of the size of the error file with option action.errorfile.maxsize, if error file already is created and has some content, the limitation is not working properly, and error file can increase to bigger size than configured.

1. Set a configuration that dumps errors, without error file size limitation
2. Log messages
3. Set a configuration that dumps errors, with file size limitation (bigger than the size of the current error file)
4. Log messages
5. It can be checked that error file increases over the maximum configured size

sarroutbi:
fixes #4821

Signed-off-by: Sergio Arroutbi <sarroutb@redhat.com>

### DIFF
--- a/action.c
+++ b/action.c
@@ -393,7 +393,7 @@ rsRetVal actionConstruct(action_t **ppThis)
 	pThis->pszName = NULL;
 	pThis->pszErrFile = NULL;
 	pThis->maxErrFileSize = 0;
-	pThis->errFileWritten = 0;
+	pThis->currentErrFileSize = 0;
 	pThis->pszExternalStateFile = NULL;
 	pThis->fdErrFile = -1;
 	pThis->bWriteAllMarkMsgs = 1;
@@ -1430,6 +1430,14 @@ actionWriteErrorFile(action_t *__restrict__ const pThis, const rsRetVal ret,
 				pThis->pszName, pThis->pszErrFile);
 			goto done;
 		}
+		if (pThis->maxErrFileSize > 0) {
+			struct stat statbuf;
+			if (fstat(pThis->fdErrFile, &statbuf) == -1) {
+				LogError(errno, RS_RET_ERR, "failed to fstat %s", pThis->pszErrFile);
+				goto done;
+			}
+			pThis->currentErrFileSize = statbuf.st_size;
+		}
 	}
 
 	for(int i = 0 ; i < nparams ; ++i) {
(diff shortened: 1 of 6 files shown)


sarroutbi:
Hello @rgerhards. May I ask this PR to be reviewed when possible? It fixes an issue with new action.errorfile.maxsize handling.

Thanks!

rgerhards:
Sure, but I am currently working on some CI issues with Solaris. After that, I can re-run the CI on all other PRs.

sarroutbi:
> Sure, but I am currently working on some CI issues with Solaris. After that, I can re-run the CI on all other PRs.

Cool. Thank you very much for  your quick response.

tests/omfwd-errfile-maxsize-filled.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+# part of the rsyslog project, released under ASL 2.0
+. ${srcdir:=.}/diag.sh init
+ERRFILE=$(mktemp)
```
rgerhards: please change ERRFILE to rsyslog's temp files, as this permits us to consistently clean out all kinds of failed runs:
```
ERRFILE="$RSYSLOG_DYNNAME.err"
```

action.c
```diff
@@ -1430,6 +1430,14 @@ actionWriteErrorFile(action_t *__restrict__ const pThis, const rsRetVal ret,
 				pThis->pszName, pThis->pszErrFile);
 			goto done;
 		}
+		if (pThis->maxErrFileSize > 0) {
+			struct stat statbuf;
+			if (fstat(pThis->fdErrFile, &statbuf) == -1) {
+				LogError(errno, RS_RET_ERR, "failed to fstat %s", pThis->pszErrFile);
+				goto done;
+			}
+			pThis->errFileWritten += statbuf.st_size;
```
rgerhards: IMHO adding to the current error file size is incorrect, at least misleading. Shouldn't it plainly set the size?

I guess it works, as the fd is only opened once, but in respect to ability to understand the code this should be changed to a pure assignment. Or am I overlooking something?

rgerhards (changes requested):

action.c
sarroutbi: The intention of the fix is to **not** write more than what is configured. Let me explain myself:
action.errfile.maxsize contains the maximum desirable size of the error file
errFileWritten contains the amount of data in error file

If errFileWritten is not incremented on error file opening, you could have a situation where the error file gets to a size bigger than configured.

Should I rename "errFileWritten" variable to "currentErrFileSize"? This way, it might be more appropriate to what it is doing ...

rgerhards: Thx for the swift reply!
well - if you fstat() the file, isn't that the "current size" - why add it to the previous "current size"? That's my question.

tests/omfwd-errfile-maxsize-filled.sh
sarroutbi: Ok. Thanks for the suggestion. In that case ... should I remove its deletion? Is $RSYSLOG_DYNNAME.err deleted automatically by test framework?

action.c
sarroutbi: Sorry, I didn't get your point. Yes, it is not necessary to add, just assignment. I changed it in updated patch.

rgerhards: I probably did not clearly enough describe it - thx!

tests/omfwd-errfile-maxsize-filled.sh
rgerhards: yes, it's automatically deleted by the framework on SUCCESSFUL text execution (this makes it easier to manually investigate, if required)

rgerhards (approved):

Issue #4818, opened 2022-03-04
100% CPU usage seen on rsyslog via GnuTLS when TLS connection is broken.


lonicerae:
Hello,

Hitting a 100% CPU usage issue when rsyslog via GnuTLS is configured.

It can be reproduced frequently. The incentive could be some broken TLS connections, and a strace log and core dump file suggests this could be a bug in rsyslog. **A timeout or subsequent checking with gnutls_record_get_direction() seems missing after gnutls_record_recv() is called.** Here's my analysis:

1. Strace shows the the thread 1352041 keeps doing recvfrom():

- rsyslogd.strace
~~~
1352037 13:47:27.980772 pselect6(0, NULL, NULL, NULL, {tv_sec=148, tv_nsec=359998684}, {[], 8} <unfinished ...>
1352047 13:47:27.980855 futex(0x7f3c71c12dec, FUTEX_WAIT_PRIVATE, 0, NULL <unfinished ...>
1352046 13:47:27.980871 futex(0x7f3c71c12d80, FUTEX_WAIT_PRIVATE, 0, NULL <unfinished ...>
1352045 13:47:27.980878 futex(0x7f3c71c12d1c, FUTEX_WAIT_PRIVATE, 0, NULL <unfinished ...>
1352044 13:47:27.980884 futex(0x7f3c71c12cb0, FUTEX_WAIT_PRIVATE, 0, NULL <unfinished ...>
1352041 13:47:27.980892 recvfrom(10<TCP:[x.x.x.x:6514->y.y.y.y:22642]>,  <unfinished ...>
1352042 13:47:27.981115 futex(0x558ad9986254, FUTEX_WAIT_PRIVATE, 0, NULL <unfinished ...>
1352041 13:47:27.981127 <... recvfrom resumed>0x7f3c5c106df3, 71, 0, NULL, NULL) = -1 EAGAIN  <0.000039>
1352040 13:47:27.981221 ppoll([{fd=7<anon_inode:inotify>, events=POLLIN}], 1, {tv_sec=0, tv_nsec=794320414}, NULL, 8 <unfinished ...>
1352041 13:47:27.981270 recvfrom(10<TCP:[x.x.x.x:6514->y.y.y.y:22642]>, 0x7f3c5c106df3, 71, 0, NULL, NULL) = -1 EAGAIN  <0.000016>
[...]

Issue #4638, opened 2021-07-21
thread pegging core at 100% in infinite EAGAIN loop


snblitz:
### Expected behavior

No threads in an infinite loop getting EAGAIN

### Actual behavior

infinite loop pegging a single core at 100% forever

### Steps to reproduce the behavior

Unknown.  seems to occur randomly, but occasionally seems to affect several different machines in a pool at the same time.

strace -f -p 526624 (pid_of_rsyslogd_pegged_at_100%) (different from pid below)

[pid 526653] recvfrom(77, 0x7f7f68052a73, 736, 0, NULL, NULL) = -1 EAGAIN (Resource temporarily unavailable)

[pid 526653] recvfrom(77, 0x7f7f68052a73, 736, 0, NULL, NULL) = -1 EAGAIN (Resource temporarily unavailable)

[pid 526653] recvfrom(77, 0x7f7f68052a73, 736, 0, NULL, NULL) = -1 EAGAIN (Resource temporarily unavailable)

forever.

The rsyslogd system looks to be working.  Data is still flowing.  Just one thread stuck at 100%.

Looking via ps -T -p 526624 I find

526624 526653 ?        00:00:00 in:imtcp

And imtcp is configured with

global(
    DefaultNetstreamDriver="gtls"
    DefaultNetstreamDriverCAFile="/etc/ssl/certs/sink-kafka-ext.openvpn.net.bundle.crt"
    DefaultNetstreamDriverCertFile="/etc/ssl/certs/sink-kafka-ext.openvpn.net.crt"
    DefaultNetstreamDriverKeyFile="/etc/ssl/private/sink-kafka-ext.openvpn.net.key"
)

module(
    load="imtcp"
    StreamDriver.Name="gtls"
    StreamDriver.Mode="1"
    StreamDriver.Authmode="anon"
    MaxSessions="2000"
)

input(
    type="imtcp"
    port="20516"
)

### Environment
- rsyslog version: 8.2001.0

- platform: Ubuntu 20.04.2 LTS

- for configuration questions/issues, include rsyslog.conf and included config files

davidelang:
We have seen this problem with TLS connections, there has been a lot of work on 
this over the last year, can you try upgrading from the Jan 2020 release to a 
current one (the latest is 8.2106, June 2021)

David Lang

snblitz:
I will give a more recent version a try.  Since the problem is intermittent it might be a while before I have anything to report back.

davidelang:
[...]

PR #4823, opened 2022-03-09
gnutls bugfix: Fix error handling in gtlsRecordRecv

alorbach:
There was a rare possibility that the E_AGAIN/E_INTERRUPT handling
could cause an infinite loop (100% CPU Usage), for example when a TLS
handshake is interrupted at a certain stage.

- After gnutls_record_recv is called, and E_AGAIN/E_INTERRUPT error
  occurs, we need to do additional read/write direction handling
  with gnutls_record_get_direction.
- After the second call of gnutls_record_recv (Expand buffer)
  we needed to also check the eror codes for E_AGAIN/E_INTERRUPT
  to do propper errorhandling.

see also: https://github.com/rsyslog/rsyslog/issues/4818

### DIFF
--- a/runtime/nsd_gtls.c
+++ b/runtime/nsd_gtls.c
@@ -551,8 +551,9 @@ gtlsRecordRecv(nsd_gtls_t *pThis)
 	DEFiRet;
 
 	ISOBJ_TYPE_assert(pThis, nsd_gtls);
-	DBGPRINTF("gtlsRecordRecv: start\n");
-
+	DBGPRINTF("gtlsRecordRecv: start (Pending Data: %zd | Wanted Direction: %s)\n",
+		gnutls_record_check_pending(pThis->sess),
+		(gnutls_record_get_direction(pThis->sess) == gtlsDir_READ ? "READ" : "WRITE") );
 	lenRcvd = gnutls_record_recv(pThis->sess, pThis->pszRcvBuf, NSD_GTLS_MAX_RCVBUF);
 	if(lenRcvd >= 0) {
 		DBGPRINTF("gtlsRecordRecv: gnutls_record_recv received %zd bytes\n", lenRcvd);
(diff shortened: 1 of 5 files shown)


runtime/nsd_gtls.c
```diff
@@ -576,14 +577,31 @@ gtlsRecordRecv(nsd_gtls_t *pThis)
...
 		}
 	} else if(lenRcvd == GNUTLS_E_AGAIN || lenRcvd == GNUTLS_E_INTERRUPTED) {
-sslerr:
-		pThis->rtryCall = gtlsRtry_recv;
-		dbgprintf("GnuTLS receive requires a retry (this most probably is OK and no error condition)\n");
-		ABORT_FINALIZE(RS_RET_RETRY);
+sslerragain:
+		/* Check if the underlaying file descriptor needs to read or write data!*/
+		wantsWriteData = gnutls_record_get_direction(pThis->sess);
+		if (wantsWriteData == 0) {
```
lonicerae: In our case, the problem is the thread keeps doing 'recvfrom()' endlessly. Could we abort the connection after some retries or some time?

lonicerae (changes requested):

runtime/nsd_gtls.c
alorbach: @rgerhards what do you think? In order to maintain existing 

> In our case, the problem is the thread keeps doing 'recvfrom()' endlessly. Could we abort the connection after some retries or some time?

Could you post a gdb stack trace?
Or even better could you enable debug output, if we could catch that error with debug log enabled that would be awesome.

These modules would be sufficient for debug logging: 
```
global( 
       debug.whitelist="on"
       debug.files=["nsd_gtls.c", "nsd_ptcp.c", "tcpsrv.c", "nsdsel_ossl.c", "nsdsel_gtls.c", "nsdpoll_ptcp.c", "dnscache.c"]
)
```

opoplawski:
I have debug output, but it's lot of data and I don't want to post it publicly.

opoplawski:
I'm afraid the debug output is 1.5GB compressed and too large for that form to upload.

alorbach:
> I'm afraid the debug output is 1.5GB compressed and too large for that form to upload.
 I have extended the form to support up to 10GB files. Can you retry?

alorbach:
> I've uploaded the segment. Hope it helps. Sorry for the delay - vacation and some technical issues.

thanks I got it 👍 , I will review the logfile ASAP.

alorbach:
> I've uploaded the segment. Hope it helps. Sorry for the delay - vacation and some technical issues.
I reviewed the debug logfile and this really looks odd. Something is not handeled right at some point, properly with a huge amount of sessions. 

I have added some more debug output and added output of the nsd pointer at more debug outputs so we can track which nsd is doing what. For some reason it never reaches the point of "nsdsel_gtls: reached end".  

@opoplawski  Can you try to run with the amended debug output changes from here? It will help me better tracking the nsd objects inside the debug logfile.

opoplawski:
I've started running it, but it's generating about 1GB of output each minute - so not sure how long I can run it.

opoplawski:
New file uploaded

alorbach:
@opoplawski Thanks a lot we are coming closer ;)

> tcpsrv.c: tcpsrv: ready to process 1 event entries
> imtcp.c        : tcpsrv.c: tcpsrv: processing item 15, pUsr 0x7f133c1bbb60, bAbortConn
> imtcp.c        : tcpsrv.c: netstream 0x7f133c1bba30 with new data
> imtcp.c        : nsd_gtls.c: gtlsRecordRecv: start
> imtcp.c        : nsd_gtls.c: gtlsRecordRecv: gnutls_record_recv received 456 bytes
> imtcp.c        : nsd_gtls.c: gtlsRecordRecv return. nsd 0x7f133c19ab60, iRet 0, lenRcvd 456, lenRcvBuf 456, ptrRcvBuf 0
> imtcp.c        : nsd_gtls.c: gtlsRcv return. nsd 0x7f133c19ab60, iRet 0, lenRcvBuf -1, ptrRcvBuf 456

The same tcp session is causing the loop over and over, however I can't see why. It looks like rsyslog is getting hammered with the same amount of data on that TCP Session over and over again. 

I have added more debug output, it will now reveal the tcp session client, causing the loop. Its ok to have debug log run for 1 minute, I don't need more data. 

Thanks a lot for helping.

opoplawski:
Uploaded new log.  This is just the final minute before the server was shutdown but probably doesn't capture the initial start of the issue.

opoplawski:
@alorbach Is this ready for testing?

alorbach:
@opoplawski : Yes I added a potential fix for the problem. However I cannot test it, if you could run it at your end it would be great. When you debug you may see "doRetry: gtlsRecordRecv returned RETRY, but there is no pending" from time to time, when this happens it could have caused the 100% loop before.

alorbach:
@opoplawski Did you have a chance to test the PR again?

opoplawski:
Yes - it's been running for nearly 2 days and I haven't seen the problem again - which is the longest I've ever seen it go.  So I think this has likely fixed the issue for me.

alorbach:
@opoplawski Thanks a lot for the feedback 👍 ! If it hasn't happened again, it most likely is fixed now. 
Can you check if any rsyslog tls errors werre logged during meantime?

alorbach:
So it appears that this PR should close the following issues: 
https://github.com/rsyslog/rsyslog/issues/4834
https://github.com/rsyslog/rsyslog/issues/4818
https://github.com/rsyslog/rsyslog/issues/4638

PR #4833, opened 2022-03-24
rsyslogd: adjust the order of doHUP() and processImInternal()

zy1306:
After call doHUP(), probably there is a internal log in the list. However, it
will not be wrote out immediately, because the mainloop will be blocked at
pselect in wait_timeout() until a long timeout or next message occur.
More deadly, the log may be lost if the deamon exits unexpectedly.

We might as well put processImInternal() after doHUP(), so that the message
will be flushed out immediately.

Fixes: 723f6fdfa6(rsyslogd: Fix race between signals and main loop timeout)
Signed-off-by: Yun Zhou <yun.zhou@windriver.com>

### DIFF
--- a/tools/rsyslogd.c
+++ b/tools/rsyslogd.c
@@ -1952,8 +1952,6 @@ mainloop(void)
 	sigaddset(&sigblockset, SIGHUP);
 
 	do {
-		processImInternal();
-
 		pthread_sigmask(SIG_BLOCK, &sigblockset, &origmask);
 		if(bChildDied) {
 			reapChild();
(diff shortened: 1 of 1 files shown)


zy1306:
Who can rebuild the the 'buildbot' pipline? It seams that it failed due to 'OSError: [Errno 12] Cannot allocate memory'

zy1306:
The second solution is to block SIGTTOU as SIGHUP, SIGCHLD and SIGTERM, so that pselect will not sleep if a internal message was generated in doHUP().

PR #4844, opened 2022-04-04
imfile: do_inotify() use only one blocking call

rgerhards:
This kind of re-opens https://github.com/rsyslog/rsyslog/pull/4445, which was closed for some (unknown) reason .

do_inotify block only once for events with timeouts. There is possible
scenario where no data is available and read() enters blocking while poll()
is released.

The change introduced here changes the racy reading to be same for the
triggered file as for the rest of the files:

current:
- Inotify is triggered
- Triggering file is read for new data
- Rest of the files are read for timer expiration
- more inotifys are processed these may be at the end of the file already as well then

new:
- Inotify is triggered
- All files are read for timer expiration, which includes the triggering file
- Triggering files is read for data, this may be at the end of the file already then
- more inotifys are processed these may be at the end of the file already as well then

Therefore the change introduced harmonises the way triggering and the future triggers
are handled, by not making the triggering one as an exceptional one, so it is easier
to test the changes as well.

### DIFF
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -2444,10 +2444,11 @@ do_inotify(void)
 	int rd;
 	int currev;
 	static int last_timeout = 0;
+	struct pollfd pollfd;
 	DEFiRet;
 
 	CHKiRet(wdmapInit());
-	ino_fd = inotify_init();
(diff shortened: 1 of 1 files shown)


rgerhards:
Assigning milestone in the hope that this PR is still mergeable - else I'll drop it, as it was abandoned beforehand.

PR #4849, opened 2022-04-12
Fix non-null-terminated-string used with strlen

bucko909:
The `failedmsg_entry` expects a null-terminated string in `key`, but
here we allocate with malloc and copy a string-with-length-n into only
the first n bytes. If the final byte is null, this is by coincidence
only.

We've observed this by means of seeing random binary data appended to
keys submitted to kafka apparently at random, and this looks like a
smoking gun.


bucko909:
(I've changed only the commit message in the above push, "length into all but" -> "length-n into only".)

PR #4850, opened 2022-04-15
MMDBLOOKUP::FIXED:: Fix parsing of data received from mmdb lib

frikilax:
### Fixed
- [MMDBLOOKUP] wrong copy of buffer following parse of libmaxminddb's return after a successful search sometimes failed to return specific field from data

### Added
- [MMDBLOOKUP] support arrays in MMDB entry
- [MMDBLOOKUP] support escaped quotes '"' in MMDB entry
- [MMDBLOOKUP] support '<' characters in MMDB entry, when in a field
- [MMDBLOOKUP] support '}' characters in MMDB entry, when in a field


rgerhards:
Thx for the PR!

PR #4881, opened 2022-05-25
omazureeventhubs: Initial implementation of new output module

alorbach:
omazureeventhubs output module uses Apache "Qpid Proton C API" which is a solid
AMQP protocol library implementation that can be integrated
very well into the rsyslog dev environment.

- Implemented Delivery with submitted and accepted state checking
- impstatscounter used in testcases
- saving of failed messages in a failed list with support of saving
  and restoring.
- Add testcases (requires ENV variables) to testbench
- Using application/octect-stream (binary) to send messages based on
  Microsoft Code Sample:
  https://docs.microsoft.com/en-us/azure/event-hubs/event-hubs-c-getstarted-send
  * Note original Microsoft Samplecode is not working anymore, we are using
  * QPID Proton Proactor based on
    https://github.com/apache/qpid-proton/blob/main/c/examples/send.c
- requires QPID-PROTON Version 0.13 or higher because of the proactor API
- Add EventProperties configuration parameters
- using internal array instead of linkedlist for better performance
- using single byte helper to store message status (ubsubmit/submitted...)
- trigger wake up if messages need to be resubmitted (rejected)
- Slow down when sender credit reaches zero (10ns).
- Add enhanced performance stress test omazureeventhubs-stress.sh
- Add support for static library linking of qpid-proton
  This is needed to build the module from source and remove
  library package dependencies.
- adjusted valgrind suppressions

### DIFF
--- a/plugins/omazureeventhubs/omazureeventhubs.c
+++ b/plugins/omazureeventhubs/omazureeventhubs.c
@@ -1,0 +1,1309 @@
+/* omazureeventhubs.c
+ * This output plugin make rsyslog talk to Azure EventHubs.
+ *
+ * Copyright 2014-2017 by Adiscon GmbH.
+ *
+ * This file is part of rsyslog.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *       -or-
+ *       see COPYING.ASL20 in the source distribution
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "config.h"
+#include <stdio.h>
+#include <stdarg.h>
+#include <stdlib.h>
+#include <string.h>
+#include <strings.h>
+#include <assert.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <pthread.h>
+#include <sys/uio.h>
+#include <sys/queue.h>
+#include <sys/types.h>
+#include <math.h>
+#ifdef HAVE_SYS_STAT_H
+#	include <sys/stat.h>
+#endif
+#include <sys/time.h>
+#include <time.h>
+
+// Include Proton headers
+#include <proton/version.h>
+#include <proton/condition.h>
+#include <proton/connection.h>
+#include <proton/delivery.h>
+#include <proton/link.h>
+#include <proton/listener.h>
+#include <proton/netaddr.h>
+#include <proton/message.h>
+#include <proton/object.h>
+#include <proton/proactor.h>
+#include <proton/sasl.h>
+#include <proton/session.h>
+#include <proton/transport.h>
+#include <proton/ssl.h>
+
+// Include rsyslog headers
+#include "rsyslog.h"
+#include "syslogd-types.h"
+#include "srUtils.h"
+#include "template.h"
+#include "module-template.h"
+#include "errmsg.h"
+#include "atomic.h"
+#include "statsobj.h"
+#include "unicode-helper.h"
+#include "datetime.h"
+#include "glbl.h"
+
+MODULE_TYPE_OUTPUT
+MODULE_TYPE_NOKEEP
+MODULE_CNFNAME("omazureeventhubs")
+
+/* internal structures
+ */
+DEF_OMOD_STATIC_DATA
+DEFobjCurrIf(glbl)
+DEFobjCurrIf(datetime)
+DEFobjCurrIf(strm)
+DEFobjCurrIf(statsobj)
+
+statsobj_t *azureStats;
+STATSCOUNTER_DEF(ctrMessageSubmit, mutCtrMessageSubmit);
+STATSCOUNTER_DEF(ctrAzureFail, mutCtrAzureFail);
+STATSCOUNTER_DEF(ctrCacheMiss, mutCtrCacheMiss);
+STATSCOUNTER_DEF(ctrCacheEvict, mutCtrCacheEvict);
+STATSCOUNTER_DEF(ctrCacheSkip, mutCtrCacheSkip);
+STATSCOUNTER_DEF(ctrAzureAck, mutCtrAzureAck);
+STATSCOUNTER_DEF(ctrAzureMsgTooLarge, mutCtrAzureMsgTooLarge);
+STATSCOUNTER_DEF(ctrAzureQueueFull, mutCtrAzureQueueFull);
+STATSCOUNTER_DEF(ctrAzureOtherErrors, mutCtrAzureOtherErrors);
+STATSCOUNTER_DEF(ctrAzureRespTimedOut, mutCtrAzureRespTimedOut);
+STATSCOUNTER_DEF(ctrAzureRespTransport, mutCtrAzureRespTransport);
+STATSCOUNTER_DEF(ctrAzureRespBrokerDown, mutCtrAzureRespBrokerDown);
+STATSCOUNTER_DEF(ctrAzureRespAuth, mutCtrAzureRespAuth);
+STATSCOUNTER_DEF(ctrAzureRespSSL, mutCtrAzureRespSSL);
+STATSCOUNTER_DEF(ctrAzureRespOther, mutCtrAzureRespOther);
+
+#define MAX_ERRMSG 1024		/* max size of error messages that we support */
+#define MAX_DEFAULTMSGS 1024	/* Initial max size of the proton message helper array */
+
+#define SETUP_PROTON_NONE 0
+#define SETUP_PROTON_AUTOCLOSE 1
+
+/* flags for writeAzure: shall we resubmit a failed message? */
+#define RESUBMIT	1
+#define NO_RESUBMIT	0
+
+/* flags for transaction Handling */
+enum proton_submission_status
+{
+	PROTON_UNSUBMITTED = 0,	// Message not submitted yet
+	PROTON_SUBMITTED,	// Message submitted to proton sender instance
+	PROTON_ACCEPTED,	// Message accepted from remote target
+	PROTON_REJECTED,	// Message rejected from remote target (zero credit?)
+};
+
+// event_property NEEDED?
+struct event_property {
+	const char *key;
+	const char *val;
+};
+
+static pn_timestamp_t time_now(void);
+
+/* Struct for Proton Messages Listitems */
+struct s_protonmsg_entry {
+	uchar* payload;
+	size_t payload_len;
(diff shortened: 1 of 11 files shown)


plugins/omazureeventhubs/omazureeventhubs.c
```diff
@@ -215,41 +146,30 @@ typedef struct _instanceData {
...
-	uchar *statsFile;
-	int fdStatsFile;        /* stats file fd or -1 if not open */
-	pthread_mutex_t mutStatsFile;
-	int bIsOpen;
-	int bIsSuspended;	/* when broker fail, we need to suspend the action */
+	int bIsConnected;		/* 1 if connected, 0 if disconnected */
+	int bIsSuspended;		/* when broker fail, we need to suspend the action */
 	pthread_rwlock_t pnLock;
-	pthread_mutex_t mut_doAction; /* make sure one wrkr instance max in parallel */
+	pthread_mutex_t mut_doAction;	/* make sure one wrkr instance max in parallel */
```
rgerhards: Is this a requirement? If not, we should drop it. Anyhow, for the first version this may be useful to ensure there are no races.

plugins/omazureeventhubs/omazureeventhubs.c
```diff
@@ -1459,27 +654,119 @@ CODESTARTtryResume
...
+ *	New Transaction action interface
  */
+BEGINcommitTransaction
+	instanceData *__restrict__ const pData = pWrkrData->pData;
+	unsigned i;
+	sbool bDone = 0;
+	protonmsg_entry* pMsgEntry = NULL;
+CODESTARTcommitTransaction
+	pthread_mutex_lock(&pData->mut_doAction);
+#ifndef NDEBUG
```
rgerhards, line 661: you do not need the ifdef, at least not really. there are configure options which turn out the DBGPRINTF macro. Usually it's a good idea to have debug info in release build as well.

plugins/omazureeventhubs/omazureeventhubs.c
```diff
@@ -1459,27 +654,119 @@ CODESTARTtryResume
...
+			// Get reference to Proton Array Helper
+			pMsgEntry = ((protonmsg_entry*)pData->aProtonMsgs[i]);
+			if (	pMsgEntry == NULL) {
+				// Send Message to Proton 
+				writeProton(pData, pParams, i);
+			} else if (	pMsgEntry->submitted == 1 &&
+					pMsgEntry->rejected == 1) {
+				// Reset Message Entry, it will be retried
+				pMsgEntry->submitted = 0;
+				pMsgEntry->accepted = 0;
```
rgerhards: wouldn't a single status field with values (submitted, accepted, rejected) be easier? I guess they are mutally excluvise?

plugins/omazureeventhubs/omazureeventhubs.c
```diff
@@ -1773,3 +1037,341 @@ pn_timestamp_t time_now(void)
...
+		if(glbl.GetGlobalInputTermState() == 1)
+			break; /* terminate input! */
+
+		/* Note: the additional 10000ns wait is vitally important. It guards rsyslog
+		 * against totally hogging the CPU if the users selects a polling interval
+		 * of 0 seconds. It doesn't hurt any other valid scenario. So do not remove.
+		 * rgerhards, 2008-02-14
+		 */
+		if(glbl.GetGlobalInputTermState() == 0)
+			srSleep(0, 100000);
```
rgerhards: is this whole section really needed or a "copy victim" form a "copy victim" that copied imfile? Can this code really hog the CPU? I see the pn_proactor-wait() call plus myself insight the comment, which smells fishy ;-)

plugins/omazureeventhubs/omazureeventhubs.c
```diff
@@ -1773,3 +1037,341 @@ pn_timestamp_t time_now(void)
...
+	case PN_LINK_REMOTE_CLOSE:
+	case PN_LINK_REMOTE_DETACH:
+		DBGPRINTF("handleProton: remote link closed for %s\n", pData->azurehost);
+		proton_check_condition(event, pn_link_remote_condition(pn_event_link(event)), 
+			"remote link closed");
+		break;
+	case PN_PROACTOR_INACTIVE:
+		DBGPRINTF("handleProton: INAKTIVE for %s\n", pData->azurehost);
+		break;
+#ifdef __GNU_C
```
rgerhards, line 1302: why the #ifdef?

rgerhards (changes requested):
I admit I have not gone through all of the code, but thought some comments might be useful. Maybe have a side discussion makes sense.

plugins/omazureeventhubs/omazureeventhubs.c
```diff
@@ -1459,27 +654,119 @@ CODESTARTtryResume
...
+				}
+			}
+			free(pData->aProtonMsgs);
+		}
+
+		// Expand our proton helper array
+		DBGPRINTF("omazureeventhubs: commitTransaction expand helper array from %d to %d for pData %p\n",
+			pData->nMaxProtonMsgs, nParams, pData);
+		pData->nMaxProtonMsgs = nParams; // Set new MAX
+		CHKmalloc(pData->aProtonMsgs = calloc(pData->nMaxProtonMsgs, sizeof(struct s_protonmsg_entry)));
```
rgerhards: remalloc() is quicker. You may even keep the proton messages, depding on their exact semantics. But in case of remalloc() you must zero out the new elemets yourself.

rgerhards (changes requested):
sry, missed this comment in initial review.

plugins/omazureeventhubs/omazureeventhubs.c
alorbach: At the moment it works like mutWrite in omfile, it makes sure commitTransaction can only run once per instance.

plugins/omazureeventhubs/omazureeventhubs.c
alorbach: I don't want to turn them off completely, just some are very verbose and only needed for dev now - later for production build they would not be needed at all.

plugins/omazureeventhubs/omazureeventhubs.c
rgerhards: Note that omfile has a hard reason to limit itself to one active instance: it is simply impossible to write to **sequential** files concurrently (well, you can, but you'll get "garbagge", aka randomly split and intermixed lines). However, this mutex is a kind of last resort and should only be used by an action, if no concurrency is technically possible. I would guess the qpid proton can handle concurrency.

That said, I still think it is a good idea to limit concurrency in the early stages, as it simplifies debugging. But at least a comment should explain if or if not this is a hard requirement.

alorbach: Alright we will decide in a later step if the mutex is really needed.

plugins/omazureeventhubs/omazureeventhubs.c
alorbach: Submitted can be accepted or rejected and the different states matter. Could use a bit mask but in this case it is better human readable this way.

plugins/omazureeventhubs/omazureeventhubs.c
alorbach: copy victim but needs to be confirmedd if needed.

plugins/omazureeventhubs/omazureeventhubs.c
alorbach: It turns of those warnings: 

`error: enumeration value ‘PN_EVENT_NONE’ not handled in switch [-Werror=switch-enum]`

I am open for other solutions

rgerhards: ok, I see. Thx! I would suggest to add this info as a comment.

plugins/omazureeventhubs/omazureeventhubs.c
alorbach: I chose calloc because of zeroing, and the message buffer expansion should not happen that often so from my point of view it is acceptable that calloc is a little less performant.

plugins/omazureeventhubs/omazureeventhubs.c
rgerhards: question is: is it possible for azure to hammer the server? Doesn't look so. Pls change at least the comment (I didn't do it ;-)) and note that this is fishy and needs to be reviewed later.

plugins/omazureeventhubs/omazureeventhubs.c
rgerhards: I thought the live cycle would be
```
unsubmitted -> submitted ----> rejected ---> processed
      ^              |            |
      |              |            |
      +--------------+------------+
```
Is there any situation where the same message is both submitted and rejected? If so, doesn't that mean it would need to be resubmitted, which means change state to unsubmitted and the to submitted again?

rgerhards: follow up: in a sense, can't we transit from "rejected" directly to "unsubmitted"?

I assume that on reject a short wait is done for rate-liminting. No matter if this happens during the initial submit or after submit when the storage engine rejects the message.

alorbach: OK thats a point, I will change it to a single variable later.

plugins/omazureeventhubs/omazureeventhubs.c
alorbach: Actually that whole block can be removed. (will be next amend)

plugins/omazureeventhubs/omazureeventhubs.c
rgerhards: It's actually more than you think. Because with calloc(), you need to free the previous array (else you get a memory leak). With relloc(), this is not necessary, as the previsou array is allocated. The important thing to note is that the free/calloc path requires two calls to system memory allocator. This is pretty expensive, because most allocators use a global lock for there operation, ending all concurrency for other threads using the malloc system. So each call makes a big difference. I am telling from experience. Also, this is one that that usually is not corrected when not done right initially. And usually it should be pretty cheap in terms of effort to do it right up from the begining. Just my 2 cts...

plugins/omazureeventhubs/omazureeventhubs.c
```diff
@@ -514,66 +505,59 @@ writeProton(instanceData *__restrict__ const pData,
...
-	} else {
-		/* Optional */
-		const int callbacksCalled = rd_kafka_poll(pData->rk, 0); /* call callbacks */
-		LogMsg(0, NO_ERRCODE, LOG_INFO, "omazureeventhubs: doHUP kafka - '%s' outqueue length: %d,"
-			"callbacks called %d\n", pData->tplName,
-			rd_kafka_outq_len(pData->rk), callbacksCalled);
-	}
-finalize_it:
-#endif
+	// Nothing todo in doHUP for the time being
```
rgerhards: IIRC this entry point can be removed if not required (framework will check that) - 85% sure ;-)

alorbach: allright will remove it next amend

rgerhards (approved):

Issue #4797, opened 2022-02-07
Imfile: Logrotation causes duplicates when processing pipeline is long


StrongestNumber9:
### Expected behavior
No data gets duplicated

### Actual behavior
Events get duplicated, amount depends on load and pipeline length.

Snip from PoC container
```sh
Successfully built 0174257d958b
Successfully tagged rsyslog-logrotate-reread:latest
Starting curl spamming
Starting logrotates
Logrotate and httpd graceful reload, round 1
Logrotate and httpd graceful reload, round 2
Logrotate and httpd graceful reload, round 3
Logrotates done
Curling done, sleeping a bit to allow rsyslog to catch up
Amount of logs on disk: 10000
Amount of logs in spool.log: 5037
<snip>
Amount of logs in spool.log: 18001
Found total of 18001 events
And 8001 were duplicates! :(
Example of duplicate line: 
2022-02-07T13:08:38.059282+00:00 2a6dd7aab40f access_log 127.0.0.1 - - [07/Feb/2022:13:08:37 +0000] "GET /request-3003 HTTP/1.1" 404 210 "-" "curl/7.29.0" -- /logs/access_log -- 98294
2022-02-07T13:09:14.213275+00:00 2a6dd7aab40f access_log 127.0.0.1 - - [07/Feb/2022:13:08:37 +0000] "GET /request-3003 HTTP/1.1" 404 210 "-" "curl/7.29.0" -- /logs/access_log -- 0
Sleeping indefinitely
```

### Steps to reproduce the behavior
Have long and slow event processing pipeline, write a lot of logs and run logrotate. 

### Environment
Newest rsyslog from adiscon repos, host is Fedora 35 and container Centos 7

```sh
# rsyslogd -v
rsyslogd  8.2112.0 (aka 2021.12) compiled with:
        PLATFORM:                               x86_64-redhat-linux-gnu

# cat /etc/redhat-release 
CentOS Linux release 7.9.2009 (Core)

```

### Ready to run container + configs

Attached are container where the behaviour can be observed. Configurations are intentionally crafted to emulate the heavily loaded machine and mileage might vary on faster/slower machines

[reread-issue.zip](https://github.com/rsyslog/rsyslog/files/8015480/reread-issue.zip)

davidelang:
what is your logrotate config? It would probably help to see your rsyslog config 
as well.

David Lang

StrongestNumber9:
[...]

PR #4883, opened 2022-05-27
testbench: add testcase for frequent imfile input file change

rgerhards:
PoC test, yet incomplete

This patch not only contains the new test but also supporting
changes to testbench tooling.

see also: https://github.com/rsyslog/rsyslog/issues/4797


rgerhards:
was a duplicate

PR #4900, opened 2022-06-15
strings.h is missing in several files

kulikjak:
We recently enabled OpenSSL stream driver on Solaris and hit two build issues where the `index()` function was missing its header. This simple PR fixes that issue.

### DIFF
--- a/runtime/nsd_ossl.c
+++ b/runtime/nsd_ossl.c
@@ -28,6 +28,7 @@
 #include <stdlib.h>
 #include <assert.h>
 #include <string.h>
+#include <strings.h>
 #include <openssl/ssl.h>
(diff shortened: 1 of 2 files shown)


kulikjak:
I presume (an hope) those test failures are not this PR related...

rgerhards:
> I presume (an hope) those test failures are not this PR related...

looks like a flake - I'll take care of it :-)

thx for the PR!

rgerhards:
CI Fail and missing runs are due to CI Updates and thus no concern.

kulikjak:
Thanks for merge!

rgerhards:
Thx for the PR! :-)

PR #4903, opened 2022-06-17
omfile: support for zstd compression

rgerhards:
The zstd library provides better and faster compression than zlib.
This patch integrates zstd as a dynamically-loadable functionality.
As such, no further dependencies need to be added to the rsyslog
base package.

Due to the increased performance, usage of zstd is highly recommended
for high-volume use cases.

### DIFF
--- a/.github/workflows/run_checks.yml
+++ b/.github/workflows/run_checks.yml
@@ -177,6 +177,11 @@ jobs:
                      --enable-imdiag --enable-impstats --enable-imfile --disable-imfile-tests \
                      --disable-fmhttp --enable-valgrind --disable-default-tests \
                      --enable-elasticsearch-tests --enable-elasticsearch"
+              export CI_MAKE_OPT='-j20'
+              export CI_MAKE_CHECK_OPT='-j8'
+              export CI_CHECK_CMD='check'
+              export CI_VALGRIND_SUPPRESSIONS="ubuntu22.04.supp" # they are still valid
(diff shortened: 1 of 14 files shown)


davidelang:
how well does it work when the file is closed and re-opened? I've sometimes had 
issues with zlib compressed files when there are frequent HUPs

Not looking for the super-inefficient but super-safe mode that we have to have 
zlib reset itself after each log entry, but we need to understand and document 
how it works under harsh conditions.

Also, what happens if you have multiple actions writing to the same file?

David Lang

rgerhards:
Zstd compression will work along the same lines and limitations as zlib compression. 

I have never seen a case where frequent HUP on a reasonable scale (unreasonable: just some seconds) causes issues. I don't say they don't exist, but I do not know about them. If we can reproduce, we should look into it.

Writing via multiple actions is indeed an issue. We should add least add this to the doc, if not already present.

The main intent of this PR is to enable far faster compression than zlib does. This is especially the case for very high workloads (zstd is at last twice as fast, depending on actual case  n times faster with n > 2). This is a big win. Especially with multiple threads.

rgerhards:
side-note: as given in the description: this PR is not yet finished. I'll work on it within the next time. It was just the initial version that I wanted to have CI runs on. I may need to restructure some code, even with zlib, but this needs to be seen in detail.

rgerhards:
doc in https://github.com/rsyslog/rsyslog-doc/pull/973

Issue #4902, opened 2022-06-17
rsyslog silenty discards any received messages for an action if "error starting up action queue" is encountered.


kortemik:
### Expected behavior
rsyslog determines that there is an "error starting up action queue" and shuts itself down to prevent any message loss. especially if ```$AbortOnUncleanConfig on``` is set.

### Actual behavior
rsyslog starts up and logs "error starting up action queue". It does not inform that an action is disabled and messages may be lost if no further actions are configured as message receptions is on.

### Steps to reproduce the behavior
reproduce a situation where "error starting up action queue" is encountered.
configure rsyslog with imrelp for reliable event reception and configure an action with disk queue for omrelp to ensure that delivery problems can be mitigated. try sending data in and see that it is nowhere to be found on the omrelp side.

### Environment
- rsyslog version: v8.2102.0
- platform: el7

code responsible for the behaviour
```
/* helper to activateActions, it activates a specific action.                                    
 */                                                                                              
DEFFUNC_llExecFunc(doActivateActions)                                                            
{                                                                                                
        rsRetVal localRet;                                                                       
        action_t * const pThis = (action_t*) pData;                                              
        localRet = qqueueStart(pThis->pQueue);                                                   
        if(localRet != RS_RET_OK) {                                                              
                LogError(0, localRet, "error starting up action queue");                         
                if(localRet == RS_RET_FILE_PREFIX_MISSING) {                                     
                        LogError(0, localRet, "file prefix (work directory?) "                   
                                        "is missing");                                           
                }                                                                                
                actionDisable(pThis);                                                            
        }                                                                                        
        DBGPRINTF("Action %s[%p]: queue %p started\n", modGetName(pThis->pMod),                  
                  pThis, pThis->pQueue);                                                         
        return RS_RET_OK; /* we ignore errors, we can not do anything either way */              
}    
```

rgerhards:
Without special config, the behavior is intended: we consider it preferable to fail one action instead of all processing. This is the original spirit of (r)syslogd for over 30 years.

That said, I see this does not fit all use cases. Not sure if "abortOnUncleanConfig" should address this (after all the config is fine) or a separate param. But this functionality must be added.

rgerhards:
@kortemik I have created https://github.com/rsyslog/rsyslog/pull/4904

CI is currently running, but manual tests look good. Doc is missing, but I will add either tomorrow or begin of week. 

I think you can already apply it. So I am a bit before mid-summer ;-)

[...]

PR #4904, opened 2022-06-18
config: add "abortOnFailedQueueStartup" global config parameter

rgerhards:
similiar to "abortONUncleanConfig", this parameter aborts rsyslog
when a queue has problems during startup. Some users perfer rsyslog
to terminate in this case. By default, nothing changes.

closes https://github.com/rsyslog/rsyslog/issues/4902

### DIFF
--- a/action.c
+++ b/action.c
@@ -64,7 +64,7 @@
  * beast.
  * rgerhards, 2011-06-15
  *
- * Copyright 2007-2019 Rainer Gerhards and Adiscon GmbH.
+ * Copyright 2007-2022 Rainer Gerhards and Adiscon GmbH.
  *
  * This file is part of rsyslog.
  *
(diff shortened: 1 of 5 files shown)


StrongestNumber9:
I noticed in my tests that setting `abortOnFailedQueueStartup="on"` and using either `$PrivDropToUser`or `$PrivDropToGroup` will cause pidfile to be written earlier in the startup process but it never gets cleaned up if action fails and rsyslog shuts down. It works as expected when I _don't_ use `$PrivDrop..`

Unfortunately I can't test it with the official version as this PR is not yet merged, my rsyslog is version `8.2102` + this patch on top of it.

rgerhards:
@kortemik sry, I overlooked the comment. Will look into it, but it will take a short while.

rgerhards:
@StrongestNumber9 I have amended the commit. IMHO removing the PID file at the point of failure looks like the best option. I would appreciate if you could give the patch another try.

rgerhards:
doc in https://github.com/rsyslog/rsyslog-doc/pull/987

PR #4905, opened 2022-06-21
testbench: Add more valgrind tests for sndrcv (omrelp/imrelp)

alorbach:
- These tests will help find race conditions hopefully
- fix diag.sh issues running second instance in valgrind mode only


alorbach:
will require https://github.com/rsyslog/librelp/pull/239 to be merged before tests can continue

PR #4921, opened 2022-06-30
MMDBLOOKUP::FIXED:: Avoid Rsyslog crashes when a mmdb file cannot be opened/read

frikilax:
### Fixed
- [MMDBLOOKUP] Handle failures to read/access one mmdb file properly and suspend the module for some time instead of crashing Rsyslog with a failed assert


rgerhards:
FYI: Ubuntu 22 CI is currently in development, fail is "ok"

Issue #4913, opened 2022-06-27
does it really memory leak in gtlsGlblInit while fail in tls negotiate?


YanAnzouyijun:
while i reset the remote frequently, the asan reports there are a lot of memory leak
### Expected behavior
no mem leak in asan
### Actual behavior
the asan reports there are a lot of memory leak
### Steps to reproduce the behavior
mem leak 1:
Indirect leak of 32 byte(s) in 1 object(s) allocated from:
    #0 0xffff9e000230 in __interceptor_calloc (/usr/lib64/libasan.so.6+0xa5230)
    #1 0xffff92d2d1c4 in gnutls_dh_params_init (/usr/lib64/libgnutls.so.30+0x511c4)
    #2 0xffff99ada8d0 in gtlsGlblInit /usr/src/debug/rsyslog-8.2110.0-7.h7..aarch64/runtime/nsd_gtls.c:749
    #3 0xffff99ada8d0 in nsd_gtlsClassInit /usr/src/debug/rsyslog-8.2110.0-7.h7..aarch64/runtime/nsd_gtls.c:2322
    #4 0xffff99ada8d0 in nsd_gtlsClassInit /usr/src/debug/rsyslog-8.2110.0-7.h7..aarch64/runtime/nsd_gtls.c:2314
    #5 0xffff99adabb0 in modInit /usr/src/debug/rsyslog-8.2110.0-7.h7..aarch64/runtime/nsd_gtls.c:2348
    #6 0xaaaada3b5e0c in doModInit /usr/src/debug/rsyslog-8.2110.0-7.h7..aarch64/runtime/modules.c:584
    #7 0xaaaada3b9598 in Load /usr/src/debug/rsyslog-8.2110.0-7.h7..aarch64/runtime/modules.c:1263
    #8 0xaaaada3b2264 in UseObj /usr/src/debug/rsyslog-8.2110.0-7.h7..aarch64/runtime/obj.c:1231
    #9 0xffff99f6846c  (<unknown module>)
    #10 0xaaaada32df9c in TCPSendInit /usr/src/debug/rsyslog-8.2110.0-7.h7..aarch64/tools/omfwd.c:812
    #11 0xaaaada32ec1c in doTryResume /usr/src/debug/rsyslog-8.2110.0-7.h7..aarch64/tools/omfwd.c:1006
    #12 0xaaaada3f9b40 in actionPrepare ../action.c:1078
[...]

PR #4929, opened 2022-07-07
Fix memory leak when globally de-initialize GnuTLS

seuzw930:
During globally de-initialize GnuTLS, server anon credentials and server DH parameters for anon mode were not null. It resulted in memory leak and this patch fixes this behaviour.

https://github.com/rsyslog/rsyslog/issues/4913


rgerhards:
Thx for the PR!

Issue #4928, opened 2022-07-07
Support for SHA-256 fingerprint authentication


codemaker219:
### Expected behavior
Support SHA-256 hashes for 'StreamDriverPermittedPeers` if `StreamDriverAuthMode="x509/fingerprint"` 
### Actual behavior
Currently, we can only use SHA-1 hashes which are already deprecated a long time ago. It would be nice to have the ability to use SHA-256 hashes. 
### Steps to reproduce the behavior
Its more like a feature request :-) I will try to implement that and create a PR for this.

rgerhards:
SHA-256 certificate fingerprints are supported by the current TLS netstream drivers. In fingerprint authentication modes, permitted peers may use the `SHA256:` prefix; both the GnuTLS and OpenSSL implementations recognize it.

Documentation and examples:
- GnuTLS: https://www.rsyslog.com/doc/concepts/ns_gtls.html
- OpenSSL: https://www.rsyslog.com/doc/concepts/ns_ossl.html

Closing as implemented.

PR #4930, opened 2022-07-07
support sha256 for StreamDriverAuthMode="x509/fingerprint"

codemaker219:
Support sha256 hashes for "x509/fingerprint" for gtls and ossl
#4928

### DIFF
--- a/runtime/nsd_ossl.c
+++ b/runtime/nsd_ossl.c
@@ -381,7 +381,7 @@ long BIO_debug_callback(BIO *bio, int cmd, const char __attribute__((unused)) *a
  * rgerhards, 2008-05-08
  */
 static rsRetVal
-GenFingerprintStr(uchar *pFingerprint, size_t sizeFingerprint, cstr_t **ppStr)
+GenFingerprintStr(uchar *pFingerprint, size_t sizeFingerprint, cstr_t **ppStr, const char* prefix)
 {
 	cstr_t *pStr = NULL;
 	uchar buf[4];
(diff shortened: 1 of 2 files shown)


runtime/nsd_ossl.c
```diff
@@ -633,6 +644,9 @@ osslChkPeerFingerprint(nsd_ossl_t *pThis, X509 *pCert)
 		if(!rsCStrSzStrCmp(pstrFingerprint, pPeer->pszID, strlen((char*) pPeer->pszID))) {
 			dbgprintf("osslChkPeerFingerprint: peer's certificate MATCH found: %s\n", pPeer->pszID);
 			bFoundPositiveMatch = 1;
+		} else if(!rsCStrSzStrCmp(pstrFingerprintSha256, pPeer->pszID, strlen((char*) pPeer->pszID))) {
+			dbgprintf("osslChkPeerFingerprint: peer's certificate MATCH found: %s\n", pPeer->pszID);
```
alorbach: Perhaps adding "Sha256" and "Sha1" to dbgprintf() ? would help to debug which matched the fingerprint

alorbach:
Thanks for the contribution. The PR looks fine to me, the only recommendation I have is to add SHA1 / SHA256 to the dbgprintf outputs, so you can see which type matched.

codemaker219:
Thanks for the feedback. I modified the log message for gtls and added them to ossl.

alorbach:
Thanks for the changes :), I believe there is a limit of commits in one PR. Can you squash the last 3 commits into one? Thanks

codemaker219:
Done

rgerhards:
Thanks for the PR, but I think we need to update the doc a bit. Please add doc to https://github.com/rsyslog/rsyslog-doc/ If you think this is not necessary, please let me know why not.

rgerhards:
> Should I update the documentation?

good catch! yes, please update!

codemaker219:
https://github.com/rsyslog/rsyslog-doc/pull/980

PR #4931, opened 2022-07-11
imptcp: restructure and optimize performance

rgerhards:


rgerhards:
This did not work out, replaced by #4937

PR #4932, opened 2022-07-12
performance-tune json generation

rgerhards:


rgerhards:
A possibly better alternative is in #4935.

rgerhards:
replaced by #4935

Issue #4938, opened 2022-07-16
rsyslogd leak memory when free action worker data table


seuzw930:
### Expected behavior
no mem leak in when free action worker data table
### Actual behavior
the asan reports there is a memory leak when an action finished
### Steps to reproduce the behavior
There is an asan report when rsyslog complete the action operation:
Direct leak of 80 byte(s) in 1 object(s) allocated from:
    #0 0xffff81b5f230 in __interceptor_calloc (/usr/lib64/libasan.so.6+0xa5230)
    #1 0xffff7da8bd0c  (<unknown module>)
    #2 0xaaaae77c9f04 in initTCP /usr/src/debug/rsyslog-8.2110.0-7.eulerosv2r11.aarch64/tools/omfwd.c:1185
    #3 0xaaaae77c9f04 in createWrkrInstance /usr/src/debug/rsyslog-8.2110.0-7.eulerosv2r11.aarch64/tools/omfwd.c:412
    #4 0xaaaae78967c0 in actionCheckAndCreateWrkrInstance ../action.c:984
    #5 0xaaaae789f6b4 in actionPrepare ../action.c:1071
    #6 0xaaaae789f6b4 in processMsgMain ../action.c:1662
    #7 0xaaaae78a0668 in doSubmitToActionQ ../action.c:1834
    #8 0xaaaae788a5f4 in execAct /usr/src/debug/rsyslog-8.2110.0-7.eulerosv2r11.aarch64/runtime/ruleset.c:209
    #9 0xaaaae788a5f4 in scriptExec /usr/src/debug/rsyslog-8.2110.0-7.eulerosv2r11.aarch64/runtime/ruleset.c:599
    #10 0xaaaae788be60 in processBatch /usr/src/debug/rsyslog-8.2110.0-7.eulerosv2r11.aarch64/runtime/ruleset.c:660
    #11 0xaaaae77c03a4 in msgConsumer /usr/src/debug/rsyslog-8.2110.0-7.eulerosv2r11.aarch64/tools/rsyslogd.c:696
    #12 0xaaaae788199c in ConsumerReg /usr/src/debug/rsyslog-8.2110.0-7.eulerosv2r11.aarch64/runtime/queue.c:2198
    #13 0xaaaae7874c9c in wtiWorker /usr/src/debug/rsyslog-8.2110.0-7.eulerosv2r11.aarch64/runtime/wti.c:428
[...]

PR #4939, opened 2022-07-18
Fix memory leak when free action worker data table

seuzw930:
During free action worker data table when action destruct, worker instance in worker data table were not null. It resulted in memory leak and this patch fixes this behaviour.

https://github.com/rsyslog/rsyslog/issues/4938


rgerhards:
thx for the PR, much appreciated.

PR #4942, opened 2022-07-27
ksi bugfix: possible crash fixed when several log files are opened

taavi-valjaots:
KSI module in async mode used to request aggregator conf. every time a log
file was opened. When several log files were opened simultaneously
corresponding amount of pointless concurrent conf. requests were posted.
Concurrent conf. requests lead to a bug in libksi, where internal count of
pending requests was not decremented correctly causing system to crash.

Fix for the issue is to optimize the frequency of conf. requests so that only
one conf. requests is handled at once. Instead of checking conf. every time
log file is opened, conf is requested periodically after conf timeout. This will
affect both sync and async mode.

New option for KSI module introduced - sig.confinterval="time, s".

### DIFF
--- a/runtime/lib_ksils12.c
+++ b/runtime/lib_ksils12.c
@@ -732,6 +732,19 @@ static void handle_ksi_config(rsksictx ctx, KSI_AsyncService *as, KSI_Config *co
 	}
 }
 
+static int
+isAggrConfNeeded(rsksictx ctx) {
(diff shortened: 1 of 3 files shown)


rgerhards:
Thanks for the PR and sorry for the late reply. Please add doc to https://github.com/rsyslog/rsyslog-doc/ so that users can discover the new option.

Issue #4940, opened 2022-07-19
remote stack overflow: mmanon strcpy() on long string


micah:
I have rsyslog running on a centralized log-collector with around 100 rsyslog clients connecting to it in order to ship logs there. It is *often* core dumping, at least once per day. Systemd notices this and restarts it, and then it core dumps again, resulting in ~6 coredumps in less than a minute before systemd decides that the service is too unstable and it stops trying to restart it. It then requires manual intervention to start it again, and then it runs fine for a while (after some high load as the queues from the remote clients are flushed).

I've attached some gdb backtraces from cores below, and you will find in the follow-up comments here, that it appears the problem appears to be in mmanon.

### Environment
```
rsyslogd  8.2206.0 (aka 2022.06) compiled with:
	PLATFORM:				x86_64-pc-linux-gnu
	PLATFORM (lsb_release -d):		
	FEATURE_REGEXP:				Yes
	GSSAPI Kerberos 5 support:		Yes
	FEATURE_DEBUG (debug build, slow code):	No
	32bit Atomic operations supported:	Yes
	64bit Atomic operations supported:	Yes
	memory allocator:			system default
	Runtime Instrumentation (slow code):	No
	uuid support:				Yes
	systemd support:			Yes
	Config file:				/etc/rsyslog.conf
	PID file:				/run/rsyslogd.pid
	Number of Bits in RainerScript integers: 64
```
- platform: The rsyslog collector server is Debian Bullseye.

Configuration on server:
```
global(
  maxMessageSize="64k"
)

module(load="imuxsock"
  SysSock.Use="off"
)

module(load="imkmsg")

module(
  load="omprog"
)

module(
  load="omrelp"
  tls.tlslib="openssl"
)

# The 'imklog' module before rsyslog 8.2008 always uses the default ruleset.
# Force 'incoming' as the default ruleset to catch kernel messages as well.
$DefaultRuleset incoming

# Use the journald syslog forwarding socket, instead of reading
# directly from the journal via the imjournal module. Some of the
# reasons for this are mentioned in
# https://www.rsyslog.com/doc/v8-stable/configuration/modules/imjournal.html
# and are related to performance and reliability.
#
# Unfortunately we lose the journald metadata this way, which would be
# useful to have (in that case all logs would be "structured" logs).
input(type="imuxsock" Socket="/run/systemd/journal/syslog" ruleset="incoming")

$IncludeConfig /etc/rsyslog.d/*.conf

[...]

PR #4943, opened 2022-07-28
mmanon: Simplified and fixed IPv4 digit detection.

alorbach:
- Fixed an issue with numbers above int64 in syntax_ipv4.
  Numbers that were up to 256 above the max of an int64
  could incorrectly be detected as valid ipv4 digit.
- Simplified the IPv4 digit detection function and renamed
  to isPosByte.
- added testcasse for malformed IPvc4 addresses

closes: https://github.com/rsyslog/rsyslog/issues/4940

### DIFF
--- a/plugins/mmanon/mmanon.c
+++ b/plugins/mmanon/mmanon.c
@@ -22,6 +22,7 @@
 #include "config.h"
 #include "rsyslog.h"
 #include <stdio.h>
+#include <ctype.h>
 #include <stdarg.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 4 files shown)


tests/mmanon_simple_mallformed_ipv4.sh
```diff
@@ -0,0 +1,37 @@
...
+generate_conf
+add_conf '
+template(name="outfmt" type="string" string="%msg%\n")
+
+module(load="../plugins/mmanon/.libs/mmanon")
+module(load="../plugins/imtcp/.libs/imtcp")
+input(type="imtcp" port="0" listenPortFileName="'$RSYSLOG_DYNNAME'.tcpflood_port" ruleset="testing")
+
+ruleset(name="testing") {
+	action(type="mmanon" ipv4.bits="32" ipv4.mode="simple")
```
micah, line 20: I dont think `mode=simple` tests the relevant code path, it should be `mode=zero`, if you look at https://github.com/rsyslog/rsyslog/blob/master/plugins/mmanon/mmanon.c#L785

alorbach: Indeed, but my main goal for this test was to make confirm that those string did not get detected as IPv4 addresses. The new syntax_ipv4 & isPosByte function implementation should makes sure valid IPv4 addresses will be valid in any case.

micah: I do agree with you, however... it might be wise to add a similar test with mode=zero, especially since it was not crashing because of that fix.

plugins/mmanon/mmanon.c
```diff
@@ -388,73 +389,83 @@ getHexVal(char c)
...
 {
-	int64_t val;
-	size_t nproc;
+	size_t nproc = 0;
 	size_t i;
 	int r = 0;
-
-	val = getPosInt(buf, buflen, &i);
-	if(val < 0 || val > 255)
+	if(	isPosByte(buf, buflen, &i) == 0) {
```
micah: Just code-quality note here, you added a space after the parens in all the changes you made, but the original code doesn't seem to have those (nor is the space balanced on the other end of the closing parens).

Note: I only point this out here, but it is the same for the other bits you added.

plugins/mmanon/mmanon.c
```diff
@@ -388,73 +389,83 @@ getHexVal(char c)
...
+	}
 	i += nproc;
 
 	*nprocessed = i;
 	r = 1;
-
 done:
+/*
+	DBGPRINTF("syntax_ipv4: %s Is IPV4=%s\n", buf, (r == 1 ? "YES" : "NO"));
+*/
```
micah: Did you mean to leave this code here commented out?

micah (changes requested):

plugins/mmanon/mmanon.c
alorbach: > 

yes it is quiet a very busy output which was helpful for debugging only. 
I have removed it along with the extra spaces

PR #4951, opened 2022-08-05
omrabbitmq: Add TLS support

21stcaveman:
Trying to enable TLS connections for omrabbitmq module.

This code is currently working on Rocky 8.6 running rsyslog 8.2102.0
Connection is made successfully to a RabbitMQ 3.10.6 server over TLS, messages flowing.

While I finish the testing, can someone more familiar with the codebase please review the code for any mistakes I might have made?

### DIFF
--- a/contrib/omrabbitmq/omrabbitmq.c
+++ b/contrib/omrabbitmq/omrabbitmq.c
@@ -4,6 +4,7 @@
  *
  * Copyright 2012-2013 Vaclav Tomec
  * Copyright 2014 Rainer Gerhards
+ * Copyright 2022 Hamid Maadani
  *
  * This program is free software: you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public License
@@ -22,7 +23,7 @@
  * Author: Vaclav Tomec
  * <vaclav.tomec@gmail.com>
  *
- * TLS Support added by:
+ * TLS & AMQP heartbeat support added by:
  * Hamid Maadani
  * <hamid@dexo.tech>
  *
@@ -140,6 +141,7 @@ typedef struct _instanceData {
 	int initOpenSSL;                /* should rabbitmq-c initialize OpenSSL? */
 	int verifyPeer;                 /* should peer be verified for TLS? */
 	int verifyHostname;             /* should hostname be verified for TLS? */
+	int heartbeat;                  /* AMQP heartbeat interval in seconds (0 means disabled, which is default) */
 	char *caCert;                   /* CA certificate to be used for TLS connection */
 
 	recover_t recover_policy;
@@ -186,6 +188,7 @@ static struct cnfparamdescr actpdescr[] = {
 	{ "host", eCmdHdlrString, 0 },
 	{ "port", eCmdHdlrInt, 0 },
 	{ "virtual_host", eCmdHdlrGetWord, 0 },
+	{ "heartbeat_interval", eCmdHdlrNonNegInt, 0 },
 	{ "user", eCmdHdlrGetWord, 0 },
 	{ "password", eCmdHdlrGetWord, 0 },
 	{ "ssl", eCmdHdlrBinary, 0 },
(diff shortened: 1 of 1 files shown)


rgerhards:
@21stcaveman Thx for the PR. It looks like we have some version dependency. Can you confirm the minimum version of RabitMQ needed? Can you do the necessary configure check (I can aid here)?

21stcaveman:
Ah, of course. This is compiled with v0.9.0. That would be the minimum requirement given that 'amqp_uninitialize_ssl_library' function was introduced in that version.

Speaking about that, is there any flag or function in rsyslog codebase to determine whether openssl has been initialized or not? Currently, I am relying on a config flag provided by the user, (init_openssl) which is admittedly not the best option.
Problem is, 'amqp_ssl_socket_new' will initialize openssl by default (unless stopped via amqp_set_initialize_ssl_library), and double initializing openssl would crash rsyslogd. Any suggestions on how to detect that?


Also, can you please put me in the right direction on the config check? what's necessary to be done there?

21stcaveman:
Fyi, I have added an additional commit to give a user the ability to enable AMQP heartbeats and set an interval for them.
I have been observing TCP connections drop client-side, but not get disconnected server-side. AMQP heartbeats seem to solve that issue.

rgerhards:
sry, working intermittently.  I don't think we have an "openssl is initialized" flag - we actually do not know if a library does do that internally :-( Maybe @alorbach  could comment on this, but he is unavailable the next ~2 weeks.

21stcaveman:
No worries at all.
A bit surprised openssl even allows double initialization, I have been looking to find a function in openssl itself that would let me know it has been already initialized with little success.
I think it would benefit rsyslog to have a static global flag for the process, which indicates if a library has initiated openssl. It should be required for each library that makes use of openssl to check that flag before initializing openssl, and set it if it does.
That being said, my tests for omrabbitmq are done. The only improvement I can think of at this point, would be moving openssl init away from the mod config.

rgerhards:
> I think it would benefit rsyslog to have a static global flag for the process, which indicates if a library has initiated openssl.

well, we often do not know what a support library does internally. I do not talk about rsyslog modules, but about libraries they use. We often have seen surprises there in the past.

rgerhards:
Forgot: That said, we can add a global flag to be set when rsyslog or a module calls it. It's not bullet-proof, but probably better ;-)

21stcaveman:
Ah, understood. That's fair, you can't control every single library.
Whichever way you guys decide is best, works for me. Let me know if you decide to go with the flag or something similar, so I can modify omrabbitmq accordingly.

alorbach:
> sry, working intermittently. I don't think we have an "openssl is initialized" flag - we actually do not know if a library does do that internally :-( Maybe @alorbach could comment on this, but he is unavailable the next ~2 weeks.

We do not have such a flag, but I think it would not hurt to look into this and once doing also implement using "OPENSSL_init_ssl" for newer OpenSSL Versions (We are using depreceated SSL_library_init).

rgerhards:
> Which tells me it is not unexpected that the init function is called multiple times.

This sounds like we do not need such a flag, or am I wrong?

21stcaveman:
> Which tells me it is not unexpected that the init function is called multiple times.

Certainly reads that way. I was looking to see what rabbitmq-c does for openssl initialization, and it seems like the call both  "OPENSSL_init_ssl" and "SSL_library_init" (see 'setup_openssl' in [here](https://github.com/alanxz/rabbitmq-c/blob/master/librabbitmq/amqp_openssl.c))
If the use of "SSL_library_init" is deprecated now, that should be changed within the rabbitmq-c library as well. I've opened an issue with them [here](https://github.com/alanxz/rabbitmq-c/issues/737)

> This sounds like we do not need such a flag, or am I wrong?

Since "SSL_library_init" is deprecated in OpenSSL 1.1.1, I would concur. If rsyslog side is converted to "OPENSSL_init_ssl", then it would be unnecessary. As you said before, it is not feasible to inspect and control each and every library. At least for omrabbitmq, this is covered by the "init_openssl" config option until they convert their side as well.

alorbach:
> > Which tells me it is not unexpected that the init function is called multiple times.
> 
> This sounds like we do not need such a flag, or am I wrong?

I don't think so, yes. But we should implement "OPENSSL_init_ssl" for newer OpenSSL versions

rgerhards:
@21stcaveman @alorbach Just to be sure: there are no concerns or future plans that prevent me from merging this PR as is? If so, I would merge. Please comment both.

alorbach:
> @21stcaveman @alorbach Just to be sure: there are no concerns or future plans that prevent me from merging this PR as is? If so, I would merge. Please comment both.

From my point of view, this can be merged.

rgerhards:
Thx for the responses!

rgerhards:
my bad... I overlooked doc is missing.

21stcaveman:
Done, please see PR [989](https://github.com/rsyslog/rsyslog-doc/pull/989)

I edited the only file I could find related to omrabbitmq. Please let me know if there is something I missed.

rgerhards:
all looks good, thx. Merging now :-) [next scheduled stable planned for October, 18th]

Issue #4953, opened 2022-08-08
Segmentation fault in close journal


t-feng:
### Expected behavior
No Segmentation fault

### Actual behavior
Core was generated by `/usr/sbin/rsyslogd -n -i/var/run/rsyslogd.pid'.
Program terminated with signal SIGSEGV, Segmentation fault.

### Steps to reproduce the behavior
NO, I found core file during daliy work

### Environment
- rsyslog version: rsyslog-8.2110.0
- platform: openEuler-x86_64
- for configuration questions/issues, include rsyslog.conf and included config files

```
# Where to place auxiliary files
global(workDirectory="/var/lib/rsyslog")

# Use default timestamp format
module(load="builtin:omfile" Template="RSYSLOG_TraditionalFileFormat")

#### MODULES ####

module(load="imuxsock"    # provides support for local system logging (e.g. via logger command)
       SysSock.Use="off") # Turn off message reception via local log socket;
                          # local messages are retrieved through imjournal now.
module(load="imjournal"             # provides access to the systemd journal
       StateFile="/run/log/imjournal.state") # File to store the position in the journal
#module(load="imklog") # reads kernel messages (the same are read from journald)
#module(load="immark") # provides --MARK-- message capability

$imjournalRatelimitInterval 0

# Include all config files in /etc/rsyslog.d/
include(file="/etc/rsyslog.d/*.conf" mode="optional")
#### RULES ####

# Log all kernel messages to the console.
# Logging much else clutters up the screen.
#kern.*                                                 /dev/console

# Log anything (except mail) of level info or higher.
# Don't log private authentication messages!
*.info;mail.none;authpriv.none;cron.none                /var/log/messages

# The authpriv file has restricted access.
authpriv.*                                              /var/log/secure

# Log all the mail messages in one place.
mail.*                                                  -/var/log/maillog


# Log cron stuff
cron.*                                                  /var/log/cron

# Everybody gets emergency messages
*.emerg                                                 :omusrmsg:*

# Save news errors of level crit and higher in a special file.
uucp,news.crit                                          /var/log/spooler

# Save boot messages also to boot.log
local7.*                                                /var/log/boot.log
```

t-feng:
I dig this issue, and got the following info:
```
(gdb) bt
#0  0x00005626ed0d8804 in ratelimitDestruct (ratelimit=0x0) at ratelimit.c:406
#1  0x00007f9b5b220c5b in afterRun () at imjournal.c:1004
[...]

PR #4954, opened 2022-08-08
Fix Segmentation fault in close journal

t-feng:
In imjournal module, we destruct ratelimiter after closing journal. But, we should check ratelimiter is NULL or not.

Fix: #4953


rgerhards (approved):
Thx for the PR!

PR #4957, opened 2022-08-10
conf: Fix #pragma GCC diagnostic not allowed inside functions

julthomas:
Commit 829a6f6 "CI: improve use of github action matrix feature;
optimization" has introduced use of #pragma GCC diagnostic in function
doNameLine() of runtime/conf.c. This is not supported by old compilers
like GCC 4.4.7 on el6.


rgerhards (changes requested):
sry for the late reply and thx for the PR.

Unfortunately, you hit the wrong spot. The right location to fix is here: https://github.com/rsyslog/rsyslog/blob/master/runtime/rsyslog.h#L94

julthomas:
Hello @rgerhards, I rebased, is it OK?

rgerhards (approved):

rgerhards:
> Hello @rgerhards, I rebased, is it OK?

looks perfect, thx!

Issue #4961, opened 2022-08-14
rsyslog leak memory when SetString


seuzw930:
### Expected behavior
no mem leak in when SetString
### Actual behavior
the asan reports there is a memory leak when SetString
### Steps to reproduce the behavior
There is an asan report when rsyslog SetString:
=================================================================
Indirect leak of 29 byte(s) in 1 object(s) allocated from:
    #0 0x7f328bb081b7 in __interceptor_malloc (/usr/lib64/libasan.so.6+0xaf1b7)
    #1 0x560cb38dbcbf in SetString /usr/src/debug/rsyslog-8.2208.0-1.x86_64/runtime/prop.c:87
    #2 0x560cb38dbcbf in SetString /usr/src/debug/rsyslog-8.2208.0-1.x86_64/runtime/prop.c:77
    #3 0x560cb38dbeb9 in CreateStringProp /usr/src/debug/rsyslog-8.2208.0-1.x86_64/runtime/prop.c:156
    #4 0x560cb38456f1 in resolveAddr /usr/src/debug/rsyslog-8.2208.0-1.x86_64/runtime/dnscache.c:334
    #5 0x560cb3846409 in addEntry /usr/src/debug/rsyslog-8.2208.0-1.x86_64/runtime/dnscache.c:361
    #6 0x560cb3846409 in findEntry /usr/src/debug/rsyslog-8.2208.0-1.x86_64/runtime/dnscache.c:415
    #7 0x560cb3846409 in dnscacheLookup /usr/src/debug/rsyslog-8.2208.0-1.x86_64/runtime/dnscache.c:452
    #8 0x7f3287a622df  (<unknown module>)
    #9 0x7f3287c0606c  (<unknown module>)
    #10 0x7f3287a8c819  (<unknown module>)
    #11 0x7f3287a8f375  (<unknown module>)
    #12 0x7f3287a918e7  (<unknown module>)
    #13 0x7f3287a9209e  (<unknown module>)
    #14 0x7f3287aa48b0  (<unknown module>)
    #15 0x560cb38f8184 in thrdStarter ../threads.c:244
    #16 0x7f328b7064f9  (/usr/lib64/libc.so.6+0x8b4f9)

SUMMARY: AddressSanitizer: 8064 byte(s) leaked in 1 allocation(s).

### Environment
- rsyslog version: 8.2208.0
- platform: centos
- for configuration questions/issues, include rsyslog.conf and included config files

rgerhards:
[...]

PR #4962, opened 2022-08-14
Fix memory leak when SetString reassign to pThis->szVal.psz

seuzw930:
During SetString reassign to pThis->szVal.psz, pThis->szVal.psz might not null. It resulted in memory leak and this patch fixes this behaviour.

https://github.com/rsyslog/rsyslog/issues/4961


rgerhards:
It looks like the patch introduces some issues. But I haven't looked deeper yet.

Did you experience a real memory leak?

Issue #4958, opened 2022-08-10
make check failure: /imfile-statefile-delete.sh FAIL: state file still exists when it should have been deleted


mbiebl:
Trying to build 8.2208.0 on Debian sid, I ran into the following `make check` failure:

```
FAIL: imfile-statefile-delete
=============================

testbench: TZ env var not set, setting it to UTC
------------------------------------------------------------
14:50:57[0]  Test: ./imfile-statefile-delete.sh
------------------------------------------------------------
config rstb_581193_af91b9aesqP3_.conf is:
     1	module(load="../plugins/imdiag/.libs/imdiag")
     2	global(inputs.timeout.shutdown="60000"
     3	       default.action.queue.timeoutshutdown="20000"
     4	       default.action.queue.timeoutEnqueue="20000")
     5	# use legacy-style for the following settings so that we can override if needed
     6	$MainmsgQueueTimeoutEnqueue 20000
     7	$MainmsgQueueTimeoutShutdown 10000
     8	$IMDiagListenPortFileName rstb_581193_af91b9aesqP3.imdiag.port
     9	$IMDiagServerRun 0
    10	$IMDiagAbortTimeout 580
    11	
    12	:syslogtag, contains, "rsyslogd"  ./rstb_581193_af91b9aesqP3.started
    13	###### end of testbench instrumentation part, test conf follows:
    14	
    15	global(workDirectory="rstb_581193_af91b9aesqP3.spool")
    16	module(load="../plugins/imfile/.libs/imfile" mode="inotify" PollingInterval="1")
    17	input(type="imfile" tag="file:" file="./rstb_581193_af91b9aesqP3.input")
    18	
    19	template(name="outfmt" type="string" string="%msg:F,58:2%\n")
    20	if $msg contains "msgnum:" then
    21		action(type="omfile" file=`echo $RSYSLOG_OUT_LOG` template="outfmt")
rsyslogd: NOTE: RSYSLOG_DEBUG_TIMEOUTS_TO_STDERR activated
main Q:Reg: worker start requested, num workers currently 0
main Q:Reg: wrkr start initiated with state 0, num workers now 1
rsyslog debug: main Q:Reg: worker 0x558275611850 started
rsyslog debug: main Q:Reg: started with state 3, num workers now 1
14:50:57[0]  rstb_581193_af91b9aesqP3:.pid found, pid 224676
14:50:57[0]  rsyslogd startup msg seen, pid  224676
[...]

PR #4993, opened 2022-10-05
testbench: fix imfile-statefile-delete.sh (INOTIFY trigger)

alorbach:
Due the patch in PR https://github.com/rsyslog/rsyslog/pull/4895 state files are deleted with a 5 second delay in order to fix missing or duplicated messages. However in INOTIFY mode, we need an INOTIFY event to trigger a poll_tree that triggers the delayed deletion. The testcase imfile-statefile-delete.sh will now create empty dummy files after 6 seconds delay in order to trigger INOTIFY events.

This fixes & closes https://github.com/rsyslog/rsyslog/issues/4958


rgerhards:
Just to clarify: in normal operations, we do not necessarily need inotify events. Polling will also occur at certain timeouts. However, in testing we do not run sufficiently long to have timeouts occur. So it is the proper cure to "help" inotify a bit in this test.

PR #4994, opened 2022-10-07
atomic_operations.m4: also test these ops on time_t

he32:
On NetBSD, time_t has for a long time now been __int64_t. On 32-bit CPUs, the compiler is not obliged to define __sync_bool_compare_and_swap_8, so instead this ends up as an undefined symbol when linking rsyslog.  This fix makes rsyslog fall back to the pthread / locking method on these systems, but at least lets the program build.


rgerhards:
thx for the patch, much appreciated!

Issue #4986, opened 2022-09-14
[RFE] Dropping capabilities on rsyslogd


Cropi:
### Problem description
Executing
```
pscap | grep rsyslogd
```
Shows that rsyslogd is running with full set of capabilities [1]:
```
1     19209 root        rsyslogd            full +
```

**Would it be possible to drop the capabilities to only the necessary set**, to minimize security exposure in case there was ever a mistake in a networking plugin or some other input resource?
Ideally, all apps running as uid 0 should drop privileges. Some can't for good reasons but many can.
In most of the scenarios, I could use the $PrivDropTo{User, Group} config directives to drop privileges by impersonating as another user and/or group after startup. However, occasionally I need to access some files that are not readable by the specified user/group e.g. when using the imfile module, so here root privileges are required but I would prefer to not give rsyslogd full set of capabilities.

### Proposed feature
Introduce ```libcap-ng``` [2] in rsyslog, which provides a clean mechanism to modify/drop capabilities. Thus, we could specify a set of caps that are needed for rsyslog to properly function, but nothing more. Enabled modules would also be included. The feature could be enabled via the "--enable-libcap-ng" configure option and would be turned off by default. Some plugins might require additional capabilities.

Let me know what are your thoughts on this topic. Thank you.

[1] https://man7.org/linux/man-pages/man7/capabilities.7.html
[2] https://github.com/stevegrubb/libcap-ng

rgerhards:
@Cropi This sounds like a cool idea. Would you be willing to craft a PR?

Cropi:
> @Cropi This sounds like a cool idea. Would you be willing to craft a PR?

Sure, I will do that. I just did not want to get into it without accepting the idea first.

rgerhards:
I think it is a very useful addition! :-)

rgerhards:
@Cropi Not sure.., do you think we need to add some specific info to the doc? If so, could you have a look?

Cropi:
> @Cropi Not sure.., do you think we need to add some specific info to the doc? If so, could you have a look?

I will do my best to add it to the documentation, or at least mention it somewhere.

rgerhards:
> I will do my best to add it to the documentation, or at least mention it somewhere.

Perfect. I think I will still merge the PR this week, so that it gets a chance to get used. In that sense, we may see some reports coming up. Just FYI.

theodoresi:
[...]

PR #4999, opened 2022-10-13
Introduce --enable-libcap-ng configure option

Cropi:
For more details, see https://github.com/rsyslog/rsyslog/issues/4986

What I am not sure about is where to place the ```libcap-ng``` related functions. So if you have a better idea on that, do not hesitate to share it with me.


Cropi:
The PR is ready for review. From code point of view, the proposed feature consists of two main parts.
### 1.Specify which caps are needed for rsyslog to function properly
After some digging in the code, I came up with the following list:
- CAP_BLOCK_SUSPEND: use of ```epoll()```
- CAP_CHOWN: use of ```chown()```
- CAP_IPC_LOCK: use of ```mmap()```
- CAP_LEASE: Establish leases on arbitrary files using ```fcntl()```
- CAP_NET_ADMIN: use of ```setsockopt()```
- CAP_NET_BIND_SERVICE: ports below 1024 are used
- CAP_PERFMON: use of ```syscall()```
- CAP_SETGID: use of setgid()
- CAP_SETUID: use of setuid()
- CAP_SYS_ADMIN: exceed /proc/sys/fs/file-max, the system-wide limit on the number of open files, in system calls that open files (e.g. accept execve), use of setns(),...
- CAP_SYS_CHROOT: use of ```chroot()```
- CAP_SYS_RESOURCE: use of ```setrlimit()```
- CAP_SYSLOG: use of ```syslog()```

### 2. Change UID and GID while retaining previously specified capabilities
Some might want to drop privileges by impersonating as another user and/or group after startup. In order to preserve defined caps, we need to let the system know. That is the reason why ```doDropPrivGid()``` and ```doDropPrivUid()``` functions had to be altered a little bit. Default behavior is preserved, the change occurs when ```ENABLE_LIBCAPNG``` is defined.

rgerhards:
Thanks for the PR. Please note that due to magnitude of the change, we will only merge it after the 8.2212.0 scheduled stable has been released. We do not want to generate some bad surprises during the holiday period. Hope that is fine with you ;-)

I'll nevertheless do code review in the mean time.

Cropi:
> Thanks for the PR. Please note that due to magnitude of the change, we will only merge it after the 8.2212.0 scheduled stable has been released. We do not want to generate some bad surprises during the holiday period. Hope that is fine with you ;-)
> 
> I'll nevertheless do code review in the mean time.

Thank you for the heads up, I am totally fine with that :) No need to rush

Cropi:
Update to the capability list: I do think it makes sense to add the CAP_DAC_OVERRIDE capability if privileges are not dropped. 
```
CAP_DAC_OVERRIDE: Bypass file read, write, and execute permission checks.  (DAC is an abbreviation of "discretionary access control".)
```

Use case: without this capability rsyslog would not be able to read from/write to logs if file permissions are not right, even if it's running as root.

mbiebl:
This broke `make check` when run as regular user.
Will file a separate bug report

theodoresi:
Hi @Cropi , I found that I can't run rsyslog as a non-root user like I used to. I asked the question on stackexchange, and it seems it is related to this code?
https://unix.stackexchange.com/questions/747224/unable-to-run-rsyslogd-as-non-root-user-on-centos-stream-9

PR #5001, opened 2022-10-13
testbench: Adjusted timing for test imfile-logrotate-async.sh

alorbach:
In some cases, the logrotate is not fast enough rotating the
logfile after sighub has been send to inputfilegen process. It appears
to happen more often on github CI machines, when it happens,
inputfilegen reopens the NOT yet rotated file and keeps writing
into it. However most messages written until the next sighub are lost
after the logfile finally is rotated.

In order to avoid this, I have added an extra sleep between file
CLOSE and OPEN after sighub with a default of 5ms which appears to work
very well. This should give logrotate enough time to perform the
actuzal rotate.

Also added another parameter to inputfilegen to keep the process
open after finish. Which helps syncing with the testcases.

### DIFF
--- a/tests/imfile-logrotate-async.sh
+++ b/tests/imfile-logrotate-async.sh
@@ -6,6 +6,10 @@ check_command_available logrotate
 export NUMMESSAGES=10000
 export RETRIES=50
 
+# Uncomment fdor debuglogs
+#export RSYSLOG_DEBUG="debug nostdout noprintmutexaction"
+#export RSYSLOG_DEBUGLOG="$RSYSLOG_DYNNAME.debuglog"
+
 # Write logrotate config file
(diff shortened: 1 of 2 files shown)


rgerhards:
Unfortunately, it looks like the Ubuntu20 test failed with the new test :-(

alorbach:
> Unfortunately, it looks like the Ubuntu20 test failed with the new test :-(

Perhaps there is another problem with the timing, going to look into this.

alorbach:
> Unfortunately, it looks like the Ubuntu20 test failed with the new test :-(

My initial assumption was wrong. There was a timing issue between inputfilegen and logrotate, it appears to be fixed now (see commit).

PR #5012, opened 2022-11-02
imfile: Fix external dir fd leaking

sakateka:
At this point there is a code [imfile.c#L919](https://github.com/rsyslog/rsyslog/blob/64ff7ee7f9d0accd9cdd88c4bbb676eccd3f25c0/plugins/imfile/imfile.c#L919) that adds an inotify observer to the parent of the symbolic link target
But there is no such code that removes this observer in the case when inotify events do not occur in the directory tree above.
This may be if the directory tree of the symbolic link target and the directory tree of the symbolic link itself are divided into different subtrees somewhere at the levels above

For example, in the rsyslog configuration, an `imfile` with the template `/var/log/containers/*.log` is configured and there is the following directory tree
```
/var/log/pods/pod-1/a/0.log
/var/log/containers/pod-1-a-0.log -> /var/log/pods/pod-1/a/0.log
```

In this example, kubernetes cron jobs will permanently delete directories at the `/var/log/pods/pod-*` level
And thus, inotify observer on the parent object of the symbolic link target (/var/log/pods/pod-1/a/0.log) looking at the directory ` /var/log/pods/pod-1/a` will constantly leak.

This is due to the fact that the list of active objects in the `edge` with path `/var/log/containers`, where the parent object of the target symbolic link is added, is not checked.
Verification and deletion will occur only in the case of an inotify event in the upper nodes of the directory tree, in `/var/log` and above

Here is our example of fd leaking from production
```
 INOTIFY   INSTANCES
   WATCHES      PER
    COUNT     PROCESS   PID USER         COMMAND
------------------------------------------------------------
    7307         1      683926 root        /usr/sbin/rsyslogd -n -f/configs/rsyslog.conf

...
rsyslogd 683926       root *322r   DIR    9,1        0     0 62128300 /var/log/pods/default_svn-test-27766379-xfg2c_91594765-2a99-48d3-b786-7387fc02a7d4/svn-test (deleted)
rsyslogd 683926       root *323r   DIR    9,1        0     0 62128242 /var/log/pods/default_svn-test-27766328-clhnq_f4d85955-116d-4a42-8067-d9f874aaf665/svn-test (deleted)
rsyslogd 683926       root *324r   DIR    9,1        0     0 62128216 /var/log/pods/default_svn-test-27766380-h2j6h_b1ce0de5-c4cc-45d4-a3ee-adf7fd57a78b/svn-test (deleted)
rsyslogd 683926       root *325r   DIR    9,1        0     0 62128252 /var/log/pods/default_svn-test-27766412-wlnpp_5b301d5c-a755-4fbd-ada1-1219c1497cc7/svn-test (deleted)
...
```
The first commit in this PR adds a test for this case, the second adds a fix.


sakateka:
Related pr #2914

rgerhards:
I don't understand why CodeFactor complains that much - looks like existing code... I'll ignore that for now (and if not, do a separate PR)

sakateka:
I corrected a few comments that `shellcheck` could not parse, perhaps after that he started doing more checks, it was written about in the messages with errors.
```
SC1073: Couldn't parse this shellcheck directive. Fix to allow more checks. shellcheck (1073) [1753, 3]
SC1072: Expected '=' after directive key. Fix any mentioned problems and try again. shellcheck (1072) [1753, 33]
```

Example of a comment before correction:
```
# shellcheck disable=SC2009  - we do not grep on the process name!
```

sakateka:
I don't understand why these tests failed
```
ubuntu_22: imfile-symlink.sh
fedora_36: imfile-wildcards-dirs-multi5.sh
```
I ran the `devtools/devcontainer.sh` 100 times for the `ubuntu_22` check with the following environment.
```
export RSYSLOG_CONTAINER_UID=""
export CFLAGS='-g'
export CC='gcc'
export USE_AUTO_DEBUG='off'
export CI_MAKE_OPT='-j20'
export CI_MAKE_CHECK_OPT='-j12'
export RSYSLOG_DEV_CONTAINER='rsyslog/rsyslog_dev_base_ubuntu:22.04'
export CI_VALGRIND_SUPPRESSIONS="ubuntu22.04.supp"
```

Of these launches, that's how many tests have fallen
```
      1 FAIL: dynstats_overflow-vg.sh
      1 FAIL: sndrcv_tls_anon_ipv6.sh
      1 FAIL: sndrcv_tls_certless_clientonly.sh
      2 FAIL: imfile-wildcards-dirs-multi5.sh
      3 FAIL: imfile-endregex-timeout-none-polling.sh
     21 FAIL: omfwd_fast_imuxsock.sh
```

But I couldn't reproduce the failure for the test imfile-symlink.sh

sakateka:
@rgerhards
If I need to correct something on my part to merge this PR, please write.
About tests: it seems to me that if you restart them, they will be successful
About CodeFactor: I can try to make a new PR with corrections in diag.sh and then rebase the current one.

sakateka:
I rolled back the correction of the spellcheck directives, now the 
CodeFactor is happy, but in fact it just doesn't work.

sakateka:
Hi, @rgerhards!
Is there any chance to merge this into the master in the foreseeable future? Or there is some kind of blocker for merging, I will accept any suggestions or comments adequately and try to finalize the patch.

Thank you!

rgerhards:
sry I lost trace of this PR due to overload :-( I have now assigend it and will work on finishing.

rgerhards:
sry again for the long wait. All looks very well. The hanging test are due to github action env changes, which is fixed in master. So I am merging now!

sakateka:
It's cool! Thank you!

PR #5023, opened 2022-11-16
ksi bugfix: sending of too many signing requests fixed.

taavi-valjaots:
As there is a bug in libksi where too many signing requests may have bene sent out the amount of signing requests will be limited by KSI module until the fix is implemented.


rgerhards:
Note: I have assigned "bug" albeit it is actually a work-around for a bug. But from the end user's perspective, it looks like a rsyslog issue (in closely related components), so I think "bug" qualifies here.

PR #5027, opened 2022-11-17
Fix ffaup thread safety

ThibaudCartegnie:
ffaup fix : The ffaup function fails to work properly when it is used with multiple workers.
The faup_handler_t struct is not supposed to be shared between threads.

This may cause memory corruptions and race conditions when used inside of actions.


frikilax:
I confirm there is a legitimate bug in the handling of the faup handler and that this development  resolves it

rgerhards:
@ThibaudCartegnie can you please amend the exact description (like here in the issue) to the commit itself. Only so way we ensure the information is tied forever to the project, no matter what fate github will have over time. Thx!

ThibaudCartegnie:
@rgerhards We found an easier fix that does not involve the __thread extension, I will push that solution instead of the one presented with a fixed commit message as you asked.

rgerhards:
Merci! Merged.

PR #5033, opened 2022-11-24
gtls: add IP address SAN support

AkshinK:
Currently, only DNS can be used as part
of subject alternative name to verify
the peers. Adding a check for IP addresses.


rgerhards:
This has been implemented by other PRs before I came back to here. So I only can close this PR. @AkshinK sorry for this and the delay.

PR #5063, opened 2023-01-11
omfile: add action parameters "rotation.*"

rgerhards:
Add new action parameters
- rotation.sizeLimit
- rotation.sizeLimitCommand provide automatic output file rotation functionality feature-wise equivalent to legacy $outchannel. This finally permits to use this feature set in rscript.


rgerhards:
doc: https://github.com/rsyslog/rsyslog-doc/pull/996

PR #5109, opened 2023-03-30
imjournal: Add FileCreateMode module parameter

Cropi:
### Description
The parameter allows to set the default file mode bits when creating new files by the module. As of now, it has only impact on the state file. Other modules that create files such as omfile, imptcp, omprog provide a similar parameter.

If accepted, I can prepare a PR in rsyslog-doc to include it in the documentation part as well.
What do you think?

[UPDATE]
I've set the default permission of the state file to 0644 to provide the same results as we currently have but personally I would limit it to 0600. I appreciate the feedback :)

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -35,6 +35,7 @@
 #include <sys/socket.h>
 #include <errno.h>
 #include <systemd/sd-journal.h>
+#include <fcntl.h>
 
 #include "dirty.h"
 #include "cfsysline.h"
@@ -73,6 +74,7 @@ struct modConfData_s {
 
 static struct configSettings_s {
 	char *stateFile;
+	int fCreateMode; /* default mode to use when creating new files, e.g. stateFile */
 	int iPersistStateInterval;
 	unsigned int ratelimitInterval;
 	unsigned int ratelimitBurst;
@@ -92,6 +94,7 @@ static rsRetVal facilityHdlr(uchar **pp, void *pVal);
 /* module-global parameters */
 static struct cnfparamdescr modpdescr[] = {
 	{ "statefile", eCmdHdlrGetWord, 0 },
+	{ "filecreatemode", eCmdHdlrFileCreateMode, 0 },
 	{ "ratelimit.interval", eCmdHdlrInt, 0 },
 	{ "ratelimit.burst", eCmdHdlrInt, 0 },
 	{ "persiststateinterval", eCmdHdlrInt, 0 },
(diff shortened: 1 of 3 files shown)


rgerhards:
Sorry, I missed the PR :-(

This sounds good to me. It's a though decision on the default. Personally, I would say this could be worth a breaking change, but we will probably cause a lot of troubles potentially. @davidelang what do you think?

rgerhards:
I fear too few people read the doc. But that trigerred an idea. How about this:

* leave the old default
* if the default is used (not explicitly set), emit a warning message highlighting the issue

That way we would break nothing, but educate people to make the system more secure. It's a bit nagging though, for new users. How does that sound? Don't be shy to say its a bad idea ;-)

davidelang:
On Mon, 17 Apr 2023, Rainer Gerhards wrote:

> I fear too few people read the doc. But that trigerred an idea. How about this:
>
> * leave the old default
> * if the default is used (not explicitly set), emit a warning message highlighting the issue
>
> That way we would break nothing, but educate people to make the system more secure. It's a bit nagging though, for new users. How does that sound? Don't be shy to say its a bad idea ;-)

I like this, if you make the nag message include instructions on how to silence 
it, it shouldn't be too annoying

and it should be a lot severity (debug or info) level message, not warning or 
something like that.

David Lang

Cropi:
Sorry for the late reply and thanks for the suggestions. I adjusted the PR according to that :)

Cropi:
I also created a PR for the documentation part, https://github.com/rsyslog/rsyslog-doc/pull/1005

Cropi:
Some of the required checks are failing but I can not even find a single log why it happens.

rgerhards (approved):

rgerhards (approved):

Cropi:
> @Cropi if possible, could you rebase to master. That will get us a clean CI. In any case, I'll merge soon. Thx again!

Done. Thanks.

plugins/imjournal/imjournal.c
```diff
@@ -1074,6 +1083,14 @@ CODESTARTsetModCnf
 		}
 	}
 
+	/* File create mode is not set */
+	if (cs.fCreateMode == -1) {
+		const int fCreateMode = 0644;
+		LogMsg(0, RS_RET_OK_WARN, LOG_WARNING, "imjournal: filecreatemode is not set, "
+			"using default %04o", fCreateMode);
+		cs.fCreateMode = fCreateMode;
```
saper, line 1091: If my analysis in https://github.com/rsyslog/rsyslog/issues/5375#issuecomment-2798063780 is correct, `cs.fCreateMode` will remain at -1 if `$ModLoad imjournal` is used, because this configuration function will not be called at all. 

Just changing the config line to `module(load="imjournal")` causes rsyslogd to generate a warning as it should.

rgerhards:
@Cropi Pls see https://github.com/rsyslog/rsyslog/issues/5375#issuecomment-3116719826

I think my concerns about the breaking change was right. I suggest that you change RH packaging to work with the patch. In general, I think it would make sense to convert a bit more of the config to current modern style.

Cropi:
Thanks for pointing out, I will prioritize this so we can fix it. And sorry for the confusion.

rgerhards:
@Cropi thx for the swift answer. It may also make sense to have some decent default for the premissions if $modLoad is used. I think this primarily affects upgrading RH users, which still have $ModLoad in use from previous version.

PR #5114, opened 2023-04-12
Send HUP on "systemctl reload rsyslog" on Debian.

XANi:
Allow `systemctl reload rsyslog` to send HUP for log rotation. I'm not sure whether it's "right" semantic for reload operation that only reopens logs 

Tested on Debian9/10/11/12, I *assume* centos one would be same change but have no machine to test it so I don't include it in patch

### DIFF
--- a/platform/debian/rsyslog.service.10
+++ b/platform/debian/rsyslog.service.10
@@ -7,6 +7,7 @@ Documentation=https://www.rsyslog.com/doc/
 [Service]
 Type=notify
 ExecStart=/usr/sbin/rsyslogd -n -iNONE
+ExecReload=/bin/kill -HUP $MAINPID
 StandardOutput=null
 Restart=on-failure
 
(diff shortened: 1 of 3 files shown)


rgerhards:
Well, it's not the right semantic, because `reload` usually is associated with a configuration reload - which does not happen on HUP. As such, I will not merge the PR.

Besides, you can send HUP via systemctl, with "kill" and "--signal=HUP" (see man systemctl for details).

XANi:
Huh, I didn't knew that, thanks for info. That's a terrible name for subcommand that "just" sends signals...

rgerhards:
Well, "terrible" depends. I am not here to defend systemd, but under Unix CLI you send signals via the `kill` command. So it looks reasonable to use the same term in systemd. I agree, though, that the original choice looks terrible today. But in the early days, kill usually meant exactly that,

PR #5120, opened 2023-04-21
Added debug info before action state changes

lonicerae:
When debugging an issue of `action.resumeInterval` and  `action.resumeIntervalMax`, it seems showing the caller of `actionSetState()` is useful. So share it here.

### DIFF
--- a/action.c
+++ b/action.c
@@ -691,6 +691,7 @@ actionSetState(action_t * const pThis, wti_t * const pWti, uint8_t newState)
  */
 static void actionCommitted(action_t * const pThis, wti_t * const pWti)
 {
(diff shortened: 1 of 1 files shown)


rgerhards:
sry for the late attention

rgerhards:
@lonicerae sry again for the late reply. After weighing arguments, I think I will not merge as-is. But you have a valid point with the caller names. I will craft a PR which does this with common code and credit you for providing the idea. I hope this is fine with you.

rgerhards:
replaced by https://github.com/rsyslog/rsyslog/pull/6449 - once again many thanks to @lonicerae for this prototype PR.

PR #5121, opened 2023-04-23
Fixed a race issue between action.resumeInterval and action.resumeInt…

lonicerae:
Fixed a race issue between action.resumeInterval and action.resumeIntervalMax

  * The sleep time of the action.resumeInterval can make a retry scheduled by
    the action.resumeInterval unable to be executed under some circumstance.


rgerhards:
merged via https://github.com/rsyslog/rsyslog/pull/5260

PR #5140, opened 2023-05-17
Fix quoting for omprog, improg, mmexternal

### LINKED ISSUE #4249: omprog: quotes are passed to argument
### Expected behavior
Quotes are removed from arguments. `argv` contains strings without quotes.
### Actual behavior
Quotes are passed to arguments. This makes not possible configs like `binary="/bin/sh -c \"echo done\""` because `sh` gets `"echo done"` in `argv[1]` not `echo done`.

- add to `rsyslog.conf`: 
```
*.err action(type="omprog" binary="/usr/local/bin/test.sh -c \"echo done\"" output="/tmp/o")
```
- `#cat /usr/local/bin/test.sh`
```
#!/bin/sh

echo "$@"
```
- `logger -p mail.err fff`
- check `/tmp/o`: `-c "echo done"`
- run from shell `/usr/local/bin/test.sh -c "echo done"`, check output: `-c echo done`

### Environment
- rsyslog version: 8.1901.0
- platform: Debian 10

paulfertser:
This changes the current behaviour from honouring the double quotes in any part of the argument but leaving them in place and passing to the executed binary to requiring quotes exactly at the beginning and at the end of a multi-word argument, and not including them in the actual call.

Testcases added to ensure the expected results.

Fixes #4249.

I will send the corresponding change to rsyslog-doc repository later if you give the green light to this.

### DIFF
--- a/runtime/srutils.c
+++ b/runtime/srutils.c
@@ -836,12 +836,25 @@ split_binary_parameters(uchar **const szBinary, char ***const __restrict__ aPara
 		iCnt = iStr = 0;
 		c = es_getBufAddr(estrParams); /* Reset to beginning */
 		while(iCnt < es_strlen(estrParams) ) {
-			if ( c[iCnt] == ' ' && !bInQuotes ) {
-				estrTmp = es_newStrFromSubStr( estrParams, iStr, iCnt-iStr);
-			} else if ( iCnt+1 >= es_strlen(estrParams) ) {
-				estrTmp = es_newStrFromSubStr( estrParams, iStr, iCnt-iStr+1);
-			} else if (c[iCnt] == '"') {
-				bInQuotes = !bInQuotes;
+			if (c[iCnt] == '"' && iCnt == iStr && !bInQuotes) {
+				bInQuotes = TRUE;
+				iStr++;
+			} else {
+				int bEOL = iCnt+1 == es_strlen(estrParams);
+				int bSpace = c[iCnt] == ' ';
+				int bQuoteEnd = bInQuotes && ((bSpace && c[iCnt-1] == '"') ||
+							      (c[iCnt] == '"' && bEOL));
+				if (bEOL || bQuoteEnd || (bSpace && !bInQuotes)) {
+					int iSubCnt = iCnt - iStr;
+					if (bEOL)
+					    iSubCnt++;
+					if (bQuoteEnd)
+					    iSubCnt--;
+					estrTmp = es_newStrFromSubStr(estrParams, iStr, iSubCnt);
+				}
+
+				if (bQuoteEnd)
+				    bInQuotes = FALSE;
 			}
 
 			if ( estrTmp != NULL ) {
--- a/tests/omprog-defaults.sh
+++ b/tests/omprog-defaults.sh
@@ -26,7 +26,8 @@ template(name="outfmt" type="string" string="%msg%\n")
 :msg, contains, "msgnum:" {
     action(
         type="omprog"
-	    binary=`echo $srcdir/testsuites/omprog-defaults-bin.sh p1 p2 p3`
+	    binary="'$srcdir'/testsuites/omprog-defaults-bin.sh \"p1 with spaces\"'\
+' p2 \"\" --p4=\"middle quote\" \"--p6=\"proper middle quote\"\" \"p7 is last\""
         template="outfmt"
         name="omprog_action"
     )
@@ -37,7 +38,14 @@ injectmsg 0 10
 shutdown_when_empty
 wait_shutdown
 
-export EXPECTED="Starting with parameters: p1 p2 p3
+export EXPECTED="Starting with parameters: p1 with spaces p2  --p4=\"middle quote\" --p6=\"proper middle quote\" p7 is last
+Next parameter is \"p1 with spaces\"
+Next parameter is \"p2\"
+Next parameter is \"\"
+Next parameter is \"--p4=\"middle\"
+Next parameter is \"quote\"\"
+Next parameter is \"--p6=\"proper middle quote\"\"
+Next parameter is \"p7 is last\"
 Received msgnum:00000000:
 Received msgnum:00000001:
 Received msgnum:00000002:
--- a/tests/omprog-if-error.sh
+++ b/tests/omprog-if-error.sh
@@ -25,6 +25,9 @@ cat $RSYSLOG_DYNNAME.othermsg
 content_check 'must be terminated with \n' $RSYSLOG_DYNNAME.othermsg
 
 export EXPECTED="Starting with parameters: p1 p2 p3
+Next parameter is \"p1\"
+Next parameter is \"p2\"
+Next parameter is \"p3\"
 Received msgnum:00000000:
 Received msgnum:00000001:
 Received msgnum:00000002:
--- a/tests/testsuites/omprog-defaults-bin.sh
+++ b/tests/testsuites/omprog-defaults-bin.sh
@@ -3,6 +3,10 @@
 outfile=$RSYSLOG_OUT_LOG
 
 echo "Starting with parameters: $@" >> $outfile
+while [ $# -gt 0 ]; do
+    echo Next parameter is \""$1"\"
+    shift
+done >> $outfile
 
 read log_line
 while [[ -n "$log_line" ]]; do


paulfertser:
@rgerhards Hey :) I really tried to understand the CI failures but they seem to be non-deterministic. Here we get one error, on this test github pull request with exactly the same code there was another error https://github.com/paulfertser/rsyslog/actions/runs/5088686392/jobs/9145401391 and it's anyway not related to the part of the code I'm changing.
I know you're afk for the coming days, would appreciate your feedback when you return, have a nice weekend.

paulfertser:
Gentle reminder about this one. @rgerhards , do you want me to rebase on current HEAD or something else needed to proceed?

paulfertser:
Hello @rgerhards ,
Can you please provide some feedback on this one? I tried hard to make proper tests and provide rationale and explanations, is anything still missing?

rgerhards:
sry, I overlooked this last may. I'll have a look...

rgerhards:
> This changes the current behaviour from honouring the double quotes in any part of the argument but leaving them in place

As I general comment, the rsyslog project does not like breaking changes. This one looks especially problematic. It would be good if you could give some hard reasoning, why this is actually needed and why there are no other ways to reach that functionality.

Even then, we would need to guard it be a config setting, because otherwise we would potentially break a lot of configurations. Rsyslog users expect that no matter how new a version, their existing config continues to work (the good ol'e datacenter paradigm when we had no fear of new releases...).

Re CI, yes it unfortunately is flaky. Timing is critical and CI machines are sometimes overwhelemed (even with other processes on the same bare metal). We could synchronize all and everything, but then the tested and executed code would not have much in common :-(

rgerhards:
To me, it looks like the PR re-implement this example from the omprog doc page:

```
module(load="omprog")

action(type="omprog"
       binary="/path/to/log.sh p1 p2 --param3=\"value 3\""
       template="RSYSLOG_TraditionalFileFormat")
```

Note the space in param3.

paulfertser:
Hello,

Thank you for taking a look!

> > This changes the current behaviour from honouring the double quotes in any part of the argument but leaving them in place
> 
> As I general comment, the rsyslog project does not like breaking changes. This one looks especially problematic. It would be good if you could give some hard reasoning, why this is actually needed and why there are no other ways to reach that functionality.

That's understandable. Let me try to explain why I thought it's an "ok" idea to offer. First of all, it fixes real bug report affecting several people, mentioned in the commit message. It doesn't seem to be possible to fix it in any other way short of providing an alternative option. Second, the current way to treat quotes in multi-word arguments is breaking assumptions, people are used to how POSIX shells treat quotes and that the commands get the arguments with the quotes removed, so I consider the current behaviour to be plain buggy, it doesn't allow calling external programs properly without additional wrappers, shell evals etc. I wonder how many existing configurations can actually be counting on such an unusual behaviour.

As an example of why it is needed, https://github.com/rsyslog/rsyslog-doc/pull/1007 I think is a good one, as it shows how a very generic and simple (thus correct and unsurprising) script applicable to all other rsyslog omprog usecases can be used without additional helpers.

> Even then, we would need to guard it be a config setting, because otherwise we would potentially break a lot of configurations. Rsyslog users expect that no matter how new a version, their existing config continues to work (the good ol'e datacenter paradigm when we had no fear of new releases...).

If it's absolutely required, OK, I can add another option to omprog action, probably `newQuoting="on"` but that would add to the bloat, retaining buggy code and interfaces for backwards compatibility is so microsoft-ish... I'm open to other ideas of course, it would be nice if you could take another fresh look at the original bug report and my omprog-deque msmtp example, as I'm naturally already too biased towards the solution I was working on.
 
> Re CI, yes it unfortunately is flaky. Timing is critical and CI machines are sometimes overwhelemed (even with other processes on the same bare metal). We could synchronize all and everything, but then the tested and executed code would not have much in common :-(

I do not mind flaky CI, it's just not outright obvious whether you expect contributors to worry about the failures reported on "github pull requests".

rgerhards:
I think I see your point, but I need to probably do a number of tests to confirm there really is a problem (we need to be crystal-clear here, especially when we declare it a bug and try to fix with a breaking change). Unfortunately, I am busy this week and out of office next. If you don't hear back by mid-may, pls ping me. I have assigned a milestone, so the issue should at least not get out of "vision".

paulfertser:
It is possible to solve the problem by keeping the existing syntax and filtering out the quotes but that would 1. still not allow embedding a literal double quote in the argument when really desired; 2. require creating a new string and filling it char-by-char rather than just calling `es_newStrFromSubStr`. It might be a reasonable trade-off between backwards compatibility and bug fixing after all. It also would still have the potential to break existing configs if someone's really counting on the quotes to be there in the passed arguments. Yet another option is to keep the code as it is and provide additional shell wrapper which would cover all the usecases somehow (probably other than embedding double quote as they would all be stripped one way or another). Or just document the limitations and clarify that if the current option handling is not suitable one must embed all the options in external script.
So yes, plenty of things to consider here, I hope at least the problem is clear enough now to make an informed decision once you have time for it.

rgerhards:
I would really like to dig a bit deeper, as I begin to think you are right its a bug. But it's several years since this code was written and I want to find out what were the reasons behind that.

rgerhards:
OK, sry for my ignorance. I had a strange misunderstanding. There indeed is a bug. Somehow I thought (even after reading multiple times) it was meant the the quotes were removed and were intended to stay...

Maybe because this is the question so often when talking about "such" templates. Anyhow, now it's clear.

rgerhards:
As this really is a bug that affects every well-behaving problem, I think we do not keep backwards-compatibility. Especially as I think it is hard to envision that someone actually depends on the quotes. I would tend to say that always a "starter bash script" was used in cases where it mattered.

rgerhards (approved):

rgerhards:
The CI issue is a flake, so it is fine to merge - will happen soon.

rgerhards:
Thanks again for the PR - and your patience and persistence! Much appreciated.

paulfertser:
I'm happy to help! Hope you won't regret the change :)
The idea to require quotes for multi word arguments exactly at the beginning and the end is a bit controversial but that's the best I could come up with without adding escaping for quotes or spaces and without complicating the code too much. At least now the arguments handling is covered by the test cases.
I'm available to discuss and tweak the corresponding change to rsyslog-doc https://github.com/rsyslog/rsyslog-doc/pull/1007 if desired.

rgerhards:
> I'm happy to help! Hope you won't regret the change :)

> The idea to require quotes for multi word arguments exactly at the beginning and the end is a bit controversial but that's the best I could come up with without adding escaping for quotes or spaces and without complicating the code too much.

I agree to this approach. I thought about it, but I did not come up with a valid case, where at the end is not an unescaped quote. In CLI, that would mean bash expects more input. If it poses a problem (I really doubt it), we may later add an error message to those cases.

FYI my ChangeLog entry for this change:
```
- 2024-04-18: [o|i]mprog/mmexternal bugfix: invalid command line parameter passing
  Quoted command line parameters were incorrectly passed to the called program. This
  resulted in unexpected behaviour. For example, in
  bash -c "echo test"
  bash got passed parameters '-c" and '"echo test"'. Note the double quotes in
  '"echo test"'. These needed to be removed by Unix/Linux standards but were instead
  passed to bash. That in turn resulted in invalid execution. With the fix, the
  behaviour is now correct. Now, bash gets passed '-c' and 'echo test'.
  This actually is a BREAKING CHANGE. However, it was outright wrong behaviour from
  the beginning. We assume that people either never noticed it (because they did not
  use quoted parameters) or used some workaroud, likely a "starter script", which
  performed the right task. As such, we do expect that only a very limited set of
  installations might be affected by the change.
  Even more important, we would need to change the default behaviour in any case,
  because the previous handling was obviosuly unacceptable. As such, there was no
  way to keep rsyslog perform the previous action.
  Thanks to Paul Fertser for the patch.
```

> At least now the arguments handling is covered by the test cases.

Well, I would tend to say that there were tests, but they were based on the same wrong assumption made during development.

> I'm available to discuss and tweak the corresponding change to rsyslog-doc [rsyslog/rsyslog-doc#1007](https://github.com/rsyslog/rsyslog-doc/pull/1007) if desired.

Looks good to me, but if you could rebase, things would be smoother. We had some CI issues at the time you submitted it, these are probably now fixed (still a lot of moving parts at the moment because we optimize the doc).

paulfertser:
> FYI my ChangeLog entry for this change:

Looks convincing to me, thank you.

> > At least now the arguments handling is covered by the test cases.
> 
> Well, I would tend to say that there were tests, but they were based on the same wrong assumption made during development.

Yes, I see, parameters were tested but not how quoting and splitting worked, `$@` just printed them all: https://github.com/rsyslog/rsyslog/commit/2847fa3aa0974c6b8400279fa176d47df7968865#diff-a9d0794f19daf54078e936787ea45fd6849bc96e61579f588d220c745dbae696R6

> > I'm available to discuss and tweak the corresponding change to rsyslog-doc [rsyslog/rsyslog-doc#1007](https://github.com/rsyslog/rsyslog-doc/pull/1007) if desired.
> 
> Looks good to me, but if you could rebase, things would be smoother. We had some CI issues at the time you submitted it, these are probably now fixed (still a lot of moving parts at the moment because we optimize the doc).

Sure, rebased and pushed. However CI is still problematic (guess two seconds wait isn't always enough), it built the first version of the patch even though it was after me triggering a new build by a force-push. I also noticed the URL for the generated artifact has a redirect to a server which seems to drop (rather than reject) TCP connections to HTTPS port so my browser "hanged" after automatically trying to force https for that (using an add-on).

PR #5142, opened 2023-05-19
imjournal: Add option to freeze journal pointer

wltu:
Add `disableJournalForwarding` option in both imjournal and omfwd. If both imjournal and omfwd action is `on`, then freeze the journal pointer to avoid reading more data when we loose outgoing connection. Once the connection resumes then it should resume the pointer too.

Tested:

owfwrd:

- Disabled forwarding at boot up if no network connection
```
rsyslogd[23301]: cannot connect to 2002:ab1:c012:0:fd01:::32722: Connection
  refused [v8.2212.0 try https://www.rsyslog.com/e/2027 ]
rsyslogd[23301]: omfwd: Disable Journal Forwarding Logs... [v8.2212.0 try
  https://www.rsyslog.com/e/2027 ]
rsyslogd[23301]: action 'action-4-builtin:omfwd' suspended (module
  'builtin:omfwd'), retry 0. There should be messages before this one giving
  the reason for suspension. [v8.2212.0 try https://www.rsyslog.com/e/2007 ]
```

- Disabled forwarding after loosing connection
```
rsyslogd[22638]: omfwd: remote server at 2002:ab1:c012:0:fd01:::32722 seems to
  have closed connection. This often happens when the remote peer (or an interim
  system like a load balancer or firewall) shuts down or aborts a connection.
  Rsyslog will re-open the connection if configured to do so (we saw a generic
  IO Error, which usually goes along with that behaviour). [v8.2212.0 try
  https://www.rsyslog.com/e/2027 ]
rsyslogd[22638]: omfwd: Disable Journal Forwarding Logs...
  [v8.2212.0 try https://www.rsyslog.com/e/2007 ]
rsyslogd[22638]: action 'action-4-builtin:omfwd' suspended (module
  'builtin:omfwd'), retry 0. There should be messages before this one giving the
   reason for suspension. [v8.2212.0 try https://www.rsyslog.com/e/2007 ]
```

- Resume forwarding after reconnected.
```
rsyslogd[23301]: omfwd: Enable Journal Forwarding Logs... [v8.2212.0 try
  https://www.rsyslog.com/e/2359 ]
rsyslogd[23301]: action 'action-4-builtin:omfwd' resumed (module
  'builtin:omfwd') [v8.2212.0 try https://www.rsyslog.com/e/2359 ]
```

imjournal:

- Stop Journal pointer
```
SECONDS=0
expected="$(md5sum /var/log/state)"
echo "Expected state ${expected}"
while [ $SECONDS -lt 300 ];
do
  new_state="$(md5sum /var/log/state)"
  if [[ "${new_state}" != "${expected}" ]];
  then
    echo "Got state ${new_state} instead of ${expected}"
      exit 1
  fi
  sleep 5
done
echo "Passed"
exit 0
```

```
Expected state 309593de34bc095e4245c1ba07a7520c  /var/log/state
Passed
```

- Resume Journal Pointer

The state file changed after the connection resumed.
```
$ md5sum /var/log/state
d11c7e46fec4f86151df204608525d89  /var/log/state
```

The message send in when the connection was down is also sent out properly.

- Journal Pointer rotating out. Once the pointer rotates out it just goes to the top of the journal with no issue. I printed something in journal and waited for it to rotate out... and it was not send out after resuming pointer.

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -85,6 +85,7 @@ static struct configSettings_s {
 	int bWorkAroundJournalBug; /* deprecated, left for backwards compatibility only */
 	int bFsync;
 	int bRemote;
+	int disableJournalForwarding;
 } cs;
 
 static rsRetVal facilityHdlr(uchar **pp, void *pVal);
@@ -103,7 +104,8 @@ static struct cnfparamdescr modpdescr[] = {
 	{ "usepid", eCmdHdlrString, 0 },
 	{ "workaroundjournalbug", eCmdHdlrBinary, 0 },
 	{ "fsync", eCmdHdlrBinary, 0 },
-	{ "remote", eCmdHdlrBinary, 0 }
+	{ "remote", eCmdHdlrBinary, 0 },
+	{ "disablejournalforwarding", eCmdHdlrBinary, 0 }
 };
 static struct cnfparamblk modpblk =
 	{ CNFPARAMBLK_VERSION,
@@ -842,6 +844,14 @@ CODESTARTrunInput
 	while (glbl.GetGlobalInputTermState() == 0) {
 		int r;
 
+		if (cs.stateFile && cs.disableJournalForwarding &&
+		     glbl.GetGlobalJournalDisableForwarding()) {
+			// disableJournalForwarding enabled.
+			// Skip the journal pointer if GetGlobalJournalDisableForwarding is
+			// 1.
+			continue;
+		}
+
 		r = sd_journal_next(journalContext.j);
 		if (r < 0) {
 			LogError(-r, RS_RET_ERR, "imjournal: sd_journal_next() failed");
@@ -1035,7 +1045,9 @@ CODESTARTsetModCnf
 	for (i = 0 ; i < modpblk.nParams ; ++i) {
 		if (!pvals[i].bUsed)
 			continue;
-		if (!strcmp(modpblk.descr[i].name, "persiststateinterval")) {
+		if (!strcmp(modpblk.descr[i].name, "disablejournalforwarding")) {
+			cs.disableJournalForwarding = (int) pvals[i].val.d.n;
+		} else if (!strcmp(modpblk.descr[i].name, "persiststateinterval")) {
 			cs.iPersistStateInterval = (int) pvals[i].val.d.n;
 		} else if (!strcmp(modpblk.descr[i].name, "statefile")) {
 			cs.stateFile = (char *)es_str2cstr(pvals[i].val.d.estr, NULL);
--- a/runtime/glbl.c
+++ b/runtime/glbl.c
@@ -93,6 +93,7 @@ char** glblDbgFiles = NULL;
 size_t glblDbgFilesNum = 0;
 int glblDbgWhitelist = 1;
 int glblPermitCtlC = 0;
+int bJournalDisableForwarding = 0;		/* global switch that journal state should stop forwarding */
 
 pid_t glbl_ourpid;
 #ifndef HAVE_ATOMIC_BUILTINS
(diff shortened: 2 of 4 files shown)


davidelang:
how should this work if you have multiple outputs? what if some of them have 
queues? how do you add this in a way that doesn't change the behavior of 
existing configs? what happens if the pointer is frozen long enough that the 
logs it points to have been deleted?

I could see freezing the journal pointer is you are blocking the input (i.e. 
your main queue is completely full), what currently happens under those 
conditions?

David Lang

rgerhards:
@wltu I probably understand your point, but there probably is a much simpler solution: bind imjournal to a ruleset containing a single omfwd action, set it to retry eternally (-1) and assign a **direct** queue to the ruleset. In that case, when the output blocks, the input does so as well.

As the PR also has  some other issues in respect to rsyslog design and principles, I will not merge it.

wltu:
Thanks for raising up the valid concerns. Since this approach is not within the rsyslog design principles. I'll look into the suggestion to see if it will serve my use case.

Basically, I just want to forward the logs to a remote server and not take in read more journal if we know the connection is down.

wltu:
It seems like input module doesn't work on `imjournal` at the moment? Do we have existing implementation of that? I played around with it and got part of it working.

I got the ruleset working with input module + imjournal. There are a few issues.
- I am currently only allowing one input module binding since we don't want to read multiple sources for different module.
- Journal pointer does not stop on failing the output

Example config.
```
ruleset(name="remote"){
  action(type="omfwd" target="IP_ADDRESS" port="PORT" protocol="tcp"
    template="journal-format"
    StreamDriver="ossl"
    # Run driver in TLS-only mode
    StreamDriverMode="0"
    StreamDriverAuthMode="anon"
    action.resumeRetryCount="-1"
    queue.type="direct"
  )
}

input(
  type="imjournal"
  ruleset="remote"
)
```

Any tip on this issue?

davidelang:
full config please (did you remember to load the module first??)

David Lang

wltu:
I have the change in https://github.com/wltu/rsyslog/commit/ecd18cd4851797da18defd463aa8d40cfe0a4fa0

Yeah, I did load the module first.

Full config is the following

## /etc/rsyslog.conf
```
# Listen custom socket.
module(load="imuxsock" SysSock.Use="off")
module(load="imjournal" StateFile="/var/log/state"
                        RateLimit.Interval="0"
                        RateLimit.Burst="2000"
)

$EscapeControlCharactersOnReceive off
```

## /etc/rsyslog.d/client.conf
```
global(
  DefaultNetstreamDriver="ossl"
  DefaultNetstreamDriverCAFile="/usr/share/production_syslog/root_cert.pem"
)

template(name="journal-format" type="string"
  string="%TIMESTAMP:::date-rfc3339% %HOSTNAME% %syslogtag% %msg%\n"
)

ruleset(name="remote"){
  action(type="omfwd" target="IP_ADDRESS" port="PORT" protocol="tcp"
    template="journal-format"
    StreamDriver="ossl"
    StreamDriverMode="0"
    StreamDriverAuthMode="anon"
    queue.type="direct"
    action.resumeRetryCount="0"
  )
}

input(
  type="imjournal"
  ruleset="remote"

```

rgerhards:
you need to set action.resumeRetryCount to **-1**. Zero means it gives up immediately. Thus no wait.

Also, I think, the ruleset is missing direct mode (at least if the main queue is not in direct mode):

```
ruleset(name="remote" queue.type="direct") {
```

wltu:
oh yeah, sorry. `resumeRetryCount` was `-1` before. I was testing out different config to see what I did wrong.

Ah. Let me try to set the rulset queue to direct. Thanks.

wltu:
Tested and the suggestion seems to work. Thanks for the help @rgerhards . :)

I'll create another PR for the imjournal input module change and discussion more detail on that.

wltu:
For anyone here that is interested: https://github.com/rsyslog/rsyslog/pull/5146

PR #5143, opened 2023-05-23
New Redis streams functionality + various fixes and improvements

frikilax:
### Added
- [TESTS] Implemented tests for existing functionalities
  - changed diag.sh to be able to start/stop/clean a redis server
  - added helper functions in diag.sh to be able to query a redis server instance
  - added new tests for imhiredis module to check
    - that the queue mode works, with both lpop and rpop
    - that the module is capable of handling a redis server going down
    - that the module is capable of handling a redis server that appears afterwards
    - that the subscribe mode works
- [IMHIREDIS] Missing redis replies' types in enumeration
- [IMHIREDIS] Add support for simple XREADs from Redis Streams (Redis >= 5.0 required)
- [IMHIREDIS] Add support for XREADGROUP from Redis Streams, allowing for user to define workers to dequeue logs in a stream
- [IMHIREDIS] stream mode can select fields to extract and insert in custom keys
- [IMHIREDIS] Add tests for the new 'stream' modes
- [OMHIREDIS] module is now able to insert entries to a Redis Stream
- [OMHIREDIS] in 'stream' mode, module can insert the message to a custom field in the entry ('msg' by default)
- [OMHIREDIS] in 'stream' mode, module can acknowledge an entry coming from imhiredis (if entry was claimed but not ACK'ed)
- [OMHIREDIS] in 'stream' mode, acknowledgements can be made from dynamic templates or static values
- [OMHIREDIS] in 'stream' mode, module can approximately cap the size of the output stream
- [OMHIREDIS] in 'stream' mode, module can delete an entry while inserting its message (useful to remove entry coming from another stream with imhiredis)
- [OMHIREDIS] new tests for 'stream' mode

### Changed
- [IMHIREDIS] factorize code for different modes
- [IMHIREDIS] Clean and improve logging lines
- [IMHIREDIS] Poll extinction state less frequently for main thread (less aggresive)
- [IMHIREDIS] Set 'key' action parameter to REQUIRED
- [IMHIREDIS] Use known message length instead of calculating it when enqueuing message

### Fixed
- [IMHIREDIS] Correctly initialize instance object, especially for redisNodesList
- [IMHIREDIS] Correctly print input mode's value in logs when set incorrectly
- [OMHIREDIS] Correctly authenticate to server before sending queries
- [OMHIREDIS] Properly ensure messages are successful in Redis, or suspend module
- [CONFIGURE.AC] Missing line to give omhiredis compilation status

--------------------------------------

associated PR for documentation : [rsyslog-doc PR #1008](https://github.com/rsyslog/rsyslog-doc/pull/1008)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2461,6 +2461,17 @@ AC_ARG_ENABLE(omhiredis,
         [enable_omhiredis=no]
 )
 
+AC_ARG_ENABLE(redis_tests,
+        [AS_HELP_STRING([--enable-redis-tests],[Enable redis tests, needs redis-server @<:@default=no@:>@])],
+        [case "${enableval}" in
+         yes) enable_redis_tests="yes" ;;
+          no) enable_redis_tests="no" ;;
+           *) AC_MSG_ERROR(bad value ${enableval} for --enable-redis-tests) ;;
+         esac],
+        [enable_redis_tests=no]
+)
+AM_CONDITIONAL(ENABLE_REDIS_TESTS, test x$enable_redis_tests = xyes)
+
 if test "x$enable_omhiredis" = "xyes" -o "x$enable_imhiredis" = "xyes" ; then
     PKG_CHECK_MODULES(HIREDIS, hiredis >= 0.10.1, [],
         [AC_SEARCH_LIBS(redisConnectWithTimeout, hiredis,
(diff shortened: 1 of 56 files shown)


frikilax:
@rgerhards do you consider that PR to be valid? (especially regarding the number of commits)
Also, some jobs are failing, but I reckon that's not due to my development, can you confirm?

rgerhards:
I thought I had asked if you could rebase to master. Unfortunately, I either did not at all or forgot to hit "Comment" ;-)

Anyhow, a rebase would help with CI. We had some issues with the github action platforms. These are solved, but not applied to older PRs. 

Re the number of commit messages, I think this is OK because they seem to cover larger blocks of (kind of) independent work. So that would work for me.

rgerhards:
merging without rebase as the CI issues are known and already fixed. Remaining checks should be sufficient. If there is some compile glitch, we'll fix afterwards (I'll do a safety check).

frikilax:
@rgerhards I forgot to mention : there were some tests added for both IM and OMhiredis, you might want to add redis tests to the testbench (I added a new `--enable-redis-tests` flag to the `./configure` to explicitely add those tests, as they require to have redis executables)

Issue #5091, opened 2023-02-22
PR #4999 breaks make check


mbiebl:
For testing purposes, I enabled the new --enable-libcap-ng feature in 8.2302.0.
libcap-ng is at version 0.8.3-1, the test environment is a Debian sid system.

When running `make check` as regular/unprivileged user (which debian builds do), it just hangs and doesn't proceed:

```
make[2]: Leaving directory '/home/michael/debian/build-area/rsyslog-8.2302.0/tests'
make  check-TESTS
make[2]: Entering directory '/home/michael/debian/build-area/rsyslog-8.2302.0/tests'
make[3]: Entering directory '/home/michael/debian/build-area/rsyslog-8.2302.0/tests'
<nothing happening at this point anymore>


michael   663328  0.0  0.0   2972  2224 pts/3    S+   18:44   0:00 make check
michael   663329  0.0  0.0   4056  2948 pts/3    S+   18:44   0:00 /bin/bash -c fail=; \ if (target_option=k; case ${target_option-} in ?) ;; *) echo "am__make_running_with_option: internal error: invalid" "target option '${target_option-}' specified" >&2; exit 1;; 
michael   663391  0.0  0.0   3732  3108 pts/3    S+   18:44   0:00 make check
michael   663393  0.0  0.0   3936  3124 pts/3    S+   18:44   0:00 make check-TESTS
michael   663399  0.0  0.0   4272  3148 pts/3    S+   18:44   0:00 /bin/bash -c set +e; bases='immark.log immark-inputname.log immark-ruleset.log immark-ruleset-custom-msg.log operatingstate-basic.log operatingstate-empty.log operatingstate-unclean.log smtradfile.lo
michael   663408  0.0  0.0   3908  3160 pts/3    S+   18:44   0:00 make test-suite.log TEST_LOGS=immark.log immark-inputname.log immark-ruleset.log immark-ruleset-custom-msg.log operatingstate-basic.log operatingstate-empty.log operatingstate-unclean.log smtradfile.
michael   663409  0.0  0.0   4088  2912 pts/3    S+   18:44   0:00 /bin/bash ../test-driver --test-name immark.sh --log-file immark.log --trs-file immark.trs --color-tests yes --enable-hard-errors yes --expect-failure no -- ./immark.sh
michael   663413  1.5  0.0   5544  4568 pts/3    S+   18:44   0:00 /bin/bash ./immark.sh
[...]

PR #5144, opened 2023-05-23
Add new global config option "libcapng.default"

### LINKED ISSUE #5096: Ryslog 8.2212.0 fails to start in docker environment
### Expected behavior

Run rsyslog in docker

### Actual behavior

Errors with:

rsyslogd 8.2302.0: running as pid 1, enabling container-specific defaults, press ctl-c to terminate rsyslog
rsyslog internal message (3,-2455): could not transfer  the  specified  internal posix  capabilities settings to the kernel, capng_apply=-5
 [v8.2302.0 try https://www.rsyslog.com/e/2455 ]

Run rsyslog 8.2212.0 within a docker container

### Environment
- rsyslog version: 8.2212.0
- platform: docker/alpine

Cropi:
The new global option defines how rsyslog should behave in case something went wrong when capabilities were to be dropped. Default value is "on", in which case rsyslog exits on a libcapng related error. If set to "off", an error message describing the problem appears at startup, nothing more. Default value is preserved for backwards compatibility.

The capability dropping had to be moved to a different place, where config parameters are already available.

Closes #5096

### DIFF
--- a/runtime/glbl.c
+++ b/runtime/glbl.c
@@ -178,7 +178,8 @@ static struct cnfparamdescr cnfparamdescr[] = {
 	{ "parser.supportcompressionextension", eCmdHdlrBinary, 0 },
 	{ "shutdown.queue.doublesize", eCmdHdlrBinary, 0 },
 	{ "debug.files", eCmdHdlrArray, 0 },
-	{ "debug.whitelist", eCmdHdlrBinary, 0 }
+	{ "debug.whitelist", eCmdHdlrBinary, 0 },
+	{ "libcapng.default", eCmdHdlrBinary, 0 }
 };
 static struct cnfparamblk paramblk =
 	{ CNFPARAMBLK_VERSION,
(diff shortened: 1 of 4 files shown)


rgerhards:
@Cropi Sry for the late request, just noticed it pre-merge: I would appreciate if you could rebase to master as this should solve some CI issues.

As we have new parameters, please also add doc to https://github.com/rsyslog/rsyslog-doc/

Cropi:
Doc PR created https://github.com/rsyslog/rsyslog-doc/pull/1009

rgerhards:
@Cropi I did not think well enough.

The libcap-ng feature actually introduced a breaking change for a number of scenarios, e.g. when running as an ordinary user. So the default setting should reflect that state before we introduced that feature. Given the fact it was only present for ~2 month, I do not see any issue with breaking compatibility for that short period of time. Will create a PR.

For an example of the initial breaking change, pls see also https://github.com/rsyslog/rsyslog/issues/5091

PR #5146, opened 2023-05-24
imjournal: Support input module

wltu:
Allow us to bind the imjournal input to an output module. Thie enable us to block journal reading if the output module is failing. Note that we will only allow one input module for imjournal since we do not have multiple input sources. The memory usage can be high with multiple outputs.

The input module support the `main` argument to select the config to use in the main thread and will be stop stopped if the output source is not accepting data. Other rules in the background will continue.

Tested:

Example Config:
```
template(name="journal-format" type="string"
  string="%TIMESTAMP:::date-rfc3339% %HOSTNAME% %syslogtag% %msg%\n"
)

ruleset(name="remote" queue.type="direct"){
  action(
    type="omfwd"
    target="IP_ADDRESS"
    port="PORT"
    protocol="tcp"
    template="journal-format"
    StreamDriver="ossl"
    StreamDriverMode="0"
    StreamDriverAuthMode="anon"
    action.resumeRetryCount="-1"
  )
}

input(
 type="imjournal"
 ruleset="remote"
 main="on"
)
```

- Stop collector network -> Stop Journal pointer
```
SECONDS=0
expected="$(md5sum /var/log/state)"
echo "Expected state ${expected}"
while [ $SECONDS -lt 300 ];
do
  new_state="$(md5sum /var/log/state)"
  if [[ "${new_state}" != "${expected}" ]];
  then
    echo "Got state ${new_state} instead of ${expected}"
      exit 1
  fi
  sleep 5
done
echo "Passed"
exit 0
```

```
Expected state 309593de34bc095e4245c1ba07a7520c  /var/log/state
Passed
```

- Resume Journal Pointer

The state file changed after the connection resumed.
```
$ md5sum /var/log/state
d11c7e46fec4f86151df204608525d89  /var/log/state
```

The message send in when the connection was down is also sent out properly.

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -33,6 +33,9 @@
 #include <time.h>
 #include <sys/poll.h>
 #include <sys/socket.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
 #include <errno.h>
 #include <systemd/sd-journal.h>
 #include <fcntl.h>
@@ -46,6 +49,7 @@
 #include "net.h"
 #include "glbl.h"
 #include "statsobj.h"
+#include "ruleset.h"
 #include "parser.h"
 #include "prop.h"
 #include "errmsg.h"
@@ -65,11 +69,23 @@ DEFobjCurrIf(glbl)
 DEFobjCurrIf(parser)
 DEFobjCurrIf(prop)
 DEFobjCurrIf(net)
+DEFobjCurrIf(ruleset)
 DEFobjCurrIf(statsobj)
+
+/* Module static data */
+typedef struct journal_etry_s {
+	pthread_t tid;	/* the worker's thread ID */
+	ruleset_t *pBindRuleset;
+	char* stateFile;
+	struct journalContext_s *journalContext;
+	struct journal_etry_s *next;
+} journal_etry_t;
+static journal_etry_t *journal_root = NULL;
+static int n_journal = 0;
 
 struct modConfData_s {
 	rsconf_t *pConf;
-	int bIgnPrevMsg;
+	instanceConf_t *root, *tail;
 };
 
 static struct configSettings_s {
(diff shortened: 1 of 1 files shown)


wltu:
Updated thanks. I copied the wrong one from before.

Seems like CI failed. I'll work on fixing those.

wltu:
um. The CI is still failing. I have fixed the mem leak issue, but it now failing with `buildbot/rsyslog solaris10x64 sunstudio`.

rgerhards:
There was some problem with the github CI. If you could rebase and push, that should heal the issue (except for some flakes we sometimes see - I'll take care of that).

Please note: I personally will unfortunately not be able to attend github the next couple of days.

wltu:
I am looking at different options of this since maybe we don't need have multiple different journal pointer.

If we uses direct queue and there are a log of messages piled up before the timeout value, are there any concern for that? Does the RAM usage goes up as more actions are queued up?

I have noticed that if the retry count is `-1` then it will prevent sending messages to other output, but if it is other values it will allow the messages to send to the other outputs. So if we want to retry for a long time while not discarding the message, is the only way setting the retry count to really high value?

davidelang:
On Tue, 30 May 2023, Willy Tu wrote:

> I am looking at different options of this since maybe we don't need have multiple different journal pointer.
>
> If we uses direct queue and there are a log of messages piled up before the timeout value, are there any concern for that? Does the RAM usage goes up as more actions are queued up?

Yes, more items in the queue will use more ram

> I have noticed that if the retry count is `-1` then it will prevent sending 
> messages to other output, but if it is other values it will allow the messages 
> to send to the other outputs. So if we want to retry for a long time while not 
> discarding the message, is the only way setting the retry count to really high 
> value?

this is where you would setup a queue (potentially with a disk assist queue) on 
the action, so that logs for that action can queue up while other actions can 
process logs

see:
https://www.rsyslog.com/doc/master/whitepapers/queues_analogy.html

David Lang

wltu:
Thanks for the resource David. I'll take a closer look into https://www.rsyslog.com/doc/master/whitepapers/queues_analogy.html and see if what's the best approach for my needs.

wltu:
I have updated the message to sync with the change. I am still working on getting on trying to optimize the memory usage, since right now each input binding will have a new journal source which can increase the allocated mem usage.

rgerhards:
@wltu just checking if you have further work planned

wltu:
Hi @rgerhards ,

yeah, I have plan to optimize the memory usage a bit more to make sure we can support multiple input bindings efficiently. 
Do you have any suggestion on limiting the memory usage from the current change?

(I won't be able to look at it again until next week).

wltu:
Sorry for the delay, I have updated and tested again to make sure there are no functional issues.

(There was a coredump before and was fixed, oops)

rgerhards:
@wltu I am sorry, I had overlooked this. But could you please add the relevant doc. 

wltu:
No worries, I'll update the doc this week. :+1:

wltu:
Added the documentation in [imjournal-input](https://github.com/wltu/rsyslog/tree/imjournal-input)

rgerhards:
thx for the patch and your persistence.

Issue #5167, opened 2023-06-28
libgcry.c, undefined behavior, and 5 failed self tests


noloader:
It appears libgcry.c may have a bit of undefined behavior, and it results in 5 failed self tests when testing with Undefined Behavior sanitizer (UBsan).

### Expected behavior

No undefined behavior in `libgcry.c`.

### Actual behavior

A sprinkle of undefined behavior in `libgcry.c`

### Steps to reproduce the behavior

Configure with Undefined Behavior sanitizer (UBsan). Add `-fsanitize=undefined -fno-sanitize-recover` to `CFLAGS` and `CXXFLAGS`:

```
./configure --enable-testbench=yes --enable-imdiag --enable-omstdout CFLAGS="-fsanitize=undefined -fno-sanitize-recover" CXXFLAGS="-fsanitize=undefined -fno-sanitize-recover"
```

### Environment

I working from the tip of master.

```
$ find . -name rsyslogd
./platform/freebsd/rsyslogd
./tools/rsyslogd

$ tools/rsyslogd -v
rsyslogd  8.2306.0 (aka 2023.06) compiled with:
        PLATFORM:                               x86_64-pc-linux-gnu
        PLATFORM (lsb_release -d):              Description:    Ubuntu 22.04.2 LTS
        FEATURE_REGEXP:                         Yes
        GSSAPI Kerberos 5 support:              No
        FEATURE_DEBUG (debug build, slow code): No
        32bit Atomic operations supported:      Yes
        64bit Atomic operations supported:      Yes
        memory allocator:                       system default
        Runtime Instrumentation (slow code):    No
        uuid support:                           Yes
        systemd support:                        Yes
        Config file:                            /etc/rsyslog.conf
        PID file:                               /var/run/rsyslogd.pid
        Number of Bits in RainerScript integers: 64
```

### Results

```
./configure ...   # as shown above
...
make -j 8 V=1
...
make check V=1
...

PASS: parsertest-snare_ccoff_udp.sh
PASS: parsertest-snare_ccoff_udp2.sh
FAIL: queue-encryption-disk.sh
FAIL: queue-encryption-disk_keyfile.sh
FAIL: queue-encryption-disk_keyprog.sh
PASS: queue-encryption-da.sh
PASS: omfile_hup-vg.sh
PASS: gzipwr_hup-vg.sh
FAIL: tcpflood_wrong_option_output.sh
PASS: imtcp-octet-framing-too-long-vg.sh
[...]

PR #5168, opened 2023-06-28
Clear undefined behavior in libgcry.c (GH #5167)

noloader:
This commit clear some undefined behavior in libgcry.c. When testing with Undefined Behavior sanitizer, it led to 5 failed self tests.

### DIFF
--- a/runtime/libgcry.c
+++ b/runtime/libgcry.c
@@ -575,7 +575,7 @@ seedIV(gcryfile gf, uchar **iv)
 		if(shift == 0) { /* need new random data? */
 			rndnum = randomNumber();
 		}
-		(*iv)[i] = 0xff & ((rndnum & (0xff << shift)) >> shift);
+		(*iv)[i] = 0xff & ((rndnum & (255u << shift)) >> shift);
 	}
 finalize_it:
 	RETiRet;


rgerhards (approved):
good catch - thx

rgerhards:
Thx for the patch and alerting us. I'll also check why UBSan in CI didn't catch this. Will leave it unmerged for now, as this eases CI testing.

rgerhards:
@noloader which version of ubsan did you try? I ask because we actually run it inside CI and it never triggerred any issues.

rgerhards:
Maybe it's also the platform. We run ubsan on Ubuntu 22.04 on AMD64.

noloader:
@rgerhards,

I tested using Ubuntu 22.04.6, amd64.

I can't find the commands I used for rsyslog in Bash history. However, another library I tested used:

```
clear && CFLAGS="-fsanitize=address -fno-sanitize-recover" CXXFLAGS="-fsanitize=address -fno-sanitize-recover" cmake ../
```

Be sure to use `-fno-sanitize-recover` to crash the test. Otherwise, some testing frameworks swallow the output and report no problems.

After testing is complete, I also perform the following to ensure the testing framewrok did not swallow something:

```
cd test
grep -IR 'runtime error'
```

If you have some failed self tests due to UBsan, then `grep` will find them.

rgerhards:
@noloader thx - but it sounds close to what we do, have a look here:

https://github.com/rsyslog/rsyslog/blob/master/.github/workflows/run_checks.yml#L120

I think I'll accept this as "well, it's like it is" (it sometimes is just so with the sanitizers) and merge the PR.

PR #5175, opened 2023-07-03
TLS CRL Support Issue 5081

darrenmoffat:


rgerhards:
Hi @darrenmoffat thx for the PR. There seem to be at least some build issues, please have a look. Also, I would appreciate if you could add all relevant information to the commit itself. We always try to keep commits self-descriptive so that they can be used without the need to any external ressources (it's convenient, and it also saves your life if the ressource goes away :-)).

Finally, it looks like we need some updates to the doc.

@alorbach can you check if you can craft a similar PR for GnuTLS? @darrenmoffat FYI: we need to provide equal functionality at the net layer, so we need the GnuTLS PR in place in order to merge.

rgerhards:
I've tentatively assigned the 2023.08 milestone, but it all depends on when we are completely ready to merge (including GnuTLS).

alorbach:
That's a very complex PR. and some issues need to be taken care of besides the support for gnutls in my opinion:
 
- a testcase for the new feature including a CRL File, a valid certificate and a revoked certificate file. The testcase should try to establish a TLS connection using a revoked certificate. The testcase would fail with a specific error message, that error message would need to be checked for. There are similar tests already like https://github.com/rsyslog/rsyslog/blob/master/tests/imtcp-tls-ossl-error-cert.sh .
- using makro checks on OPENSSL_VERSION_NUMBER, as far as I can see X509_STORE_load_file API is new since OpenSSL 3.0.0 and we have many build environments with much older OpenSSL Versions.
- GNUTLS, there are API's available: I think it can be done: https://www.gnutls.org/manual/html_node/PKIX-certificate-revocation-lists.html

@rgerhards : Some very specific enhancements, perhaps we could take it from here and add the necessary changes to the PR if  @darrenmoffat agrees?

darrenmoffat:
I agree it is complex and also note quite complete. I'm happy for you to take this over.  
An update to Oracle Solaris 11.4 with the changes I wrote will be released very soon (likely in the 11.4.60 release) since we need them but if any of the UI/config options change when you take this over that is OK, we can adapt since it won't be enabled by default.

alorbach:
> I agree it is complex and also note quite complete. I'm happy for you to take this over. An update to Oracle Solaris 11.4 with the changes I wrote will be released very soon (likely in the 11.4.60 release) since we need them but if any of the UI/config options change when you take this over that is OK, we can adapt since it won't be enabled by default.

Alright I will take over your PR and amend necessary changes, so that the PR can continue to stay.

In future I would recommend that you checkout a custom branch first and do the changes there, because you will most likely have to reset your local master branch after the PR has been merged in order to continue.

darrenmoffat:
Many thanks for taking this on.

Thanks also for the hint on how best to do the development with git.  I’m primarily a Mercurial user and the way we work in Solaris is we the stacked diffs and rebase on the default branch method so I’m not used to using branches for fix/feature development.  It is time I learned how to use git branching for this!.

Darren

From: Andre Lorbach ***@***.***>
Sent: 06 July 2023 09:36
To: rsyslog/rsyslog ***@***.***>
Cc: Darren Moffat ***@***.***>; Mention ***@***.***>
Subject: [External] : Re: [rsyslog/rsyslog] TLS CRL Support Issue 5081 (PR #5175)


I agree it is complex and also note quite complete. I'm happy for you to take this over. An update to Oracle Solaris 11.4 with the changes I wrote will be released very soon (likely in the 11.4.60 release) since we need them but if any of the UI/config options change when you take this over that is OK, we can adapt since it won't be enabled by default.

Alright I will take over your PR and amend necessary changes, so that the PR can continue to stay.

In future I would recommend that you checkout a custom branch first and do the changes there, because you will most likely have to reset your local master branch after the PR has been merged in order to continue.

alorbach:
@rgerhards @darrenmoffat I think this PR can be considered as ready. I was able to add support for older OpenSSL Versions and for GnuTLS driver. Fun fact, an old commit of 2008 for GnuTLS (https://github.com/rsyslog/rsyslog/commit/b0d63ea8f26f525bbfd177aaa6a1294b0d94f1f9) already contained the necessary API gnutls_certificate_set_x509_crl_file but was commented out.

I also added a new test certificate which is revoked and two testcases to check if the CRL check is working.

I will update the doc later and refer to this PR

alorbach:
Added PR for DOC: 
https://github.com/rsyslog/rsyslog-doc/pull/1012

PR #5191, opened 2023-07-25
CI: update zookeper download to newer version

rgerhards:
Old version is no longer available.


rgerhards:
The opensuse CI builder has a know local problem. So it does not prevent merge.

Issue #5192, opened 2023-07-26
Increase default timeout in diag.sh function "issue_HUP"


alorbach:
This most likely fix timing issues with tests using issue_HUP and temporary name resolution issues resolving the local hostname.

PR #5193, opened 2023-07-26
testbench: make waiting for HUP processing more reliable

rgerhards:
The previous approach was more or less delay based. We have now changed the code to enable imdiag to detect if HUP is underway and wait until it is completed. The new method still employs some kind of timeout, but is now quite reliable. Most importantly, it works great with long-running HUP processing, which can happen e.g. when querying the system name takes long or some actions need longer time to persist their HUP processing.

The new approach will most likely reduce CI flakes and also speed up testbench runs. The speedup happens from not having to wait a full delay in cases where we detect HUP is completed (plus reduced timeout when we cannot clearly detect this - see code comments why the new method is still considered more reliable than the old one).

closes https://github.com/rsyslog/rsyslog/issues/5192

### DIFF
--- a/dirty.h
+++ b/dirty.h
@@ -5,7 +5,7 @@
  * yet a runtime library, because it depends on some functionality
  * residing somewhere else.
  *
- * Copyright 2007-2014 Rainer Gerhards and Adiscon GmbH.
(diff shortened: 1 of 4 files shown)


rgerhards:
The ElasticSearch CI failure is unrelated and being addressed in another context. So we are ready to merge.

PR #5220, opened 2023-08-31
Avoid crash on restart in imrelp SIGTTIN handler

### LINKED ISSUE #5219: rsyslog crashes on restart in imrelp SIGTTIN
### Expected behavior

Should not crash

### Actual behavior

It crashes occasionally

To easily reproduce run the following script

`#!/bin/bash

FILE=/coredumps/core-*
count=0
while true ; do
        count=$((count+1))
        echo "restart rsyslog count = $count"
	sudo systemctl restart rsyslog
        if [ -f "$FILE" ]; then
                echo "Found coredump!!!!"
                break
        fi
	sleep 5
done`

While sending SIGTTIN continuously

`while :; do pgrep rsyslogd | xargs kill -21; done`

There are two issues, depending on the configuration

1. debug.unloadModules="off"
2. debug.unloadModules="on" (default)

With configuration (1), the crash is caused by a double free in doSIGTTIN of pRelpEngine
With configuration (2), the signal handler sometimes is called when it is already been unmapped from memory with the following backtrace

`#0  0x00007fbd42518c20 in ?? ()
(gdb) bt
#0  0x00007fbd42518c20 in ?? ()
#1  <signal handler called>
#2  _fini () at ../sysdeps/x86_64/crti.S:80
#3  0x00007fbd42479478 in _dl_fini () at dl-fini.c:161
#4  0x00007fbd416c79c9 in __run_exit_handlers (status=0, listp=0x7fbd41a2c540 <__exit_funcs>, run_list_atexit=run_list_atexit@entry=true)
    at exit.c:82
#5  0x00007fbd416c7a15 in __GI_exit (status=<optimized out>) at exit.c:104
#6  0x00007fbd416b0acc in __libc_start_main (main=0x40fd00 <main>, argc=3, argv=0x7ffcec0f1538, init=<optimized out>, 
    fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7ffcec0f1528) at libc-start.c:323
#7  0x0000000000410179 in _start () at ../sysdeps/x86_64/start.S:118`

I'm preparing a pull request with patch to fix both issues.

### Environment
rsyslog [v8.2308.0]

Alix82:
While existing, if at specific time rsyslog receives a SIGTTIN, it crashes due to 2 issues.

1. debug.unloadModules="off" a double free of pRelpEngine
2. debug.unloadModules="on" it crashes because the signal handler has been unmapped from memory.

This patch covers both issues and fixes #5219.

### DIFF
--- a/plugins/imrelp/imrelp.c
+++ b/plugins/imrelp/imrelp.c
@@ -839,7 +839,7 @@ static void
 doSIGTTIN(int __attribute__((unused)) sig)
 {
 	const int bTerminate = ATOMIC_FETCH_32BIT(&bTerminateInputs, &mutTerminateInputs);
-	if(bTerminate) {
+	if(bTerminate && (pRelpEngine != NULL)) {
 		relpEngineSetStop(pRelpEngine);
 	}
 }
@@ -882,6 +882,12 @@ ENDafterRun
 
 BEGINmodExit
 CODESTARTmodExit
+	struct sigaction newAct;
+	memset(&newAct, 0, sizeof (newAct));
+	sigemptyset(&newAct.sa_mask);
+	newAct.sa_handler = SIG_IGN;
+	sigaction(SIGTTIN, &newAct, NULL);
+
 	if(pRelpEngine != NULL)
 		iRet = relpEngineDestruct(&pRelpEngine);
 


rgerhards (approved):

rgerhards (changes requested):
I do not yet know why, but CI consistently shows problems. While we are used to some testbench flakes, this points to a real issue. We need to investigate.

rgerhards:
Yes, I think this is caused by the new signal handler settings:
```
==102808== Syscall param rt_sigaction(act->sa_mask) points to uninitialised byte(s)
==102808==    at 0x489A4F6: __libc_sigaction (sigaction.c:58)
==102808==    by 0x4861892: modExit (imrelp.c:887)
==102808==    by 0x46DBC2: modPrepareUnload (modules.c:551)
==102808==    by 0x46DACE: modUnlinkAndDestroy (modules.c:1004)
==102808==    by 0x46C3AF: modUnloadAndDestructAll (modules.c:1029)
==102808==    by 0x413609: deinitAll (rsyslogd.c:2265)
==102808==    by 0x41229C: main (rsyslogd.c:2344)
==102808==  Address 0x1ffefff558 is on thread 1's stack
==102808==  in frame #0, created by __libc_sigaction (sigaction.c:43)
==102808== 
```

Still not sure why it occurs, but it breaks at least CI testing.

plugins/imrelp/imrelp.c
```diff
@@ -882,6 +882,10 @@ ENDafterRun
 
 BEGINmodExit
 CODESTARTmodExit
+	struct sigaction newAct;
```
rgerhards, line 885: It looks like this struct needs to be static. If so, we do not get the valgrind violations. I haven't checked the API semantics yet, but thought I let you know @Alix82

alisarctl: Could be because I don't memset newAct, will update the PR, thanks.

rgerhards: > Could be because I don't memset newAct, will update the PR, thanks.

yes, I just went a bit further. The mem init is missing, thus the issue. When static, the loader zeros out.

alisarctl: Updated the PR, thank you.

rgerhards (approved):

PR #5223, opened 2023-09-12
omprog bugfix: Add CAP_DAC_OVERRIDE to the bounding set

Cropi:
The ```omprog``` module uses the ```execve()``` function to execute a third party program. 
Some required capabilities were not preserved in the bounding set [1]. This caused problems, e.g. the program could not write to files even if rsyslog was executed as root and privileges were not dropped (default behavior).

As of now, only the ```CAP_DAC_OVERRIDE``` capability is added to the bounding set. Others could be added later, if there is justification behind that.
```
    [1] The capability bounding set is a security mechanism that
    can be used to limit the capabilities that can be gained
    during an execve(2). During an execve, the capability
    bounding set is ANDed with the file permitted capability
    set, and the result of this operation is assigned to the
    thread's  permitted  capability  set. The capability
    bounding set thus places a limit on the permitted
    capabilities that may be granted by an executable file.
```

### DIFF
--- a/tools/rsyslogd.c
+++ b/tools/rsyslogd.c
@@ -1590,23 +1590,24 @@ initAll(int argc, char **argv)
 		int capability; /* capability code */
 		const char *name; /* name of the capability to be displayed */
 		sbool present; /* is the capability present that is needed by rsyslog? if so we do not drop it */
+		capng_type_t type;
 	} capabilities_t;
 
 	capabilities_t capabilities[] = {
-		#define CAP_FIELD(code) { code, #code,  0 }
-		CAP_FIELD(CAP_BLOCK_SUSPEND),
-		CAP_FIELD(CAP_CHOWN),
-		CAP_FIELD(CAP_IPC_LOCK),
-		CAP_FIELD(CAP_LEASE),
-		CAP_FIELD(CAP_NET_ADMIN),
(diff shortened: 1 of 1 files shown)


rgerhards:
Thx - that explains a lot. I had overlooked that subtle issue!

mbiebl:
It appears this also affects omfile.
I just had an rsyslog user report that the following rule failed:
```
news.crit                       action(type="omfile" file="/var/log/news/news.crit"   fileOwner="news" fileGroup="news" createDirs="off")

logger -p news.crit bar

Dez 31 15:21:13 pluto rsyslogd[446498]: file '/var/log/news/news.crit': open error: Permission denied [v8.2312.0 try https://www.rsyslog.com/e/2433 ]
Dez 31 15:21:13 pluto rsyslogd[446498]: action 'action-2-builtin:omfile' suspended (module 'builtin:omfile'), retry 0. There should be messages before this one giving the reason for suspension. [v8.2312.0 try https://www.rsyslog.com/e/2007 ]
Dez 31 15:21:13 pluto rsyslogd[446498]: action 'action-2-builtin:omfile' resumed (module 'builtin:omfile') [v8.2312.0 try https://www.rsyslog.com/e/2359 ]
```

Adding CAP_DAC_OVERRIDE fixed that issue.

PR #5230, opened 2023-09-26
fix startup issue on modern systemd systems

rgerhards:
When we startup AND are told to auto-background ourselfs, we must close all unneeded file descriptors. Not doing this has some security implications. Traditionally, we do this by iterating over all possible file descriptor values. This is fairly compatible, because we need no OS-specific method. However, modern systemd configs tend to not limit the number of fds, so there are potentially 2^31 fds to close. While this is OKish, it takes some time and makes systemd think that rsyslog did not properly start up.

We have now solved this by using the /proc filesystem to obtain our currently open fds. This works for Linux, as well as Cygwin, NetBSD, FreeBDS and MacOS. Where not available, we fall back to the traditional method. In our opionion, this fallback is unproblematic, as on these platforms there is no systemd and in almost all cases a decent number of fds to close.

Very special thanks go out to Brennan Kinney, who clearly described the issue to us on github and also provided ample ways to solve it. What we did is just implement what we think is the best fit from rsyslog's PoV.

closes https://github.com/rsyslog/rsyslog/issues/5158


polarathene:
> Very special thanks go out to Brennan Kinney, who clearly described the issue to us on github and also provided ample ways to solve it.
> What we did is just implement what we think is the best fit from rsyslog's PoV.

Thanks for resolving the issue so quickly! ❤️ 

---

> However, modern systemd configs tend to not limit the number of fds, so there are potentially 2^31 fds to close.

Just to clarify, by default since systemd v240 (2018Q4), that should be `1024:524288` limit. As in the soft limit is the expected `1024`.

The problem is other software shipping misconfiguration in systemd services that overrides this to something silly like `LimitNOFILE=infinity`.
- Which will map to the sysctl `fs.nr_open` (_a value systemd v240 also raises from `2^20` to 2^30`, some distro like Debian are known to opt-out via patch for the `fs.nr_open` change_).
- With the biggest issue there being that the soft limit was also set to `infinity` instead of their software requesting to raise the soft limit to a higher value that the hard limit permits. `infinity` isn't at all sane though.
- The known source of this misconfiguration is container software such as Docker and `containerd` (_which would often sync with the systemd `.service` config from the Docker daemon `dockerd.service`_).

---

> We have now solved this by using the `/proc` filesystem to obtain our currently open fds.

Just as additional reference, linking to my two detailed comments from the associated issue:
- https://github.com/rsyslog/rsyslog/issues/5158#issuecomment-1724558068
- https://github.com/rsyslog/rsyslog/issues/5158#issuecomment-1724755490

There is a caveat cited for the `/proc` approach (_from the referenced Fail2Ban commit in the first comment link_):

> _This wouldn't work on some platforms, or if `proc/fdescfs` not mounted, or a chroot environment, then it'd raise a `FileExistsError`._

I've only used Docker personally and the Fail2Ban approach works in that environment. I'm not sure if any others use a `chroot` environment. I am aware of some software like Postfix being configured to run within a chroot environment while also inside a container.

Both Fedora and openSUSE package managers have features to install at a different root where `chroot` is also used IIRC. I recall an issue with openSUSE where it was failing (_at least when run in a container_) due to `/proc` + `/dev` usage involved by a package install scripts. [Here's the link about that](https://github.com/openSUSE/libzypp/issues/444#issuecomment-1498617139) for reference if it becomes a problem, `/dev/fd` can symlnk to `/proc/fd` it mentions.

I don't know how likely this caveat would affect users. Probably best to go forward and see if any new issues are raised regarding it 😅 This is the same approach RedHat used AFAIK, although Fedora images running `dnf` last I checked still misbehaved under `LimitNOFILE=infinity`, I had not investigated further to see if that was using `/proc/fd` where it was stalling.

rgerhards:
> I don't know how likely this caveat would affect users. Probably best to go forward and see if any new issues are raised regarding it 😅 This is the same approach RedHat used AFAIK, although Fedora images running `dnf` last I checked still misbehaved under `LimitNOFILE=infinity`, I had not investigated further to see if that was using `/proc/fd` where it was stalling.

That's actually the path we currently have decided to take. If we receive new issues for that problem, we'll update the code. Especially in `chroot` environment, I am very sceptic of the security issues of not ensuring everything is closed. We would probably add a config setting to let the user specify up to which max fd we will close.

Using the newer system calls makes little sense ATM to be honest. We've seen far to few distros where we could actually use them.

So all in all, we think the current approach will solve the issue. Especially, as you said, because we are already working around other system component's bugs, which will hopefully become fixed soon.

Any more comments are appreciated. I have also added your problem case statement to the commit text.

tools/rsyslogd.c
```diff
@@ -41,6 +41,9 @@
 #ifdef ENABLE_LIBCAPNG
 	#include <cap-ng.h>
 #endif
+#if defined(HAVE_CLOSE_RANGE)
+#	include <linux/close_range.h>
+#endif
```
polarathene: Just to confirm, this is decided at build time right? Will that cause any issues on systems when it's not available at runtime?

I'm not familiar with C, just curious if the logic below fails gracefully if that's the case. Containers can be built with software on one system and run on many others where the kernel is different.

rgerhards: Well, the include is build-time only. The systemcall obviously is runtime. If the loader can't load rsyslog due to missing kernel/libc symbols, that's something the packagers need to decide. If the symbol loads and the call fails, we again go to the loop.

PR #5234, opened 2023-10-03
Add new global config option "libcapng.enable"

Cropi:
Defines whether rsyslog should drop capabilities at startup or not. By default, it is set to "on". Until this point, if the project was compiled with ```--enable-libcap-ng``` option, capabilities were automatically dropped. This is configurable now.


rgerhards:
Hi @Cropi thx for the PR, look quite useful.

But can you please shorten a couple of lines to keep within our codestyle? I can also do that if you have no time right now.

https://build.rsyslog.com/#/builders/178/builds/4962

Cropi:
> Hi @Cropi thx for the PR, look quite useful.
> 
> But can you please shorten a couple of lines to keep within our codestyle? I can also do that if you have no time right now.
> 
> https://build.rsyslog.com/#/builders/178/builds/4962

I've updated the PR based on your suggestion. Thanks.

PR #5235, opened 2023-10-04
Add CAP_NET_RAW to the capability set

Cropi:
The `CAP_NET_RAW` ensures the use of RAW and PACKET sockets,  which is utilized by the `omudpspoof` module, more precisely the `libnet_init` function.
Would it be possible to include this patch before the `8.2310.0` release? Thanks.

PR https://github.com/rsyslog/rsyslog/pull/5234 is needed to be merged before this can be done. The only thing that was added in the change is the CAP_NET_RAW capability.

Source: https://github.com/rsyslog/rsyslog/blob/master/plugins/omudpspoof/omudpspoof.c#L563


rgerhards:
> Would it be possible to include this patch before the `8.2310.0` release?

Definitely, I just keep it inside the queue. A doc issue seems to have popped up again, will comment in that other tracker.

mbiebl:
The reason I'm asking is because I'm wondering whether to enable libcapng support in Debian or if I should simply use systemd's `CapabilityBoundingSet=` feature in rsyslog.service.

The one benefit I see in libcapng support in rsyslog directly, is that it could adjust the needed capabilities depending on the loaded modules and the used configuration.  If the list is simply static, I think I can just utilize systemd here without requiring a library dependency on libcapng.

PR #5237, opened 2023-10-11
Drop CAP_IPC_LOCK capability

mbiebl:
It doesn't seem to be necessary.
```
$ grep -RE "(mlock|mmap|memfd|shmctl)"
<empty>
```


mbiebl:
@Cropi ^

Cropi:
Hello, sorry for the late reply but I was in a pretty bad health condition for ~2 weeks. I will look into it and provide you more details soon.

PR #5246, opened 2023-10-18
use logind instead of utmp for wall messages with systemd

tblume:
Future SUSE versions will get rid of utmp due to a 32bit time_t counter overflow in 2038.
See details at:

https://github.com/thkukuk/utmpx/blob/main/Y2038.md

On systemd based systems logind is an alternative to utmp.


tblume:
oops, obviously that needs rework

PR #5251, opened 2023-10-20
Remove no longer used --with-systemdsystemunitdir configure switch

mbiebl:
This is a clean up following the removal of the service unit in cfd07503ba055100a84d75d1a78a5c6cceb9fdab


mbiebl:
Fwiw, I still find it unfortunate, that the upstream provided .service file was removed.

PR #5252, opened 2023-10-20
Update Debian sample rsyslog.service unit

mbiebl:


rgerhards:
thx for the PR. Looks like CI needs an update, will do and re-run affected builders. Just so that everything looks clean.

mbiebl:
The current version of the file can be found at
https://salsa.debian.org/debian/rsyslog/-/blob/debian/master/debian/rsyslog.service

mbiebl:
Good question.
I notice, that platform/debian is not included in the dist tarball (contrary to other platforms, not sure if that is intentional).
I'm personally fine with whatever your deem best.

mbiebl:
not sure how many users (of git) make use of the files in platform/ ?

rgerhards:
> not sure how many users (of git) make use of the files in platform/ ?

I would tend to say *very* few. I moved the files to over there when we got questions that the samples did not work for their distros. That was years ago (actually quite some time, don't remember exactly when. git can tell, but I think "old" is sufficient).

rgerhards:
> I'm personally fine with whatever your deem best.

my personal gut feeling is that it hurts nobody to remove them. After all, there are many distros and we only have potentially outdated stuff. IIRC that is also the reason we keep it only in git.

I can run the question on the mailing list.

mbiebl:
I'm fine with removing the file and replacing it with a link to the file in Debian's gitlab

rgerhards:
As of the feedback on the rsyslog mailing list, I will now:

* merge this PR to get a clean state
* do a follow-up PR that replaces the actual service file with a readme and link to the debian definitions

PR #5254, opened 2023-10-22
Fix forking issue do to close_range call

### LINKED ISSUE #5241: Fork pipe error during start up on FreeBSD
Hi.

After updating Rsyslog to the latest version 8.2310.0 on FreeBSD, I see an error during start up:

3284.449390867:main thread    : rsyslogd.c: rsyslogd: in child, finalizing initialization
3284.449464175:main thread    : rsyslogd.c: closes unneeded files: opendir failed for /proc/self/fd
3284.449487278:main thread    : rsyslogd.c: closes unneeded files: opendir failed for /proc/fd
3284.449519181:main thread    : rsyslogd.c: rsyslogd: select() returns 1: OK
3284.449563086:main thread    : rsconf.c: telling modules to activate config (before dropping privs) 0x800e12000
3284.449571887:main thread    : rsconf.c: pre priv drop activating config 0x800e12000 for module imuxsock
3284.449584088:main thread    : imuxsock.c: imuxsock: allocating memory for 0 listeners
3284.449606490:main thread    : rsyslogd.c: rsyslog parent startup failure: error reading "fork pipe": No such file or directory
rsyslog startup failure: error reading "fork pipe": No such file or directory

Looks like it's trying to find the /proc filesystem, but FreeBSD doesn't have one :)

nhuff:
fixes #5241 
After fork if the child process uses close_range to close open file descriptors it has no way to exempt the parentPipeFD causing a failure to signal successful startup to the parent process.  This causes failures on all systems that aren't Linux that implement close_range.  This patch removes the use of close_range and reverts to just looping over possible file descriptors on those platforms.

### DIFF
--- a/tools/rsyslogd.c
+++ b/tools/rsyslogd.c
@@ -460,19 +460,24 @@ prepareBackground(const int parentPipeFD)
 		/* try MacOS, FreeBSD */
 		if(close_unneeded_open_files("/proc/fd", beginClose, parentPipeFD) != 0) {
 			/* did not work out, so let's close everything... */
-			const int endClose = getdtablesize();
-#		if defined(HAVE_CLOSE_RANGE)
-			if(close_range(beginClose, endClose, 0) != 0) {
-				dbgprintf("errno %d after close_range(), fallback to loop\n", errno);
-#		endif
-				for(int i = beginClose ; i <= endClose ; ++i) {
-					if((i != dbgGetDbglogFd()) && (i != parentPipeFD)) {
-						  aix_close_it(i); /* AIXPORT */
-					}
+			int endClose = (parentPipeFD > dbgGetDbglogFd()) ? parentPipeFD : dbgGetDbglogFd();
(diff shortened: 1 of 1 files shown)


rgerhards (changes requested):
Thx for the analysis and PR.

I think we should try to keep usage of close_range due to the runtime benefits. In almost all cases, all "regular" fds should be within the same range, and often consecutively. We could check that and use close_range for the remainder.

Would you be up to add this? If not, I can amend and/or add a follow-up PR.

nhuff:
I updated the pull request to do the fd closing in two phases.  

1. Loop through file descriptors between beginClose and MAX(parentPipeFD,dbgGetDbglogFd()) making sure not to close those two file descriptors.
2. Potentially use close_range to close all file descriptors above MAX(parentPipeFD,dbgGetDbglogFd())

rgerhards (approved):
thx for changing the method

PR #5255, opened 2023-10-23
replace debian sample systemd service file by readme

rgerhards:
The sample service file is as of 2023-10-23 no longer provided by the rsyslog project because service settings strongly depend on distribution policies.

For Debian, the distro current version can be found at

https://salsa.debian.org/debian/rsyslog/-/blob/debian/master/debian/rsyslog.service

Note: We provided a sample service files in the early days of systemd, when almost nobody was using systemd. That made sense because it eased transition for those trying out systemd. Now that systemd is mainstream, there is no real need any longer for a sample.

@mbiebl FYI


kochtopf:
The new service file adds CapabilityBoundingSet=CAP_BLOCK_SUSPEND CAP_CHOWN CAP_LEASE CAP_NET_ADMIN CAP_NET_BIND_SERVICE CAP_SYS_ADMIN CAP_SYS_RESOURCE CAP_SYSLOG. This will result in an open error: Permission denied [v8.2312.0 try https://www.rsyslog.com/e/2433 for action(type="omfile" file="/var/log/nginx/access.log" flushOnTXEnd="off" ioBufferSize="8k").
Fix: append CAP_DAC_OVERRIDE to CapabilityBoundingSet.

rgerhards:
You need to report this to the maintainer of the distro in question. The rsyslog project does not recommend any settings.

Issue #5160, opened 2023-06-21
Kernel timestamp in log message despite of default `ParseKernelTimestamp` off


opty77:
### Expected behavior
Log message without kernel timestamp e.g. in `/var/log/syslog`:

```
2023-06-21T10:09:55.693976+02:00 kiwi kernel - - - [IPTABLES DROP]: IN=venet0 OUT= MAC=<snip>
```

### Actual behavior
Log message **with** kernel timestamp e.g. in `/var/log/syslog`:

```
2023-06-21T10:09:55.693976+02:00 kiwi kernel - - - [1154069.463446] [IPTABLES DROP]: IN=venet0 OUT= MAC=<snip>
```

### Steps to reproduce the behavior
Run rsyslog and wait for a kernel message e.g. in `/var/log/syslog`. :-)

### Environment
- rsyslog version:

```
opty@kiwi:~$ rsyslogd -v
rsyslogd  8.2102.0 (aka 2021.02) compiled with:
        PLATFORM:                               x86_64-pc-linux-gnu
        PLATFORM (lsb_release -d):
        FEATURE_REGEXP:                         Yes
        GSSAPI Kerberos 5 support:              Yes
        FEATURE_DEBUG (debug build, slow code): No
        32bit Atomic operations supported:      Yes
        64bit Atomic operations supported:      Yes
        memory allocator:                       system default
        Runtime Instrumentation (slow code):    No
        uuid support:                           Yes
        systemd support:                        Yes
        Config file:                            /etc/rsyslog.conf
        PID file:                               /run/rsyslogd.pid
        Number of Bits in RainerScript integers: 64

See https://www.rsyslog.com for more information.
```

- platform: Debian GNU/Linux 11.7 (Bullseye) in an LXC container
- configuration:

```
opty@kiwi:~$ grep -vE '^(#|$)' /etc/rsyslog.conf
module(load="imuxsock") # provides support for local system logging
input(type="imuxsock" socket="/dev/log" useSpecialParser="off")
module(load="imklog")   # provides kernel logging support
template(name="SyslogProtocol23FormatWoPriVerMLS" type="string" string="%TIMESTAMP:::date-rfc3339% %HOSTNAME% %APP-NAME% %PROCID% %MSGID% %STRUCTURED-DATA%%msg:::sp-if-no-1st-sp%%msg%\n")
$ActionFileDefaultTemplate SyslogProtocol23FormatWoPriVerMLS
$FileOwner root
$FileGroup adm
[...]

PR #5262, opened 2023-10-30
imklog bugfix: keepKernelTimestamp=off config param did not work

rgerhards:
... at least not as expected. It was only honored for kernel-level messages and only when parseKernelTimestamp was "on". Otherwise, the kernel timestamp was always kept inside the message.

closes https://github.com/rsyslog/rsyslog/issues/5160


rgerhards:
Solaris seems to hang in CI. However, this patch has no effect on Solaris, so it is save to merge in any case.

Issue #5161, opened 2023-06-21
RFE: `ParseKernelTimestamp` or `KeepKernelTimestamp` in imkmsg


opty77:
Please add support for at least `ParseKernelTimestamp` configuration parameter or even `KeepKernelTimestamp` too in imkmsg.

(Template skipped due to RFE.)

rgerhards:
Definitely sounds like a good idea. I've seen the other messages and have the whole situation on my focus. Feel free to ping me if you do not hear back by end of the month.

opty77:
@rgerhards: Ping. :-)

rgerhards:
@opty77 sry, it looks like I was severely distracted (complicated year...).

Can you brush me up on the related issues? I can find https://github.com/rsyslog/rsyslog/issues/5160, but it looks like there is more.

rgerhards:
I finally found "the other one", at least I think. It is https://github.com/rsyslog/rsyslog/issues/4561

Based on that issue, I think we do not need options but the path is probably not to rely on the timestamp, except on boot. I'll look closer at the situation, any comments are welcome.

opty77:
> @opty77 sry, it looks like I was severely distracted (complicated year...).
> 
> [...]

@rgerhards: No problem, similar here. :-)

rgerhards:
@opty77 I have now looked into the imkmsg contributed module. It actually parses the timestamp returned from `/dev/kmsg` and times look correct to me even after hibernate. Could it be that the kmsg driver does some magic behind the scenes? In any case, I cannot confirm any bug in that regard.

Also, the integer timestamp is stored in json alongside the message. So with a proper template, you should be able to add it by configuration. I'll post a template somewhat later.

As #5160 talks about **`imklog`** I wonder if you are confusing the two here?

opty77:
@rgerhards: How did you check? I haven't tried imkmsg yet due to #4561 but using imklog I can see the difference between `/var/log/syslog`:

```
2023-10-31T12:54:04.639584+01:00 kiwi kernel - - - [2498588.930168] [IP6TABLES DROP]: IN=venet0 OUT= MAC=<snip> SRC=<snip> DST=<snip> LEN=60 TC=0 HOPLIMIT=236 FLOWLBL=0 PROTO=TCP SPT=58774 DPT=8001 WINDOW=65535 RES=0x00 SYN URGP=0
```

and `dmesg --time-format iso`:

```
[...]

Issue #4561, opened 2021-03-23
imkmsg is not calculating timestamps correctly


wizeman:
The imkmsg module is calculating timestamps incorrectly.
It is off by a few seconds on my Raspberry Pi and off by almost 10 hours on my laptop.

I've tracked down the issue to these offending lines:
https://github.com/rsyslog/rsyslog/blob/0be64ea70947ba60565cf503d575b479b4bdf358/contrib/imkmsg/kmsg.c#L132-L150

I suspect that the code I linked above might not be taking into account small time corrections (by an NTP daemon such as chronyd) and/or possibly the time elapsed while the computer has been suspended (potentially hours/days), but I haven't tested this.

This is affecting both the `timereported` and the `timegenerated` properties, so it seems that there is no way to get correct timestamps (with say, up to 1 second delay on an otherwise idle system).

The following patch solves the issue for me, at the expense of ignoring the timestamps in `/dev/kmsg` and just using the time when the messages are processed:
```patch
diff --git a/contrib/imkmsg/kmsg.c b/contrib/imkmsg/kmsg.c
index beb4076ef..379eceb99 100644
--- a/contrib/imkmsg/kmsg.c
+++ b/contrib/imkmsg/kmsg.c
@@ -57,8 +57,6 @@ static void
 submitSyslog(uchar *buf)
 {
    long offs = 0;
-   struct timeval tv;
-   struct sysinfo info;
    unsigned long int timestamp = 0;
    char name[1024];
    char value[1024];
@@ -129,27 +127,7 @@ submitSyslog(uchar *buf)
        json_object_object_add(json, name, jval);
    }
 
-   /* calculate timestamp */
-   sysinfo(&info);
-   gettimeofday(&tv, NULL);
-
-   /* get boot time */
-   tv.tv_sec -= info.uptime;
-
-   tv.tv_sec += timestamp / 1000000;
-   tv.tv_usec += timestamp % 1000000;
-
-   while (tv.tv_usec < 0) {
-       tv.tv_sec--;
-       tv.tv_usec += 1000000;
-   }
-
-   while (tv.tv_usec >= 1000000) {
-       tv.tv_sec++;
-       tv.tv_usec -= 1000000;
-   }
-
-   Syslog(priority, (uchar *)msg, &tv, json);
+   Syslog(priority, (uchar *)msg, NULL, json);
 }
```

I don't know if this is the correct solution and I'm not really interested in spending a lot of time on this so I didn't submit a PR.
[...]

PR #5264, opened 2023-10-31
imkmsg: add params parseKernelTimestamp and fixKernelTimeStamp

rgerhards:
... as module parameters. Here, "parseKernelTimestamp" is modelled after the same param from imklog. In essence, it permits to ignore all kernel time stamps and use current time instead.

Parameter fixKernelTimeStamp uses the kernel time stamp during the initial read loop of /dev/kmsg, but replaced it later ignores it for later reads.

The reasoning at least for the later is given in a sysklogd commit by Joachim Wiberg, which we reproduce here to have a stable reference. The commit itself can be found for example at:

https://github.com/troglobit/sysklogd/commit/9f6fbb3301e571d8af95f8d771469291384e9e95

============== QUOTE ===============
The spec[1] says the /dev/kmsg timestamp is a monotonic clock and in microseconds.  After a while you realize it's also relative to the boot of the system, that fact was probably too obvious to be put in the spec. However, what's *not* in the spec, and what takes a while to realize, is that this monotonic time is *not* adjusted for suspend/resume cycles ...

On a frequently used laptop this can manifest itself as follows.  The kernel is stuck on Nov 15, and for the life of me I cannot find any to adjust for this offset:

    $ dmesg -T |tail -1; date
    [Mon Nov 15 01:42:08 2021] wlan0: Limiting TX power to 23 (23 - 0) dBm as advertised by 18:e8:29:55:b0:62
    Tue 23 Nov 2021 05:20:53 PM CET

Hence this patch.  After initial "emptying" of /dev/kmsg when syslogd starts up, we raise a flag (denoting done with backlog), and after this point we ignore the kernel's idea of time and replace it with the actual time we have now, the same that userspace messages are logged with.

Sure, there will be occasions where there's a LOT of kernel messages to read and we won't be able to keep track.  Yet, this patch is better than the current state (where we log Nov 15).

[1]: https://www.kernel.org/doc/Documentation/ABI/testing/dev-kmsg ===========END QUOTE ===============

closes https://github.com/rsyslog/rsyslog/issues/4561 closes https://github.com/rsyslog/rsyslog/issues/5161


opty77:
Regarding old messages: https://github.com/troglobit/sysklogd/issues/29 and beyond.

@rgerhards: Would you mind opening an issue by myself? O:-)

(Note to self: Do not comment commits because of https://github.com/isaacs/github/issues/1526.)

(Copied [my comment](https://github.com/rgerhards/rsyslog/commit/a6a60c300df6a758605396652f536712e33dd14b#commitcomment-131406893) from https://github.com/rgerhards/rsyslog/commit/a6a60c300df6a758605396652f536712e33dd14b and extended.)

wizeman:
Here's my initial experience report with `FixKernelTimeStamp=on` (and `ParseKernelTimestamp` left as default, i.e. `off`). 

I will provide some "before vs after" comparisons, as well as a few of the most important log messages while booting my laptop, especially those related to rsyslog and what the timestamps look like. I will provide my conclusions at the end of this long comment.

However, please note the following:
1. The "before" in my system log corresponds to a configuration equivalent to `ParseKernelTimestamp=off` and `FixKernelTimestamp=off`. i.e. if you take a look at #4561, you will see that I was using a patch that basically disabled the parsing of the kernel timestamps, which means that all the timestamps in my "before" log corresponded to when `rsyslog` parsed those messages, rather than when they were actually emitted during boot.

2. I will provide these log messages in chronological order as they appear in my log (which unfortunately, does not always correspond to timestamp order).

3. I should also mention that while I have `chronyd` enabled to set my system time from NTP servers, it did not perform any "step jumps" in the system time during this boot, so we can discard that as being a confounding factor. So when I say "backwards time jump" or "forwards time jump", I only mean that in the sense of the timestamps in the log. The actual system time should not have had any actual "time jumps".

4. For reference, you can see my config below. It is especially important to know that I have configured `journald` to forward messages to `/dev/kmsg`, which are later picked up by `rsyslog`, who actually writes those messages to my actual log file(s).

<details>
<summary>My <code>journald.conf</code></summary>

```ini
[Journal]
Storage=persistent
RateLimitInterval=0
RateLimitBurst=0

ReadKMsg=false
ForwardToKMsg=true
MaxLevelKMsg=debug
```
</details>

<details>
<summary>My <code>syslog.conf</code></summary>

```C
$WorkDirectory /var/spool/rsyslog

module(load="imkmsg" fixkerneltimestamp="on")
module(load="immark" interval="300")

template(name="DefaultFormat" type="list") {
    property(name="timegenerated" dateFormat="rfc3339" date.inUTC="on")
    constant(value=" ")
    property(name="hostname")
    constant(value=" ")
    property(name="syslogfacility-text")
    constant(value=".")
    property(name="syslogseverity-text")
    property(name="msg" spifno1stsp="on" )
    property(name="msg" droplastlf="on" )
    constant(value="\n")
}

template(name="KernelFormat" type="list") {
    property(name="timegenerated" dateFormat="rfc3339" date.inUTC="on")
    constant(value=" ")
    property(name="hostname")
    constant(value=" ")
    property(name="syslogfacility-text")
    constant(value=".")
    property(name="syslogseverity-text")
    constant(value=" ")
    property(name="syslogtag")
    property(name="msg" spifno1stsp="on" )
    property(name="msg" droplastlf="on" )
    constant(value="\n")
}

$ActionFileDefaultTemplate DefaultFormat

mail.*              action(type="omfile" file="/var/log/mail.log" FileCreateMode="0600")
auth.*;authpriv.*   action(type="omfile" file="/var/log/auth.log" FileCreateMode="0600")
*.*;kern.none       action(type="omfile" file="/var/log/messages.log" FileCreateMode="0600")
*.warning;kern.none action(type="omfile" file="/var/log/warn.log" FileCreateMode="0600")
kern.warning        action(type="omfile" file="/var/log/warn.log" FileCreateMode="0600" Template="KernelFormat")
kern.*              action(type="omfile" file="/var/log/kernel.log" FileCreateMode="0600" Template="KernelFormat")
&                   action(type="omfile" file="/var/log/messages.log" FileCreateMode="0600" Template="KernelFormat")
```
</details>

Regarding the very early boot, in which there are only kernel messages, you can already see the difference. Before, this is what it looked like:

```python
2023-10-31T11:46:32.915586+00:00 laptop kern.info kernel: software IO TLB: area num 16.
2023-10-31T11:46:32.915594+00:00 laptop kern.info kernel: Memory: 22312356K/22952252K available (12289K kernel code, 29381K rwdata, 3276K rodata, 1852K init, 12480K bss, 639636K reserved, 0K cma-reserved)
```
After this PR:
```python
2023-11-01T14:24:34.574240+00:00 laptop kern.info kernel: software IO TLB: area num 16.
2023-11-01T14:24:37.518006+00:00 laptop kern.info kernel: Memory: 22312356K/22952252K available (12289K kernel code, 29381K rwdata, 3276K rodata, 1852K init, 12480K bss, 639636K reserved, 0K cma-reserved)
```
As you can see, it is now possible to observe that there is a ~3 second delay during which the Linux kernel initializes memory.

This all seems to continue as expected until I get the prompt to input my LUKS password. Before:
```python
2023-10-31T11:46:32.931373+00:00 laptop daemon.info systemd-tty-ask-password-agent[219]: Starting password query on /dev/tty1.
2023-10-31T11:46:32.937340+00:00 laptop daemon.info systemd-tty-ask-password-agent[219]: Password query on /dev/tty1 finished successfully.
```
After:
```python
2023-11-01T14:24:39.577611+00:00 laptop daemon.info systemd-tty-ask-password-agent[222]: Starting password query on /dev/tty1.
2023-11-01T14:24:59.035118+00:00 laptop daemon.info systemd-tty-ask-password-agent[222]: Password query on /dev/tty1 finished successfully.
```
As you can see, before this PR, you couldn't tell how long the laptop was waiting for me to input my password, but now you can see that I finished entering my password about ~20 seconds after the prompt appeared.

Now, at this point in time I also noticed something weird happening after this PR was applied. In my log, you can see the kernel outputting some USB initialization messages just before systemd asks for my LUKS password:

```python
2023-11-01T14:24:39.378331+00:00 laptop kern.info kernel: usb 3-1: New USB device strings: Mfr=0, Product=0, SerialNumber=0
2023-11-01T14:24:39.397048+00:00 laptop kern.info kernel: hub 1-1:1.0: USB hub found
2023-11-01T14:24:39.565694+00:00 laptop daemon.info systemd[1]: Started Dispatch Password Requests to Console.
2023-11-01T14:24:39.577611+00:00 laptop daemon.info systemd-tty-ask-password-agent[222]: Starting password query on /dev/tty1.
2023-11-01T14:24:59.035118+00:00 laptop daemon.info systemd-tty-ask-password-agent[222]: Password query on /dev/tty1 finished successfully.
```

systemd then continues booting as expected, but after a while you can see what appears to be the timestamps "jumping into the past":
```python
2023-11-01T14:25:04.684935+00:00 laptop daemon.info systemd[1]: Stopped Apply Kernel Variables.
2023-11-01T14:25:04.685622+00:00 laptop daemon.info systemd[1]: systemd-modules-load.service: Deactivated successfully.
2023-11-01T14:25:04.685763+00:00 laptop daemon.info systemd[1]: Stopped Load Kernel Modules.
2023-11-01T14:25:04.686389+00:00 laptop daemon.info systemd[1]: systemd-tmpfiles-setup.service: Deactivated successfully.
2023-11-01T14:24:39.397556+00:00 laptop kern.info kernel: hub 1-1:1.0: 5 ports detected
2023-11-01T14:24:39.469374+00:00 laptop kern.info kernel: usb 2-1: new SuperSpeed Plus Gen 2x1 USB device number 2 using xhci_hcd
2023-11-01T14:24:39.483728+00:00 laptop kern.info kernel: usb 2-1: New USB device found, idVendor=0bda, idProduct=0487, bcdDevice= 1.49
```

However, notice how the timestamps in the past appear to come from the kernel (rather than systemd) and they are a continuation of the USB initialization process, with almost the same timestamps as the earlier kernel USB messages!

So it kinda seems that systemd must have somehow temporarily prevented the kernel from adding messages to `/dev/kmsg`, but the kernel then later appended those messages with the proper earlier timestamps, causing what appears to be jumps in the system time (although, I'm pretty sure the system time didn't actually change). My guess is that this has nothing to do with this PR, i.e. it's probably an artifact of the interaction between systemd and the kernel and how they append messages to `/dev/kmsg`, although this is just my guess.

These kernel USB messages, along with other kernel messages, continue to appear in the log with similar timestamps, until we see another apparent "time jump forward", when the log transitions again from kernel messages to systemd messages:

```python
2023-11-01T14:25:00.924310+00:00 laptop kern.notice kernel: Key type encrypted registered
2023-11-01T14:25:04.692017+00:00 laptop daemon.info systemd[1]: Stopped Create Volatile Files and Directories.
2023-11-01T14:25:04.692755+00:00 laptop daemon.info systemd[1]: Stopped target Local File Systems.
```

At this point in time, the boot continues normally. I can see many further transitions between kernel and systemd messages (and back), but this time there are no "backwards time jumps" and the timestamps look just like I would expect them to look, i.e. without any unexplained gaps in time.

However, later in the boot process, `rsyslog` finally starts. At this point, you can see `rsyslog` emitting a few log messages (in the middle of many other services starting as well). However, soon after, just as the kernel starts emitting a bunch of additional messages, there appears to be another "backwards time jump" as well, of about ~1.4 seconds:

```python
2023-11-01T14:25:08.004484+00:00 laptop daemon.info systemd[1]: Started Syslog Daemon.
(...)
2023-11-01T14:25:08.073766+00:00 laptop syslog.info rsyslogd[2293]: [origin software="rsyslogd" swVersion="8.2304.0" x-pid="2293" x-info="https://www.rsyslog.com"] start
2023-11-01T14:25:08.074534+00:00 laptop daemon.info dbus-daemon[2296]: dbus[2296]: Unknown username "nm-openconnect" in message bus configuration file
2023-11-01T14:25:08.081843+00:00 laptop daemon.info bluetoothd[2250]: Bluetooth daemon 5.66
2023-11-01T14:25:08.086296+00:00 laptop daemon.info dbus-daemon[2296]: dbus[2296]: Unknown username "pulse" in message bus configuration file
2023-11-01T14:25:08.095638+00:00 laptop daemon.notice zed[2362]: eid=2 class=config_cache_write pool='rpool'
2023-11-01T14:25:08.099206+00:00 laptop daemon.notice zed[2365]: eid=1 class=history_event pool='rpool'
2023-11-01T14:25:06.729061+00:00 laptop kern.info kernel: [drm] amdgpu: 10942M of GTT memory ready.
2023-11-01T14:25:06.729531+00:00 laptop kern.info kernel: [drm] GART: num cpu pages 262144, num gpu pages 262144
```

As you can see, this backwards time jump also corresponds to a transition between systemd messages and kernel messages.

When the log transitions back to systemd messages, there's an almost imperceptible ~0.1s "forwards time jump". I didn't even notice it at first because it's so small, but it's possible to see that there was indeed a forwards time jump because the timestamps of the systemd messages do continue where there they left off (I will again show a couple of the previous systemd messages for clarity):

```python
2023-11-01T14:25:08.095638+00:00 laptop daemon.notice zed[2362]: eid=2 class=config_cache_write pool='rpool'
2023-11-01T14:25:08.099206+00:00 laptop daemon.notice zed[2365]: eid=1 class=history_event pool='rpool'
(...)
2023-11-01T14:25:07.907125+00:00 laptop kern.info kernel: RPC: Registered tcp NFSv4.1 backchannel transport module.
2023-11-01T14:25:08.021605+00:00 laptop kern.info kernel: input: TPPS/2 ALPS TrackPoint as /devices/platform/i8042/serio1/serio2/input/input28
2023-11-01T14:25:08.121297+00:00 laptop daemon.notice zed[2386]: eid=8 class=history_event pool='bpool'
2023-11-01T14:25:08.122149+00:00 laptop daemon.info dbus-daemon[2296]: dbus[2296]: Unknown username "systemd-timesync" in message bus configuration file
```

I would have liked to see the message `DBGPRINTF("imkmsg: initial read done, changing to non-blocking mode\n")` being output to the log to see exactly where this transition happened, but unfortunately I hadn't enabled this debug message so it does not appear in my log. In fact, I don't even know how to enable it, I'd have to consult the documentation...

My guess is that this event could have happened here, because you can see an unexplained backwards time jump in-between a whole bunch of docker messages (i.e. outside of a systemd<->kernel message transition):
```python
2023-11-01T14:25:11.230323+00:00 laptop daemon.info dockerd[3028]: time="2023-11-01T14:25:11.232821021Z" level=info msg="loading plugin \x5c"io.containerd.internal.v1.tracing\x5c"..." type=io.containerd.internal.v1
2023-11-01T14:25:11.230538+00:00 laptop daemon.info dockerd[3028]: time="2023-11-01T14:25:11.233023702Z" level=error msg="failed to initialize a tracing processor \x5c"otlp\x5c"" error="no OpenTelemetry endpoint: skip plugin"
2023-11-01T14:25:10.249729+00:00 laptop daemon.info dockerd[3028]: time="2023-11-01T14:25:11.242225728Z" level=info msg=serving... address=/var/run/docker/containerd/containerd-debug.sock
2023-11-01T14:25:10.252115+00:00 laptop daemon.info dockerd[3028]: time="2023-11-01T14:25:11.243672574Z" level=info msg=serving... address=/var/run/docker/containerd/containerd.sock.ttrpc
```

However, I'm not entirely sure if that's what happened there, because after the above messages, you can see some normal transitions between kernel and systemd messages (without any time jumps) but then you can see another apparent "backwards time jump" in another one of these transitions, which as far as I understand should not have happened if indeed `rsyslog` had already performed the initial read:

```python
2023-11-01T14:25:12.147932+00:00 laptop daemon.info syncthing[2699]: [X7LZF] 2023/11/01 14:25:12 INFO: Completed initial scan of sendreceive folder "<redacted>" (<redacted>)
2023-11-01T14:25:12.152182+00:00 laptop daemon.info syncthing[2699]: [X7LZF] 2023/11/01 14:25:12 INFO: Completed initial scan of sendreceive folder "<redacted>" (<redacted>)
2023-11-01T14:25:11.565909+00:00 laptop kern.info kernel: r8169 0000:03:00.0: invalid VPD tag 0x00 (size 0) at offset 0; assume missing optional EEPROM
2023-11-01T14:25:11.821583+00:00 laptop kern.warning kernel: serial 0000:03:00.1: VPD access failed.  This is likely a firmware bug on this device.  Contact the card vendor for a firmware update
```

So, from a user perspective, here's what I can conclude:

1. Using `FixKernelTimeStamp=on` definitely provides a lot more information about the actual message timestamps than I had before (in which I had the equivalent of `ParseKernelTimestamp=off` and `FixKernelTimestamp=off`).

2. However, there are a few backwards and forwards jumps in the timestamps corresponding to "kernel log" <-> "systemd log" transitions which can make the interpretation of these timestamps confusing.

3. In addition, there's something quite weird happening in these last two snippets of my log. The first one shows an apparent backwards time jump in-between docker messages. It could have corresponded to the time `rsyslog` finished the initial processing of the `/dev/kmsg` log, but even if that is so... then why do we see another backwards time jump later (in-between systemd and kernel log messages)?

I was going to test this PR on one of my Raspberry Pis, which doesn't have a battery-backed clock and therefore always boots with the system clock set to the wrong date/time, but those logs would probably be even more confusing, so perhaps we should sort this one out first?

Thanks!

wizeman:
Oh damn, I just noticed that my `chronyd` is being started with `-s`, which I believe causes `chronyd` to set the system time from the RTC as a side effect. This might explain the jump in the timestamps in-between `dockerd` messages. Sorry for the confusion!

wizeman:
I am now seeing a more significant problem: I have suspended my laptop for about a minute and now all the timestamps in my log are running about 1 minute behind the system time of my laptop...

So it appears the `fixkerneltimestamp="on"` is having no effect. I wonder if I misconfigured this option?

contrib/imkmsg/kmsg.c
```diff
@@ -220,7 +246,13 @@ readkmsg(void)
 			continue;
 		} else {
 			/* something went wrong - error or zero length message */
-			if (i < 0 && errno != EINTR && errno != EAGAIN) {
+			if(i < 0 && (errno == EAGAIN || errno == EWOULDBLOCK)) {
+				DBGPRINTF("imkmsg: initial read done, changing to non-blocking mode\n");
```
wizeman: I think this message should actually be "changing to blocking mode" (rather than "non-blocking")?

wizeman:
> So it appears that `fixkerneltimestamp="on"` is not having the desired effect. I wonder if I misconfigured this option?

Indeed, I had misconfigured the option. The problem is that with this PR, `rsyslog` interprets this config:
 `module(load="imkmsg" fixkerneltimestamp="on")`
to be the same thing as this config:
`module(load="imkmsg")`
(or this one: `module(load="imkmsg" parsekerneltimestamp="on")`.

In other words, `parsekerneltimestamp="on"` (which is the default) overrides `fixkerneltimestamp="on"`, thus causing `fixkerneltimestamp="on"` to have no effect.

Once I changed my config to `module(load="imkmsg" parsekerneltimestamp="off" fixkerneltimestamp="on")`, it now behaves as expected. That is, the early kernel boot log shows the delays in-between messages (i.e. the timestamps as when they were generated, rather than when rsyslog parsed them), but now the timestamps after suspending/resuming the laptop are also in sync with the system time!

Thanks!

rgerhards:
@opty77 

> Would you mind opening an issue by myself?

No problem. We had a public holiday over here yesterday and I was catching up on other things today. Hope to come back to this topic tomorrow.

Re message dup, I think about a somewhat diff approach. But we can discuss that in the new tracker, it's a good idea.

contrib/imkmsg/kmsg.c
rgerhards: indeed (Y)

rgerhards:
> Indeed, I had misconfigured the option. The problem is that with this PR, `rsyslog` interprets this config:
> `module(load="imkmsg" fixkerneltimestamp="on")`
> to be the same thing as this config:
> `module(load="imkmsg")`
> (or this one: `module(load="imkmsg" parsekerneltimestamp="on")`.

sound like I need to do a bit more error-checking on config parameters and their interdependency. Thx.

and special thanks for the elaborte testing!!!

wizeman:
I just tried the latest version of this PR with the `ParseKernelTimestamp="startup"` option explicitly set and it also seems to be working just fine, both in terms of parsing the startup timestamps and also in terms of keeping accurate timestamps after the initial parsing (e.g. after suspending/resuming my laptop).

Thanks!

PR #5268, opened 2023-11-04
imdocker: Add image name to metadata fields

aliask:
## What
Exposes the image name and tag of the source container (what the Docker API simply calls `Image`) as an additional metadata value alongside the existing ones.

## Why
While the ImageID is available today, a SHA hash is not a particularly user-friendly way to identify what is actually generating the logs.

The image name provides a simple, easy-to-read identifier for the common scenario where an unmodified application image is used.

## Testing
A test for this additional property has been written, however the `devtools/devcontainer.sh` environment does not include Docker, so attempting to run the test locally crashes with `docker: command not found`. 

Modifying this development environment to add Docker is probably beyond the scope of this PR, so I hope that the CI tests will suffice.

## Reference
Docker [API documentation](https://docs.docker.com/engine/api/v1.27/#tag/Container) reference.

### DIFF
--- a/contrib/imdocker/imdocker.c
+++ b/contrib/imdocker/imdocker.c
@@ -72,6 +72,7 @@ extern int Debug;
 
 #define DOCKER_CONTAINER_ID_PARSE_NAME "Id"
 #define DOCKER_CONTAINER_NAMES_PARSE_NAME "Names"
+#define DOCKER_CONTAINER_IMAGE_PARSE_NAME "Image"
 #define DOCKER_CONTAINER_IMAGEID_PARSE_NAME "ImageID"
 #define DOCKER_CONTAINER_CREATED_PARSE_NAME "Created"
 #define DOCKER_CONTAINER_LABELS_PARSE_NAME "Labels"
@@ -124,6 +125,7 @@ typedef struct imdocker_req_s {
 
 typedef struct docker_container_info_s {
     uchar *name;
+    uchar *image;
     uchar *image_id;
     uint64_t created;
     /* json string container labels */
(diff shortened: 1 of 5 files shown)


rgerhards:
@aliask Due to my sluggishness, this PR unfortunately went way out of sync with the current codebase. My apologies. I can see if I can rebase it, but it would probably be better to extract the changes and integrate them into the current codebase. I will look at this, and I can commit with the proper --author applied.

If you prefer, we can also work together to get it back on track.

aliask:
Thanks for taking the time to pick this up @rgerhards.

I've rebased the branch against main and resolved the conflicts, let me know if this is okay!

rgerhards:
@aliask thx for your work! Unfortunately, there is a problem left. Looking at CI and AI review, I guess the AI nailed it.

rgerhards:
devtools/format-code.sh should resolve the formatting change. If it sticks, there is a problem w/ clang-format (I recently noticed that it works somewaht different between versions :-( - we use v18).

aliask:
> devtools/format-code.sh should resolve the formatting change. If it sticks, there is a problem w/ clang-format (I recently noticed that it works somewaht different between versions :-( - we use v18).

I noticed that on Ubuntu 25.04, when following the prompt in `format-code.sh` to run `apt install clang-format` it will install v20. However v18 is also available by running `apt install clang-format-18` instead.

rgerhards:
> I noticed that on Ubuntu 25.04, when following the prompt in `format-code.sh` to run `apt install clang-format` it will install v20. However v18 is also available by running `apt install clang-format-18` instead.

OK, this is something we need to look into - it's unfortunate that clang-format does not keep the canonical format (the whole point of format change was to have no flip/flop of different formatting in git history). I can reformat it with v18 if that is OK for you.

aliask:
I've already run the formatting script with `clang-format-18` and now also squashed the branch, but I wonder if the message in the script might be better if it:
1. Calls `/usr/bin/clang-format-18`
2. Recommends installing `clang-format-18` - this package name appears to be present since at least Ubuntu 20.04 and should hopefully keep things a little more consistent

I'm happy to include this change in this PR or raise a separate one if you think it's appropriate.

rgerhards:
> I'm happy to include this change in this PR or raise a separate one if you think it's appropriate.

I'd appreciate a second PR, that would be great.

rgerhards:
gemini comments sound acceptable to me, so let's finally merge before it goes out of sync. We may want to apply some changes later on.

PR #5278, opened 2023-11-21
imkmsg: add params "readMode" and "expectedBootCompleteSeconds"

rgerhards:
These parameters permit to control when imkmsg reads the full kernel log upon startup.

Parameter "readMode" provides the following options:

  * full-boot - (default) read full klog, but only "immediately" after boot. "Immediately" is hereby meant in seconds of system uptime given in "expectedBootCompleteSeconds"
  * full-always - read full klog on every rsyslog startup. Most probably causes messag duplication
  * new-only - never emit existing kernel log message, read only new ones.

Note that some message loss can happen if rsyslog is stopped in "full-boot" and "new-only" read mode. The longer rsyslog is inactive, the higher the message loss probability and potential number of messages lost. For typical restart scenarios, this should be minimal. On HUP, no message loss occurs as rsyslog is not actually stopped.

The default value for "expectedBootCompleteSeconds" is 90.

see also https://github.com/rsyslog/rsyslog/issues/5161

### DIFF
--- a/contrib/imkmsg/imkmsg.c
+++ b/contrib/imkmsg/imkmsg.c
@@ -72,7 +72,9 @@ static int bLegacyCnfModGlobalsPermitted;/* are legacy module-global config para
 
 /* module-global parameters */
 static struct cnfparamdescr modpdescr[] = {
-	{ "parsekerneltimestamp", eCmdHdlrGetWord, 0 }
+	{ "parsekerneltimestamp", eCmdHdlrGetWord, 0 },
+	{ "readmode", eCmdHdlrGetWord, 0 },
+	{ "expectedbootcompleteseconds", eCmdHdlrPositiveInt, 0 }
 };
 static struct cnfparamblk modpblk =
 	{ CNFPARAMBLK_VERSION,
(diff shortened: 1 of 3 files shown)


opty77:
@rgerhards: It seems I failed to open an issue in time, sorry. :-/

I guess you don't like tracking message sequence number then? (No message loss or duplication should happen, see https://github.com/troglobit/sysklogd/commit/92a4fb3318709b0daed4701d6710abc485107460 and https://github.com/troglobit/sysklogd/commit/b0d4e4cc3f8b8a923aa8c256b0817d8c52a9af11.)

Your solution looks good enough for regular use, though. (Like HUP after rotating logs weekly.)

rgerhards:
> I guess you don't like tracking message sequence number then? (No message loss or duplication should happen, see [troglobit/sysklogd@92a4fb3](https://github.com/troglobit/sysklogd/commit/92a4fb3318709b0daed4701d6710abc485107460) and [troglobit/sysklogd@b0d4e4c](https://github.com/troglobit/sysklogd/commit/b0d4e4cc3f8b8a923aa8c256b0817d8c52a9af11).)

I am not sure it's worth yet another state file. I also fear that we may run into situations where after a restart a tracking number is "recreated" and we get hard-to-diagnose issue reports.

opty77:
@rgerhards: Not sure what exactly you mean by "recreated" but okay, why not to try another (yours) approach.

rgerhards:
> Not sure what exactly you mean by "recreated"

I am not sure how the sequence numbers are generated by the kmsg device. So it may be possible that we run into the same number (that's inside the state file) again, which can confuse the algo. On the other hand, as in the quoted commits, one can use os boot cleanup of the state file. But that may in turn interfere with privilege drop and/or capabilities.

In any case, let's first try this approach. If you are unhappy, I am willing to implement state-file caching of last message read. I could do it actually in any case, but then the question is which mode to use as default.

Issue #5211, opened 2023-08-04
DTLS Support with rsyslog


alorbach:
Look into DTLS Support for rsyslog udp sending.

For OpenSSL see: https://www.openssl.org/docs/man1.1.1/man3/DTLS_server_method.html

PR #5280, opened 2023-11-24
Initial implementation of imdtls and omdtls modules

alorbach:
Initial implementation of imdtls and omdtls modules

- Extracted basic OpenSSL helper functions into own module net_ossl.h/net_ossl.c
  Both are compiled into lmnsd_ossl.
- Cleanup of OpenSSL code, fixed minor compiler and linking issues.
- Added DTLS Sender option DTLS into tcpflood for testbench.
- Add initial implementation of imdtls input module. Added to configure and makefile
- Add initial implementation of omdtls output module. Added to configure and makefile
- Add multiple basic tests for imdtls receiving data by using tcpflood.
- Add multiple send-receive test for imdtls and omdtls based on existing tls tests.
- Add timeout and sessionbreak tests for imdtls stress testing.

closes: https://github.com/rsyslog/rsyslog/issues/5211

### DIFF
--- a/plugins/imdtls/imdtls.c
+++ b/plugins/imdtls/imdtls.c
@@ -1,0 +1,1164 @@
+/**
+ * The dtls input module, uses OpenSSL as library to implement DTLS.
+ *
+ * \author  Andre Lorbach <alorbach@adiscon.com>
+ *
+ * Copyright (C) 2023 Adiscon GmbH.
+ *
+ * This file is part of rsyslog.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *       -or-
+ *       see COPYING.ASL20 in the source distribution
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "config.h"
+
+#include <stdio.h>
+#include <arpa/inet.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <sys/errno.h>
+#include <poll.h>
+#include <assert.h>
+#include <time.h>
+
+// --- Include openssl headers as well
+#include <openssl/ssl.h>
+#include <openssl/x509v3.h>
+#include <openssl/err.h>
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L && !defined(LIBRESSL_VERSION_NUMBER)
+#	include <openssl/bioerr.h>
+#endif
+#include <openssl/engine.h>
+// ---
+
+#include "rsyslog.h"
+#include "dirty.h"
+#include "module-template.h"
+#include "cfsysline.h"
+#include "msg.h"
+#include "errmsg.h"
+#include "glbl.h"
+#include "srUtils.h"
+#include "msg.h"
+#include "parser.h"
+#include "datetime.h"
+#include "net.h"
+#include "net_ossl.h"
+#include "prop.h"
+#include "ruleset.h"
+#include "statsobj.h"
+#include "unicode-helper.h"
+
+MODULE_TYPE_INPUT
+MODULE_TYPE_NOKEEP
+MODULE_CNFNAME("imdtls")
+
+/* defines */
+#define MAX_DTLS_CLIENTS 1024
+#define MAX_DTLS_MSGSIZE 65536
+#define DTLS_LISTEN_PORT "4433"
+// 1800 seconds = 30 minutes
+#define DTLS_DEFAULT_TIMEOUT 1800
+
+/* Module static data */
+DEF_OMOD_STATIC_DATA
+DEFobjCurrIf(glbl)
+DEFobjCurrIf(datetime)
+DEFobjCurrIf(prop)
+DEFobjCurrIf(ruleset)
+DEFobjCurrIf(statsobj)
+DEFobjCurrIf(net)
+DEFobjCurrIf(net_ossl)
+
+#define DTLS_MAX_RCVBUF 8192	/* Maximum DTLS packet is 8192 bytes which fits into Jumbo Frames. If not enabled
+				*  message fragmentation (Ethernet MTU of ~ 1500 bytes) can occur.*/
+
+/* config settings */
+typedef struct configSettings_s {
+	uchar *pszBindRuleset;		/* name of Ruleset to bind to */
+} configSettings_t;
+static configSettings_t cs;
+
+struct dtlsClient_s {
+	SSL* sslClient;			/* DTSL (SSL) Client */
+	time_t lastActivityTime;	/* Last Activity Time */
+	int clientfd;			/* ClientFD */
+};
+
+// Use typedef to define a type 'dtlsClient_t' based on 'struct dtlsClient_s'
+typedef struct dtlsClient_s dtlsClient_t;
+
+struct instanceConf_s {
+	/* Network properties */
+	uchar *pszBindAddr;		/* Listening IP Address */
+	uchar *pszBindPort;		/* Port to bind socket to */
+	int timeout;			/* Default timeout for DTLS Sessions */
+	/* Common properties */
+	uchar *pszBindRuleset;		/* name of ruleset to bind to */
+	uchar *pszInputName;
+	prop_t *pInputName;		/* InputName in property format for fast access */
+	ruleset_t *pBindRuleset;	/* ruleset to bind listener to (use system default if unspecified) */
+	sbool bEnableLstn;		/* flag to permit disabling of listener in error case */
+	statsobj_t *stats;		/* listener stats */
+	STATSCOUNTER_DEF(ctrSubmit, mutCtrSubmit)
+	/* OpenSSL properties */
+	uchar *tlscfgcmd;			/* OpenSSL Config Command used to override any OpenSSL Settings */
+	permittedPeers_t *pPermPeersRoot;	/* permitted peers */
+	int CertVerifyDepth;			/* Verify Depth for certificate chains */
+	/* Instance Variables */
+	char *pszRcvBuf;
+	int lenRcvBuf;
+	/**< -1: empty, 0: connection closed, 1..NSD_OSSL_MAX_RCVBUF-1: data of that size present */
+	int ptrRcvBuf;			/**< offset for next recv operation if 0 < lenRcvBuf < NSD_OSSL_MAX_RCVBUF */
+
+	/* OpenSSL and Config Cert vars inside net_ossl_t now */
+	net_ossl_t *pNetOssl;		/* OSSL shared Config and object vars are here */
+	int nClients;			/* */
+	dtlsClient_t **dtlsClients;	/* Array of DTSL (SSL) Clients */
+	int sockfd;			/* UDP Socket used to bind to */
+	struct sockaddr_in server_addr;	/* Server Sockaddr */
+	int port;			/* Server Port as integer */
+
+	int id;				/* Thread ID */
+	thrdInfo_t *pThrd;		/* Thread Instance Info */
+	pthread_t tid;			/* the instances thread ID */
+
+	struct instanceConf_s *next;
+	struct instanceConf_s *prev;
+};
+
+/* config variables */
+struct modConfData_s {
+	rsconf_t *pConf;		/* our overall config object */
+	instanceConf_t *root, *tail;
+	uchar *pszBindRuleset;		/* default name of Ruleset to bind to */
+	AuthMode drvrAuthMode;		/* authenticate peer if no other name given */
+};
+
+static modConfData_t *loadModConf = NULL;/* modConf ptr to use for the current load process */
+static modConfData_t *runModConf = NULL;/* modConf ptr to use for the current load process */
+
+static prop_t *pInputName = NULL;
+
+/* module-global parameters */
+static struct cnfparamdescr modpdescr[] = {
+	{ "ruleset", eCmdHdlrGetWord, 0 },
+	{ "tls.authmode", eCmdHdlrString, 0 },
+};
+static struct cnfparamblk modpblk =
+	{ CNFPARAMBLK_VERSION,
+	  sizeof(modpdescr)/sizeof(struct cnfparamdescr),
+	  modpdescr
+	};
+
+/* input instance parameters */
+static struct cnfparamdescr inppdescr[] = {
+	{ "port", eCmdHdlrString, CNFPARAM_REQUIRED },
+	{ "address", eCmdHdlrString, 0 },
+	{ "timeout", eCmdHdlrPositiveInt, 0 },
+	{ "name", eCmdHdlrString, 0 },
+	{ "ruleset", eCmdHdlrString, 0 },
+	{ "tls.permittedpeer", eCmdHdlrArray, 0 },
+	{ "tls.authmode", eCmdHdlrString, 0 },
+	{ "tls.cacert", eCmdHdlrString, 0 },
+	{ "tls.mycert", eCmdHdlrString, 0 },
+	{ "tls.myprivkey", eCmdHdlrString, 0 },
+	{ "tls.tlscfgcmd", eCmdHdlrString, 0 }
+};
+static struct cnfparamblk inppblk =
+	{ CNFPARAMBLK_VERSION,
+	  sizeof(inppdescr)/sizeof(struct cnfparamdescr),
+	  inppdescr
+	};
+#include "im-helper.h" /* must be included AFTER the type definitions! */
+
+/* create input instance, set default parameters, and
+ * add it to the list of instances.
+ */
+static rsRetVal
+createInstance(instanceConf_t **pinst)
+{
+	instanceConf_t *inst;
+	DEFiRet;
+	CHKmalloc(inst = malloc(sizeof(instanceConf_t)));
+	inst->next = NULL;
+
+	inst->pszBindAddr = NULL;
+	inst->pszBindPort = NULL;
+	inst->timeout = 1800;
+	inst->pszBindRuleset = loadModConf->pszBindRuleset;
+	inst->pszInputName = NULL;
+	inst->pBindRuleset = NULL;
+	inst->bEnableLstn = 0;
+
+	inst->tlscfgcmd = NULL;
+	inst->pPermPeersRoot = NULL;
+	inst->CertVerifyDepth = 2;
+
+	/* node created, let's add to config */
+	if(loadModConf->tail == NULL) {
+		loadModConf->tail = loadModConf->root = inst;
+	} else {
+		loadModConf->tail->next = inst;
+		loadModConf->tail = inst;
+	}
+
+	// Construct pNetOssl helper
+	CHKiRet(net_ossl.Construct(&inst->pNetOssl));
+	inst->pNetOssl->authMode = loadModConf->drvrAuthMode;
+
+	*pinst = inst;
+finalize_it:
+	RETiRet;
+}
+
+
+/* function to generate an error message if the ruleset cannot be found */
+static inline void
+std_checkRuleset_genErrMsg(__attribute__((unused)) modConfData_t *modConf, instanceConf_t *inst)
+{
+	LogError(0, NO_ERRCODE, "imdtls[%s]: ruleset '%s' not found - "
+			"using default ruleset instead",
+			inst->pszBindPort, inst->pszBindRuleset);
+}
+
+static void
+DTLSCloseSocket(instanceConf_t *inst) {
+	DBGPRINTF("imdtls: DTLSCloseSocket for %s:%d\n", inst->pszBindAddr, inst->port);
+	// Close UDP Socket
+	close(inst->sockfd);
+	inst->sockfd = 0;
+}
+
+static rsRetVal
+DTLSCreateSocket(instanceConf_t *inst) {
+	DEFiRet;
+	int optval = 1;
+	int flags;
+
+	struct in_addr ip_struct;
+	DBGPRINTF("imdtls: DTLSCreateSocket for %s:%d\n", inst->pszBindAddr, inst->port);
+	
+	// Create UDP Socket
+	inst->sockfd = socket(AF_INET, SOCK_DGRAM, 0);
+	if (inst->sockfd < 0) {
+		LogError(0, NO_ERRCODE, "imdtls: Unable to create DTLS listener,"
+				" failed to create socket, "
+				" ignoring port %d bind-address %s.",
+				inst->port, inst->pszBindAddr);
+		ABORT_FINALIZE(RS_RET_ERR);
+	}
+	setsockopt(inst->sockfd, SOL_SOCKET, SO_REUSEADDR, &optval, sizeof(optval));
+	setsockopt(inst->sockfd, SOL_SOCKET, SO_REUSEPORT, &optval, sizeof(optval));
+
+	// Set NON Blcoking Flags
+	flags = fcntl(inst->sockfd, F_GETFL, 0);
+	fcntl(inst->sockfd, F_SETFL, flags | O_NONBLOCK);
+
+	// Convert IP Address into numeric
+	if (inet_pton(AF_INET, (char*) inst->pszBindAddr, &ip_struct) <= 0) {
+		LogError(0, NO_ERRCODE, "imdtls: Unable to create DTLS listener,"
+				" invalid Bind Address, "
+				" ignoring port %d bind-address %s.",
+				inst->port, inst->pszBindAddr);
+		ABORT_FINALIZE(RS_RET_ERR);
+	}
+
+	// Set Server Address
+	memset(&inst->server_addr, 0, sizeof(struct sockaddr_in));
+	inst->server_addr.sin_family = AF_INET;
+	inst->server_addr.sin_port = htons(inst->port);
+	inst->server_addr.sin_addr.s_addr = htonl(ip_struct.s_addr);
+
+	// Bind UDP Socket
+	if (bind(inst->sockfd, (struct sockaddr*)&inst->server_addr, sizeof(struct sockaddr_in)) < 0) {
+		LogError(0, NO_ERRCODE, "imdtls: Unable to create DTLS listener,"
+				" unable to bind, "
+				" ignoring port %d bind-address %s.",
+				inst->port, inst->pszBindAddr);
+		ABORT_FINALIZE(RS_RET_ERR);
+	}
+finalize_it:
+	RETiRet;
+}
+
+/* Verify Callback for X509 Certificate validation. Force visibility as this function is not called anywhere but
+*  only used as callback!
+*/
+static int
+imdtls_verify_callback(int status, SSL* ssl)
+{
+	DEFiRet;
+	X509 *certpeer;
+	instanceConf_t *inst = NULL;
+
+	dbgprintf("imdtls_verify_callback: get SSL [%p]\n", (void *)ssl);
+	inst = (instanceConf_t*) SSL_get_ex_data(ssl, 2);
+
+	/* Continue check if certificate verify was valid */
+	if (status == 1) {
+		if (	ssl != NULL &&
+			inst != NULL) {
+			/* call the actual function based on current auth mode */
+			switch(inst->pNetOssl->authMode) {
+				case OSSL_AUTH_CERTNAME:
+					/* if we check the name, we must ensure the cert is valid */
+					certpeer = net_ossl_getpeercert(inst->pNetOssl, ssl, NULL);
+					dbgprintf("imdtls_verify_callback: Check peer certname[%p]=%s\n",
+						(void *)ssl, (certpeer != NULL ? "VALID" : "NULL"));
+					CHKiRet(net_ossl_chkpeercertvalidity(inst->pNetOssl, ssl, NULL));
+					CHKiRet(net_ossl_chkpeername(inst->pNetOssl, certpeer, NULL));
+					break;
+				case OSSL_AUTH_CERTFINGERPRINT:
+					certpeer = net_ossl_getpeercert(inst->pNetOssl, ssl, NULL);
+					dbgprintf("imdtls_verify_callback: Check peer fingerprint[%p]=%s\n",
+						(void *)ssl, (certpeer != NULL ? "VALID" : "NULL"));
+					CHKiRet(net_ossl_chkpeercertvalidity(inst->pNetOssl, ssl, NULL));
+					CHKiRet(net_ossl_peerfingerprint(inst->pNetOssl, certpeer, NULL));
+					break;
+				case OSSL_AUTH_CERTVALID:
+					certpeer = net_ossl_getpeercert(inst->pNetOssl, ssl, NULL);
+					dbgprintf("imdtls_verify_callback: Check peer valid[%p]=%s\n",
+						(void *)ssl, (certpeer != NULL ? "VALID" : "NULL"));
+					CHKiRet(net_ossl_chkpeercertvalidity(inst->pNetOssl, ssl, NULL));
+					break;
+				case OSSL_AUTH_CERTANON:
+					dbgprintf("imdtls_verify_callback: ANON[%p]\n", (void *)ssl);
+					FINALIZE;
+					break;
+			}
+		} else {
+			LogMsg(0, RS_RET_NO_ERRCODE, LOG_WARNING, "imdtls_verify_callback: MISSING ssl or inst!");
+		}
+	}
+finalize_it:
+	if(iRet != RS_RET_OK) {
+		DBGPRINTF("imdtls_verify_callback: FAILED\n");
+		status = 0;
+	}
+	return status;
+}
+
+
+static rsRetVal
+addListner(modConfData_t __attribute__((unused)) *modConf, instanceConf_t *inst)
+{
+	uchar statname[64];
+	DEFiRet;
+	DBGPRINTF("imdtls: addListner ENTER\n");
+
+	if(!inst->bEnableLstn) {
+		DBGPRINTF("imdtls: DTLS Listener not started because it is disabled by config error\n");
+		FINALIZE;
+	}
+
+	inst->pszInputName = ustrdup((inst->pszInputName == NULL) ?  UCHAR_CONSTANT("imdtls") : inst->pszInputName);
+	CHKiRet(prop.Construct(&inst->pInputName));
+	CHKiRet(prop.SetString(inst->pInputName, inst->pszInputName, ustrlen(inst->pszInputName)));
+	CHKiRet(prop.ConstructFinalize(inst->pInputName));
+
+	/* Init defaults */
+	if (inst->pszBindPort == NULL) {
+		CHKmalloc(inst->pszBindPort = ustrdup((uchar*) DTLS_LISTEN_PORT));
+	}
+
+	/* Init SSL Handles! */
+	CHKmalloc(inst->dtlsClients = (dtlsClient_t**) calloc(MAX_DTLS_CLIENTS, sizeof(dtlsClient_t*)));
+	for (int i = 0; i < MAX_DTLS_CLIENTS; ++i) {
+		CHKmalloc(inst->dtlsClients[i] = (dtlsClient_t*) calloc(1, sizeof(dtlsClient_t)));
+	}
+	inst->nClients = 0;
+
+	/* support statistics gathering */
+	CHKiRet(statsobj.Construct(&(inst->stats)));
+	snprintf((char*)statname, sizeof(statname), "%s(%s)",
+		 inst->pszInputName, inst->pszBindPort);
+	statname[sizeof(statname)-1] = '\0'; /* just to be on the save side... */
+	CHKiRet(statsobj.SetName(inst->stats, statname));
+	CHKiRet(statsobj.SetOrigin(inst->stats, (uchar*)"imdtls"));
+	STATSCOUNTER_INIT(inst->ctrSubmit, inst->mutCtrSubmit);
+	CHKiRet(statsobj.AddCounter(inst->stats, UCHAR_CONSTANT("submitted"),
+		ctrType_IntCtr, CTR_FLAG_RESETTABLE, &(inst->ctrSubmit)));
+	CHKiRet(statsobj.ConstructFinalize(inst->stats));
+	/* end stats counters */
+
+	// Init OpenSSL Context with DTLS_server_method
+	CHKiRet(net_ossl.osslCtxInit(inst->pNetOssl, DTLS_method()));
+
+#	if OPENSSL_VERSION_NUMBER >= 0x10002000L
+	// Init OpenSSL Cookie Callbacks
+	CHKiRet(net_ossl.osslCtxInitCookie(inst->pNetOssl));
+#	endif
+	// Run openssl config commands in Context
+	CHKiRet(net_ossl_apply_tlscgfcmd(inst->pNetOssl, inst->tlscfgcmd));
+
+	// Init Socket
+	CHKiRet(DTLSCreateSocket(inst));
+finalize_it:
+	if(iRet != RS_RET_OK) {
+		LogError(0, NO_ERRCODE, "DTLS Listener for thread failed to create UDP socket "
+				"for thread %s is not functional!", inst->pszInputName);
+	} else {
+		DBGPRINTF("imdtls: DTLS Listener for thread %s added\n", inst->pszInputName);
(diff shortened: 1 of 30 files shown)


plugins/imdtls/imdtls.c
```diff
@@ -0,0 +1,886 @@
...
+DEFobjCurrIf(glbl)
+DEFobjCurrIf(datetime)
+DEFobjCurrIf(prop)
+DEFobjCurrIf(ruleset)
+DEFobjCurrIf(statsobj)
+DEFobjCurrIf(net_ossl)
+
+#define DTLS_MAX_RCVBUF 1380	/* Maximum DTLS packet 1380 bytes to avoid fragmentation (smaller than the common
+				*  Ethernet MTU of 1,500 bytes to accommodate potential IP and UDP headers). */
+// #define DTLS_MAX_RCVBUF 16 * 1024 + 1/* TLS RFC 8449: max size of buffer for message reception */
```
rgerhards: remove comment -- or is there any reason we can go above 1380 bytes?

Looking further down in buffer handling, I actually think larger messages are possible, especially if "jumbo frames" are enabled at the network layer (some background: https://www.networkworld.com/article/745164/mtu-size-issues.html)

plugins/imdtls/imdtls.c
```diff
@@ -0,0 +1,886 @@
...
+	STATSCOUNTER_DEF(ctrSubmit, mutCtrSubmit)
+
+	AuthMode authMode;		/* authenticate peer if no other name given */
+	uchar *tlscfgcmd;		/* OpenSSL Config Command used to override any OpenSSL Settings */
+	struct {
+		int nmemb;
+		uchar **name;
+	} permittedPeers;
+
+	int bHaveSess;			/* True if DTLS session is established */
```
rgerhards: `bHaveSess` variable is not used anywhere in imdtls.c

plugins/imdtls/imdtls.c
```diff
@@ -0,0 +1,886 @@
...
+	inst->nClients = 0;
+
+
+	/* support statistics gathering */
+	CHKiRet(statsobj.Construct(&(inst->stats)));
+	snprintf((char*)statname, sizeof(statname), "%s(%s)",
+		 inst->pszInputName, inst->pszBindPort);
+	statname[sizeof(statname)-1] = '\0'; /* just to be on the save side... */
+	CHKiRet(statsobj.SetName(inst->stats, statname));
+	CHKiRet(statsobj.SetOrigin(inst->stats, (uchar*)"imrelp"));
```
rgerhards: wrong name `imrelp`

plugins/imdtls/imdtls.c
```diff
@@ -0,0 +1,886 @@
...
+	CHKiRet(submitMsg2(pMsg));
+	STATSCOUNTER_INC(inst->ctrSubmit, inst->mutCtrSubmit);
+finalize_it:
+
+	RETiRet;
+}
+
+static void
+DTLSHandleSessions(instanceConf_t *inst) {
+	fd_set readfds;
```
rgerhards: general note: we should replace `select()` with `poll()` an platforms that support it. This is done e.g. in imudp. `select()` is slow and, more important, has a number of issues (e.g. on some platforms fds > 1024 are not supported IIRC).

plugins/imdtls/imdtls.c
```diff
@@ -0,0 +1,886 @@
...
+	for (int i = 0; i < MAX_DTLS_CLIENTS; ++i) {
+		if (inst->sslClients[i] == NULL) {
+			return;
+		}
+		int fd = BIO_get_fd(SSL_get_wbio(inst->sslClients[i]), NULL);
+		if (FD_ISSET(fd, &readfds)) {
+			DBGPRINTF("imdtls: Read Client activity on index %d.\n", i);
+
+			// Existing client Finish handshake
+			int ret = SSL_accept(inst->sslClients[i]);
```
rgerhards: question: this looks like an `accept()` is done for every read (packet). Do I understand that correctly? If so, is this really needed?

plugins/imdtls/imdtls.c
```diff
@@ -0,0 +1,886 @@
...
+
+			// If SSL_accept succeeded, proceed to read data
+			char buf[MAX_DTLS_MSGSIZE];
+			int len = 0;
+
+			do {
+				len = SSL_read(inst->sslClients[i], buf, sizeof(buf) - 1);
+				if (len > 0) {
+					buf[len] = '\0';
+					// Process Message
```
rgerhards: comment not needed

plugins/imdtls/imdtls.c
```diff
@@ -0,0 +1,886 @@
...
+			  "param '%s'\n", inppblk.descr[i].name);
+		}
+	}
+	
+	/* check if no port is set. If so, we use DEFAULT of 433 */
+	if(inst->pszBindPort == NULL) {
+		CHKmalloc(inst->pszBindPort = (uchar*)strdup("443"));
+	}
+
+	/*
```
rgerhards: q: why is this commented out?

plugins/imdtls/imdtls.c
```diff
@@ -0,0 +1,886 @@
...
+	*/
+
+	inst->bEnableLstn = -1; /* all ok, ready to start up */
+
+finalize_it:
+CODE_STD_FINALIZERnewInpInst
+	cnfparamvalsDestruct(pvals, &inppblk);
+	if(iRet != RS_RET_OK) {
+		if(inst != NULL) {
+			// free(inst->tlscfgcmd);
```
rgerhards: this looks like the two ìf`s can be deleted - or do we still need to free something?

plugins/imdtls/imdtls.c
```diff
@@ -0,0 +1,886 @@
...
+
+BEGINactivateCnfPrePrivDrop
+	instanceConf_t *inst;
+CODESTARTactivateCnfPrePrivDrop
+	runModConf = pModConf;
+	DBGPRINTF("imdtls: activate addListners for dtls\n");
+	for(inst = runModConf->root ; inst != NULL ; inst = inst->next) {
+		addListner(pModConf, inst);
+	}
+//finalize_it:
```
rgerhards: remove commented out code

plugins/imdtls/imdtls.c
```diff
@@ -0,0 +1,886 @@
...
+	pthread_attr_t wrkrThrdAttr;
+CODESTARTrunInput
+	pthread_attr_init(&wrkrThrdAttr);
+	pthread_attr_setstacksize(&wrkrThrdAttr, 4096*1024);
+
+	DBGPRINTF("imdtls: create dtls handling threads\n");
+	for(inst = runModConf->root ; inst != NULL ; inst = inst->next) {
+		if(inst->bEnableLstn) {
+			// inst.pThrd = pThrd;
+			pthread_create(&inst->tid, &wrkrThrdAttr, startDtlsHandler, inst);
```
rgerhards, line 1076: if code not needed, remove

runtime/net_ossl.c
```diff
@@ -0,0 +1,743 @@
...
+#include "module-template.h"
+#include "parse.h"
+#include "srUtils.h"
+#include "obj.h"
+#include "errmsg.h"
+#include "net_ossl.h"
+#include "nsd_ptcp.h"
+#include "rsconf.h"
+
+// MODULE_TYPE_LIB
```
rgerhards: remove unneeded code

runtime/net_ossl.c
```diff
@@ -0,0 +1,743 @@
...
+// MODULE_TYPE_NOKEEP
+
+/* static data */
+DEFobjStaticHelpers
+DEFobjCurrIf(glbl)
+DEFobjCurrIf(nsd_ptcp)
+
+/*--------------------------------------MT OpenSSL helpers ------------------------------------------*/
+static MUTEX_TYPE *mutex_buf = NULL;
+static sbool openssl_initialized = 0; // Avoid multiple initialization / deinitialization
```
rgerhards, line 57: just a question: is it ensure that both DTLS TCP/TLS initialize openssl just once when they are both configured?

runtime/net_ossl.c
```diff
@@ -0,0 +1,743 @@
...
+
+/* set up support functions for openssl multi-threading. This must
+ * be done at library initialisation. If the function fails,
+ * processing can not continue normally. On failure, 0 is
+ * returned, on success 1.
+ */
+int opensslh_THREAD_setup(void)
+{
+	int i;
+	if (openssl_initialized == 1) {
```
rgerhards, line 110: don't we need to guard this by a mutex (BTW: this may be a bug in existing code.

If so, we need to

```
do
    lock mutex
    if  not initialzed
       do init
   unlock mutex
continue if not initialized
```
*Beware: Pseudo-code just a rough sketch to convey the idea - may not work exactly as given!*

runtime/net_ossl.c
```diff
@@ -0,0 +1,743 @@
...
+RSYSLOG_BIO_debug_callback_ex(BIO *bio, int cmd, const char __attribute__((unused)) *argp,
+			   size_t __attribute__((unused)) len, int argi, long __attribute__((unused)) argl,
+			   int ret, size_t __attribute__((unused)) *processed)
+#else
+static long
+RSYSLOG_BIO_debug_callback(BIO *bio, int cmd, const char __attribute__((unused)) *argp,
+			int argi, long __attribute__((unused)) argl, long ret)
+#endif
+{
+	long ret2 = ret; // Helper value to avoid printf compile errors long<>int
```
rgerhards: The comment and code makes little sense to me - why use %ld in the dbgprintf's below, when we have an int value (%d would work nicely)?

runtime/net_ossl.c
```diff
@@ -0,0 +1,743 @@
...
+	BIO_set_callback(conn, RSYSLOG_BIO_debug_callback);
+#endif
+}
+/* ------------------------------ End OpenSSL Callback set helpers -----------------------------*/
+
+
+/* Standard-Constructor */
+BEGINobjConstruct(net_ossl) /* be sure to specify the object type also in END macro! */
+	DBGPRINTF("net_ossl_construct: [%p]\n", pThis);
+//	iRet = nsd_ptcp.Construct(&pThis->pTcp);
```
rgerhards: remove commented-out code

runtime/net_ossl.c
```diff
@@ -0,0 +1,743 @@
...
+ */
+BEGINObjClassExit(net_ossl, OBJ_IS_CORE_MODULE) /* CHANGE class also in END MACRO! */
+CODESTARTObjClassExit(net_ossl)
+	DBGPRINTF("netosslClassExit\n");
+	/* release objects we no longer need */
+	objRelease(nsd_ptcp, LM_NSD_PTCP_FILENAME);
+	objRelease(glbl, CORE_COMPONENT);
+	/* shut down OpenSSL */
+	osslGlblExit();
+ENDObjClassExit(net_ossl)
```
rgerhards, line 1198: this is also a generic question: do we track the openssl users (imtcp, imdtls, ...) und make `osslGlblExit()` only do a real exit when the last users terminates?

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+#include "module-template.h"
+#include "glbl.h"
+#include "errmsg.h"
+#include "atomic.h"
+#include "statsobj.h"
+#include "unicode-helper.h"
+#include "datetime.h"
+#include "net_ossl.h"
+
+// --- Include openssl headers as well
```
rgerhards: the system headers should go before user (rsyslog) headers - just a best practices point

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+DEFobjCurrIf(glbl)
+DEFobjCurrIf(datetime)
+DEFobjCurrIf(statsobj)
+DEFobjCurrIf(net_ossl)
+
+statsobj_t *dtlsStats;
+STATSCOUNTER_DEF(ctrDtlsSubmit, mutCtrDtlsSubmit);
+STATSCOUNTER_DEF(ctrDtlsFail, mutCtrDtlsFail);
+
+#define DTLS_MAX_RCVBUF 1380	/* Maximum DTLS packet 1380 bytes to avoid fragmentation (smaller than the common
```
rgerhards: see releated comment(s) on packet sizes in imdtls.c

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+
+} wrkrInstanceData_t;
+
+#define INST_STATSCOUNTER_INC(inst, ctr, mut) \
+	do { \
+		if (inst->stats) { STATSCOUNTER_INC(ctr, mut); } \
+	} while(0);
+
+// QPID Proton Handler functions
+//static rsRetVal dtls_run_thread(wrkrInstanceData_t *pWrkrData);
```
rgerhards: remove commented-out code.
QPID looks like a left-over

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+ENDcheckCnf
+
+BEGINactivateCnfPrePrivDrop
+	instanceData *inst;
+CODESTARTactivateCnfPrePrivDrop
+	runModConf = pModConf;
+	DBGPRINTF("omdtls: activateCnfPrePrivDrop\n");
+
+	for(inst = runModConf->root ; inst != NULL ; inst = inst->next) {
+		// Init OpenSSL Context with DTLS_method
```
rgerhards: this comment looks somewhat unneeded ;-)

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+	pWrkrData->bIsConnecting = 0;
+	pWrkrData->bIsConnected = 0;
+	pWrkrData->bIsSuspended = 0;
+
+	CHKiRet(pthread_rwlock_init(&pWrkrData->pnLock, NULL));
+
+	pWrkrData->bThreadRunning = 0;
+	pWrkrData->tid = 0;
+
+	// Run Proton Worker Thread now
```
rgerhards: * remove invalid comment
* remove commented out code

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+
+	free(pData->tplName);
+	free(pData->statsName);
+ENDfreeInstance
+
+BEGINfreeWrkrInstance
+CODESTARTfreeWrkrInstance
+	DBGPRINTF("freeWrkrInstance[%p]: ENTER\n", pWrkrData);
+
+	/* Closing azure first! */
```
rgerhards: remove invalid comment

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+	pthread_rwlock_wrlock(&pWrkrData->pnLock);
+
+	// Close Proton Connection
+	dtls_close(pWrkrData);
+
+	// Stop Proton Handle Thread
```
rgerhards: remove invalid comment/code

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+ENDfreeWrkrInstance
+
+
+BEGINdbgPrintInstInfo
+CODESTARTdbgPrintInstInfo
+ENDdbgPrintInstInfo
+
+BEGINtryResume
+CODESTARTtryResume
+#ifndef NDEBUG
```
rgerhards: I would tend to do the dbgprintf also in production builds. Makes troubleshooting easier

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+*/
+BEGINcommitTransaction
+	unsigned i;
+	sbool bDone = 0;
+CODESTARTcommitTransaction
+#ifndef NDEBUG
+	DBGPRINTF("omdtls[%p]: commitTransaction [%d msgs] ENTER\n", pWrkrData, nParams);
+#endif
+	do {
+		// Put unsubmitted messages into Proton
```
rgerhards: remove invalid comment

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+		for(i = 0 ; i < nParams ; ++i) {
+			// Send Message to Proton
```
rgerhards: remove invalid comment

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+			CHKiRet(dtls_send(pWrkrData, pParams, i));
+		}
+		bDone = 1;
+	} while (bDone == 0);
+finalize_it:
+	/* TODO: Suspend Action if broker problems were reported in error callback */
```
rgerhards: remove invalid comment

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+	if (pWrkrData->bIsSuspended) {
+		DBGPRINTF("omdtls[%p]: commitTransaction failed to send messages, suspending action\n",
+			pWrkrData);
+		iRet = RS_RET_SUSPENDED;
+	}
+	if(iRet != RS_RET_OK) {
+		DBGPRINTF("omdtls[%p]: commitTransaction failed with status %d\n", pWrkrData, iRet);
+// TODO CHECK!
```
rgerhards: is this TODO valid?

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+	CHKiRet(OMSRsetEntry(*ppOMSR, 0, (uchar*)strdup((pData->tplName == NULL) ?
+						"RSYSLOG_FileFormat" : (char*)pData->tplName),
+						OMSR_NO_RQD_TPL_OPTS));
+
+	if (pData->statsName) {
+		CHKiRet(statsobj.Construct(&pData->stats));
+		CHKiRet(statsobj.SetName(pData->stats, (uchar *)pData->statsName));
+		CHKiRet(statsobj.SetOrigin(pData->stats, (uchar *)"omdtls"));
+
+		/* Track following stats */
```
rgerhards, line 507: It looks like these counters will never be incremented. That needs to be added.

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+CODEqueryEtryPt_STD_CONF2_QUERIES
+CODEqueryEtryPt_STD_CONF2_setModCnf_QUERIES
+CODEqueryEtryPt_STD_CONF2_PREPRIVDROP_QUERIES
+CODEqueryEtryPt_STD_CONF2_OMOD_QUERIES
+ENDqueryEtryPt
+
+
+BEGINmodInit()
+CODESTARTmodInit
+//	uchar *pTmp;
```
rgerhards: remove commented out code

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+{
+	DEFiRet;
+	int iErr, sslerr;
+	size_t lenSent = -1;
+	instanceData *const pData = (instanceData *const) pWrkrData->pData;
+
+	// Get Msg String
+	const char* pszParamStr = (const char*)actParam(pParam, 1 /*pData->iNumTpls*/, iMsg, 0).param;
+	size_t tzParamStrLen = actParam(pParam, 1 /*pData->iNumTpls*/, iMsg, 0).lenStr;
+	DBGPRINTF("dtls_send[%p]: writeProton for msg %d msg:'%.*s%s'\n",
```
rgerhards: invalid "proton" !

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+		(tzParamStrLen > 64 ? "..." : ""));
+
+	iErr = SSL_write(pWrkrData->sslClient, pszParamStr, tzParamStrLen);
+	if(iErr > 0) {
+		DBGPRINTF("dtls_send[%p]: Successfully send message '%s' with %ld bytes to %s:%s\n",
+			pWrkrData, pszParamStr, tzParamStrLen, pData->target, pData->port);
+	} else {
+		sslerr = SSL_get_error(pWrkrData->sslClient, iErr);
+		if (sslerr == SSL_ERROR_SYSCALL) {
+			dbgprintf("dtls_send[%p]: SSL_write failed with SSL_ERROR_SYSCALL(%s)"
```
rgerhards, line 599: q: would it make sense to output these error messages also to the user? I mean, could it be a useful trobuleshooting/operational monitoring aid? Or does this happen frequently and does not indicate a a problem (actually does not look so to me).

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+		}
+	}
+finalize_it:
+	DBGPRINTF("dtls_send[%p]: Exit with %d for %s:%s\n", pWrkrData, iRet, pData->target, pData->port);
+	if(iRet != RS_RET_OK) {
+		dtls_close(pWrkrData);
+	}
+	RETiRet;
+
+	// Return sent
```
rgerhards: remove comment

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+
+	// Create new SSL 
+	pWrkrData->sslClient = SSL_new(pData->pNetOssl->ctx);
+	if(!pWrkrData->sslClient) {
+		dbgprintf("dtls_connect[%p]: SSL_new failed failed\n", pWrkrData);
+		net_ossl_lastOpenSSLErrorMsg(pData->target, 0, pWrkrData->sslClient, LOG_WARNING, "omdtls", "SSL_new");
+		ABORT_FINALIZE(RS_RET_ERR);
+	}
+
+// MOVE TO SETUP
```
rgerhards: valid TODO item?

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+			"dtls_connect", "BIO_new_dgram");
+		ABORT_FINALIZE(RS_RET_ERR);
+	}
+	BIO_ctrl(bio_client, BIO_CTRL_DGRAM_SET_CONNECTED, 0, &pWrkrData->dtls_client_addr);
+	SSL_set_bio(pWrkrData->sslClient, bio_client, bio_client);
+
+	/* Set debug Callback for conn BIO as well! */
+	net_ossl_set_bio_callback(bio_client);
+
+	/* Blocking socket */
```
rgerhards: remove commented-out code

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+			"dtls_connect", "SSL_connect");
+		ABORT_FINALIZE(RS_RET_ERR);
+	}
+
+	// Print Cipher info
+	sslCipher = SSL_get_current_cipher(pWrkrData->sslClient);
+	dbgprintf("dtls_connect[%p]: Cipher Version: %s Name: %s\n", pWrkrData,
+		SSL_CIPHER_get_version(sslCipher), SSL_CIPHER_get_name(sslCipher));
+/*
+	// Print Peer Certificate info
```
rgerhards: remove commented-out code

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+dtls_init(wrkrInstanceData_t *pWrkrData)
+{
+	DEFiRet;
+	int iErr;
+	instanceData *const pData = (instanceData *const) pWrkrData->pData;
+	struct addrinfo *addrResolved = NULL;
+	struct addrinfo hints;
+	DBGPRINTF("dtls_init[%p]: setup for %s:%s\n", pWrkrData, pData->target, pData->port);
+
+	// Close Before
```
rgerhards: it would be good to enhance the comment so that one knows why a close is necessary. This check in init looks a bit strange, but I guess there is a good reason for it.

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+		ABORT_FINALIZE(RS_RET_ADDRESS_UNKNOWN);
+	}
+	pWrkrData->dtls_rcvr_addrinfo = addrResolved;
+	addrResolved = NULL;
+
+	// Init Socket Connection (Which technically does not connect but prepares socket for DTLS)
+	DBGPRINTF("dtls_init[%p]: Init Session to %s:%s\n", pWrkrData, pData->target, pData->port);
+
+	pWrkrData->sockout = socket(AF_INET, SOCK_DGRAM, 0);
+	// Connect the UDP socket to the server's address
```
rgerhards: reconsider comment? We are acting as client here, not as server. Is the comment or the code wrong?

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+			"dtls_init[%p]: Failed to connect to hostname '%s':'%s': %s",
+			pWrkrData, pData->target, pData->port, gai_strerror(iErr));
+		ABORT_FINALIZE(RS_RET_ERR);
+	}
+	
+	// Socket Connect successfull, no continue with TLS
+	CHKiRet(dtls_connect(pWrkrData));
+
+	// We reached this position means we are ready to send!
+	pWrkrData->bIsConnecting = 0;
```
rgerhards: do we need these two state variables?

I mean could we replace it wiht a single one with multiple values:

* disconnected
* connecting
* connected

I guess that these exact states must be gone through.

IIRC correctly, I suggested a similar replacement in some azure module, and you agreed to and implemented it.

plugins/omdtls/omdtls.c
```diff
@@ -0,0 +1,885 @@
...
+			iErr = SSL_shutdown(pWrkrData->sslClient);
+			if (iErr <= 0){
+				/* Shutdown not finished, call SSL_read to do a bidirectional shutdown, see doc:
+				*	https://www.openssl.org/docs/man1.1.1/man3/SSL_shutdown.html
+				*/
+				char rcvBuf[DTLS_MAX_RCVBUF];
+				SSL_read(pWrkrData->sslClient, rcvBuf, DTLS_MAX_RCVBUF);
+			}
+			SSL_free(pWrkrData->sslClient);
+			pWrkrData->sslClient = NULL;
```
rgerhards, line 828: this looks like we need to loop and retry the shutdown?

davidelang:
On Tue, 28 Nov 2023, Rainer Gerhards wrote:

> I reviewed the most important files. While I found a lot, it's mostly nits (with comment issues being the top issues). However, there are some more substantial comments, but nothing really bad :-)
>
>> +#define MAX_DTLS_MSGSIZE 65536
> +#define DTLS_LISTEN_PORT "4433"
> +
> +
> +/* Module static data */
> +DEF_OMOD_STATIC_DATA
> +DEFobjCurrIf(glbl)
> +DEFobjCurrIf(datetime)
> +DEFobjCurrIf(prop)
> +DEFobjCurrIf(ruleset)
> +DEFobjCurrIf(statsobj)
> +DEFobjCurrIf(net_ossl)
> +
> +#define DTLS_MAX_RCVBUF 1380	/* Maximum DTLS packet 1380 bytes to avoid fragmentation (smaller than the common
> +				*  Ethernet MTU of 1,500 bytes to accommodate potential IP and UDP headers). */
> +// #define DTLS_MAX_RCVBUF 16 * 1024 + 1/* TLS RFC 8449: max size of buffer for message reception */
>
> remove comment -- or is there any reason we can go above 1380 bytes?
>
> Looking further down in buffer handling, I actually think larger messages are possible, especially if "jumbo frames" are enabled at the network layer (some background: https://www.networkworld.com/article/745164/mtu-size-issues.html)

fragmented UDP packets are a thing, they make the messages far more fragile in 
congested enviornments, but they do work to much larger messages, even without 
jumbo frames.

not saying they are a good idea, but they can work.

if the size is set too large, I would suggest a warning on the topic, rather 
than just disallowing.

David Lang

plugins/imdtls/imdtls.c
alorbach: Did little research and it looks like that we go up to 8kb is safe for the DTLS Protocol if jumbo frames are enabled. Without jumbo frames, you may have fragmentation if the message exceeds the MTU. However I think this is acceptable, DTLS would not be a useful protocol to transmit large packages.

rgerhards: > DTLS would not be a useful protocol to transmit large packages.

fully agree. But folks may still use it. That's especially a concern with mixed senders, where one (non-rsyslog) sender may use them.

plugins/imdtls/imdtls.c
alorbach: removed

plugins/imdtls/imdtls.c
alorbach: Fixed

plugins/imdtls/imdtls.c
alorbach: Indeed, will fix

plugins/imdtls/imdtls.c
alorbach: removed

plugins/imdtls/imdtls.c
alorbach: not needed, done in net_ossl code
will remove

plugins/imdtls/imdtls.c
alorbach: not needed, will delete

plugins/imdtls/imdtls.c
alorbach: will do so

plugins/imdtls/imdtls.c
alorbach: is needed for DTLS Session handling

runtime/net_ossl.c
alorbach: will do so

runtime/net_ossl.c
alorbach: this is primary for the OpenSSL library init regardless of the ContextTypes later used (TLS / DTLS), so it should only be done once.

runtime/net_ossl.c
alorbach: it is used inside openssl (see CRYPTO_set_locking_callback(locking_function). I would not touch this, It was originally copied from OpenSSL sample code and is needed for multithreading.

runtime/net_ossl.c
alorbach: removed comment. %ld is needed because ret2 is long and needs to be because of the callback definition,

runtime/net_ossl.c
alorbach: removed

runtime/net_ossl.c
alorbach: It was just moved from nsd_ossl.c here: https://github.com/rsyslog/rsyslog/blob/84539a71a6b77ebe4787f6b83b9cbb3601544b17/runtime/nsd_ossl.c#L2331 to net_ossl. It should work the same as before.

plugins/omdtls/omdtls.c
alorbach: fixed

plugins/omdtls/omdtls.c
alorbach: also fixed

plugins/omdtls/omdtls.c
alorbach: removed

plugins/omdtls/omdtls.c
alorbach: removed

plugins/omdtls/omdtls.c
alorbach: removed

plugins/omdtls/omdtls.c
alorbach: removed

plugins/omdtls/omdtls.c
alorbach: removed

plugins/omdtls/omdtls.c
alorbach: I agree but in this case

plugins/omdtls/omdtls.c
alorbach: removed

plugins/omdtls/omdtls.c
alorbach: removed

plugins/omdtls/omdtls.c
alorbach: invalid, removed.

plugins/omdtls/omdtls.c
alorbach: Indeed, needs to be added.

plugins/omdtls/omdtls.c
alorbach: removed

plugins/omdtls/omdtls.c
alorbach: fixed

plugins/omdtls/omdtls.c
alorbach: LogMsg is used in line below inside net_ossl_lastOpenSSLErrorMsg function.

plugins/omdtls/omdtls.c
alorbach: removed

plugins/omdtls/omdtls.c
alorbach: real one, done - but comment is invalid - core has to be in connect

plugins/omdtls/omdtls.c
alorbach: done

plugins/omdtls/omdtls.c
alorbach: done, but not removed - useful for debugging

plugins/omdtls/omdtls.c
alorbach: dtls_init is already called if connect is necessar, in case of previopus connect, dtls_close will close ressources if necessary,

plugins/omdtls/omdtls.c
alorbach: Actually this is not true. As we are in connectionless UDP mode, we need an incoming and outgoing UDP Socket for sending and receiving. The handling is done inside the OpenSSL library.

plugins/omdtls/omdtls.c
alorbach: Sure we can change intro a single variable with state values.

plugins/omdtls/omdtls.c
alorbach: We do a single SSL_read which should be fine for protocol handling if necessary. We do not expect data to be received from the server.

plugins/imdtls/imdtls.c
rgerhards: so you need to "comment it in"?

runtime/net_ossl.c
rgerhards: I guess if we have e.g. imtcp, omfwd, imdtls and omdtls running on a system, we may get at least 4 concurrent calls into that code? If so, it's a long-standing issue, it just boils up in this PR.

runtime/net_ossl.c
rgerhards: I am not convinced yet, but let's take up the discussion in a later state again.

runtime/net_ossl.c
rgerhards: and maybe before it was already incorrect (see related comments). Let's discuss this later in depth.

plugins/omdtls/omdtls.c
rgerhards: what's special in this case?

plugins/omdtls/omdtls.c
rgerhards: shouldn't we then just guard it by an if and enable if debugging is turned on? Makes much more sense to me. Once rsyslog packages are deployed, you can no longer change compile time settings at all ;-)

plugins/omdtls/omdtls.c
rgerhards: I admit I do not understand your comment. Do you mean to say the comment  should actually read "this has already been closed by the caller"?

plugins/omdtls/omdtls.c
rgerhards: The word "server" trigerred me. It has the co-notation of being used only in the server case. I strongly suggest to connect that comment with what you wrote here in the conversation. That's a much better and useful description. And it's not obvious if you only read the code.

plugins/omdtls/omdtls.c
rgerhards: > Sure we can change intro a single variable with state values.

We should definitely do that - it's much cleaner and also will prevent hard-to-find errors if the current two vars for some reason get out of sync (by later maintenance).

plugins/omdtls/omdtls.c
rgerhards: *This* is worth a comment, as it is not obvious.

plugins/omdtls/omdtls.c
alorbach: Alright I will change the comment

plugins/omdtls/omdtls.c
alorbach: Will amend to the existing comment

plugins/imdtls/imdtls.c
alorbach: Implemented poll() conditionally with 
ifdef HAVE_EPOLL_CREATE

plugins/omdtls/omdtls.c
alorbach: the comment is little missleading. I can change it to "Close previous ressources if any left open"

plugins/omdtls/omdtls.c
alorbach: yes this makes sense, will add if(Debug) {}

plugins/imdtls/imdtls.c
alorbach: my bad, misunderstood the comment. Will remove comment code.

runtime/net_ossl.c
alorbach: You mean we could initialize the library multiple times at the same time?

runtime/net_ossl.c
alorbach: I get it, we should create a separated issue for that.

runtime/net_ossl.c
alorbach: indeed, I would put it into the same issue as the init mutex one.

plugins/omdtls/omdtls.c
alorbach: Sry, I meant: I agree in this case

plugins/omdtls/omdtls.c
alorbach: removed

rgerhards:
Made announcement of this effort: https://www.rsyslog.com/elevating-syslog-security-rsyslog-introduces-dtls-plugins-for-udp/

davidelang:
I haven't dug through things yet, but given that out-of-order recovery and fragmentation handling can be used for DOS purposes (valid handshake, then start sending everything but the first packet, per the dtls rfc the receiver it supposed to hang on to the later packets waiting for the first one)

there should be configuration for timeouts, but for retransmission and also for throwing away later packets in the sequence.

alorbach:
> I haven't dug through things yet, but given that out-of-order recovery and fragmentation handling can be used for DOS purposes (valid handshake, then start sending everything but the first packet, per the dtls rfc the receiver it supposed to hang on to the later packets waiting for the first one)
> 
> there should be configuration for timeouts, but for retransmission and also for throwing away later packets in the sequence.

Good point. I will add timeout handling to imdtls module. I would suggest a default timeout of 30 minutes for a DTLS session, do you agree?

alorbach:
> This is less about timing out an entire session than timing out individual chunks. the rfc suggests a 1s timeout for lost chunks (before you are supposed to retransmit them) but says nothing about timing out n+x chunks when you haven't received chunk N yet (it says if you receive future chunks you are supposed to hold on to them and process them after you receive chunk N). If you have a proper sender who will retry chunk N quickly, this isn't bad (and avoids having to retransmit all newer chunks because one arrives out of order), but if you have a malicous sender you need to have a short timeout (single digit seconds) that you hang on to these out-of-order chunks before you throw them away and force the sender to retransmit them David Lang
> […](#)
> On Thu, 11 Jan 2024, Andre Lorbach wrote: > I haven't dug through things yet, but given that out-of-order recovery and fragmentation handling can be used for DOS purposes (valid handshake, then start sending everything but the first packet, per the dtls rfc the receiver it supposed to hang on to the later packets waiting for the first one) > > there should be configuration for timeouts, but for retransmission and also for throwing away later packets in the sequence. Good point. I will add timeout handling to imdtls module. I would suggest a default timeout of 30 minutes for a DTLS session, do you agree?

I think this is protocol handling and it is done inside OpenSSL. And I assume it should be properly handled inside OpenSSL. We use the DTLS_method() context and the same OpenSSL API's to receive by DTLS as we use in the nsd_ossl.c code (SSL_read).

rgerhards:
> I think this is protocol handling and it is done inside OpenSSL. And I assume it should be properly handled inside OpenSSL.

I would check if openSSL accepts config parameters for the timeouts @davidelang mentioned. Actually I would assume that such params exist.

PR #5335, opened 2024-02-23
imfile - proposed fix for file handle leaks

n2yen:
- FIle handler leaks can occur if the deleted file still has a lingering symbolic link (pointing to it). Reason is, even if the file has actually been deleted, `fstat` will return successfully, and thus this block never gets a chance to release the file handle.


n2yen:
hi @rgerhards ,
Wonder if this is one of the possible edge cases discussed: https://github.com/rsyslog/rsyslog/blob/4c11c58b519b7da8b36177e506165026c26aaa71/plugins/imfile/imfile.c#L840-L842

mind having a look when you get a chance? thank you!

rgerhards:
code factor is false positive

PR #5341, opened 2024-02-28
MMDBLOOKUP::FIXED:: Don't suspend action on incompatible IP lookups

frikilax:
Mmdblookup module used to suspend after erroring on an IPv6 IP lookup on IPv4-only DBs
This commit replaces the suspension of the module by a simple log message, allowing it to keep working for future lookups


rgerhards:

PR #5366, opened 2024-04-19
Do not free the uninitialized cstring.

lonicerae:
* Better deal with corrupted queue messages

### DIFF
--- a/runtime/stringbuf.c
+++ b/runtime/stringbuf.c
@@ -219,7 +219,7 @@ finalize_it:
 
 void rsCStrDestruct(cstr_t **const ppThis)
 {
-	free((*ppThis)->pBuf);
+	if ((*ppThis)->pBuf) free((*ppThis)->pBuf);
 	RSFREEOBJ(*ppThis);
 	*ppThis = NULL;
 }
--- a/runtime/stringbuf.c
+++ b/runtime/stringbuf.c
@@ -219,7 +219,7 @@ finalize_it:
 
 void rsCStrDestruct(cstr_t **const ppThis)
 {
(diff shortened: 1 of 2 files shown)


runtime/stringbuf.c
```diff
@@ -219,7 +219,7 @@ rsCStrConstructFromCStr(cstr_t **const ppThis, const cstr_t *const pFrom)
 
 void rsCStrDestruct(cstr_t **const ppThis)
 {
-	free((*ppThis)->pBuf);
+	if ((*ppThis)->pBuf) free((*ppThis)->pBuf);
```
Cropi: I think there is no need to check the condition. If `(*ppThis)->pBuf` is `NULL`, then no operation is performed by the `free()` function as per `man 3 free`.

Cropi:
I do confirm that the patch fixes the issue. Another approach could be to change `step <= 2` to `step <= 3` in `objDeserializeProperty`:
```
if(iRet != RS_RET_OK && iRet != RS_RET_NO_PROPLINE) {
  if(step <= 2) {
	  pProp->varType = VARTYPE_NONE;
  }
}
```

One thing that we should also investigate is how the queue got malformed. Is there a possibility that there were 2 concurrent writes to the same queue file?

runtime/stringbuf.c
lonicerae: Ah, yes. Let's keep it unchanged.

Cropi (approved):

Cropi:

rgerhards:
> Can we add this to the next milestone? Thanks. @rgerhards

thx, I seem to have overlooked this

PR #5377, opened 2024-05-08
Fix passing the value to EscapeChar.

lonicerae:
* Properly handle the EscapeChar after refactor.

### DIFF
--- a/runtime/cfsysline.c
+++ b/runtime/cfsysline.c
@@ -59,7 +59,7 @@ linkedList_t llCmdList; /* this is NOT a pointer - no typo here ;) */
  * HINT: check if char is ' and, if so, use 'c' where c may also be things
  * like \t etc.
  */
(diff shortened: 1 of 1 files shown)


rgerhards:
do you have a sample conf that trigges the behavior?

Cropi:
I see that in `doGetChar`, the handler function `setParserControlCharacterEscapePrefix` is called with the appropriate escape character, which is type of `uchar`. However, `setParserControlCharacterEscapePrefix` expects it to be `uchar *`, that's where the segfault happens.

I would rather use a different type than `uid_t`, as it's usually used to hold a used ID. Ideally a replacement could be `uchar`.

Cropi (approved):
LGTM!
I tried out the both `parser.controlcharacterescapeprefix` and `$controlcharacterescapeprefix` versions, it works. Thanks for the patch.

Cropi:

rgerhards:
thx for the reminder AND the PR. Now merged.

PR #5381, opened 2024-05-14
Refactor the rscryutil tool

Cropi:
The `rscryutil` tool is fully dependent on the `libgcrypt` library. We need to refactor it in case there will ever be a new crypto provider added. I've developed a patch introducing support for an alternative crypto provider utilizing symmetric encryption from the OpenSSL library. This addition not only maintains the existing functionality but also enhances it with additional features, as outlined in my discussion on the mailing list. I'll be submitting another pull request that incorporates this new cryptographic provider library.

This PR does not change the functionality at all.

### DIFF
--- a/runtime/libgcry_common.c
+++ b/runtime/libgcry_common.c
@@ -25,7 +25,6 @@
 #include "config.h"
 #endif
 #include <stdio.h>
-#include <gcrypt.h>
 #include <sys/stat.h>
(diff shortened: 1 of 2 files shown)


Cropi:
I am closing this PR as the changes are also available in https://github.com/rsyslog/rsyslog/pull/5382 . Both PRs are tightly connected and it will be more visible in a single one.

PR #5382, opened 2024-05-16
New crypto provider via OpenSSL library

Cropi:
This new `ossl` crypto provider encapsulates the ability to use log encryption in the `omfile` module. It provides the same functionality as `gcry`, additionaly:
- Algorithm are not hardcoded. We simply provide everything what can be fetched via the `EVP_CIPHER_fetch()` call. The same applies to the mode parameter.
- It does not add extra dependency to a new package. Distro maintainers who are already using openssl do not need add anything in particular. By default, if rsyslog is compiled with `--enable-openssl`, the ossl crypto provider will also be included in both `omfile` and also `rscryutil`.
- Side-channel vulnerability in libgcrypt - the Marvin Attack
   https://lists.gnupg.org/pipermail/gcrypt-devel/2024-March/005607.html (see the whole thread)
   Some might be worried about other side-channel vulnerabilities against it, like the
   Minerva Attack[1], and the Raccoon Attack[2], given the way they  handled one of the oldest, well known, and
   most straightforward side-channel attacks. They did downgrade their threat model as a result of Marvin too:  https://gnupg.org/documentation/security.html. I don't see it as a proper way to deal with potential threats.

[1] https://minerva.crocs.fi.muni.cz/
[2] https://raccoon-attack.com/

Default behavior is preserved. I've tested a few different scenarios and the two crypto providers can work in tandem, but I don't suggest to mix them.
In order to put it into action, add `cry.provider="ossl"` to an omfile action. Generate some syslog messages, and finally use the the `rscryutil` with `--lib=ossl`. The only difference if altering the algorithm/mode then you must do it in one step, e.g.:
- `./tools/rscryutil -k keyfile -l ossl -a AES-256-CBC encryptedfile` for ossl
- `./tools/rscryutil -k keyfile -l gcry -a AES256 -m CBC encryptedfile` for gcry

The documentation is currently missing, I will create it if the patch is accepted. Thanks.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1126,6 +1126,15 @@ if test "x$enable_openssl" = "xyes"; then
 fi
 AM_CONDITIONAL(ENABLE_OPENSSL, test x$enable_openssl = xyes)
 
+enable_openssl_crypto_provider="no";
+if test "x$enable_openssl" = "xyes"; then
+	AC_CHECK_LIB(
+		[crypto],
+		[EVP_CIPHER_get_block_size],
+		[enable_openssl_crypto_provider="yes"],
+		[AC_MSG_WARN([The ossl encryption provider will not be enabled (EVP_CIPHER_get_block_size is missing)])]
+	)
+fi
 
 # GnuTLS support
 AC_ARG_ENABLE(gnutls,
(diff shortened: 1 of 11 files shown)


Cropi:
Hello @rgerhards , is there anything I can do to move this PR forward?

rgerhards:
> Hello @rgerhards , is there anything I can do to move this PR forward?

As I wrote in a (related?) thread, I was busy, and things get better this week. The PR is targeted for next Tuesday scheduled stable and you should now see it moving forward. Again, this applies to other stalled-looking PRs.

BTW: Thanks for the fine PR.

rgerhards:
While I am at it and now having done any real review: do we have a doc PR for the new provider? If not, that would be a blocker.

rgerhards (approved):

rgerhards:
The PR looks good, but indeed the doc is currently the blocker. So please go-ahead and create it. If done by the weekend, I can merge into v8.2408.0 scheduled stable.

Cropi:
Thanks for the quick reaction and for the feedback. I created https://github.com/rsyslog/rsyslog-doc/pull/1060 to cover the missing documentation.

> The PR looks good, but indeed the doc is currently the blocker. So please go-ahead and create it. If done by the weekend, I can merge into v8.2408.0 scheduled stable.

rgerhards:
The current CI failures are known (and being worked-on) environment issues. So save to merge.

rgerhards:
@Cropi thx again for the PR and your patience!

Issue #5390, opened 2024-05-28
The rsyslogd process fails to be started repeatedly


HappyDrink-okk:
### Expected behavior
rsyslogd process start nomally
### Actual behavior
rsyslogd process fails to be started repeatedly, see this log output
```
starting pid 16174, tty '': '/usr/sbin/rsyslogd'
rsyslogd: error reading pid file, cannot start up
rsyslogd: run failed with error -3000 (see rsyslog.h or try https://www.rsyslog.com/e/3000 to learn what that number means)
svrmonit:"/usr/sbin/rsyslogd" not find right pid!
process '/usr/sbin/rsyslogd' (pid 16174) exited. Scheduling for restart.
starting pid 16182, tty '': '/usr/sbin/rsyslogd'
rsyslogd: error reading pid file, cannot start up
rsyslogd: run failed with error -3000 (see rsyslog.h or try https://www.rsyslog.com/e/3000 to learn what that number means)
svrmonit:"/usr/sbin/rsyslogd" not find right pid!
process '/usr/sbin/rsyslogd' (pid 16182) exited. Scheduling for restart.
starting pid 16183, tty '': '/usr/sbin/rsyslogd'
rsyslogd: error reading pid file, cannot start up
rsyslogd: run failed with error -3000 (see rsyslog.h or try https://www.rsyslog.com/e/3000 to learn what that number means)
svrmonit:"/usr/sbin/rsyslogd" not find right pid!
...(repeat the above printing all the time)
```
### Steps to reproduce the behavior
1. configuring rsyslogd in /etc/inittab
```
# cat /etc/inittab
::respawn:/usr/sbin/rsyslogd
```
2. use this shell script, and then problem recurs
```
#!/bin/bash

killall /usr/sbin/rsyslogd

NOW_SIZE=$(df /var | grep "var" | awk -F ' ' '{print $2}')
NOW_USED=$(df /var | grep "var" | awk -F ' ' '{print $3}')
REMOUNT_SIZE=$((${NOW_USED} + 1024))

mount -o remount,size=${REMOUNT_SIZE}K /var

dd if=/dev/urandom of=/var/test1 bs=${REMOUNT_SIZE}K count=1

sleep 2

mount -o remount,size=${NOW_SIZE}K /var
rm /var/test1
killall /usr/sbin/rsyslogd

```
### Environment
- rsyslog version: 8.2110.0
```
rsyslogd  8.2110.0 (aka 2021.10) compiled with:
[...]

PR #5391, opened 2024-05-28
Use file lock instead of rename to handle concurrency

HappyDrink-okk:
While starting up the rsyslogd process, executing remount /var may cause the /var/run/rsyslogd.pid file to become an empty file. Using file locks instead of the original rename solution can solve this problem.

see issue#5390


HappyDrink-okk:
#5390

rgerhards:
Doesn't the lock prevent unmount of the file system? I need to dig into the code, but there was a reason for doing it the way we do it currently.  We need to check commit history.

HappyDrink-okk:
> Doesn't the lock prevent unmount of the file system? I need to dig into the code, but there was a reason for doing it the way we do it currently. We need to check commit history.

I used this and the problem went away, but I don't know why the original code caused the pid file to become empty.

rgerhards:
/gemini review - espeially: can this prevent volume unmount?

rgerhards:
@HappyDrink-okk Sorry for the long silence, but I finally have taken the opportunity to review this in detail. Even with the patch, an empty pid file can happen (system error etc) - this is not primarily a concurrency bug. From that point on, I have crafted a different patch which tries to make the write ultra-reliable and aborts rsyslog if the pid file cannot be written. This is probably the best action to be done if a pid file is really needed (it often is no longer). This is a change of behaviour, but I think an acceptable one. The new patch will be provided as PR very soon.

rgerhards:
replaced by https://github.com/rsyslog/rsyslog/pull/6462

PR #5415, opened 2024-07-18
ossl: Fix #pragma GCC diagnostic not allowed inside functions

julthomas:
Some old compilers, eg. GCC 4.4.7 on el6 do not support #pragma GCC diagnostic in functions. Rsyslog has PRAGMA_* macros to handle those cases, so let's use them.

### DIFF
--- a/runtime/net_ossl.c
+++ b/runtime/net_ossl.c
@@ -220,8 +220,8 @@ osslGlblInit(void)
 	ERR_load_crypto_strings();
 #endif
 
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+PRAGMA_DIAGNOSTIC_PUSH
+PRAGMA_IGNORE_Wdeprecated_declarations
 
 	// Initialize OpenSSL engine library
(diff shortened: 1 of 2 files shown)


Cropi:
Hello,

GCC 4.4.7 was released 10+ years ago.
Despite the significant efforts by rsyslog developers to ensure backwards compatibility, I believe it is not wise to implement changes for such an outdated version of GCC.

julthomas:
@Cropi You are not wrong, tell me if you want me to close the pull request. Note there are places where rsyslog PRAGMA_* are used but not in that new code. Just for consistency / uniformization accross the project, you should consider using the macros defined by rsyslog.

rgerhards:
I think it is OK to add this support. These versions are still out there and actually the macros *should* be used and may be of value in the future as well.

I am currently unfortunately not able to review in depth and merge some PRs, but will be so around the mid of August when I craft a new (and exciting, let me say!) scheduled stable release.

rgerhards (approved):

rgerhards:
current CI failures are known issues inside CI, which are being worked on. So save to merge.

PR #5433, opened 2024-08-13
Adjust documentation regarding rscryutil

Cropi:
Add documentation regarding rscryutil changes as well.

In a previous PR, I made a mistake (sorry), I accidentally added an auto-generated file into the repository. This PR intends to remove unnecessary `tools/rscryutil.1` files


Cropi:
The CI looks much better after the rebase. Only one builder got timed out

rgerhards:
The solaris build worker is thankfully provided by OpenCSW. They host quite some projects on these machines and sometimes they are quite overloaded. I guess that was the case, I have re-initiated the run.

rgerhards (changes requested):
I notice that rscryutil is still present. I think you must explicitely purge it from the commit, if it already is part of it.

Cropi:
> I notice that rscryutil is still present. I think you must explicitely purge it from the commit, if it already is part of it.

I am not able to see it. The commit shows that the file has been deleted. This is about removing `tools/rscryutil.1`, which is basically an auto-generated doc from `tools/rscryutil.rst` . Maybe I overlooked something.

rgerhards:
> Maybe I overlooked something.

no, you are right. I am doing too much in parallel ;-)

rgerhards (approved):

PR #5434, opened 2024-08-13
Refactor ossl crypto provider

Cropi:
Very simple refactoring a use case, where `--enable-openssl` is defined but `--enable-libgcrypt` is not:

- Move common defines to a common file that is used by both `ossl` and `gcry`. I haven't realized it sooner, but some defines were put into a wrong file and if rsyslog was compiled without `gcry` support then those specific defines used by `ossl` were not present.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1131,7 +1131,9 @@ if test "x$enable_openssl" = "xyes"; then
 	AC_CHECK_LIB(
 		[crypto],
 		[EVP_CIPHER_get_block_size],
-		[enable_openssl_crypto_provider="yes"],
+		[enable_openssl_crypto_provider="yes"
+		AC_DEFINE([ENABLE_OPENSSL_CRYPTO_PROVIDER], [1], [Indicator that openssl(EVP_CIPHER_get_block_size) is present])
+		],
 		[AC_MSG_WARN([The ossl encryption provider will not be enabled (EVP_CIPHER_get_block_size is missing)])]
 	)
 fi
--- a/runtime/libcry_common.h
+++ b/runtime/libcry_common.h
@@ -22,6 +22,15 @@
 #define INCLUDED_LIBCRY_COMMON_H
 #include <stdint.h>
 
+/* error states */
+#define RSGCRYE_EI_OPEN 1 /* error opening .encinfo file */
+#define RSGCRYE_OOM 4 /* ran out of memory */
+
+#define EIF_MAX_RECTYPE_LEN 31 /* max length of record types */
+#define EIF_MAX_VALUE_LEN 1023 /* max length of value types */
+#define RSGCRY_FILETYPE_NAME "rsyslog-enrcyption-info"
+#define ENCINFO_SUFFIX ".encinfo"
+
 int cryGetKeyFromFile(const char* const fn, char** const key, unsigned* const keylen);
 
 int cryGetKeyFromProg(char* cmd, char** key, unsigned* keylen);
--- a/runtime/libgcry.c
+++ b/runtime/libgcry.c
@@ -52,6 +52,7 @@
 #include "srUtils.h"
 #include "debug.h"
 #include "libgcry.h"
+#include "libcry_common.h"
 
 #define READBUF_SIZE 4096	/* size of the read buffer */
 
@@ -703,7 +704,7 @@ rsgcryEncrypt(gcryfile pF, uchar *buf, size_t *len)
 {
 	int gcryError;
 	DEFiRet;
-	
+
 	if(*len == 0)
 		FINALIZE;
 
@@ -729,7 +730,7 @@ rsgcryDecrypt(gcryfile pF, uchar *buf, size_t *len)
 {
 	gcry_error_t gcryError;
 	DEFiRet;
-	
+
 	if(pF->bytesToBlkEnd != -1)
 		pF->bytesToBlkEnd -= *len;
 	gcryError = gcry_cipher_decrypt(pF->chd, buf, *len, NULL, 0);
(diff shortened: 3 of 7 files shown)


rgerhards:
obviously, I overlooked this as well :-)

I'll merge this week. But maybe I wait a day or two in the hope the CI nit can be sorted out in the meantime.

rgerhards (changes requested):
The error here is valid:

https://github.com/rsyslog/rsyslog/actions/runs/10373039895/job/28717195683?pr=5434#step:3:1153

(The other ones are displayed, but are actually disabled via macros).

rgerhards:
That's strange... I do not see why the build now fails, but it fails - and without the patch it succeeds. @Cropi do you have an idea?

rgerhards:
> But I don't know why it fails in the PR and not in the previous ones (strange).

yeah, that's my point. I'll do a dummy PR to check. The fun fact is that the CI builders didn't change at all, too....

Cropi:
To me it seems that the pragmas used in the code are not respected at all:
In net_ossl.c: 223 we have
```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wdeprecated-declarations"

	// Initialize OpenSSL engine library
	ENGINE_load_builtin_engines();
...
```
But  the build fails with the following
```
net_ossl.c:227:2: error: 'ENGINE_load_builtin_engines' is deprecated [-Werror,-Wdeprecated-declarations]
        ENGINE_load_builtin_engines();
        ^
/usr/include/openssl/engine.h:358:1: note: 'ENGINE_load_builtin_engines' has been explicitly marked deprecated here
OSSL_DEPRECATEDIN_3_0 void ENGINE_load_builtin_engines(void);
```

rgerhards:
That's a good hint - I'll check. Interestingly my test PR shows the same symptoms, and it has this patch. Also, this one should have it, because github is configured to merge based on current master before running the tests.

As a side-note: your "upgrade" commit is also fine. Feel free to submit it as PR. But I'll look into the root cause of the current problem first.

rgerhards:
I found the culprit: https://github.com/rsyslog/rsyslog/pull/5437

I guess I have overlooked the problem during the merge of that PR (due to the currently existing other issues in CI). Once https://github.com/rsyslog/rsyslog/pull/5437 is merged, all should be well. I'll re-init builders after merge.

Thanks again for pointing out the macros!

Cropi:
> I found the culprit: #5437
> 
> I guess I have overlooked the problem during the merge of that PR (due to the currently existing other issues in CI). Once #5437 is merged, all should be well. I'll re-init builders after merge.
> 
> Thanks again for pointing out the macros!

That's great news! I did a rebase and moved the libfastjson stuff into a separate PR, https://github.com/rsyslog/rsyslog/pull/5436.

rgerhards:
The two solaris-builders are known issues, I have disabled them for now. So this is no merge stopper.

rgerhards (approved):

Cropi:
I gave a moment to troubleshoot `ubuntu_22_san` issue with the `omhttp-batch-retry-metadata.sh` test. The test results are not deterministic. Sometimes it passes, sometimes it fails. The `omhttp-validate-response.py` script tries to parse the json input line by line, but sometimes the message and response fields are on separate lines, thus it fails to parse.

```
{ "message": {"msgnum":"00038963"}, "response": { "code": 207, "body": "INTERNAL ERROR", "batch_index": 94 } }
{ "message": {"msgnum":"00038964"}, "response": { "code": 207, "body": "INTERNAL ERROR", "batch_index": 95 } }
{ "message": {"msgnum":"00038965"}, "response": { "code": 207, "body": "INTERNAL ERROR", "batch_index": 96 } }
{ "message": {"msgnum":"00038966"}, "response": { "code": 207, "body": "INTERNAL ERROR", "batch_index": 97 } }
{ "message": {"msgnum":"00038967"}, "response": { "code": 207, "body": "INTERNAL ERROR", "batch_index": 98 } }
{ "message": {"msgnum":"00038968"}, "response": { "code": 207, "body": "INTERNAL ERROR", "batch_index": 99 } }
{ "message": {"msgnum":"00029129"}
, "response": { "code": 207, "body": "INTERNAL ERROR", "batch_index": 0 } }
```

And it fails because `{ "message": {"msgnum":"00029129"}` is not a valid json:
```
msgnum: 00038968 code: 207 body: INTERNAL ERROR batch_index: 99
line:  { "message": {"msgnum":"00029129"}

Traceback (most recent call last):
  File "/home/alakatos/github/rsyslog/tests/./omhttp-validate-response.py", line 28, in <module>
    json_obj = json.loads(line)
               ^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/json/decoder.py", line 353, in raw_decode
    obj, end = self.scan_once(s, idx)
```

Whilst the test requires investigation, I think this is not connected with the patch proposed here.

rgerhards:
> Whilst the test requires investigation, I think this is not connected with the patch proposed here.

definitely - and thx for the analysis. I just began to look into it. It's contributed and I'll push fixing as a post-release step. At least we now know why it is flaky :-)

rgerhards:
The elasticsearch fail was actually caused by a VM which did no proper cleanup. So builder issue, only. This has been solved.

rgerhards:
Solaris, as I wrote, is currently dead, the env ceased to work. I asked for a re-push because there is no other way to get them out of the test results. But I'll merge in any case, it's just cosmetic. so no more need to re-push.

PR #5436, opened 2024-08-14
Replace deprecated json_object_object_get with fjson_object_object_get_ex

Cropi:


rgerhards:
The two solaris-builders are known issues, I have disabled them for now. So this is no merge stopper.

PR #5438, opened 2024-08-15
minor: flag function as OK for undefined behaviour

rgerhards:
The undefined bahavior sanitizer deteced an integer overflow inside some code. However, this is hash code and the interger overflow is perfectly fine.


rgerhards:
solaris 10 does not work ATM, so merge OK

Issue #5453, opened 2024-09-20
runConf NULL pointer refence when call systemd restart rsyslog in early startup of system.


apple-ouyang:
### Expected behavior
systemd restart rsyslog OK

### Actual behavior
systemd restart rsyslog let rsyslog segmentation fault.

### Steps to reproduce the behavior
when start of the OS, add systemd restart rsyslog in init script

### Environment
when we use rsyslog 8.2110.0, it is OK.
But when we upgrade to 8.2210, the bug come out.
```
rsyslogd  8.2110.0 (aka 2021.10) compiled with:
        PLATFORM:                               aarch64-euler-linux-gnu
        PLATFORM (lsb_release -d):              Description:    SUSE Linux Enterprise Server 12 SP5
        FEATURE_REGEXP:                         Yes
        GSSAPI Kerberos 5 support:              No
        FEATURE_DEBUG (debug build, slow code): No
        32bit Atomic operations supported:      Yes
        64bit Atomic operations supported:      No
        memory allocator:                       system default
        Runtime Instrumentation (slow code):    No
        uuid support:                           Yes
        systemd support:                        Yes
        Config file:                            /etc/rsyslog.conf
        PID file:                               /var/run/rsyslogd.pid
        Number of Bits in RainerScript integers: 64

See https://www.rsyslog.com for more information.
```

```
rsyslogd  8.2210.0 (aka 2022.10) compiled with:
        PLATFORM:                               arm-eulermllib32-linux-gnueabi
        PLATFORM (lsb_release -d):              Description:    SUSE Linux Enterprise Server 12 SP5
        FEATURE_REGEXP:                         Yes
        GSSAPI Kerberos 5 support:              No
        FEATURE_DEBUG (debug build, slow code): No
        32bit Atomic operations supported:      Yes
        64bit Atomic operations supported:      No
        memory allocator:                       system default
        Runtime Instrumentation (slow code):    No
        uuid support:                           Yes
        systemd support:                        Yes
        Config file:                            /etc/rsyslog.conf
        PID file:                               /var/run/rsyslogd.pid
        Number of Bits in RainerScript integers: 64

See https://www.rsyslog.com for more information.
```

apple-ouyang:
Here is the stack trace:
pc:0x0000aaaacdd4c8a4 (null)(rsyslogd+0x188a4)
lr:0x0000aaaacdd4c868 (null)(rsyslogd+0x18868)

---stack information begin---
[...]

PR #5454, opened 2024-09-20
Fix runConf NULL pointer refence

apple-ouyang:
`systemd restart rsyslog` in the early start of OS will let rsyslog segmentation fault. This cmd will send sigTerm to rsylogd, and rsyslogd will handle the signal in rsyslogdDoDie. If the rsyslogd havn't parse the conf, the runConf will be NULL So check the pointer before reference it.

#5453


rgerhards (approved):

rgerhards:
Thx for the PR and sry for the long time to merge.

PR #5458, opened 2024-09-25
rainerscript: do not try to call a function if it does not exist

Cropi:
Having a config snippet like the following causes **segmentation fault**, because the function can not be found.
```
if nonExistentFunction(1) then {
        stop
}
```
Output:
```
rsyslogd: version 8.2410.0.master, config validation run (level 1), master config /etc/rsyslog.conf
rsyslogd: error during parsing file /etc/rsyslog.d/crash.conf, on or before line 5: function 'nonExistentFunction' not found [v8.2410.0.master try https://www.rsyslog.com/e/2207 ]
rsyslogd: error during parsing file /etc/rsyslog.d/crash.conf, on or before line 5: Invalid function nonExistentFunction [v8.2410.0.master try https://www.rsyslog.com/e/2207 ]
[1]    123378 segmentation fault  ./tools/rsyslogd -N1
```

I intentionally removed the `rainerscript: internal error: NULL...` part because it was causing too much noise in the logs. If a function does not exist, it is reported to be missing when doing a config only check, so I think it's fine.
With patch applied, segfault does not happen on function call.

### DIFF
--- a/grammar/rainerscript.c
+++ b/grammar/rainerscript.c
@@ -2901,11 +2901,6 @@ doFuncCall(struct cnffunc *__restrict__ const func, struct svar *__restrict__ co
 		free(fname);
 	}
 	if(func->fPtr == NULL) {
-		char *fname = es_str2cstr(func->fname, NULL);
-		LogError(0, RS_RET_INTERNAL_ERROR,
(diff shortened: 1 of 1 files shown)


Cropi:

rgerhards (approved):

PR #5460, opened 2024-10-07
clean(imhiredis): Fix incorrect variable definitions and missing function prototype

frikilax:
Simple clean of imhiredis to remove some warnings given by GCC


rgerhards (approved):
sry for the late attention to this PR

Issue #5456, opened 2024-09-23
minitcpsrvr fails on mipsel: setsockopt failed: Protocol not available


mbiebl:
Version: 8.2408.0

Since 8.2408.0, `make check` started to fail on mips64el:
https://buildd.debian.org/status/logs.php?pkg=rsyslog&ver=8.2408.0-1&arch=mips64el

The issue is this:
```
 ./tests/minitcpsrv -p 8000 -t 127.0.0.1 -f out.file
writing to file out.file
setsockopt failed: Protocol not available
minitcpsrv ABORTS!!!
```

Afaics, this was introduced by 1c0f9bba505d32af8a0421b495027f726cb0aa4c

mbiebl:
From an IRC discussion on #debian-devel
```
mbiebl>
 any idea why this code fails on mips64el: https://salsa.debian.org/debian/rsyslog/-/blob/debian/master/tests/minitcpsrvr.c?ref_type=heads#L92

[tests/minitcpsrvr.c · debian/master · Debian / rsyslog · GitLab](https://salsa.debian.org/debian/rsyslog/-/blob/debian/master/tests/minitcpsrvr.c?ref_type=heads#L92) - GitLab
rsyslog packaging
https://buildd.debian.org/status/logs.php?pkg=rsyslog&ver=8.2408.0-1&arch=mips64el

[Build logs for rsyslog_8.2408.0-1 on mips64el](https://buildd.debian.org/status/logs.php?pkg=rsyslog&ver=8.2408.0-1&arch=mips64el)
Debian Package Auto-Building Build logs for rsyslog_8.2408.0-1 on mips64el Package: Tracker – Changelog – Bugs – packages.d.o – Source Build logs for rsyslog _8.2408.0-1

zeha>
i wonder if its supposed to work on other archs in the firstplace
i don't think SO_ constants are meant to be ORed together
the SO_... constants are arch-specific. looking at the defines for amd64 and mips, i think it just does something different by accident on the archs where it 'succeeds'
(but probably not what is intended)
mbiebl>
: ^
wRAR>
do you need two setsockopt calls instead?
mbiebl>
hm.. https://codesearch.debian.net/search?q=SO_REUSEADDR+%7C+SO_REUSEPORT&literal=1

[Debian Code Search: SO_REUSEADDR | SO_REUSEPORT](https://codesearch.debian.net/search?q=SO_REUSEADDR+%7C+SO_REUSEPORT&literal=1)
[...]

PR #5466, opened 2024-10-13
testbench: correctly apply socket options in minitcpsrvr

mbiebl:
The option name in setsockopts is not a bitmask, so SO_REUSEADDR and SO_REUSEPORT can't be ORed together. Instead apply the options via separate calls.

Fixes: #5456

Thanks: Chris Hofstaedtler <zeha@debian.org>

### DIFF
--- a/tests/minitcpsrvr.c
+++ b/tests/minitcpsrvr.c
@@ -89,8 +89,11 @@ createListenSocket(void)
 	}
 	// Set SO_REUSEADDR and SO_REUSEPORT options
 	int opt = 1;
-	if (setsockopt(listen_fd, SOL_SOCKET, SO_REUSEADDR | SO_REUSEPORT, &opt, sizeof(opt)) < 0) {
-		errout("setsockopt failed");
+	if (setsockopt(listen_fd, SOL_SOCKET, SO_REUSEADDR, &opt, sizeof(opt)) < 0) {
+		errout("setsockopt failed for SO_REUSEADDR");
+	}
+	if (setsockopt(listen_fd, SOL_SOCKET, SO_REUSEPORT, &opt, sizeof(opt)) < 0) {
+		errout("setsockopt failed for SO_REUSEPORT");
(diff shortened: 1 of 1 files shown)


rgerhards:
Thx for the PR. It looks like there is an issue on Solaris, I'll see that I can update the PR to fix that.

mbiebl:
This solaris build failure doesn't be a regression due to this PR though.
See https://build.rsyslog.com/#/builders/220/builds/1955

rgerhards:
The solaris build env did not properly work for some time. so it might got unnoticed. so just for clarity, I'll merge now and do the fix separately.

mbiebl:
https://docs.oracle.com/cd/E88353_01/html/E37843/setsockopt-3c.html
→ SO_EXCLBIND, SO_PASSIVE_CONNECT, SO_REUSEPORT, SO_TIMESTAMP
available since 11.0.0

seems the solaris build environment is simply too old.
maybe this code needs to be made conditional if you want to support older solaris releases.

mbiebl:
> The solaris build env did not properly work for some time. so it might got unnoticed.

commit https://github.com/rsyslog/rsyslog/commit/1c0f9bba505d32af8a0421b495027f726cb0aa4c

should also be the root cause for the solaris build failures.

rgerhards:
yes, it's a relatively straightforward fix, using #ifdef. The builder is intentionally Solaris 10, because a lot of solaris folks seem to depend on that version (as far as I have been told by the solaris package maintainers). This will also lead to a number of CI tests which cannot be executed on Solaris 10, as this option is vital for them.

rgerhards:
> commit [1c0f9bb](https://github.com/rsyslog/rsyslog/commit/1c0f9bba505d32af8a0421b495027f726cb0aa4c)
> 
> should also be the root cause for the solaris build failures.

yes, SO_REUSEPORT is the problem with solaris 10.

PR #5477, opened 2024-10-25
imjournal: Fix first message loss after journal rotation

Cropi:
Previously, upon detecting a journal rotation, the pointer did not correctly advance to the next log entry, causing the first new message to be missed. This update ensures that the cursor is advanced correctly, allowing all entries to be read without skipping the first message after rotation.

### Technical details
There were two calls to `sd_journal_next()` — one in `handleRotation` and another in `doRun` caused the first message after rotation to be skipped." This is because sd_journal_next advances the read pointer into the journal by one entry. I deleted the call from `handleRotation()`.

### Reproducer
```
# cat writer.sh 
#!/bin/bash
for i in {1..15}; do
	logger "test$i"
done
exit 0
```

```
# cat /etc/systemd/journald.conf
MaxRetentionSec=10s
MaxFileSec=10s
Audit=false
```

Run `systemctl restart systemd-journald; sleep 11; ./writer.sh` and observe that some logs will not be collected, more precisely the very first log after rotation. The reproducer is not 100% reliable but works cca 2-3/15. It depends on which message is the first to be retrieved from the journal.

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -676,16 +676,9 @@ static rsRetVal
 handleRotation(struct journalContext_s *journalContext, char* stateFile)
 {
 	DEFiRet;
-	int r;
 
 	LogMsg(0, RS_RET_OK, LOG_NOTICE, "imjournal: journal files changed, reloading...\n");
 	STATSCOUNTER_INC(statsCounter.ctrRotations, statsCounter.mutCtrRotations);
-	closeJournal(journalContext);
-
-	iRet = openJournal(journalContext);
-	if (iRet != RS_RET_OK) {
-		ABORT_FINALIZE(RS_RET_ERR);
-	}
 
 	/* outside error scenarios we should always have a cursor available at this point */
 	if (!journalContext->cursor)
@@ -706,14 +699,8 @@ handleRotation(struct journalContext_s *journalContext, char* stateFile)
 		iRet = RS_RET_ERR;
 	}
 	journalContext->atHead = 0;
-	/* Need to advance because cursor points at last processed message */
-	if ((r = sd_journal_next(journalContext->j)) < 0) {
-		LogError(-r, RS_RET_ERR, "imjournal: sd_journal_next() failed");
-		iRet = RS_RET_ERR;
-	}
 
 finalize_it:
-	journalContext->reloaded = 1;
 	RETiRet;
 }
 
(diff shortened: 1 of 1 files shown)


christianhorn:
A partner looked at the patch, and thinks that a further fix is needed in loadJournalState. Without that fix, this issue is reproduced (i.e. "test1" is lost) after starting rsyslog if rsyslog is stopped before writing the logs.

```
    - imjournal: Fix first message loss after journal rotation #5477
    https://github.com/rsyslog/rsyslog/pull/5477/commits/567b315d69c7219e15b9e4c8f81aa1002362d908#diff-8b495bcb58deb75da0fc4e2949cf13850dec264d96e6726250958a701e530f41R962-R965
    
    - the fix for loadJournalState()
    -----
         * but if cursor has been intentionally compromised it could stop logging even
         * with persistent journal.
         * */
    -    if ((r = sd_journal_get_cursor(journalContext.j, &tmp_cursor)) < 0) {
    +    if ((r = sd_journal_get_cursor(journalContext.j, &tmp_cursor)) < 0
    +              || ((r = sd_journal_test_cursor(journalContext.j, readCursor)) <= 0)) {
                 LogError(-r, RS_RET_IO_ERROR, "imjournal: "
                 "loaded invalid cursor, seeking to the head of journal\n");
                 if ((r = sd_journal_seek_head(journalContext.j)) < 0) {
    -----

```

Our partner confirmed the patch on RHEL8.2. FWIW, I used this slightly easier to use reproduce steps on affected versions:
```
terminal #1: while :; do systemctl restart systemd-journald; sleep 11; ./writer.sh; logger donetest1; done
terminal #2: tail -f /var/log/messages|egrep 'test1$'

Example output：
Oct 22 08:01:05 amd82 root[1740]: test1
Oct 22 08:01:06 amd82 root[1755]: donetest1
Oct 22 08:01:19 amd82 root[1765]: test1
Oct 22 08:01:20 amd82 root[1781]: donetest1
Oct 22 08:01:33 amd82 root[1790]: test1
Oct 22 08:01:34 amd82 root[1806]: donetest1
Oct 22 08:01:47 amd82 root[1818]: test1
Oct 22 08:01:47 amd82 root[1834]: donetest1
Oct 22 08:02:01 amd82 root[1859]: donetest1
Oct 22 08:02:14 amd82 root[1869]: test1
Oct 22 08:02:15 amd82 root[1885]: donetest1
Oct 22 08:02:29 amd82 root[1910]: donetest1
Oct 22 08:02:43 amd82 root[1938]: donetest1
```

rgerhards:
Note: we are working on CI, so you may see failures that are obviously not related to the patch. Pls ignore those. I'll make me heard if I need to bring anything to your attention.

And thx for the patch!

PR #5486, opened 2024-11-07
Fix build on musl

quincyf467:
Adding signal.h to plugins/imdtls/imdtls.c fixes this issue: https://bugs.gentoo.org/942294

### DIFF
--- a/plugins/imdtls/imdtls.c
+++ b/plugins/imdtls/imdtls.c
@@ -24,6 +24,7 @@
 #include "config.h"
 
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


rgerhards (approved):
Thx for the PR, much appreciated.

rgerhards:
Note that we are currently working on CI upgrades, that can lead to false positives and/or flakes. I'll handle that. If possible, I would like to merge with "pass", for cosmetic reasons (obviously, the patch is fine).

quincyf467:
okay. sounds good

PR #5491, opened 2024-11-15
CI: user newer OS versions in CI workflow

rgerhards:


rgerhards:
This did not work out and was replaced with a simpler PR.

Issue #5507, opened 2024-11-26
Build failure with GCC 15 (-std=gnu23)


kostadinsh:
rsyslog fails to build with upcoming GCC 15 that defaults to -std=gnu23. This can be reproduced on older GCC and Clang versions by adding `-std=gnu23` to CFLAGS. Below are only some of the errors, since I don't wanna make the bug report too long.

Downstream Gentoo bug report: https://bugs.gentoo.org/943899

```
linkedlist.c:45:10: error: conflicting types for 'llInit'; have 'rsRetVal(linkedList_t *, rsRetVal (*)(void *), rsRetVal (*)(void *), int (*)(void *, void *))' {aka 'enum rsRetVal_(struct linkedList_s *, enum rsRetVal_ (*)(void *), enum rsRetVal_ (*)(void *), int (*)(void *, void *))'}
   45 | rsRetVal llInit(linkedList_t *pThis, rsRetVal (*pEltDestructor)(void*), rsRetVal (*pKeyDestructor)(void*),
      |          ^~~~~~

In file included from rsconf.c:38:
rsconf.c: In function 'rsconfInitialize':
rsconf.c:292:53: error: passing argument 2 of 'llInit' from incompatible pointer type [-Wincompatible-pointer-types]
  292 |         CHKiRet(llInit(&pThis->rulesets.llRulesets, rulesetDestructForLinkedList,
      |                                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                     |
      |                                                     rsRetVal (*)(void *) {aka enum rsRetVal_ (*)(void *)}

In file included from rsconf.c:38:
rsconf.c: In function 'rsconfInitialize':
rsconf.c:292:53: error: passing argument 2 of 'llInit' from incompatible pointer type [-Wincompatible-pointer-types]
  292 |         CHKiRet(llInit(&pThis->rulesets.llRulesets, rulesetDestructForLinkedList,
      |                                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                     |
      |                                                     rsRetVal (*)(void *) {aka enum rsRetVal_ (*)(void *)}
In file included from msg.c:52:
msg.c: In function 'msgClassInit':
[...]

PR #5510, opened 2024-11-28
CI: add check for compile with -std=gnu23 gcc option

rgerhards:
see also https://github.com/rsyslog/rsyslog/issues/5507


rgerhards:
The new gnu23 checks turn out to be heavily overdone. They trigger on perfectly legal, albeit complex code. I could refactor to something more complex which has more or less the same properties (maybe a bit slower), but that involves a lot of work on things that haven't been touched for 10+ years and are not meant to be touched again (code for legacy constructs). I am not sure if we should really try to "solve" this.

rgerhards:
fixed via https://github.com/rsyslog/rsyslog/pull/5514

PR #5516, opened 2024-12-10
testbench: Updated legacy configs in tls tests

alorbach:
Tests should work faster now as some were running longer than necessary.


rgerhards:
@alorbach we need to check if some legacy config tests remain. If we convert everything to new-style, we do no longer test that we do not break legacy config.

PR #5540, opened 2025-01-27
omfwd: align some parameters with imtcp names

rgerhards:
Parameter names have historically evolved and are sometimes inconsistent between modules (while still having the same semantics and similiar name).

This commit creates three aliases in omfwd to keep stream driver parameter names consistent with the names used in imtcp. Note that we do not create an alias for
"streamdriverpermittedpeers" because the name differ more considerably and we would also need to create an alias in imtcp as well. We will do this only on request.

Note: aliases help, but are not a great solution. They may cause confusion if both names are used together in a single config. So care must be taken when using an alias.

### DIFF
--- a/tools/omfwd.c
+++ b/tools/omfwd.c
@@ -235,8 +235,11 @@ static struct cnfparamdescr actpdescr[] = {
 	{ "conerrskip", eCmdHdlrNonNegInt, 0 },
 	{ "gnutlsprioritystring", eCmdHdlrString, 0 },
 	{ "streamdriver", eCmdHdlrGetWord, 0 },
+	{ "streamdriver.name", eCmdHdlrGetWord, 0 }, /* alias for streamdriver */
 	{ "streamdrivermode", eCmdHdlrInt, 0 },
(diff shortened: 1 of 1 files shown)


davidelang:
> Note: aliases help, but are not a great solution. They may cause confusion if both names are used together in a single config. So care must be taken when using an alias.

In the documentation, please group the alias with main parameter (I would put 
them at an equal level, over time we may want to migrate to the alias as the 
preferred name for consistancy)

and it would be good to have the config check flag the case where both names are 
used.

David Lang

rgerhards:
> and it would be good to have the config check flag the case where both names are used.

I agree, but that would be 1 to 5 days of work - which I think would not really be appreciated by most folks. To do so, I would need to refactor part of the parameter system to really introduce aliases as an object.

> In the documentation, please group the alias with main parameter

I'll add them into the same section, doc PR will be upcoming soon.

rgerhards:
doc PR: https://github.com/rsyslog/rsyslog-doc/pull/1076

Issue #5439, opened 2024-08-16
flaky omhttp test


rgerhards:
from @Cropi 

I gave a moment to troubleshoot `ubuntu_22_san` issue with the `omhttp-batch-retry-metadata.sh` test. The test results are not deterministic. Sometimes it passes, sometimes it fails. The `omhttp-validate-response.py` script tries to parse the json input line by line, but sometimes the message and response fields are on separate lines, thus it fails to parse.

```
{ "message": {"msgnum":"00038963"}, "response": { "code": 207, "body": "INTERNAL ERROR", "batch_index": 94 } }
{ "message": {"msgnum":"00038964"}, "response": { "code": 207, "body": "INTERNAL ERROR", "batch_index": 95 } }
{ "message": {"msgnum":"00038965"}, "response": { "code": 207, "body": "INTERNAL ERROR", "batch_index": 96 } }
{ "message": {"msgnum":"00038966"}, "response": { "code": 207, "body": "INTERNAL ERROR", "batch_index": 97 } }
{ "message": {"msgnum":"00038967"}, "response": { "code": 207, "body": "INTERNAL ERROR", "batch_index": 98 } }
{ "message": {"msgnum":"00038968"}, "response": { "code": 207, "body": "INTERNAL ERROR", "batch_index": 99 } }
{ "message": {"msgnum":"00029129"}
, "response": { "code": 207, "body": "INTERNAL ERROR", "batch_index": 0 } }
```

And it fails because `{ "message": {"msgnum":"00029129"}` is not a valid json:
```
msgnum: 00038968 code: 207 body: INTERNAL ERROR batch_index: 99
line:  { "message": {"msgnum":"00029129"}

Traceback (most recent call last):
  File "/home/alakatos/github/rsyslog/tests/./omhttp-validate-response.py", line 28, in <module>
    json_obj = json.loads(line)
               ^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/json/decoder.py", line 353, in raw_decode
[...]

PR #5544, opened 2025-01-27
testbench: fix flaky omhttp-batch-retry-metadata.sh test

rgerhards:
The test results were not deterministic. Sometimes it passed, sometimes it failed. The omhttp-validate-response.py script tried to parse the json input line by line, but sometimes the message and response fields are on separate lines, thus it failed to parse.

This patch updates the check program so that it reads in complete lines when parsing fails, thus removing the condition for flakiness.

closes https://github.com/rsyslog/rsyslog/issues/5439


rgerhards:
@Cropi it looks there may be a problem in imhttp in addition to what you mentioned in #5439 - in case you are interested in digging deeper.

At least the test seems now to be somewhat less flaky.

PR #5568, opened 2025-02-19
Imtcp mt refactor alternate approach

rgerhards:
Alternate approach to https://github.com/rsyslog/rsyslog/pull/5497 where I got stuck with the thread pool. I hope it were too many changes and thus too hard to find root issue. Usually a "restart" works.


rgerhards:
This branch has created a good solution for the worker thread pool. Now I will continue with the original feature branch https://github.com/rsyslog/rsyslog/pull/5497.

Issue #5572, opened 2025-02-24
CI: fix missing symbolizer in Ubuntu 22.04 CI Environment


rgerhards:
We generally need this for CI, but it is very useful to have for #5497 .

PR #5573, opened 2025-02-24
CI: use Ubuntu 24.04 for ASAN, UBSAN etc sanitizer runs

rgerhards:
Ensure we run the newest and most capable sanitizers.

This also fixes the issue of the current Ubuntu 22.04 environment not providing a proper symbolizer.

closes https://github.com/rsyslog/rsyslog/issues/5572


rgerhards:
we just need to check - I think it is done, maybe not 100%

rgerhards:
this needs to be re-done, too much evolved since I created it.

PR #5582, opened 2025-02-26
CI: Switch to clang-18 (Ubuntu 24 dev container)

alorbach:
closes: https://github.com/rsyslog/rsyslog/issues/5581


rgerhards (changes requested):
wrong file / test

PR #5591, opened 2025-03-03
Update man page dates

Cropi:
A small man page update to signal that man pages are up to date. Some users might misinterpret it if there are still old dates and versions.


rgerhards:
thx for the pr, much appreciated!

Issue #5592, opened 2025-03-04
tcpflood_wrong_option_output.sh fails


billie-alsup:
tcpflood_wrong_option_output.sh is failing for fc41.aarch64 (fedora 41, with glibc 2.40) using standard ./configure

The expected output is missing

```
-z and -Z must also be specified
```

The reason it is missing is because the standard ./configure for this environment resulted in the following output from **_tcpflood -Ttls_**

```
compiled without gnutls/openssl TLS support: "-Ttls" not supported!
```

Should the test support such output and PASS in such cases?  or perhaps SKIP in such cases?
Or perhaps it can proactively determine whether TLS is configured and SKIP the test without running it?

--enable-openssl can be enabled with the optional package openssl-devel-engine, so I'll try that.

rgerhards:
I think there should be configure flags and conditionals we can pick up in the testbench and not run the test at all in these cases. I'll have a look.

rgerhards:
@billie-alsup I would appreciate if you could git https://github.com/rsyslog/rsyslog/pull/5595 a try in your environment.

billie-alsup:
> [@billie-alsup](https://github.com/billie-alsup) I would appreciate if you could git [#5595](https://github.com/rsyslog/rsyslog/pull/5595) a try in your environment.

This works, however, I've added some comments on the PR.

PR #5595, opened 2025-03-05
testbench bugfix: do not execute tcpflood_wrong_option_output.sh with…

rgerhards:
…out TLS driver

when neither GTLS or OSSL TLS drivers are enabled, this test cannot pass. We now check for this condition and execute it only it at least one of the two is configured.

closes https://github.com/rsyslog/rsyslog/issues/5592

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -513,11 +513,19 @@ TESTS +=  \
 	queue-encryption-da.sh
 endif # ENABLE_LIBGCRYPT
 if HAVE_VALGRIND
+# note if the same test is added multiple times, it still is only executed once!
+if ENABLE_GNUTLS_TESTS
+TESTS +=  \
+	tcpflood_wrong_option_output.sh
+endif # ENABLE_GNUTLS_TESTS
+if ENABLE_OPENSSL
+TESTS +=  \
+	tcpflood_wrong_option_output.sh
+endif # ENABLE_OPENSSL
 TESTS +=  \
 	omusrmsg-noabort-vg.sh \
 	omfile_hup-vg.sh \
 	gzipwr_hup-vg.sh \
-	tcpflood_wrong_option_output.sh \
 	imtcp-octet-framing-too-long-vg.sh \
 	smtradfile-vg.sh \
 	dnscache-TTL-0-vg.sh \


rgerhards:
Thx for the comments!

> This solves the problem, however, it seems to me that this test was in the wrong section to begin with. This section should have just been for the *-vg.sh tests, right? 

well, for anything that needs valgrind. This test uses it, so it needs to go in here. Otherweise, we get a FAIL if the build environment does not support it (e.g. because ASAN is enabled, which is incompatible).

> As long as you are making changes, would it be better to put your new code outside of the HAVE_VALGRIND check?

It has nothing to do with new or old code. We have test that use valgrind, we have tests that do not.  For example, we run valgrind test and we also run another worker with ASAN enabled. Both are not perfect and sometimes one finds a bug where the other does not. 

> It is unclear to me if the two ENABLE_* options are mutually exclusive or not, so I wonder if this could result in the test running twice (although I guess that isn't really a problem). 

That's in the comment immediately above, but that the problem with comments ;-) In short words: it's not mutually exclusive, but autoconf just runs it once.  Do to autotools restrictions I can neither indent the if blocks nor can I place the comment on the same line as the if :-(

> I note that the tcpflood_wrong_option_output.sh actually runs two tests, one of which is independent of these flags. So we are getting rid of both tests, when only one of them was problematic. Perhaps the test case should be split in two? One tests "-t" and one tests "-Ttls" (and that's the problematic one).

Valid point, but it just checks the testing tool. If it malfunctions, we most probably see a test fail. Also, in CI, there are many builders that have the requirements. So I do not see much value in splitting it. Plus, a test has some minimal overhead in itself, and we already have some builders who run roughly an hour, sometime even a bit longer.

But you are right, two tests would be cleaner.

> Finally, I don't think the test should be removed from EXTRA_DISTS, should it?

Yup! Typo! The "make distcheck" runs in CI did also catch it. I have corrected this now.

And thanks again for your thoughts! I actually feel a bit tempted to split the test, but...

billie-alsup (approved):
I see that it does invoke valgrind directly.  I was initially only looking for USE_VALGRIND and presuming valgrind was only invoked from within the diag.sh functions.  This looks fine then.

PR #5596, opened 2025-03-05
imptcp bugfix: potential socket leak

rgerhards:
imptcp connection is accepted and right away rsyslog is stopped

detected by Coverity Scan. Not sure if it ever happened in practice.


rgerhards:
@Cropi I think the underlaying coverity report is a false positive. In any case, the fix can close the socket prematurely. Can you give some hints what the report actually complained about?

rgerhards:
I updated the patch, as cases are different. I guess it would change the coverity issue (based on the commit comment).

PR #5606, opened 2025-03-17
imuxsock: Add statistics counter for discarded messages

Cropi:
This patch introduces a counter to track the number of discarded messages. Additionally, it fixes a bug where the submitted message count was incremented even when the message was discarded due to exceeding the allowed message rate within a given interval.


Cropi:
Hello @rgerhards , would you accept such a patch in the future? Do you need me to update the patch in any way?

rgerhards:
sorry for the delay. It's a challenging time, but I think we finally near great outcome. Expect to see more activity over here!

And, yes, pls continue to submit PRs, they are much appreciated.

It would be great, however, if you could add related user doc.  In the future this hopefully will be easier, I am evaluating placing the reference documentation either right into the source files or "nearby".

Issue #5607, opened 2025-03-22
Omjournal module set false value of log priority


anis-bougrine:
Omjournal module use the journald api to log to journal: 
	sd_journal_send("MESSAGE=%s", getMSG(pMsg),
		"PRIORITY=%d", sev,
		"SYSLOG_FACILITY=%d", pMsg->iFacility,
		"SYSLOG_IDENTIFIER=%s", tag,
		NULL);

We can see PRIORITY value is set to severity. However, based on syslog PRIORITY = FACILITY | SEVERITY

rgerhards:
fixed #5608

PR #5608, opened 2025-03-22
omjournal: Fix priority value.

anis-bougrine:
Based on [RFC](https://datatracker.ietf.org/doc/html/rfc5424#section-6.2.1) PRIORITY = FACILITY | SEVERITY.
However, omjournal sets PRIORITY = SEVERITY.
So, this PR fix the issue #5607 "Omjournal module set false priority value"


davidelang:
the correct math is actually facility <<3 |severity so that the two don't overlap.
Severity is 0-7 so facility 1 needs to be 8

anis-bougrine:
You are correct David there is overlapping risk here. However, I've just checked the [NFC](https://datatracker.ietf.org/doc/html/rfc5424#section-6.2.1) and i've realised that don't shift the facility value in order to avoid overlap, instead we have to multiply facility value by 8. This is a quote from the NFC:
"The Priority value is calculated by first multiplying the Facility
   number by 8 and then adding the numerical value of the Severity.  For
   example, a kernel message (Facility=0) with a Severity of Emergency
   (Severity=0) would have a Priority value of 0.  Also, a "local use 4"
   message (Facility=20) with a Severity of Notice (Severity=5) would
   have a Priority value of 165."
So to summarise, this is the final formula: PRIORITY = FACILITY*8 | severity

davidelang:
multplying by 8 is the same as shifting 3 bits to the left.

David Lang

PR #5609, opened 2025-03-25
Add journald namespace support for omjournal

anis-bougrine:
Starting from systemd V256, systemd API supports writing to journal in a specific namespace. So this PR make use of this feature in rsyslog.


anis-bougrine:
Hello @rgerhards, I've made changes lately and all checks pass successively, so I think this worth reviewing. Thank you.

anis-bougrine:
@rgerhards, I hope this is ready for review. Thank you

PR #5621, opened 2025-04-17
tcpsrv bugfix: do not busy wait on io events

rgerhards:
Depending on circumstances, tcpsrv worker threads did effectively busy-wait on io events to handle. Not always, but often. This was caused be improperly re-arming the inotify subsystem.

This effected overall system performance, but not general rsyslog stability. The bug was introduced on March 1st 2025 into the daily stable build.

closes https://github.com/rsyslog/rsyslog/issues/5623

### DIFF
--- a/plugins/imdiag/imdiag.c
+++ b/plugins/imdiag/imdiag.c
@@ -132,13 +132,13 @@ doOpenLstnSocks(tcpsrv_t *pSrv)
 
 
 static rsRetVal
-doRcvData(tcps_sess_t *pSess, char *buf, size_t lenBuf, ssize_t *piLenRcvd, int *oserr)
+doRcvData(tcps_sess_t *pSess, char *buf, size_t lenBuf, ssize_t *piLenRcvd, int *oserr, unsigned *nextIODirection)
 {
 	assert(pSess != NULL);
 	assert(piLenRcvd != NULL);
 
 	*piLenRcvd = lenBuf;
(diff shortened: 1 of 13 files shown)


billie-alsup:
Thanks.  I believe this will resolve https://github.com/rsyslog/rsyslog/issues/5623

rgerhards:
> I'm not sure how epoll reacts in that case if you poll for writes but the write half is shutdown.

This should be save, as epoll is supposed to emit an error/HUP event in such cases. Also, we now control the io direction, so we should always see that this happens. Any hang would also be detected by the testbench. Previously, it ran through because "everthing worked" - it "just" was doing a busy wait, which the testbench did not detect and also I overlooked during manual testing.

rgerhards:
the ci fails were induced by problems in the CI environment. We are working on that.

billie-alsup:
> If you like, you can apply the patch right now and supply feedback. It looks rather complete, I am shortly before merging. I just want to see if I can add a few manually created tests for better coverage (not yet sure).

Now that it is merged, I'll revert back to working with master branch and apply my changes (two sets of changes now, one still a pending PR, and one not opened yet).

My upcoming changes also support namespaces (as well as abstract AF_UNIX socket addressses), and my first PR was to support namespaces, and there is a 3rd module that already supports namespaces.  I don't want  to cut&paste code, so my latest changes add a generic ns_socket call.  I want to put this into an existing object that can be shared (net perhaps?), and that way it can use a common configuration scheme across users, e.g. net.namespace or similar.  What do you think?

PR #5625, opened 2025-04-24
ossl tls driver: better error reporting

rgerhards:
Replace some generic error codes with more specific counterparts. There is still some more work to do, but this covers important spots.


rgerhards:
solaris still has CI env issues

Issue #5622, opened 2025-04-17
aarch64 fedora-42: master branch: make distcheck shows 67 failures


billie-alsup:
I updated my AARCH64 VM from fedora-41 to fedora-42.    I am also working on a new PR off of master branch (7b79492624372300f585be207c1dad03ac5cc9ce).   I find that make distcheck shows 67 failures on the unmodified master branch.  It is unclear if this reflects instability in master branch, or if it is something new with fedora-42.

[test-suite.log](https://github.com/user-attachments/files/19801423/test-suite.log)

billie-alsup:
Seems to be a single problem related to an implementation change within glibc-2.41.  With a single new suppression, the build is clean:

```
============================================================================
Testsuite summary for rsyslog 8.2504.0.master
============================================================================
# TOTAL: 581
# PASS:  573
# SKIP:  8
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
```

```
fedora:~/rsyslog.master$ git diff
diff --git a/tests/known_issues.supp b/tests/known_issues.supp
index 9bb8fa39e..6daafc4eb 100644
--- a/tests/known_issues.supp
+++ b/tests/known_issues.supp
@@ -60,3 +60,22 @@
    fun:exit
    fun:(below main)
 }
+{
+   <issue with Fedora-42 glibc-2.41>
+   Memcheck:Leak
+   match-leak-kinds: definite
+   fun:malloc
+   fun:UnknownInlinedFun
+   fun:_dl_map_object_deps
+   fun:dl_open_worker_begin
+   fun:_dl_catch_exception
+   fun:dl_open_worker
+   fun:_dl_catch_exception
+   fun:_dl_open
+   fun:do_dlopen
+   fun:_dl_catch_exception
+   fun:_dl_catch_error
+   fun:dlerror_run
+   fun:__libc_dlopen_mode
+}
fedora:~/rsyslog.master$ 
```

billie-alsup:
A fix is included in https://github.com/rsyslog/rsyslog/pull/5630

billie-alsup:
Closed with merge of #6121 and new suppressions for glibc 2.40 and 2.41

PR #5519, opened 2024-12-18
imtcp support for NetworkNamespace

billie-alsup:
This adds a NetworkNamespace parameter to imtcp, based on the omfwd implementation.

This builds on "PR#6121 net: Add NetworkNamespace APIS" to add Network Namespace support to imtcp module.  This extends imtcp to support a wider range of Unix/Linux environments (or any environment supporting network namespaces).
    
The imtcp module is enhanced to accept a NetworkNamespace parameter, both as a default at the module level, and on a per-instance basis.
    
The tcpsrv module is enhanced to allow the NetworkNamespace to be applied to a listener's configuration parameters.
    
Finally, the netstrm module is enhanced to switch namespaces before invoking the downstream (driver specific) LstnInit function.
    
A new test imtcp-netns (and associated imtcp-netns-vg) is added to test this functionality.  This must be run as root (technically it must be run by a user with CAP_SYS_ADMIN capabilities, as network namespace creating/change is required).
    
A slight change to diag.sh is made to allow passing $RS_REDIR to valgrind (as $RS_REDIR is used in the imtcp-netns.sh test for some negative cases).

PR #5630, opened 2025-05-05
New functionality for omuxsock

billie-alsup:
This builds on "PR#6121 net: Add NetworkNamespace APIS" to add Network Namespace support for omuxsock.  In addition, new functionality is added to support abstract socket names, as well as connected sockets.
    
Changes are isolated to omuxsock to provide the new functionality:
    
a. Abstract unix socket names (including network namespaces)
b. Connected socket support (SOCK_STREAM, SOCK_SEQPACKET)
c. Support for load, and action configuration
d. Bug fix related to closing/reopening sockets: This bug would cause the socket to be closed and reopened for every message sent.
    
New tests are added as follows:
    
a. uxsock_multiple.sh tests basic functionality with multiple output instances.
b. uxsock_multiple_netns.sh is similar but using multiple namespaces.
c. uxsock_simple_abstract.sh mirrors the existing uxsock_simple.sh but tests with abstract socket names.
    
The uxsockrcvr test program was modified to support these new tests.

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -1597,11 +1597,13 @@ endif
 if ENABLE_OMUXSOCK
 TESTS += uxsock_simple.sh \
 	 uxsock_simple_abstract.sh \
-	 uxsock_simple_abstract-vg.sh \
 	 uxsock_multiple.sh \
+	 uxsock_multiple_netns.sh
+if HAVE_VALGRIND
+TESTS += uxsock_simple_abstract-vg.sh \
 	 uxsock_multiple-vg.sh \
-	 uxsock_multiple_netns.sh \
 	 uxsock_multiple_netns-vg.sh
+endif # HAVE_VALGRIND
 endif
 
 if ENABLE_RELP


billie-alsup:
Includes suppression for aarch64 fedora-42 issue as described in https://github.com/rsyslog/rsyslog/issues/5622

rgerhards:
Sry for the further delay, I need some more time to review and think of implications.

billie-alsup:
The gen kafka distcheck failure is related to the a new file which is meant to be linked into other modules.  I thought I had addressed the distcheck issue myself, but I will doublecheck.

```

Makefile:548: ../../runtime/.deps/omuxsock_la-ns_socket.Plo: No such file or directory
make[2]: *** No rule to make target '../../runtime/.deps/omuxsock_la-ns_socket.Plo'.  Stop.
```

Two failures seem related to the build environment themselves.  These are for solaris10x64 rsyslog and sunstudio.

```
../libtool: line 1714: ar: command not found
```

A solaris11x64 rsyslog failed due to the uxsock_simple_abstract.sh test failing.  I suspect I need to put in a linux check, as solaris does not support abstract unix socket addresses.

plugins/omuxsock/omuxsock.c
```diff
@@ -243,9 +343,85 @@ BEGINfreeInstance
...
+			 * Should we really transfer ownership, or should we allow it to be the default
+			 * by calling ustrdup?
+			 */
+			pData->sockName = cs.sockName;
+			pData->bAbstract = cs.bAbstract;
+			pData->sockType = cs.sockType;
+			pData->bConnected = cs.bConnected;
+			cs.sockName = NULL; /* pData is now owner and will fee it */
+		}
+	}
```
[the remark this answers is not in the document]
billie-alsup: This is possible, but I explicitly did not want these semantics.  If you are using the new functionality offered by this PR, then I explicitly want you to configure the four-tuple together, whether at the pData level, or the loadModConf level.  The four-tuple goes together.  Obviously, you can leave pieces out to get defaults, but those are not the defaults from loadModConf.  loadModConf has a four-tuple default if no four-tuple is given for pData.  I think it would be confusing to actually have the 4 individual parameters be individual defaults when not specified otherwise for pData. @rgerhards please indicate which behavior you would like to see.

tests/uxsockrcvr.c
```diff
@@ -151,40 +207,71 @@ main(int argc, char *argv[])
 	/*      Set up address structure for server socket      */
 	memset(&addr, 0, sizeof(addr));
 	addr.sun_family = AF_UNIX;
-	strcpy(addr.sun_path, sockName);
+	strncpy(addr.sun_path + abstract, sockName, sizeof(addr.sun_path) - abstract);
```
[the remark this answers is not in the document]
billie-alsup: I'll take care of this.  I guess I get lazy with test code.

billie-alsup:
I am perplexed by the distcheck failure.  I do not see a 'make distcheck' failure in my environment.

```
Making distclean in plugins/omuxsockmake[2]: Entering directory '/rsyslog/rsyslog-8.2508.0.daily/_build/sub/plugins/omuxsock'
Makefile:550: ../../runtime/.deps/omuxsock_la-ns_socket.Plo: No such file or directory
make[2]: *** No rule to make target '../../runtime/.deps/omuxsock_la-ns_socket.Plo'.  Stop.
make[2]: Leaving directory '/rsyslog/rsyslog-8.2508.0.daily/_build/sub/plugins/omuxsock'
```

I can avoid it by actually #include'ing the external source, but that just seems wrong.  @rgerhards do you have any pointers?

billie-alsup:
The buildbox/rsyslog solaris11x64 rsyslog failure is in an unrelated test pmrfc3164-json.sh that uses imtcp and omfile with tcpflood.  This particular PR only affects omuxsock.  Note that pmrfc3164-json.sh passes locally for me.

billie-alsup:
> Sry for the further delay, I need some more time to review and think of implications.

The goal is to be completely backward compatible with existing configurations.  They will continue to use sendto, unconnected sockets with SOCK_DGRAM (although it is legal to use connect on them as well; it just caches the target address), and filesystem based unix socket addresses.

I had problems with one of the workflow tasks when referencing a 2nd source file for omuxsock.c, so I just changed omuxsock.c to #include the source file (ns_socket.c).  I have another PR for imtcp.c #5519 that is waiting on this one.  Once this is in, then I'll update PR #5519 to use the same ns_socket.c functions, instead of duplicating them.  Right now it is simply waiting to be updated (due to the clang-format changes).  So I will update it once this one is in.  Or if desired, i can break out the ns_socket.[ch] changes into a separate PR.  Ideally I'd like both PRs in the same release to bring back to my OpenEmbedded recipes.

rgerhards:
sry, i needed to shuffle scheduled stable milestone, but will be able to clear up my pr backlog this week w/ real comments

plugins/omuxsock/omuxsock.c
```diff
@@ -278,19 +442,32 @@ static rsRetVal openSocket(instanceData *pData) {
...
     pData->addr.sun_family = AF_UNIX;
-    strncpy(pData->addr.sun_path, (char *)pData->sockName, sizeof(pData->addr.sun_path));
-    pData->addr.sun_path[sizeof(pData->addr.sun_path) - 1] = '\0';
+    strncpy(pData->addr.sun_path + pData->bAbstract, (char *)pData->sockName,
+            sizeof(pData->addr.sun_path) - pData->bAbstract);
+    if (pData->addr.sun_path[sizeof(pData->addr.sun_path) - 1]) {
+        pData->addr.sun_path[sizeof(pData->addr.sun_path) - 1] = '\0';
+        LogError(0, RS_TRUNCAT_TOO_LARGE, "Unix socket path truncated to %s", pData->addr.sun_path + pData->bAbstract);
+        ABORT_FINALIZE(RS_TRUNCAT_TOO_LARGE);
+    }
```
ImanolCoding: It looks like this check on sun_path[sizeof-1] is trying to catch truncation, but the behavior differs between pathname and abstract Unix sockets.

For pathname sockets, a clearer approach would be to check nlen >= sizeof(sun_path) before copying, and then explicitly add a terminating '\0'.

For abstract sockets, there shouldn’t be a trailing '\0' at all, so modifying the last byte here could change the intended name.

To make the intent clearer and avoid accidental truncation, it might be simpler to split the handling explicitly:

```suggestion
	pData->addr.sun_family = AF_UNIX;
	if (pData->bAbstract) {
		/* Abstract: sun_path[0] must be '\0', no trailing NUL in addrLen 
		 * - sun_path[0] must be '\0'
		 * - name starts at sun_path[1], raw bytes (not necessarily NUL-terminated)
		 * - addrLen = offsetof(...) + 1 (leading NUL) + name length
		 */
		size_t nlen = strlen((char*)pData->sockName);
		if (nlen > sizeof(pData->addr.sun_path) - 1) {
			LogError(0, RS_TRUNCAT_TOO_LARGE, "Abstract Unix socket name too long");
			ABORT_FINALIZE(RS_TRUNCAT_TOO_LARGE);
		}
		memcpy(pData->addr.sun_path + 1, (char*)pData->sockName, nlen);
		pData->addrLen = offsetof(struct sockaddr_un, sun_path) + 1 + nlen;
	} else {
		/* Pathname socket (file system based):
		 * - sun_path contains the pathname and must be NUL-terminated
		 * - addrLen = offsetof(...) + path length + 1 (terminating NUL)
		 */
		size_t nlen = strlen((char*)pData->sockName);
		if (nlen >= sizeof(pData->addr.sun_path)) {
			LogError(0, RS_TRUNCAT_TOO_LARGE, "Unix socket path truncated");
			ABORT_FINALIZE(RS_TRUNCAT_TOO_LARGE);
		}
		memcpy(pData->addr.sun_path, (char*)pData->sockName, nlen);
		pData->addr.sun_path[nlen] = '\0';
		pData->addrLen = offsetof(struct sockaddr_un, sun_path) + 1 + nlen;
	}
```

billie-alsup: The suggestion seems unnecessarily verbose.  Is this coming from an AI review?  The issue itself is debatable.  Yes, the code as given does not support the maximum length supported by the kernel for a given release for an abstract socket.  For most implementations, that means we only support 106 byte abstract socket path, versus a 107 byte abstract socket.  But we are also adding other restrictions as well, such as forcing the path to be all displayable characters (afaik there is no way to enter control characters, or characters with high bit set).  that includes disallowing null characters within the path (except for the required \0 at the start which we manually insert).  So which restrictions are ok and which restrictions are not ok?  To me, since this is a new feature, merely documenting that we leave room for a trailing null is sufficient.  The goal of supporting abstract sockets was to remove dependence on the filesystem and the necessity to pre-create socket nodes in the filesystem.  I don't think limiting the length by 1 byte is an issue here, particularly since the previous code was passing the wrong length entirely (the kernel would adjust the length internally, so reading back the socket would give you the correct length).

ImanolCoding: Thanks for clarifying. My intention was to make the code clearer and more consistent: pathname sockets need an explicit trailing NUL, while abstract sockets rely on a leading NUL. Even if the extra byte isn’t critical in practice, handling each case explicitly keeps the code aligned with kernel semantics and easier to maintain.

rgerhards:
@billie-alsup My intent is to merge this today (finally...) - Thx for your great work and patience.

Note that I'll do one AI check to see if it (maybe invalidly!) points into some areas that I should look especially at. Just so that you do not wonder. If there is anything that needs to be addressed, I'll tell you. So ignore any AI comments for now if they come up.

 I already did some manual review, but need to dig a bit deeper.
 
 If I had not voiced earlier: I agree it would generally be good to abstract this functionality.

rgerhards:
yeah, that AI review comment was something that I also thought. Did you have any issues with just giving the name? The build system adds the runtime directory as include dir in compilation, so that path should not be needed. But it is needed in the Makefile to trigger rebuild on that headers change.

rgerhards:
@billie-alsup  I removed the path name in a local clone, and it works as expected. I see why you are including the C file. It's not something I would normally accept, **but** as an interim measure and in interest of functionality, I would be ready to accept it, given that:

1. it is temporary measure to get feature quickl
2. re-use in at #5519
3. clean, concrete plan towards clean integration- pls elaborate on that
4. an issue with the description from 3 is created

I also had a look at ns_socket.c. I think it is not hard to add it the the existing helper libs. So we should go for that, maybe even before #5519 merge.

Please let me know your thoughts.

billie-alsup:
I was originally thinking it should be functionality available in net.c., basically via an ns_socket API that was socket with an optional namespace argument.  It seemed quite straightforward from an omuxsock perspective.  However, when I started looking at what it would take for imtcp, I got a bit lost in the various indirections.  There the socket call is at least two levels below imtcp (tcpsrv/stream/netstrm/?) In my mind, socket addressing should simply include an optional namespace parameter, and that should be done in a central place (the entire set of parameters to provide a socket "address" should be in one place, and not spread throughout the plugins).   So ideally, the configuration callback would pass parameters down to other modules for possible parsing.  If we took this approach, then the namespace could be parsed by net (and possibly the entire set of parameters describing the address), but it has to be in a net specific (socket specific) instance.  It is perhaps a bigger change than I could effect.  So the second choice would be to simply offer some utility functions within net.  that is the simplest solution for omuxsock.  I would need to investigate further how to allow imtcp to take advantage, since it doesn't make the socket call itself.  You'll note that imtcp PR is not merge-able right now.  I have rebased it locally, but it is failing a few unrelated tests and I haven't had time to investigate.  I figured I would wait until we had a final solution for the simpler omuxsock first.

To be clear, I don't like #include'ing a C source file either.  That was my fallback when the original attempt at putting it into its own library failed for solaris.  The distcheck was working locally for me (fedora 42/aarch64) with a library, but the automation showed it failing the solaris builds.

rgerhards:
@billie-alsup I have an all-day meeting today, but how about this:

* we merge this PR as-is, but not the other
* we join forces to tackle generic implementation
   - I can provide the rough sketch based on your code, but I am heavily loaded, so I may be sluggish except in pre-agreed focus phases (usually, that's not real problem)
   - I cannot do intensive testing - that would be a task for you and others
   - code fixes/improvements can be done by everyone in the team
   - we possibly need to enter some tech discussion
   - I think there is at least one other person (check PRs) that might be interested in that effort
   - use case doc would be great to make this feature really appealing
 * once done with the team effort, we can roll out
 * I would look at one prototype, maybe your other PR, maybe some existing code 
 * ... and probably some other points

I guess you get the spirit. Not a quick task, due to time constraints. But rewarding I think and hopefully a great experience for everyone (if I guess motivations right).

rgerhards:
I forgot - we could / should setup a project board

billie-alsup:
I am new to contributing to open source projects, so am not familiar with the ways of working here.  I certainly would prefer to move this ns_socket.c into a library (perhaps an existing library, otherwise a new library).  I tried creating a static library in a previous upload, and that was working locally, but failing workflow checks  (solaris builds as i recall), where one of the libtools was not available (ranlib?).  I was also getting a distcheck failure for some architecture.  If you're willing to live with the #include of a c file until we can come up with another solution, then we can commit this as is.  I agree that it is an ugly temporary solution.  Note there is still an outstanding workflow failure with solaris11x64 failing pmrfc3164-json.sh.  That test does not use omuxsock however.  I have moved the imtcp PR to draft mode. I can help with the implementation if we can agree on the approach.  I think you know the architecture and flow best here.  As I said, I got lost in my first attempt to follow the flow!

plugins/omuxsock/omuxsock.c
```diff
@@ -43,6 +44,8 @@
 #include "glbl.h"
 #include "errmsg.h"
 #include "unicode-helper.h"
+#include "ns_socket.h"
+#include "../../runtime/ns_socket.c"
```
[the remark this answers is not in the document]
billie-alsup: This is a temporary solution due to workflow issues with solaris, including distcheck failures, when ns_socket was put into its own static library.  The goal is to have these functions shareable across multiple plugins (there is already another plugin with similar functionality that could take advantage of this ns_socket.c APIs).  There is a third plugin (imtcp) that needs similar functionality.  If we can resolve the workflow issues with the static library approach, or if we can add this file to an existing rsyslogd library, then I am all for it.  Long term, the goal is to integrate this into the rsyslogd infrastructure, such that any plugin that uses sockets can take advantage of a network namespace parameter.  That is future work.

billie-alsup:
Please merge #6121  first.  This moves the network namespace APIs out of omfwd into net module, and will be used by the next update to this PR (omuxsock) as well as the other PR (imtcp)

billie-alsup:
@rgerhards This is ready for you again.

billie-alsup:
ubuntu_22_san seems to be a problem with valgrind itself aborting, and perhaps ubuntu_24_tsan is similar.

```
==219826== Memcheck, a memory error detector
==219826== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==219826== Using Valgrind-3.18.1 and LibVEX; rerun with -h for copyright info
==219826== Command: ../tools/rsyslogd -C -n -irstb_188342_8d30a64dicZr:.pid -M../runtime/.libs:../.libs -frstb_188342_8d30a64dicZr_.conf
==219826== 
### unhandled dwarf2 abbrev form code 0x25
### unhandled dwarf2 abbrev form code 0x25
### unhandled dwarf2 abbrev form code 0x25
### unhandled dwarf2 abbrev form code 0x1b
### unhandled dwarf2 abbrev form code 0x25
### unhandled dwarf2 abbrev form code 0x25
### unhandled dwarf2 abbrev form code 0x25
### unhandled dwarf2 abbrev form code 0x1b
==219826== Valgrind: debuginfo reader: ensure_valid failed:
==219826== Valgrind:   during call to ML_(img_get)
==219826== Valgrind:   request for range [924677157, +4) exceeds
==219826== Valgrind:   valid image size of 8807048 for image:
==219826== Valgrind:   "/rsyslog/tools/rsyslogd"
==219826== 
==219826== Valgrind: debuginfo reader: Possibly corrupted debuginfo file.
==219826== Valgrind: I can't recover.  Giving up.  Sorry.
==219826== 
```

billie-alsup:
@rgerhards Any updates on this one?  The dependent PR was merged quite a while back now, and this one adapted to take advantage (reduce code duplication with respect to namespace support).

rgerhards:
@billie-alsup I do not understand the valgrind failures. I will create a testing merge PR to dig deeper into what is happening here.

rgerhards:
@billie-alsup I unfortunately see the CI failures also consistently in my "merge prototype" PR #6324. That means I cannot merge as-as, because all future PRs would fail. Given my long delays while working with you on this PR, I will try to find the root cause in my PR version. Anyhow, if you have ideas, I am all ears!

tests/Makefile.am
```diff
@@ -1607,7 +1595,13 @@ endif
 endif
 
 if ENABLE_OMUXSOCK
-TESTS += uxsock_simple.sh
+TESTS += uxsock_simple.sh \
+	 uxsock_simple_abstract.sh \
+	 uxsock_simple_abstract-vg.sh \
```
rgerhards: valgrind tests need to be run only if valgrind is enabled. This is causing at least the Ubuntu 22.04 failure.

Proposed fix:
https://github.com/rsyslog/rsyslog/pull/6324/commits/c49be8e86a082e66c2089e5ec8fa164eaf76dfd8

(CI is still running on this testing branch at the time of this writing).

rgerhards (changes requested):
Makefile.am needs to be changed. @billie-alsup if you are ok with that, I can simply merge my fix in and squash. Feel also free to grab my patch if you prefer everything from the original source.

rgerhards:
@billie-alsup yes, the fix works, no CI fail in my branch.

tests/Makefile.am
billie-alsup: Doh!  It did not even occur to me that valgrind was a configurable option on a platform!

billie-alsup: Yes, please merge your fix. Thank you!

billie-alsup:
> @billie-alsup yes, the fix works, no CI fail in my branch.

Yes, please merge your fix. Thank you!

rgerhards:
@billie-alsup thx for the feedback, will merge either today (but it's already late over here) or tomorrow.

billie-alsup:
omuxsock documentation changes are staged in a draft PR #6330 
It will need version number changes, and I haven't yet verified that it generates nice html.

billie-alsup:
> omuxsock documentation changes are staged in a draft PR #6330 It will need version number changes, and I haven't yet verified that it generates nice html.

I think the documentation PRs are ready now.  The style has changed since use of rsyslog-doc repository, but I think I got the gist of it.  Both did build for me (with warnings from doc/source/reference/parameters/omlibdbi-db.rst) and looked ok under firefox.  #6330 and #6331

tests/Makefile.am
rgerhards: > Doh! It did not even occur to me that valgrind was a configurable option on a platform!

That's necessary because of incompatiblities. valgrind and sanitizers (like tsan, ubsan, ...) cannot be enabled together. That's the main point why valgind is disabled in parts of the CI. Just FYI.

rgerhards:
@billie-alsup thx again, now finally merged. Will look at the doc PRs soon. Special thanks for your patience and persistence.

I plan to highlight this contribution so that folks will get aware of it (on my blog, social and probably rsyslog.com). If you like, feel free to add some notes. if so, mail me at rgerhards@hq.adiscon.com.

A big thanks again!

PR #5653, opened 2025-06-04
imhttp: add simple http health check functionality

rgerhards:
This also includes a basic Prometheus scrape entry point, which currently can only be used for health checking.

We plan futher enhancement in this direction.

### DIFF
--- a/contrib/imhttp/imhttp.c
+++ b/contrib/imhttp/imhttp.c
@@ -1,7 +1,9 @@
 /* imhttp.c
  * This is an input module for receiving http input.
  *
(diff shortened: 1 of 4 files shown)


rgerhards:
/gemini review - defects should be fixed now

rgerhards:
Solaris CI failure is unrelated and known https://github.com/rsyslog/rsyslog/issues/5678

rgerhards:
/gemini review - thx for the hint, I changed that and also added our portability macros instead of the plain __attribute__

PR #5661, opened 2025-06-06
omusrmsg nitfix: invalid data type

rgerhards:
Not a real problem per se, but prevented clean compile.


rgerhards:
invalid patch, correct one upcoming

Issue #5666, opened 2025-06-07
Merge rsyslog-doc into main repository as subtree


rgerhards:
## Summary

We propose to integrate the `rsyslog-doc` repository into the main `rsyslog` repository via a Git subtree.

## Motivation

The original decision to host `rsyslog-doc` as a separate repository was based on the assumption that we would gain new documentation contributors with limited access rights. This did not happen — in practice, documentation changes are almost exclusively made by the same maintainers as the codebase or via regular PRs.

Merging the documentation into the main repository simplifies maintenance, improves discoverability, and resolves common issues for tools and AI agents which currently struggle to locate the documentation repo. It also simplifies build pipelines and reduces external dependencies.

## Additional Benefits

- Simplifies CI setup for release building (especially tarballs and package generation)
- Improves cross-referencing between code and documentation
- Easier to align doc updates with code changes in a single PR

## Downsides & Mitigation

- Slightly increases repository size — not expected to be significant
- Need to distinguish doc-only and code-only changes in CI (see below)
- doc-deployment scripts (e.g. to rsyslog web site) need to be changed

## CI Considerations

Our CI for documentation is currently **not** fully functional, so integration won't break any production flows.

However, we must carefully separate CI execution for doc vs. code changes:

- Code tests are highly resource intensive, involving 50+ runners with durations up to 45 minutes.
- We should avoid running full code CI on doc-only PRs.
- Conversely, doc CI need not be run on code-only PRs (except for user-facing changes).

This may require better change detection logic in the CI configuration.

## Implementation Approach

We recommend a Git subtree merge of the existing `rsyslog-doc` history. This retains commit history and allows continued modularity if needed in the future.

The suggested location for integration is:

```
rsyslog/
├── doc/           <- new location of rsyslog-doc content
├── modules/
├── tests/
├── ...
```

After subtree merge, we can mark the original repo as read-only or archive it.

rgerhards:
done

PR #5670, opened 2025-06-07
ci: skip expensive workflows on doc-only changes

rgerhards:
see also https://github.com/rsyslog/rsyslog/issues/5666


rgerhards:
CI Fail is known false positive.

Issue #5615, opened 2025-04-15
New (?) queue.highWaterMark checks


billie-alsup:
Testing with a March 4 version of master branch 352d3d9c1b4e1e00b57ce7ac5424d23a5706a98d

I notice messages that have not previously been seen.  Is there a new configuration requirement?  Or is this related to something else possibly going on in my system (perhaps an unreachable endpoint for example)?

> rsyslogd[5460]: queue "action-0-builtin:omfile queue": queue.highWaterMark is set below queue size. It has been automatically been adjusted to 900. In any case, we strongly recommend to review the queue definition and resolve inconsistencies to guarantee the config really matches your intent. [v8.2504.0.master try https://www.rsyslog.com/e/2357 ]
> 
> rsyslogd[5460]: queue "action-0-builtin:omfile queue": queue.lowWaterMark is set below queue size or highWaterMark. It has been automatically been adjusted to 900. In any case, we strongly recommend to review the queue definition and resolve inconsistencies to guarantee the config really matches your intent. [v8.2504.0.master try https://www.rsyslog.com/e/2357 ]
> 
> rsyslogd[5460]: queue "action-1-builtin:omfile queue": queue.highWaterMark is set below queue size. It has been automatically been adjusted to 900. In any case, we strongly recommend to review the queue definition and resolve inconsistencies to guarantee the config really matches your intent. [v8.2504.0.master try https://www.rsyslog.com/e/2357 ]
> 
> ...

(through action-11).  There are no explicit queues configured, so I would hope that default settings would be consistent.

p.s. note the grammar issue with "_been automatically been_"

davidelang:
please post your full config. It sure looks like action 0 and action 1 both have 
queues set.

David Lang

billie-alsup:
> please post your full config. It sure looks like action 0 and action 1 both have
> queues set.
> 

I presume there are implicit queues configured.  I just would have expected the defaults to be consistent and avoid such warnings in the log.  There is no explicit "queue" keyword in the configuration file.

I think this was the config for this instance.  Note the 12 filenames specified in the two rulesets.  These would presumably correspond to the action-0 through action-11 error messages.  These are just simple filenames, and the queues backing them are implicitly configured I believe.

```
global(
    libcapng.enable="off"
    workDirectory="/var/run/rsyslog"
    umask="0022"
    maxMessageSize="8192"
    debug.onShutdown="on"
    debug.logFile="/var/log/rsyslog/debug.log"
)
module(
[...]

PR #5676, opened 2025-06-09
Revert "queue: emit better warning messages on queue param mismatch"

rgerhards:
Reverts rsyslog/rsyslog#5586

see https://github.com/rsyslog/rsyslog/issues/5615


rgerhards:
/gemini - you are wrong, I verified manually. This is the complete revert, `queue.fullDelayMark` was not affected by the reverted patch.

So we are good to merge.

PR #5679, opened 2025-06-10
cleanup unneded semicolons - fixes build warnings on solaris

rgerhards:
We had double semicolons in data definitions. This is fixed now.


rgerhards:
needs to be adapted to interim work - if needed in any longer at all

PR #5686, opened 2025-06-12
Removed the redundant *_la_LDFLAGS

Cropi:
Whilst this did not cause issues when building the project, it produced warnings. Let's remove unnecessary code, so it becomes a bit more readable.


rgerhards (approved):
great find, thx!

PR #5687, opened 2025-06-12
Fix unsafe sprintf use in Azure event hubs module

rgerhards:
## Summary
- replace `sprintf` with `snprintf` in `url_encode`

## Testing
- `devtools/check-codestyle.sh`
- `make check TESTS=tests/errfile-basic.sh -j4` *(fails: fatal - cannot create tests/errfile-basic.trs)*

------
https://chatgpt.com/codex/tasks/task_e_684b2fef64f48332ba98d2f24e1d9f7d


rgerhards:
/gemini this is not ready for merge imho. It looks like it fixes a symptom, but the broader picture of the buffer that snprintf is being used on must be considered.

PR #5699, opened 2025-06-14
some cleanup

rgerhards:


rgerhards:
google code review comments will be addressed in a separate PR - this here is definitely a big move into the right direction AND does not do anything worse than previous. The remaining cases described are very unlikely to occur anytime soon.

Issue #5700, opened 2025-06-14
remove -Wredundant-decls warnings


rgerhards:
While they cause no real harm, they should be removed.

Note: this happens only with `gcc`, be sure to set the right compiler.

PR #5702, opened 2025-06-14
fix duplicate definitions

rgerhards:
see #5700


rgerhards:
this needs to be re-done, it was an early idea, much has changed in the interim.

PR #5717, opened 2025-06-17
doc: first omsendertrack reference manual entry

rgerhards:


rgerhards:
gemini concerns are valid, but will be addressed in later iteration (already scheduled)

PR #5718, opened 2025-06-17
CI: add more tests for omsendertrack

rgerhards:


rgerhards:
did not work out, will check this in an overall new approach

Issue #5176, opened 2023-07-06
imfile not tracking updates in all files in directory using wildcard


andrewpbarry:
### Expected behavior
rsyslogd is configured to read in all of the log files generated by kubernetes. These are in /var/log/containers/*.log, which are symlinks kubernetes creates to the actual log files, which reside in /var/log/pods/$podname/$containername/X.log
Files in /var/log/pods, and the matching symlinks in /var/log/containers come and go, though they exist for a minimum of many hours. I would expect with this configuration that ALL files would be tracked by rsyslogd.

### Actual behavior
I've found cases where rsyslogd is not collecting data from some of the log files in the configured directory. 
log_files=`ls -l /var/log/containers/|tr -s ' '|cut -d' ' -f11|sort|uniq`
rsyslog_fds=`ls -l /proc/$rsyslog_pid/fd/ |grep -e '.log$'|tr -s ' ' |cut -d' ' -f11|sort|uniq`
On a system with 352 container logs, rsyslogd only has open filedescriptors for 324 of them. 
e.g. /var/log/pods/services_cray-cfs-batcher-8f4d587fc-8hhnj_ab979fb1-e6ad-4349-aea0-a7ae827a55f9/istio-proxy/0.log
Rsyslogd has been in this state for several hours. 

### Steps to reproduce the behavior
Create several hundred symlinks in a directory, linked to real log files in the filesystem, and ask rsyslogd to collect them with imfile. 
This is not simple to recreate. Out of 32 nodes, each with several hundred log files, one node has 23 untracked files, one node has 22, one has 15, and all of the rest have 0 untracked log files. kubernetes and rsyslogd have been running for several weeks. 
Looking for ideas on what's going wrong and what to do to get rsyslogd to rescan the file list. 

### Environment
- rsyslog version: 8.2304.0
- platform: hostos SLES15 ; container image ubuntu:22.04
- for configuration questions/issues, include rsyslog.conf and included config files
# rsyslog.conf
$MaxMessageSize 64k
module(load="imfile")
module(load="mmjsonparse")
module(load="mmnormalize")
#module(load="mmkubernetes")
module(load="impstats" interval="300")

global(workDirectory="/var/log/cray/collector")

input(type="imfile"
       File="/var/log/containers/*.log"
       PersistStateInterval="256"
       Tag="docker_container"
       addMetadata="on"
       Severity="debug"
       ruleset="collected")

template(name="json_data" type="list") {
  constant(value=" { ")
[...]

PR #5767, opened 2025-07-11
Fix imfile inode check

rgerhards:
## Summary
- imfile: check file descriptor when verifying inode

Fixes issue https://github.com/rsyslog/rsyslog/issues/5176

## Testing
- `python3 devtools/rsyslog_stylecheck.py plugins/imfile/imfile.c`

------
https://chatgpt.com/codex/tasks/task_e_6871640d371883328778c0871905c1fc


rgerhards:
this does not look right, we need to investigate from the root cause again.

Issue #2547, opened 2018-03-16
imtcp with gtls don't expose full certificate chain to clients


cvuillemez:
### Expected behavior
```
# echo | openssl s_client -connect localhost:11514 2>/dev/null |awk '/Certificate chain/,/---/'
Certificate chain
 0 s:/C=FR/O=Foo/OU=bar/CN=bar.net/emailAddress=foo@bar.net 
   i:/C=FR/O=Foo/OU=bar/CN=intermediate Server
 1 s:/C=FR/O=Foo/OU=bar/CN=intermediate Server
   i:/C=FR/O=Foo/OU=bar/CN=Root CA
---
```

### Actual behavior
```
$ echo | openssl s_client -connect localhost:11514 2>/dev/null |awk '/Certificate chain/,/---/'
Certificate chain
 0 s:/C=FR/O=Foo/OU=bar/CN=bar.net/emailAddress=foo@bar.net 
   i:/C=FR/O=Foo/OU=bar/CN=intermediate Server
---
```

### Steps to reproduce the behavior
Put the intermediate cert in /etc/pki/rsyslog/ca.pem & /etc/pki/rsyslog/cert.pem

### Environment
- rsyslog version: 8.33.1
- platform: Ubuntu 16.04
- rsyslog.conf:
```
# provides TLS TCP syslog reception
$DefaultNetstreamDriverCAFile /etc/pki/rsyslog/ca.pem
$DefaultNetstreamDriverCertFile /etc/pki/rsyslog/cert.pem
$DefaultNetstreamDriverKeyFile /etc/pki/rsyslog/key.pem

module(load="imtcp"
       streamdriver.name="gtls"
       streamdriver.mode="1"
       streamdriver.authmode="anon"
       gnutlsprioritystring="SECURE256:SECURE128:-VERS-TLS1.0:-VERS-TLS1.1:-RSA:-CAMELLIA-128-GCM:-CAMELLIA-256-GCM:-CAMELLIA-128-CBC:-CAMELLIA-256-CBC:-AES-128-CBC:-AES-256-CBC"
#       permittedpeer="*.net"
)
[...]
input(type="imtcp" port="11514" ruleset="incominglogs")
```

*Note: rsyslog documentation is available here*:
  [pre-release](http://www.rsyslog.com/doc/master/),
  [current stable release](http://www.rsyslog.com/doc/v8-stable/)

ajrpayne:
Rsyslog 8.33.0 from from git checkout.
[...]

PR #5772, opened 2025-07-12
Document GTLS certificate chain behavior

rgerhards:
## Summary
- clarify that the GTLS driver sends all certificates in the configured cert file
- note this resolves [issue #2547](https://github.com/rsyslog/rsyslog/issues/2547)

## Testing
- `python3 devtools/rsyslog_stylecheck.py doc/source/configuration/modules/imtcp.rst`

------
https://chatgpt.com/codex/tasks/task_e_68716f81b39c8332bafeabccf7516389

### DIFF
--- a/doc/source/configuration/modules/imtcp.rst
+++ b/doc/source/configuration/modules/imtcp.rst
@@ -500,6 +500,15 @@ https://docs.openssl.org/1.1.1/man3/SSL_CTX_set_verify/
 For GnuTLS, the default is 5 - see the doc for more:
 https://www.gnutls.org/manual/gnutls.html
 
+.. note::
+
+   The GnuTLS driver sends all certificates contained in the file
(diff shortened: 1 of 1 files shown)


Cropi:
Hello @rgerhards,
I was troubleshooting a similar issue with intermediate certificates. Am I correct in understanding that the requirement to provide the full certificate chain applies to both directions? In other words, it’s not only the server that needs to include intermediate certificates, but also the client (e.g., omfwd) when mutual TLS is used, right?

rgerhards:
> In other words, it’s not only the server that needs to include intermediate certificates, but also the client (e.g., omfwd) when mutual TLS is used, right?

I think so, because the client must also be able to validate the cert, and it usually needs to full chain (95% sure). @alorbach pls jump in if I am wrong.

PR #5794, opened 2025-07-15
Investigate and resolve rsyslog issue 3872

alorbach:
Fix: Initialize `offMSG` to 0 to prevent undefined behavior

The `offMSG` field in `smsg_t` was initialized to -1. This can lead to undefined behavior, crashes, or memory corruption when `offMSG` is used as an array offset (e.g., with `memcpy`) before being explicitly set to a valid value.

This change initializes `offMSG` to 0, ensuring it is always a valid offset and preventing potential issues. No existing code relies on `offMSG` being -1.

closes: https://github.com/rsyslog/rsyslog/issues/3872


rgerhards (changes requested):
binary test_offmsg is included in commit. This cannot be merged and needs to be removed.

alorbach:
> binary test_offmsg is included in commit. This cannot be merged and needs to be removed.

Removed, the commits needs to be sqashed, I will modify my prompt so cursor doesn't add binary or testfiles again.

PR #5834, opened 2025-07-19
Tests autobuild

rgerhards:


rgerhards:
FYI: I originally posted this patch together with another one. Corrected it quickly. Thus the first invalid Gemini infos.

Issue #5864, opened 2025-07-24
[Question]: imjournal: open() failed for path: '/var/lib/rsyslog/imjournal.state.tmp': Operation not permitted


kevgrig:
### 📝 Your Question / What you need help with:

Related to issue #5375.

```
rsyslogd[1284]: imjournal: open() failed for path: '/var/lib/rsyslog/imjournal.state.tmp': Operation not permitted [v8.2412.0-3.fc42 try https://www.rsyslog.com/e/2433 ]
```

### 💻 Environment:

* **Rsyslog Version:** `8.2412.0-3.fc42`
* **Platform:** `Linux Fedora 42 x64`
* **Kernel Version:** `6.15.7-200.fc42.x86_64`
* **Deployment Method:** `Digital Ocean`

### ⚙️ Configuration (if applicable):

```
$ cat /etc/rsyslog.conf
$ModLoad imjournal # provides access to the systemd journal
$ModLoad imudp
$UDPServerRun 514
$UDPServerAddress 127.0.0.1
$WorkDirectory /var/lib/rsyslog
$ActionFileDefaultTemplate RSYSLOG_TraditionalFileFormat
$IncludeConfig /etc/rsyslog.d/*.conf
$IMJournalStateFile imjournal.state
$IMJournalIgnorePreviousMessages on
*.info;mail.none;authpriv.none;cron.none                /var/log/messages
authpriv.*                                              /var/log/secure
mail.*                                                  -/var/log/maillog
cron.*                                                  /var/log/cron
*.emerg                                                 :omusrmsg:*
uucp,news.crit                                          /var/log/spooler
local7.*                                                /var/log/boot.log
$ cat /etc/rsyslog.d/*.conf
:syslogtag, isequal, "[CLOUDINIT]" /var/log/cloud-init.log
& stop
```

### ➕ Additional Context:

Resolved with:

```
#$ModLoad imjournal # provides access to the systemd journal
module(load="imjournal")
```

rgerhards:
This seems to be an incarnation of what is being fixed here: https://github.com/rsyslog/rsyslog/pull/5871

PR #5871, opened 2025-07-28
imjournal: fix vulnerability in state file creation

### LINKED ISSUE #5375: BUG in RH packaging: rsyslogd  8.2310.0-4.el9 throws imjournal: open() failed for path: '/var/lib/rsyslog/imjournal.state.tmp': Operation not permitted [v8.2310.0-4.el9 try https://www.rsyslog.com/e/2433 ]
### Expected behavior
No errors

### Actual behavior
Upon upgrading rsyslog from a previous working version (rsyslog-8.2102.0-117.el9.x86_64.rpm) to the latest offered within Oracle 9 repos (just recently added) rsyslog-8.2310.0-4.el9.x86_64.rpm results in:
```
imjournal: open() failed for path: '/var/lib/rsyslog/imjournal.state.tmp': Operation not permitted [v8.2310.0-4.el9 try https://www.rsyslog.com/e/2433 ]
```

No idea what the real impact of this is either.

Upgrade to rsyslog-8.2310.0 using yum/dnf.

### Environment
- rsyslog version: 
BROKEN:
```
rsyslogd  8.2310.0-4.el9 (aka 2023.10) compiled with:
        PLATFORM:                               x86_64-redhat-linux-gnu
        PLATFORM (lsb_release -d):
        FEATURE_REGEXP:                         Yes
        GSSAPI Kerberos 5 support:              Yes
        FEATURE_DEBUG (debug build, slow code): No
        32bit Atomic operations supported:      Yes
        64bit Atomic operations supported:      Yes
        memory allocator:                       system default
        Runtime Instrumentation (slow code):    No
        uuid support:                           Yes
        systemd support:                        Yes
        Config file:                            /etc/rsyslog.conf
        PID file:                               /var/run/rsyslogd.pid

See https://www.rsyslog.com for more information.

```

WORKING:
```
rsyslogd  8.2102.0-117.el9 (aka 2021.02) compiled with:
        PLATFORM:                               x86_64-redhat-linux-gnu
        PLATFORM (lsb_release -d):
        FEATURE_REGEXP:                         Yes
        GSSAPI Kerberos 5 support:              Yes
        FEATURE_DEBUG (debug build, slow code): No
        32bit Atomic operations supported:      Yes
        64bit Atomic operations supported:      Yes
        memory allocator:                       system default
        Runtime Instrumentation (slow code):    No
        uuid support:                           Yes
        systemd support:                        Yes
        Config file:                            /etc/rsyslog.conf
        PID file:                               /var/run/rsyslogd.pid

See https://www.rsyslog.com for more information.

```
- platform: Oracle Linux 9
- for configuration questions/issues, include rsyslog.conf:
```
# rsyslog configuration file

# For more information see /usr/share/doc/rsyslog-*/rsyslog_conf.html
# If you experience problems, see http://www.rsyslog.com/doc/troubleshoot.html

#### MODULES ####

# The imjournal module bellow is now used as a message source instead of imuxsock.
$ModLoad imuxsock # provides support for local system logging (e.g. via logger command)
$ModLoad imjournal # provides access to the systemd journal
#$ModLoad imklog # reads kernel messages (the same are read from journald)
#$ModLoad immark  # provides --MARK-- message capability

# Provides UDP syslog reception
#$ModLoad imudp
#$UDPServerRun 514

# Provides TCP syslog reception
#$ModLoad imtcp
#$InputTCPServerRun 514


#### GLOBAL DIRECTIVES ####

# Where to place auxiliary files

# create mode of files - CIS Item 4.2.1.3
$FileCreateMode 0640

#Disablerate-limitingof log entries
$SystemLogRateLimitInterval 60
$SystemLogRateLimitBurst 1500000

# Use default timestamp format
$ActionFileDefaultTemplate RSYSLOG_TraditionalFileFormat

# File syncing capability is disabled by default. This feature is usually not required,
# not useful and an extreme performance hit
#$ActionFileEnableSync on

# Include all config files in /etc/rsyslog.d/

# Turn off message reception via local log socket;
# local messages are retrieved through imjournal now.
$OmitLocalLogging on

# File to store the position in the journal

# changes to journal rate limits - see packer build
$imjournalRatelimitInterval 60
$imjournalRatelimitBurst 1500000

#### RULES ####

# Log all kernel messages to the console.
# Logging much else clutters up the screen.
#kern.*                                                 /dev/console

# Log anything (except mail) of level info or higher.
# Don't log private authentication messages!
*.info;mail.none;authpriv.none;cron.none                /var/log/messages

# The authpriv file has restricted access.
authpriv.*                                              /var/log/secure

# Log all the mail messages in one place.
mail.*                                                  -/var/log/maillog


# Log cron stuff
cron.*                                                  /var/log/cron

# Everybody gets emergency messages
*.emerg                                                 :omusrmsg:*

# Save news errors of level crit and higher in a special file.
uucp,news.crit                                          /var/log/spooler

# Save boot messages also to boot.log
local7.*                                                /var/log/boot.log

```

Cropi:
Initialize cs.fCreateMode to 0644 instead of -1 to prevent state files from being created with all permission bits set when legacy configuration is used. The setModCnf function is only called with modern configuration, leaving cs.fCreateMode at -1 for legacy configs, which results in wrong open() calls.

By default, the state file will now use a hardcoded permission of 0644 for legacy syntax, while remaining configurable through the new syntax. Since rsyslog is the sole consumer of this file (at least should be), I will maintain FileCreateMode at 0644 to avoid introducing additional issues.

Side note: another approach would be to introduce the legacy version of this parameter via:
```
CHKiRet(omsdRegCFSLineHdlr((uchar *)"filecreatemode", 0, eCmdHdlrFileCreateMode, NULL, &cs.fCreateMode, STD_LOADABLE_MODULE_ID));
```
However, adding new legacy options is not being considered.

What do you think? (for some reason I am not able to mention anyone who left a comment in #5375)
Fixes #5375 
Fixes https://github.com/rsyslog/rsyslog/issues/5864

### DIFF
--- a/plugins/imjournal/imjournal.c
+++ b/plugins/imjournal/imjournal.c
@@ -1039,7 +1039,7 @@ BEGINbeginCnfLoad
     cs.bIgnoreNonValidStatefile = 1;
     cs.iPersistStateInterval = DFLT_persiststateinterval;
     cs.stateFile = NULL;
-    cs.fCreateMode = -1;
+    cs.fCreateMode = 0644;
     cs.ratelimitBurst = 20000;
     cs.ratelimitInterval = 600;
     cs.iDfltSeverity = DFLT_SEVERITY;
@@ -1314,16 +1314,6 @@ BEGINsetModCnf
                 "param '%s' in beginCnfLoad\n",
                 modpblk.descr[i].name);
         }
-    }
-
-    /* File create mode is not set */
-    if (cs.fCreateMode == -1) {
-        const int fCreateMode = 0644;
-        LogMsg(0, RS_RET_OK_WARN, LOG_WARNING,
-               "imjournal: filecreatemode is not set, "
-               "using default %04o",
-               fCreateMode);
-        cs.fCreateMode = fCreateMode;
     }
 
 finalize_it:


rgerhards:
> Side note: another approach would be to introduce the legacy version of this parameter via:

pls not - I am right now phasing out legacy use in the doc. I reviewed of a lot of postings on various places has really shown that folks do not understand the different config formats and a suggested single one would be best. In that light, it would also not hurt to replace any $xxx directives in the default config with the much-easier-to-understand modern counterpart.

alorbach:
@Cursor review

rgerhards:
@Cropi do you like to remove the dead code mentioned above or shall I do this (makes me become the commitor, as you know)?

Cropi:
> @Cropi do you like to remove the dead code mentioned above or shall I do this (makes me become the commitor, as you know)?

You mean the dead code that is referenced by Gemini? I am not sure where it points to, because I've already removed that specific `if (cs.fCreateMode == -1) {` part. It was part of the initial PR. So please feel free to correct it yourself in case I am missing something. I am fine with those consequences.

rgerhards:
@Cropi my bad, sry - I seem to have overlooked the follow-up commit push

PR #5873, opened 2025-07-28
ossl: Add support for PrioritizeSAN option to OpenSSL network stream driver

csiltala:
Add OpenSSL equivalent of existing GnuTLS prioritizeSAN functionality implemented in https://github.com/rsyslog/rsyslog/commit/937e278fdf51c3e21ebd3acf05f9d1b8649ce2c5

https://docs.openssl.org/1.1.1/man3/X509_check_host/#description
> The X509_CHECK_FLAG_NEVER_CHECK_SUBJECT flag causes the function to never consider the subject DN even if the certificate contains no subject alternative names of the right type (DNS name or email address as appropriate); the default is to use the subject DN when no corresponding subject alternative names are present.

`0x10100004L` corresponds to OpenSSL 1.1.0-pre4 in which `X509_CHECK_FLAG_NEVER_CHECK_SUBJECT` first appeared, unsure if it would be better to compare against the official 1.1.0 release

### DIFF
--- a/doc/source/concepts/ns_ossl.rst
+++ b/doc/source/concepts/ns_ossl.rst
@@ -74,3 +74,14 @@ Supported Authentication Modes
       anon cipher will be used.
 
 
+PrioritizeSAN
+=============
+
+-  **off** - by default this binary argument is turned off, which means
+   that validation of names in certificates goes per older RFC 5280 and either
+   Subject Alternative Name or Common Name match is good and connection is
+   allowed.
+
+-  **on** - if you turn this option on, it will perform stricter name checking
+   as per newer RFC 6125, where, if any SAN is found, contents of CN are 
+   completely ignored and name validity is decided based on SAN only. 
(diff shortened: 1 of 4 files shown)


alorbach:
@Cursor review

rgerhards:
@csiltala Thx for the PR. There seem to be some issues that need your attention. It would be great if you could check.

csiltala:
It's not totally clear to me whether the construction/initialization macros were or were not already zero-initializing the new variable but I think setting it here to satisfy Cursor should be fine?

As far as the `OPENSSL_VERSION_NUMBER` check, the flag would not be present in `0x10100001L`-`0x10100003L` (although the odds of anyone using 1.1.0-pre1 through 1.1.0-pre3 are slim to none), I think it would make more sense to leave it as-is or change to `0x1010000fL` (1.1.0 release) rather than `0x10100000L`.

rgerhards:
I asked @alorbach to have a look as he wrote this module. I think we should try to evaluate the other feature checks, but am not 100% sure myself.

csiltala:
@rgerhards In the case that I can address them before his review, what do you mean by other feature checks? I also believe the `buildbot/rsyslog centos7-5` failure to be transient but would appreciate a re-run in any case

Issue #5880, opened 2025-07-31
[General]: omfwd-lb-1target-retry-1_byte_buf test is flaky


mbiebl:
The Debian package runs `make check` during build.

This has caused the build to fail sporadically.

One instance is e.g. https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1094069 where `omfwd-lb-1target-retry-1_byte_buf` failed.

Another example is

https://debusine.debian.net/debian/developers/artifact/2145996/

It doesn't happen often that  `omfwd-lb-1target-retry-1_byte_buf` fails, but often enough that every now and then it shows up on my radar.

Any ideas why the test might fail and how to make it more robust?

mbiebl:
```
FAIL: omfwd-lb-1target-retry-1_byte_buf-TargetFail
==================================================

testbench: TZ env var not set, setting it to UTC
------------------------------------------------------------
20:14:27[0]  Test: ./omfwd-lb-1target-retry-1_byte_buf-TargetFail.sh
------------------------------------------------------------
waiting for file rstb_708704_ca200ac7VdBE.minitcpsrvr_port1
writing to file rstb_708704_ca200ac7VdBE.out.log
listen on target port 0
### background minitcpsrv process id is 23826 port 43515 ###
Note: intentionally not started any local TCP receiver!
config rstb_708704_ca200ac7VdBE_.conf is:
     1	module(load="../plugins/imdiag/.libs/imdiag")
     2	global(inputs.timeout.shutdown="60000"
     3	       default.action.queue.timeoutshutdown="20000"
     4	       default.action.queue.timeoutEnqueue="20000"
     5	       debug.abortOnProgramError="on")
     6	# use legacy-style for the following settings so that we can override if needed
     7	$MainmsgQueueTimeoutEnqueue 20000
     8	$MainmsgQueueTimeoutShutdown 10000
     9	$IMDiagListenPortFileName rstb_708704_ca200ac7VdBE.imdiag.port
    10	$IMDiagServerRun 0
    11	$IMDiagAbortTimeout 580
    12	
    13	:syslogtag, contains, "rsyslogd"  ./rstb_708704_ca200ac7VdBE.started
    14	###### end of testbench instrumentation part, test conf follows:
    15	
    16	$MainMsgQueueTimeoutShutdown 10000
    17	$MainMsgQueueDequeueBatchSize 100
    18	
    19	global(allMessagesToStderr="on")
    20	
[...]

PR #5881, opened 2025-07-31
testbench: fix some omfwd-lb-1target* tests - they were very flaky

rgerhards:
It looks like at some point a change purely intend for local testing went into git, rendering the test incorrect. Still needs to be finally confirmed.

potentially fixes https://github.com/rsyslog/rsyslog/issues/5880


rgerhards:
Let's bring this to daily use and see how it works in practice. Any new issues will surface quickly and can be followed up.

Issue #5886, opened 2025-08-01
[Bug]: tests: fix GnuTLS session cleanup bug in tcpflo...


alorbach:
### Stash Information
- **Stash Ref:** `stash@{1}`
- **Label:** `bug`
- **Created:** 2025-08-01 09:32:11 UTC

### Changed Files
```
tests/tcpflood.c
```

### Stash Diff
```diff
diff --git a/tests/tcpflood.c b/tests/tcpflood.c
index 2586e0cc5..e51a44b5b 100644
--- a/tests/tcpflood.c
+++ b/tests/tcpflood.c
@@ -907,6 +907,9 @@ int sendMessages(struct instdata *inst) {
                     offsSendBuf = 0;
                 }
                 ++nConnDrops;
+                if (transport == TP_TLS) {
+                    closeTLSSess(socknum);
+                }
                 close(sockArray[socknum]);
                 sockArray[socknum] = -1;
             }
@@ -1711,6 +1714,13 @@ static void initTLS(void) {
     #pragma GCC diagnostic ignored "-Wint-to-pointer-cast"
 static void initTLSSess(const int i) {
     int r;
+
+    /* Clean up any existing session before creating a new one */
+    if (sessArray[i] != NULL) {
+        gnutls_deinit(sessArray[i]);
+        sessArray[i] = NULL;
+    }
+
     gnutls_init(sessArray + i, GNUTLS_CLIENT);
 
     /* Use default priorities */
@@ -1755,8 +1765,11 @@ static int sendTLS(int i, char *buf, size_t lenBuf) {
 }
 
 static void closeTLSSess(int i) {
-    gnutls_bye(sessArray[i], GNUTLS_SHUT_RDWR);
-    gnutls_deinit(sessArray[i]);
+    if (sessArray[i] != NULL) {
+        gnutls_bye(sessArray[i], GNUTLS_SHUT_RDWR);
+        gnutls_deinit(sessArray[i]);
+        sessArray[i] = NULL;
+    }
 }
 #else /* NO TLS available */
 static void initTLS(void) {}
```

---
*Issue created automatically from stash using stash2issue.sh*

PR #5887, opened 2025-08-01
tests: fix GnuTLS session cleanup bug in tcpflood …

alorbach:
Causing macOS ARM64 crashes Fix a critical memory corruption bug in tcpflood that was causing segmentation faults on macOS 14 ARM64 during TLS connection re-establishment. 

Problem: 
- When connections were randomly dropped, tcpflood properly closed sockets but failed to cleanup corresponding GnuTLS sessions in sessArray 
- On connection re-establishment, initTLSSess() would call gnutls_init() which overwrote old session pointers without proper cleanup 
- This led to memory corruption and crashes in gnutls_record_send() 
- Core dump showed fault address 0xf8 indicating corrupted session structure Solution: - Add proper GnuTLS session cleanup when dropping connections randomly 
- Add safety check in initTLSSess() to cleanup existing sessions before creating new ones 
- Enhance closeTLSSess() with null checks and proper session invalidation This resolves crashes in TLS-related tests on macOS 14 ARM64, particularly the imtcp-drvr-in-input-basic test that was failing with core dumps.

Fixes: https://github.com/rsyslog/rsyslog/issues/5886

### DIFF
--- a/tests/tcpflood.c
+++ b/tests/tcpflood.c
@@ -119,6 +119,8 @@
 #include <sys/resource.h>
 #include <sys/time.h>
 #include <errno.h>
+#include <execinfo.h>
+#include <ucontext.h>
 #ifdef ENABLE_GNUTLS
     #include <gnutls/gnutls.h>
     #if GNUTLS_VERSION_NUMBER <= 0x020b00
@@ -148,6 +150,7 @@ GCRY_THREAD_OPTION_PTHREAD_IMPL;
     #endif
 
 #endif
+
 
 char *test_rs_strerror_r(int errnum, char *buf, size_t buflen) {
 #ifndef HAVE_STRERROR_R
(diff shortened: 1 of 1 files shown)


tests/tcpflood.c
```diff
@@ -273,6 +276,62 @@ static struct sockaddr_in udpRcvr; /* remote receiver in UDP mode */
...
+    fprintf(stderr, "TCPFLOOD: Connections: %d\n", numConnections);
+    fprintf(stderr, "TCPFLOOD: Messages to send: %d\n", numMsgsToSend);
+
+    fprintf(stderr, "=== END TCPFLOOD CRASH REPORT ===\n");
+    fflush(stderr);
+
+    /* Exit with the signal number */
+    exit(sig);
+}
+
```
rgerhards, line 334: I am not sure this is a good idea - yrs ago we had this for main rsyslog, but core dumps are much better. Given tcpflood is a stesting too, we always know exactly what it did. But I do no object adding it, just a gentle reminder to think it is worth the extra maintenance in the future.

rgerhards:
overall, this looks like a valuable patch to me!

tests/tcpflood.c
alorbach: Ok but core dump of tcpflood are not handled in diag.sh which was the reason I created that PR.
Shall we rather detect and analyze coredumps of tcpflood in diag.sh then?

rgerhards: as I said, I do not object it. Not sure if it helps so much here. Addinig auto-coredump for tcpflood would indeed a good thing to have, but it must be carefully implemented (we need to differentiate between tcpflood and rsyslog as core generator)

alorbach: Shall we give it a try? When I created this PR, I had one race crash today which is the reason to have a core dump handler added into tcpflood.

rgerhards: we can - but did it provide you any extra value?

PR #5908, opened 2025-08-11
queue engine: implement proper batch state handling

alorbach:
For transactional actions

Potential fixe for issue #2420 where committed batches with unknown transmission status could lead to message loss. When a transactional commit fails or has an uncertain outcome, all messages in the current batch are now reset to BATCH_STATE_RDY to ensure at-least-once delivery.

Changes:
* action.c: Modified processBatchMain() to stage transactional messages as BATCH_STATE_SUB during processing, then transition to COMM on successful commit or RDY on failure/unknown status
* action.h: Updated actionCommitAllDirect() signature to return status
* runtime/ruleset.c: Added batch state reset logic for direct queue commit failures to maintain consistency across queue types
* tests/omprog-transactions-failed-messages.sh: Enhanced test robustness to handle environment issues while preserving validation logic

Resolves: #2420


alorbach:
> This "smells" a bit suspicious, especially the introduction of `OMOD_TX_OLD_RETRY_SEMANTICS`. It would be good to have reasoning for that. My gut feeling is that it is counter rsyslog policies (which are maybe undocumented) and should be done in a different way - if at all.

OMOD_TX_OLD_RETRY_SEMANTICS might be dropped completely - something open for further discussion. There might be cases where modules need the old behavior.

It is an approach which most likely will be discarded but gives some ideas for possible fixes for much more complex issue than thought before.

PR #5935, opened 2025-08-14
ommysql: remove long-standing non-needed TODO item

rgerhards:
This item was create in 2007 and has proven in ~20yrs of practice to be no issue. Even more so, this is now in legacy code that will no longer be updated at all. So the TODO no longer applies and could be removed.


rgerhards:
CI are (obviously) unrelated flakes

PR #5942, opened 2025-08-14
Add namespace support for omjournal

anis-bougrine:


rgerhards:
thx for the PR, but please also update the omjournal doc under doc/source/configuration/modules/omjournal.rst..

Otherwise this feature is not discoverable and cannot be merged. If you have problems with adding the doc, please provide a plain text writeup and I will see that I amend the commit based on that writeup. A config sample would be very useful.

rgerhards (changes requested):
pls also review the AI agent comments - they look valid.

I would appreciate if you could fix the "ommongodb" copy&paste error while you are at it.

anis-bougrine:
@rgerhards I made the required changements and I've added documentation. I hope it's clear.

PR #5958, opened 2025-08-17
Various mongo-c-driver v2 related fixes for ommongodb

mbiebl:


mbiebl:
@jjourdin ^ see commit b8c1e19

mbiebl:
@rgerhards hm, I see that for *very* old versions of libmongoc , there is no `<mongoc/mongoc.h>` but only `<mongoc.h>`. See the build failures of centos7 and ubuntu18.

I did check debian oldoldstable, i.e. https://packages.debian.org/bullseye/amd64/libmongoc-dev/filelist and it already had this header reorganization.

Is ommongodb support for centos7 and ubuntu18 important to you? If so I can rework the configure check and add a fallback.

mbiebl:
@rgerhards the alternative to the above commit e434df2  would be to bump the minimum required version of libmongoc.
Code-wise it would be a bit cleaner but I'll leave the decision up to you.

mbiebl:
I did a bit of digging: The header reorganization happened in mongo-c-driver 1.13.0

```
  * Restructure of install directory. All mongoc headers are under mongoc/
    and all bson headers are under bson/. The preferred way of including the
    headers are mongoc/mongoc.h and bson/bson.h respectively.
    Forwarding headers in the root are provided for backwards compatibility.
    (**Update**: These headers were removed in the 2.0 release)
```
Source https://github.com/mongodb/mongo-c-driver/releases/tag/1.13.0

Released Sep 17, 2018

rgerhards:
> . See the build failures of centos7 and ubuntu18.

will see other postings soon (on the road), but ubuntu 18 is not that important, I'll probably drop it very soon. centos 7 unfortunately still is relevant

mbiebl:
> centos 7 unfortunately still is relevant

Ok, got it. I've rebased the PR a bit accordingly to drop the intermediate commit.

rgerhards:
@mbiebl I probably need to update some of the CI containers. Can you give me the debian package names which I need to install? That would be convenient for me, but I can also dig it up if you are busy. Thx. :-)

mbiebl:
The package names for the the mongo-c-driver: libmongoc-dev

rgerhards:
> The package names for the the mongo-c-driver: libmongoc-dev

thx, but as it looks I was under a wrong impression, no other library to install (sometime there are).

mbiebl:
A bit of warning: While those minimal changes were sufficient to recompile against mongo-c-driver v2, I could not actually test if the resulting ommongodb works as expected due to lack of a test environment.

The compilation did not throw any deprecation warnings. So I'm a bit concerned about this warning from gemini https://github.com/rsyslog/rsyslog/pull/5958#pullrequestreview-3126270761

```
However, the implementation is incomplete and will lead to a build failure when using mongo-c-driver v2. 
A critical part of the C code still uses a function that has been removed in v2.
 I have provided a critical review comment with a suggested fix to handle different driver versions correctly,
 which will prevent the build from breaking.
```

I have not seen such a comment how to update the code for v2, did you @rgerhards ?

rgerhards:
> I have not seen such a comment how to update the code for v2, did you @rgerhards ?

mmmhhh... No.

I am actually not a mongodb expert. While the module historically is in plugins and I may have created the first version, it is very seldom used and we seem to do not even have tests for it. My memory is also very weak.

Let me propose this: I'll first see if we can integrate at least one, better more, tests for ommongodb. This strongly depends on how we can package mongo :-( Only after this is done, we touch the code itself.

The backdraw is that I have a lot of high priority tasks now that are pressing, so it could takes weeks to complete. What do you think @mbiebl ?

mbiebl:
Fwiw, I did apply the patch to Debian (since there has been a switch to mongo-c-driver v2 in Debian unstable).
It did compile successfully.

But yes, we only have very few users of the rsyslog-mongodb package and I don't know what percentage comes from Debian unstable

rgerhards:
if you do not object, I would not immediately merge but take time to implement at least a base test. Does that sound good for you? As I said, it may take some weeks...

mbiebl:
> if you do not object, I would not immediately merge but take time to implement at least a base test. Does that sound good for you? As I said, it may take some weeks...

No hurry

rgerhards:
I just talked to @alorbach this morning and we seem to have issues w/ RHEL7 where no package build succeeds. This might be a call to retire centos 7 as well (at least in regard to mongodb).

Issue #5207, opened 2023-08-02
testbench: create additional test for NetstreamDriverCaExtraFiles 


alorbach:
- Create multiple intermediate CA certificates and multiple new client / server certificates ordinated from those intermediate CA's.
- Create a testcase with a server / client testcase where the client presents the intermediate CA to the server. The server only needs to know about the rootCA.

see also: https://github.com/rsyslog/rsyslog/pull/4889

PR #5966, opened 2025-08-18
runtime: fix NetstreamDriverCAExtraFiles parsing

alorbach:
Multiple intermediate CAs were not honored because strtok() modified
the original config value at the first comma. This led to only the
first file being considered. This change preserves the original value
and validates files using a duplicate buffer.

Impact: Users can reliably specify multiple CA intermediates via
NetstreamDriverCAExtraFiles; TLS chains that require intermediates
now validate as expected. Tests added.

Before: parsing mutated the stored string, effectively truncating at
the first comma and ignoring subsequent CA files.

After: we strdup() for validation, leave the original string intact,
and set it only if all files are accessible. Error handling follows
existing RS_RET patterns; temporary buffer is freed on all paths.

Additionally, add a test (OpenSSL backend) that generates a root,
two intermediates, and leaf certs on both client and server. The test
verifies chains with openssl, aligns CN/PermittedPeer, and exercises
a full send/receive path to guard against regressions. Makefile is
updated to include the new test in TESTS and EXTRA_DIST.

Additionally, in devtools/devcontainer.sh, switched printing to
printf 'container cmd: %s\n' \"$*\" and, most importantly, passed
the command as \"$@\" to preserve argument boundaries and
embedded quotes/newlines.

Fixes: https://github.com/rsyslog/rsyslog/issues/5207

### DIFF
--- a/.github/workflows/run_checks.yml
+++ b/.github/workflows/run_checks.yml
@@ -69,7 +69,8 @@ jobs:
           case "${{ matrix.config }}" in
           'centos_7')
               export RSYSLOG_DEV_CONTAINER='rsyslog/rsyslog_dev_base_centos:7'
-              export RSYSLOG_CONFIGURE_OPTIONS_EXTRA='--disable-elasticsearch-tests --disable-kafka-tests --disable-snmp-tests'
+              export RSYSLOG_CONFIGURE_OPTIONS_EXTRA="--disable-elasticsearch-tests \
+                     --disable-kafka-tests --disable-snmp-tests"
               export CI_VALGRIND_SUPPRESSIONS='centos7.supp'
               ;;
           'centos_8')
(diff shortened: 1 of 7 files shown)


.github/workflows/run_checks.yml
```diff
@@ -107,11 +108,15 @@ jobs:
...
                      --disable-kafka-tests --enable-debug"
+              # openssl CLI is required by some tests; ensure it's present in Fedora images
+              export CI_FEDORA_OPENSSL='1'
               ;;
           'fedora_36')
               export RSYSLOG_DEV_CONTAINER='rsyslog/rsyslog_dev_base_fedora:36'
               export RSYSLOG_CONFIGURE_OPTIONS_EXTRA="--disable-elasticsearch-tests \
                      --disable-kafka-tests --enable-debug --enable-imdtls --enable-omdtls"
+              # openssl CLI is required by some tests; ensure it's present in Fedora images
+              export CI_FEDORA_OPENSSL='1'
```
rgerhards: As of the container / CI interaction design, such settings should be made in the container. This permits to apply them automatically to wherever it is correct.

rgerhards (changes requested):
I am not sure if the gemini comments are outdated. If not, one should look at them. I re-init a gemini review. Also note my other comment!

rgerhards:
@alorbach as it looks, the gemini review was recent. Pls check.

.github/workflows/run_checks.yml
alorbach: > As of the container / CI interaction design, such settings should be made in the container. This permits to apply them automatically to wherever it is correct.

Indeed I will handle that manually

rgerhards (changes requested):
pls fix valid AI review point. We should never use strtok() in new code.

alorbach:
> ## Code Review
> This pull request correctly addresses a bug in parsing `NetstreamDriverCAExtraFiles` by using `strdup` to prevent modification of the original string. The new implementation is robust, handling whitespace and potential memory allocation failures properly. The addition of a comprehensive test case for intermediate CAs is excellent and ensures the fix is well-covered. The shell script improvements in `devtools/devcontainer.sh` are also a welcome enhancement for argument handling. My review includes one suggestion to refactor the validation logic in `runtime/glbl.c` for better modularity and maintainability.

Change not necessary - Only suggestion: refactor validation logic in runtime/glbl.c for modularity. 
All other gemini issues fixed

PR #5983, opened 2025-08-19
imhttp: expose stats via Prometheus endpoint

rgerhards:
## Summary
- export rsyslog statistics in Prometheus text format
- document metricsPath/healthCheckPath options, security notes, and usage example
- handle stats retrieval failures by returning HTTP 500
- add optional Basic Auth for metrics and health endpoints and describe metrics-only setups

## Testing
- `autoreconf -i`
- `./configure`
- `make -C contrib/imhttp CFLAGS="-I/usr/include/apr-1.0"`

------
https://chatgpt.com/codex/tasks/task_e_68a47b513b70833298b4d17baa9da363

### DIFF
--- a/contrib/imhttp/imhttp.c
+++ b/contrib/imhttp/imhttp.c
@@ -99,6 +99,8 @@ struct modConfData_s {
     int nOptions;
     char *pszHealthCheckPath;
     char *pszMetricsPath;
+    char *pszHealthCheckAuthFile;
+    char *pszMetricsAuthFile;
 };
 
 struct instanceConf_s {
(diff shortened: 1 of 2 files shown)


rgerhards:
**We need to think about TSAN** The data race is present, but it is acceptable in rsyslog policy, as stats are not 100% precise if that would conflict with high throughput, which is the case here. We somehow need to suppress these TSAN failures reliably (I have not yet looked into that.

rgerhards:
TSAN report on known-racy data access has been suppressed by clang/TSAN specific suppressions.

PR #5984, opened 2025-08-20
omfwd: document and enforce pool-level suspend semantics

alorbach:
Inline code docs clarify when omfwd should suspend an action in pooled
TCP forwarding. This improves maintainability and reduces ambiguity for
future changes and reviews.

Impact: No user-visible behavior change; semantics documented.

Before/After: Before, a per-target flush failure could be read as
suspending the action; after, SUSPENDED is returned only when the pool
has zero active targets.

Technically, commitTransaction no longer marks the action suspended on
a single-target flush failure. We destruct the failing target, keep the
commit OK, and rely on pool stats plus `countActiveTargets()` to decide
pool-wide availability. Only when `nActiveTargets == 0` do we return
RS_RET_SUSPENDED, engaging the action engine's retry. Buffered frames
for a failed target remain with that target and are flushed after
`doTryResume()` re-establishes the connection in a later transaction.
This preserves existing omfwd pool semantics while documenting them in
code comments for clarity.

Refs: no issue

### DIFF
--- a/tools/omfwd.c
+++ b/tools/omfwd.c
@@ -1417,12 +1417,30 @@ BEGINcommitTransaction
                        "Remaining messages will be sent when it is online again.",
                        pWrkrData->wrkrID, pWrkrData->target[j].target_name, pWrkrData->target[j].port);
                 DestructTCPTargetData(&(pWrkrData->target[j]));
+                /* Note: do not return RS_RET_SUSPENDED here. We only return SUSPENDED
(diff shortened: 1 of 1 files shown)


rgerhards (changes requested):
this does not look correct. RS_RET_SUSPENDED must only be returned when no pool member is left active. I do not see where this condition is checked. If it is, please provide a clear inline doxygen comment on how this condition is checked.

Also, mention clearly in commit message why the previous approach did not work.

PR #6042, opened 2025-08-27
doc: correct PermittedPeer example for ns_ossl driver

rgerhards:
## Summary
- replace incorrect StreamDriverPermittedPeers line with a full example using PermittedPeer
- clarify fingerprint description for the OpenSSL network stream driver

## Testing
- `pip install -r doc/requirements.txt`
- `make -C doc html`
- `./devtools/format-code.sh`

------
https://chatgpt.com/codex/tasks/task_e_68af50502dac83328de6f961f758ca00

### DIFF
--- a/doc/source/concepts/ns_ossl.rst
+++ b/doc/source/concepts/ns_ossl.rst
@@ -38,11 +38,21 @@ Supported Authentication Modes
 -  **x509/fingerprint** - certificate fingerprint authentication as
    described in IETF's draft-ietf-syslog-transport-tls-12 Internet draft.
    The fingerprint must be provided as the SHA1 or the SHA256 hex string of
-   the certificate. Multiple values must be separated by comma (,).
-   A valid configuration would be e.G.
+   the certificate. Multiple values must be separated by a comma (,).
+   A sample configuration might look like:
    ::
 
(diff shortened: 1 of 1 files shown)


csiltala:
The previous example isn't necessarily incorrect but incomplete, I'm unsure of the fingerprint case but the `omfwd` module uses `StreamDriverPermittedPeers` as in the following excerpt from the [omfwd documentation](https://github.com/rsyslog/rsyslog/blob/main/doc/source/configuration/modules/omfwd.rst) (and I believe would be a single comma-separated string for multiple values):
```
# Forward logs securely using TLS (RFC5425)
action(
    type="omfwd"
    target="logs.example.com"
    port="6514"                 # Standard port for TLS syslog
    protocol="tcp"
    StreamDriver="ossl"         # OpenSSL for TLS
    StreamDriverMode="1"        # TLS-only mode
    StreamDriverAuthMode="x509/name"
    StreamDriverPermittedPeers="logs.example.com"
)
```

rgerhards:
Upon closer thinking, this needs further review. Not sure it the change is correct as-is.

rgerhards:
this seems to be incorrect, closing for now. If needed, another PR will follow (unlikely)

PR #6055, opened 2025-08-28
tools/rsyslog.conf.5: fix example Port= value

00gh:
This will fix the example given where a Port number is configured. The Port number (514) should be configured as a string value: "514".

### DIFF
--- a/tools/rsyslog.conf.5
+++ b/tools/rsyslog.conf.5
@@ -295,7 +295,7 @@ the omfwd module.
 
 .B Example:
 .RS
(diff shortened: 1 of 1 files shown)


rgerhards (approved):

rgerhards:
@00gh Thanks for your PR, much appreciated. But there seems to be a merge commit part of it, so it needs to be rebased to current main branch. Pls do - or let me know if I should rebase (this needs to be done under my account, so it will have both of our accounts in different roles on the commit). Pls advise.

rgerhards:
I could automatically rebase via github, so everything is fine! Thx again for the PR.

Issue #5957, opened 2025-08-17
[Feature]: Refactor `omhttp` to modern state; upgrade to project-supported module


rgerhards:
### 💡 Feature Description:

Refactor and modernize `omhttp` along three axes:

1. **Transaction API:** Migrate from the older `doAction()`-based transaction interface to the recommended `commitTransaction()` interface used by the action subsystem. The older interface is emulated via a compatibility layer and may subtly diverge in semantics and performance compared to native `commitTransaction()`.

2. **Retry handling (core-native):** Replace the current “secondary **retryRuleset**” pattern with **core-standard retry/backoff** as the default (no per-action retry ruleset required). **Do not deprecate** `retryRuleset`; keep it **optional** for exotic/advanced use-cases and backward compatibility, but treat it as a non-default path that we evaluate critically for blocking behavior in high-volume setups.

3. **HTTP status code semantics (clear, safe defaults):** Align module behavior with HTTP/common-sense delivery semantics and recent core fixes:

   * Treat **1xx/2xx** as success.
   * Treat **3xx** as failure (non-retriable by default; consider follow-redirect support in future).
   * Treat **4xx** as **permanent failure** (non-retriable), unless explicitly listed in `httpignorablecodes`.
   * Treat **5xx** (and transport errors / status 0) as **retriable** by default, consistent with `httpretrycodes` defaults.
     This dovetails with the recent core changes around transaction suspension/retry behavior and test updates that clarified 4xx as non-retriable and pushed retry focus to 5xx. ([[GitHub](https://github.com/rsyslog/rsyslog/pull/5945)][1])

Additionally:

* **Loki:** Acknowledge that Loki is already *partially supported* (e.g., `batch.format=lokirest`), but confirm templates, label handling, and batching details during the refactor to ensure first-class Loki support (we plan to use `omhttp` for Loki in containerized/docker profiles).
* **Generic “profile” configs (meta-config):** Introduce simple pre-baked **profiles** to set multiple parameters at once, e.g., `"profile=hec:<vendor>"` (Splunk, OpenSearch HEC-like, etc.) or `"profile=loki"`. This avoids module forks such as the **HEC-specific** module proposal and streams users toward a single, well-supported HTTP output with profiles. Reference: PR **#5756** proposing `omsplunkhec`. ([[GitHub](https://github.com/rsyslog/rsyslog/pull/5756)][2])

### ❓ Problem it Solves:

[...]

PR #6056, opened 2025-09-01
omhttp: improve batching byte-limit and gzip performance

alorbach:
Summary
- Improves omhttp batching accuracy and throughput with two internal changes:
  - Precise batch size prediction for maxbytes enforcement across formats
  - Reuse of the zlib compressor between requests to reduce CPU overhead

Motivation
- Address “Batching & performance” concerns in omhttp (see linked issue).

Changes
- Batch size prediction:
  - Reworked computeBatchSize() to return accurate on-the-wire size (no terminating NUL).
  - Added computeNextBatchSize() to predict size when adding the next payload, including per-format delimiter/wrapper overhead (newline, jsonarray, kafkarest, lokirest).
  - Updated commitTransaction() to use the exact payload length and computeNextBatchSize() for maxbytes decisions (fixes an off-by-one undercount).
- Compression performance:
  - Reuse the zlib stream per worker with deflateReset() instead of deflateEnd/deflateInit2() on every request; keep final cleanup in freeWrkrInstance.

Backward compatibility
- No config or interface changes.
- Behavior is preserved; maxbytes is now enforced precisely at the boundary (no undercount).
- Batching formats and defaults unchanged.

Testing
- No new tests added (non-functional/internal change). Existing omhttp tests continue to cover batching behavior.
- Formatting normalized via ./devtools/format-code.sh.

Documentation
- No /doc updates required; user-facing behavior and configuration remain the same.

Housekeeping
- Updated AGENTS.md to require running ./devtools/format-code.sh before commit/push and added clang-format installation notes.

closes: https://github.com/rsyslog/rsyslog/issues/5957

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -40,14 +40,13 @@ AI agents should follow this process:
 5.  Git blame hygiene
     Formatting-only commits listed in .git-blame-ignore-revs.
 
+Prerequisite: install `clang-format` so the formatter can run
+
+- Ubuntu/Debian: `sudo apt-get update && sudo apt-get install -y clang-format`
+- macOS (Homebrew): `brew install clang-format`
+- Fedora/RHEL: `sudo dnf install clang-tools-extra`
+
 AI Agent Note: Always run `./devtools/format-code.sh` as the final step before commit and push. CI will reject PRs with formatting violations.
-
-Tooling prerequisites (formatting)
-
-- Ensure `clang-format` is installed to run the formatter.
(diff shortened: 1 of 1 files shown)


rgerhards (commented):
Once done, the change to agents.md should be clearly in its own commit with strong reasoning. I do not object to carry on here as a testbed, but that change shall not be hidden in some other context.

alorbach:
> Once done, the change to agents.md should be clearly in its own commit with strong reasoning. I do not object to carry on here as a testbed, but that change shall not be hidden in some other context.

I will split it from this branch

PR #6070, opened 2025-09-03
Implement omhttp batching and performance improvements

alorbach:
- Improve batch sizing: predict serialized size including format overhead and separators to honor batch.maxbytes precisely, minimizing premature flushes and preventing oversize payloads.
- Add computeDeltaExtraOnAppend to check thresholds before appending each record.
- Fix dynrestpath batching: flush only when rest path changes; reinitialize and re-seed path for the next batch; free old path to avoid leaks.
- Preserve backward compatibility: existing config keys and formats unchanged; default retry via core suspend/resume; retry.ruleset supported for advanced per-message retry handling.
- Update module docs to reflect refined flush conditions, dynrestpath semantics, and default retry behavior.

see also: https://github.com/rsyslog/rsyslog/issues/5957


alorbach:
The latest Gemini review comments in this PR don’t seem particularly relevant to the changes and aren’t very helpful in this case.

Issue #5566, opened 2025-02-17
rsyslog: ftbfs with GCC-15


mbiebl:
Forwarded from https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1097810

```
Package: src:rsyslog
Version: 8.2412.0-1
Severity: important
Tags: sid forky
User: debian-gcc@lists.debian.org
Usertags: ftbfs-gcc-15

[This bug is NOT targeted to the upcoming trixie release]

Please keep this issue open in the bug tracker for the package it
was filed for.  If a fix in another package is required, please
file a bug for the other package (or clone), and add a block in this
package. Please keep the issue open until the package can be built in
a follow-up test rebuild.

The package fails to build in a test rebuild on at least amd64 with
gcc-15/g++-15, but succeeds to build with gcc-14/g++-14. The
severity of this report will be raised before the forky release.

The full build log can be found at:
http://qa-logs.debian.net/2025/02/16/amd64exp/rsyslog_8.2412.0-1_unstable_gccexp.log.gz
The last lines of the build log are at the end of this report.

To build with GCC 15, either set CC=gcc-15 CXX=g++-15 explicitly,
or install the gcc, g++, gfortran, ... packages from experimental.

  apt-get -t=experimental install g++ 

GCC 15 now defaults to the C23/C++23 standards, exposing many FTBFS.
Other Common build failures are new warnings resulting in build failures
with -Werror turned on, or new/dropped symbols in Debian symbols files.
For other C/C++ related build failures see the porting guide at
http://gcc.gnu.org/gcc-15/porting_to.html

[...]
      |                                                                                                 ~~~~~~~~~~~^~~~~~~~
glbl.c:549:1: note: 'setParserEscapeControlCharactersOnReceive' declared here
  549 | setParserEscapeControlCharactersOnReceive(void __attribute__((unused)) *pVal, int pNewVal) {
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
glbl.c:1530:9: error: passing argument 4 of 'regCfSysLineHdlr' from incompatible pointer type [-Wincompatible-pointer-types]
 1530 |         setParserSpaceLFOnReceive, NULL, NULL));
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~
      |         |
      |         rsRetVal (*)(void *, int) {aka enum rsRetVal_ (*)(void *, int)}
[...]

PR #6078, opened 2025-09-03
CI: add gcc 15 to compile checks

rgerhards:
This shall detect gcc 15 specific warnings, which will cause the build to error out.

see also: https://github.com/rsyslog/rsyslog/issues/5566


rgerhards:
handled differently

PR #5397, opened 2024-06-12
config script: add b64_decode function

KGuillemot:
### Description

This PR provides a new Rainerscript function : **b64_decode**.

This function is based on [RFC4648](https://datatracker.ietf.org/doc/html/rfc4648).

Part of the code used in this PR comes from here : https://en.wikibooks.org/wiki/Algorithm_Implementation/Miscellaneous/Base64#C_2.

PR #6080, opened 2025-09-04
config script: add b64_decode function

rgerhards:
This PR provides a new Rainerscript function: b64_decode.

This function is based on RFC4648.

This replaces https://github.com/rsyslog/rsyslog/pull/5397 which was stalled because rebase was needed, original author had no time to respond and maintainer edits were not permitted. So I create a new branch to finally merge.

I will also add some missing parts, like doc and Makefile integration.

### DIFF
--- a/grammar/rainerscript.c
+++ b/grammar/rainerscript.c
@@ -2887,6 +2887,137 @@ static int ATTR_NONNULL()
     return r;
 }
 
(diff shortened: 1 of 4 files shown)


rgerhards:
@KGuillemot This is the replacement PR for base 64. It looks like there are a couple of issues with it, including test failures. I will see that I address them.

PR #6084, opened 2025-09-04
CI: fix strange issue in codecov config file

rgerhards:


rgerhards:
This is most probably unnecessary - but the codecov yml validator seems to cause strange output

rgerhards:
the root issues seems to have somehow resolved itself. Closing this here for now.

PR #6105, opened 2025-09-09
doc: update doc structure doc to match current thinking

rgerhards:


rgerhards:
mistake on my side, will redo correctly

PR #6121, opened 2025-09-11
net: Add NetworkNamespace APIs

billie-alsup:

This moves NetworkNamespace functionality into the net module. This allows the same code to be reused across multiple tools and plugins. The first usage is with omfwd, which is changed to use the common net implementation.  Note the net implementation is based on the original omfwd implementation.  Subsequent PRs will be opened for integrating this into omuxsock and imtcp.

### Notes

The original test case tcp_forwarding_ns_tpl.sh was broken due to use of single quotes rather than double quotes, thus preventing the proper port number argument to be passed to the listener.  Note this test must be run as root.

Development has occurred across Fedora41 and Fedora42, which uses glibc 2.40 and glibc2.41 respectively.  The valgrind suppressions are updated to handle new glibc issues accordingly.

Four new functions are callable through the net module. These allow one to save a handle to the current namespace, switch to a new namespace by name, and restore the namespace from the saved handle.  A fourth higher level wrapper is used to open a socket in a named network namespace, and handles the invocation of the lower level functions. Ideally this would be the only public function, however it simplifies integration into imtcp in the future (which doesn't directly open sockets).  This may change in the future as network namespaces are integrated into more modules and plugins.

### DIFF
--- a/runtime/net.h
+++ b/runtime/net.h
@@ -27,6 +27,7 @@
 #include <arpa/inet.h>
 #include <netinet/in.h>
 #include <sys/socket.h> /* this is needed on HP UX -- rgerhards, 2008-03-04 */
+#include "netns_socket.h"
 
 typedef enum _TCPFRAMINGMODE {
     TCP_FRAMING_OCTET_STUFFING = 0, /* traditional LF-delimited */
@@ -170,8 +171,65 @@ BEGINinterface(net) /* name must also be changed in ENDinterface macro! */
     /* v8 cvthname() signature change -- rgerhards, 2013-01-18 */
     /* v9 create_udp_socket() signature change -- dsahern, 2016-11-11 */
     /* v10 moved data members to rsconf_t -- alakatos, 2021-12-29 */
+
+    /* v11 netns functions -- balsup, 2025-09-11 */
+    /*
+     * @brief Open a socket on the given namespace
+     * @param fdp A place to store the descriptor.  This must not be NULL.
+     *            A failure will store -1 here.
+     * @param domain The communication domain argument to the underlying socket call
+     * @param type The type argument to the underlying socket call
+     * @param protocol The protocol argument to the underlying socket call
+     * @param ns The desired namespace.  This may be NULL or the empty string if
+     *           the current namespace is desired.
+     * @return RS_RET_OK on success, otherwise a failure code.
+     * @details This is a wrapper to socket, allowing one to open a socket in
+     *          a given namespace. For platforms that do not support network
+     *          namespaces, an error will be returned if the ns parameter
+     *          is not NULL or the empty string.
+     */
+    rsRetVal (*netns_socket)(int *fdp, int domain, int type, int protocol, const char *ns);
+
+    /*
+     * @brief Switch to the given network namespace
+     * @param ns The desired namespace.  If this is NULL or the empty string, then
+     *           this function is a no-op.
+     * @return RS_RET_OK on success, otherwise a failure code.
+     * @details For platforms that do not support network namespaces, an error will
+     *          be returned if the ns parameter is not NULL or the empty string.
+     */
+    rsRetVal (*netns_switch)(const char *ns);
+
+    /*
+     * @brief Save a descriptor to the current network namespace
+     * @param fd The location to store the descriptor for the current
+     *           namespace.  This must not not be NULL, and the
+     *           descriptor must be pre-initialized to -1, i.e. *fd == -1
+     *           is a precondition.  This style is to prevent inadvertent
+     *           descriptor leaks that might arise by overwriting a valid
+     *           descriptor.
+     * @return RS_RET_OK on success, otherwise a failure code.
+     * @details For platforms that do not support network namespaces, this
+     *          function is a no-op and will always return RS_RET_OK.
+     */
+    rsRetVal (*netns_save)(int *fd);
+
+    /*
+     * @brief Restore the original network namespace
+     * @param fd A pointer to a descriptor associated with the original
+     *           namespace.  This must not not be NULL.  If the descriptor
+     *           is -1, then this function is a no-op.  A valid descriptor
+     *           is always closed as a side-effect of this function,
+     *           with the descriptor being updated to -1.
+     * @return RS_RET_OK on success, otherwise a failure code.
+     * @details For platforms that do not support network namespaces, this
+     *          function cannot change the network namespace.  However, if
+     *          presented with an fd that is not -1, it will still close
+     *          that fd and reset the value to -1.
+     */
+    rsRetVal (*netns_restore)(int *fd);
 ENDinterface(net)
-#define netCURR_IF_VERSION 10 /* increment whenever you change the interface structure! */
+#define netCURR_IF_VERSION 11 /* increment whenever you change the interface structure! */
 
 /* prototypes */
 PROTOTYPEObj(net);
--- a/runtime/netns_socket.c
+++ b/runtime/netns_socket.c
@@ -1,0 +1,166 @@
+/* Implementation for netns_socket API
+ *
+ * This file is part of rsyslog.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *       -or-
+ *       see COPYING.ASL20 in the source distribution
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "config.h"
+#include <stdio.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <sched.h>
+#include <string.h>
+#include <sys/socket.h>
+#include <sys/types.h>
+
+#include "rsyslog.h"
+#include "debug.h"
+#include "errmsg.h"
+#include "netns_socket.h"
+
+
+/* Change to the given network namespace.
+ * This function based on previous implementation
+ * of tools/omfwd.c function changeToNs.
+ */
+rsRetVal netns_switch(const char *ns) {
+    DEFiRet;
+#ifdef HAVE_SETNS
+    int ns_fd = -1;
+    char *nsPath = NULL;
+
+    if (ns && *ns) {
+        /* Build network namespace path */
+        if (asprintf(&nsPath, "/var/run/netns/%s", ns) == -1) {
+            // Some implementations say nsPath would be undefined on failure
+            nsPath = NULL;
+            LogError(0, RS_RET_OUT_OF_MEMORY, "%s: asprintf failed", __func__);
+            ABORT_FINALIZE(RS_RET_OUT_OF_MEMORY);
+        }
+
+        /* Open file descriptor of destination network namespace */
+        ns_fd = open(nsPath, O_RDONLY);
+        if (ns_fd < 0) {
+            LogError(errno, RS_RET_IO_ERROR, "%s: could not open namespace '%s'", __func__, ns);
+            ABORT_FINALIZE(RS_RET_IO_ERROR);
+        }
+        /* Change to the destination network namespace */
+        if (setns(ns_fd, CLONE_NEWNET) != 0) {
+            LogError(errno, RS_RET_IO_ERROR, "%s: could not change to namespace '%s'", __func__, ns);
+            ABORT_FINALIZE(RS_RET_IO_ERROR);
+        }
+        dbgprintf("%s: changed to network namespace '%s'\n", __func__, ns);
+    }
+finalize_it:
+    free(nsPath);
+    if ((ns_fd >= 0) && (close(ns_fd) != 0)) {
+        LogError(errno, RS_RET_IO_ERROR, "%s: failed to close namespace '%s'", __func__, ns);
+    }
+#else  // ndef HAVE_SETNS
+    if (ns && *ns) {
+        LogError(ENOSYS, RS_RET_VALUE_NOT_SUPPORTED, "%s: could not change to namespace '%s'", __func__, ns);
+        ABORT_FINALIZE(RS_RET_VALUE_NOT_SUPPORTED);
+    }
+finalize_it:
+#endif  // ndef HAVE_SETNS
+    RETiRet;
+}
+
+
+/* Return to the startup network namespace.
+ * This function based on code in tools/omfwd.c
+ */
+rsRetVal ATTR_NONNULL() netns_restore(int *fd) {
+    DEFiRet;
+
+#ifdef HAVE_SETNS
+    if (*fd >= 0) {
+        if (setns(*fd, CLONE_NEWNET) != 0) {
+            LogError(errno, RS_RET_IO_ERROR, "%s: could not return to startup namespace", __func__);
+            ABORT_FINALIZE(RS_RET_IO_ERROR);
+        }
+        dbgprintf("%s: returned to startup network namespace\n", __func__);
+    }
+
+finalize_it:
+#endif  // def HAVE_SETNS
+    if (*fd >= 0 && close(*fd) != 0) {
+        LogError(errno, RS_RET_IO_ERROR, "%s: could not close startup namespace fd", __func__);
+    }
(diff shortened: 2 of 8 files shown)


runtime/netns_socket.c
```diff
@@ -0,0 +1,164 @@
...
+        }
+        dbgprintf("%s: returned to startup network namespace\n", __func__);
+    }
+
+finalize_it:
+#endif // def HAVE_SETNS
+    if (*fd >= 0 && close(*fd) != 0) {
+        LogError(0, RS_RET_IO_ERROR, "%s: could not close startup namespace fd: %s", __func__, strerror(errno));
+        *fd = -1;
+    }
```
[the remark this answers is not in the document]
billie-alsup: Fixed.

runtime/netns_socket.c
```diff
@@ -0,0 +1,164 @@
...
+#ifdef HAVE_SETNS
+    int ns_fd = -1;
+    char *nsPath = NULL;
+
+    if (ns && *ns) {
+        /* Build network namespace path */
+        if (asprintf(&nsPath, "/var/run/netns/%s", ns) == -1) {
+            LogError(0, RS_RET_OUT_OF_MEMORY, "%s: asprintf failed", __func__);
+            ABORT_FINALIZE(RS_RET_OUT_OF_MEMORY);
+        }
```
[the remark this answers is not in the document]
billie-alsup: I guess it depends on what man page you look at (MacOs for example says a failure explicitly sets it to NULL, and other implementations simply leave it untouched, thus it is initialized to NULL).  For safety, I will reset it as suggested.

runtime/netns_socket.c
```diff
@@ -0,0 +1,168 @@
...
+
+        /* Open file descriptor of destination network namespace */
+        ns_fd = open(nsPath, 0);
+        if (ns_fd < 0) {
+            LogError(0, RS_RET_IO_ERROR, "%s: could not change to namespace '%s': %s", __func__, ns, strerror(errno));
+            ABORT_FINALIZE(RS_RET_IO_ERROR);
+        }
+        /* Change to the destination network namespace */
+        if (setns(ns_fd, CLONE_NEWNET) != 0) {
+            LogError(0, RS_RET_IO_ERROR, "%s: could not change to namespace '%s': %s", __func__, ns, strerror(errno));
```
rgerhards: pass errno instead of the initial zero. Then LogError will automatically add strerror(errno) at the end of message with ": " in front of it. this is recommended practice.

runtime/netns_socket.c
```diff
@@ -0,0 +1,168 @@
...
+            // Some implementations say nsPath would be undefined on failure
+            nsPath = NULL;
+            LogError(0, RS_RET_OUT_OF_MEMORY, "%s: asprintf failed", __func__);
+            ABORT_FINALIZE(RS_RET_OUT_OF_MEMORY);
+        }
+
+        /* Open file descriptor of destination network namespace */
+        ns_fd = open(nsPath, 0);
+        if (ns_fd < 0) {
+            LogError(0, RS_RET_IO_ERROR, "%s: could not change to namespace '%s': %s", __func__, ns, strerror(errno));
```
rgerhards: see other comment on LogError

runtime/netns_socket.c
```diff
@@ -0,0 +1,168 @@
...
+    }
+finalize_it:
+    free(nsPath);
+    if ((ns_fd >= 0) && (close(ns_fd) != 0)) {
+        LogError(0, RS_RET_IO_ERROR, "%s: failed to close namespace '%s': %s", __func__, ns, strerror(errno));
+    }
+#else  // ndef HAVE_SETNS
+    if (ns && *ns) {
+        LogError(0, RS_RET_VALUE_NOT_SUPPORTED, "%s: could not change to namespace '%s': %s", __func__, ns,
+                 strerror(ENOSYS));
```
rgerhards: see other comment on LogError - note: I will not flag more instances, I guess you'll do a search in any case. :-)

tests/known_issues.supp
```diff
@@ -71,3 +71,21 @@
    fun:exit
    fun:(below main)
 }
+{
+   <glibc 2.4[01] failures for fc4[12].aarch64>
+   Memcheck:Leak
+   match-leak-kinds: definite
+   fun:malloc
```
rgerhards, line 78: I am not sure if some supressions are too broad - like malloc(). I admit I did not check we already had similiar ones.

runtime/netns_socket.c
```diff
@@ -0,0 +1,168 @@
...
+            LogError(0, RS_RET_IO_ERROR, "%s: could not change to namespace '%s': %s", __func__, ns, strerror(errno));
+            ABORT_FINALIZE(RS_RET_IO_ERROR);
+        }
+        dbgprintf("%s: changed to network namespace '%s'\n", __func__, ns);
+    }
+finalize_it:
+    free(nsPath);
+    if ((ns_fd >= 0) && (close(ns_fd) != 0)) {
+        LogError(0, RS_RET_IO_ERROR, "%s: failed to close namespace '%s': %s", __func__, ns, strerror(errno));
+    }
```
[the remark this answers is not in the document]
rgerhards: I tend to say the current method is overall best as error message is emitted

rgerhards:
@billie-alsup pls also have a look at last AI review

runtime/netns_socket.c
```diff
@@ -0,0 +1,168 @@
...
+        /* Build network namespace path */
+        if (asprintf(&nsPath, "/var/run/netns/%s", ns) == -1) {
+            // Some implementations say nsPath would be undefined on failure
+            nsPath = NULL;
+            LogError(0, RS_RET_OUT_OF_MEMORY, "%s: asprintf failed", __func__);
+            ABORT_FINALIZE(RS_RET_OUT_OF_MEMORY);
+        }
+
+        /* Open file descriptor of destination network namespace */
+        ns_fd = open(nsPath, 0);
```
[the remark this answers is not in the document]
billie-alsup: Fixed in next push.

runtime/netns_socket.c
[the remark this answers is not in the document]
billie-alsup: Yes, I was going to say so as well.  There is nothing one can do about a close error, and a descriptor on these special kernel files has no write buffer outstanding that could be lost.  So it is extremely unlikely that the close can fail for legitimate reasons, and issuing a log in such cases seems sufficient.  A close failure here most likely means there is some faulty logic elsewhere and the file descriptor was inadvertently closed elsewhere by mistake.

runtime/netns_socket.c
billie-alsup: Oh, I didn't realize this functionality existed!  That will clean up the logging considerably.  I will fix all 3 PRs.

runtime/netns_socket.c
billie-alsup: All LogError's addressed in changed files.

runtime/netns_socket.c
billie-alsup: All LogError's addressed in changed files.

tests/known_issues.supp
billie-alsup: This is a suppression only for the specific backtrace to dl_open.

runtime/netns_socket.c
rgerhards: Yeah, we hide it very well ;-)

I'll see that I update at least the header, so it might become more visible in some IDEs. Also, I'll see if I can teach AI to flag it. Finally, I am sure we have some of these in existing code as well (which make the AI problem harder).

tests/known_issues.supp
rgerhards: oh, my bad. Looks like I can no longer read suppression files ;-)

rgerhards: I forgot: I do a new review after next push.

billie-alsup:
New version is pushed. I'm not sure what to do about some test failures on other platforms. They seem unrelated possibly.  For example, es-basic-es7.14.sh failure which doesn't seem to use anything I have touched.

rgerhards:
> They seem unrelated possibly. For example, es-basic-es7.14.sh failure which doesn't seem to use anything I have touched.

we have flakes, unfortunately. I'll look at them, let you know if I think you need to look at something. ES 7 btw needs replacement in CI.

runtime/net.h
```diff
@@ -170,8 +171,13 @@ BEGINinterface(net) /* name must also be changed in ENDinterface macro! */
     /* v8 cvthname() signature change -- rgerhards, 2013-01-18 */
     /* v9 create_udp_socket() signature change -- dsahern, 2016-11-11 */
     /* v10 moved data members to rsconf_t -- alakatos, 2021-12-29 */
+    /* v11 netns functions -- balsup, 2025-09-11 */
+    rsRetVal (*netns_socket)(int *fdp, int domain, int type, int protocol, const char *ns);
```
rgerhards, line 191: from what I learned on LogError(), it may be useful to add doxygen header comments on top of these definitons (I'll post a new PR for LogError soon). I just wanted to mention - I can also do it afterwards. But it's definitely a good and helpful practice, especially for folks who use an IDE (vs. me with vi ;-)).

runtime/net.h
```diff
@@ -170,8 +171,13 @@ BEGINinterface(net) /* name must also be changed in ENDinterface macro! */
...
+    rsRetVal (*netns_switch)(const char *ns);
+    rsRetVal (*netns_save)(int *fd);
+    rsRetVal (*netns_restore)(int *fd);
```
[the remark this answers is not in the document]
billie-alsup: Invalid comment.  They are already together, it's just that switch comes before in the existing code, and gemini moves it to the end.

runtime/netns_socket.c
```diff
@@ -0,0 +1,166 @@
...
+    if ((ns_fd >= 0) && (close(ns_fd) != 0)) {
+        LogError(errno, RS_RET_IO_ERROR, "%s: failed to close namespace '%s'", __func__, ns);
+    }
+#else  // ndef HAVE_SETNS
+    if (ns && *ns) {
+        LogError(ENOSYS, RS_RET_VALUE_NOT_SUPPORTED, "%s: could not change to namespace '%s'", __func__, ns);
+        ABORT_FINALIZE(RS_RET_VALUE_NOT_SUPPORTED);
+    }
+finalize_it:
+#endif  // ndef HAVE_SETNS
```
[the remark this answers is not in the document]
billie-alsup: No, gemini is incorrect.  It is closing the #else // ndef HAVE_SETNS, and so the existing #endif // ndef HAVE_SETNS is correct.

runtime/netns_socket.c
```diff
@@ -0,0 +1,166 @@
...
+    }
+#else /* def HAVE_SETNS */
+    rsRetVal iRet_restore;
+    int ns_fd = -1;
+
+    if (ns && *ns) {
+        CHKiRet(netns_save(&ns_fd));
+        CHKiRet(netns_switch(ns));
+    }
+#endif /* def HAVE_SETNS */
```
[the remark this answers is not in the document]
billie-alsup: Again, gemini is incorrect.  The #endif is closing the #else /* def HAVE_SETNS */

runtime/net.h
billie-alsup: I can add them now.  Give me 30 minutes or so.

rgerhards: great!

runtime/netns_socket.c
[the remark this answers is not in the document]
rgerhards: yeah, Feel free to ignore 2nd+ note, I keep it on my mind.

BUT: it is better if we can correct the "AI Issue" - because that prevents false positives and also helps it better understand  the code when it checks and generates. So I this case, I would be tempted to try a 

```
#endif /* end #else of #ifdef HAVE_SETNS */
```

I know - it's more verbose, but IMHO this really often not only helps AI but also human junior developers which are highly motivated but may not know every trick. Just my 2cts ;-) (and no guarantee it will work).

runtime/net.h
billie-alsup: This is pushed.

runtime/netns_socket.c
billie-alsup: I think we should add a DBGPRINTF_E, dbgprintf_e, or similar that has the additional parameter as well.  That would cleanup some code as well.  Perhaps that is already an open issue waiting for someone to take on.

billie-alsup:
I generally let the originator of a conversation mark the conversation as resolved.  If you want me to mark things resolved when I fix them, let me know.  I don't think you'd see my comments by default if I marked them resolved (you'd have to click 'show resolved' or similar).

rgerhards:
@billie-alsup It's getting late at my location right now. Pls bear with me, further responses over the weekend or monday. I am convinced we are very short of merging. Thx for all your effort.

runtime/netns_socket.c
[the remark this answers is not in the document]
billie-alsup: If there is only an #else, and no other #elif, I can see indicating that you're ending the #else, even though it is more verbose.  It is not something I would normally do though.  But in more complex situations, what would the rule be?  i.e. #if condition / #elif another_condition / #elif another_condition / ... 

It seems more appropriate simply to comment the single block that you are ending, and not try to put a comment describing the entire chain of conditions.  This may simply be a training issue, or perhaps someone has incorrectly added a rule to Gemini indicating that the comment on the #endif should match the first #if.  imo, it should not if there are intervening #else or #elif clauses.

runtime/netns_socket.c
```diff
@@ -0,0 +1,168 @@
...
+            ABORT_FINALIZE(RS_RET_IO_ERROR);
+        }
+        dbgprintf("%s: returned to startup network namespace\n", __func__);
+    }
+
+finalize_it:
+#endif  // def HAVE_SETNS
+    if (*fd >= 0 && close(*fd) != 0) {
+        LogError(0, RS_RET_IO_ERROR, "%s: could not close startup namespace fd: %s", __func__, strerror(errno));
+    }
```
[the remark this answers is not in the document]
billie-alsup: This was previously responded to.  This is a side-effect of force-pushing, we lose context of previous comments.  Perhaps you could reconsider your stance on how commits are merged :-)  My preference is to allow multiple commits for the PR.  That allows the reviewer to look at all the diffs, or just the diffs since their last review.  But they should be squashed before merging to mainline.

billie-alsup:
> @billie-alsup It's getting late at my location right now. Pls bear with me, further responses over the weekend or monday. I am convinced we are very short of merging. Thx for all your effort.

No worries. It's actual a Cisco Day-For-Me (a free day off for the company), so I'll check again next week.  I'll wait to rebase my other two PRs until this one is in its final form and merged. p.s. I'm in Pacific timezone, so it's morning for me.

runtime/netns_socket.c
rgerhards: That's a good idea, and interestingly there is no such issue yet.

runtime/netns_socket.c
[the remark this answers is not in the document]
rgerhards: As I said, I am happy with it as it is. But I think I will take this out into a discussion, as it is an overall theme on how the AI works.

runtime/netns_socket.c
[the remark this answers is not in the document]
rgerhards: That's a valid point. I think I need to change the policy here. But we need to squash before merge, as you say.

### What is the actual goal?

We want a single, consistent commit after merge in git history:

* easy to cherry-pick for distro maintainers for backporting
* incomplete commits will break `git bisect` which is a very important tool when diagnosing regressions


### What does that mean?

Usually, this would result in me doing the final squash. That has some consequences:

* I also appear author-ish on the commit although I am not
* this could diminish credit for contributor
* I need to write the final commit message, maybe missing nuances important to author

Alternatively, I can ask contributor to squash when we are done. This is an extra round-trip and can be received as negative attitude towards  contibutions.

### Best Solution?

I guess we could make this policy more explicit in the warning message generated.
I need the failing commit structure check because otherwise is tend to accidently merge a set of commits that is not consistnet (I admit even with it I have sometimes been too quick).

### Feedback?

What do, @billie-alsup think? 
I'll also probably try to bring up a discussion on that topic, too.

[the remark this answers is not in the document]
rgerhards: see https://github.com/rsyslog/rsyslog/discussions/6134

rgerhards (approved):
@billie-alsup thx - perfect with the new doxygen comments -- this really defines the new standard.

rgerhards:
@billie-alsup thx again for your patience and the high-quality work. Looking forward to the other PRs. I think once everything is done, we should announce it on the rsyslog site. I would love you if could also add some bits to it (of course only if you like!).

PR #5406, opened 2024-07-02
WiP: Replace libgcrypt-config with pkg-config

schrotthaufen:
Libgcrypt 1.11.0 deprecated libgcrypt-config in favor for gpgrt-config, which doesn't emit -lgcrypt. pkg-config does, though.

See also:
 * https://dev.gnupg.org/T7165
 * https://git.gnupg.org/cgi-bin/gitweb.cgi?p=libgcrypt.git;a=commit;h=2db5b5e995c21c5bd9cd193c2ed1109ba9b1a440

This should fix #5405

I only tested this on Arch Linux.

PR #6137, opened 2025-09-15
dev container: do not work-around libgcrypt dev interface change

rgerhards:
there was "workaround" introduced in the container which prevented to detect some issues in libgrypt-related code. This was probably done to get newer version in dev env without the need to adapt code to broken libgrypt API level in new version.

That prevented issues in https://github.com/rsyslog/rsyslog/pull/5406 to be deteced.


rgerhards:
verified to work, also with #6136 (@zhw2101024 thx for your help!)

PR #6155, opened 2025-09-19
CI: fix zookeeper install errors

rgerhards:
Provide different file name for zookeeper tar - it looks like apache changed the name (not sure, but it worked previously).

Also now cached the file on rsyslog.com, so that we have reliable access even on name change or apache download rate-limiting.


rgerhards:
this seems not to solve all issues, but at least some. Current CI failures are other flakes.

PR #6158, opened 2025-09-20
CI: remove outdated ElasticSearch v6 testing

rgerhards:
This is a first step in getting the omelasticsearch testbench in better shape. For now, the ES 6 download requirements are just commented out, to check if that is at least a feasible measure. We also need a quick fix to get CI running again.

Once the initial problem is solved, follow-up patches will be used to modernize the omelasticsearch testbench.


rgerhards:
all gemini comments are explicitely address ion commit message, so safe to ignore.

rgerhards:
this did not really work out. I'll replace it soon with another PR.

PR #6171, opened 2025-09-23
omusrmsg bugfix: potential null pointer dereference causes segfault

century6:
If there is no user in the data base with user ID uid, getpwuid() returns a null pointer value. Therefore, it is necessary to check the return value of the function to prevent segfault caused by dereference of null pointer.

### DIFF
--- a/tools/omusrmsg.c
+++ b/tools/omusrmsg.c
@@ -304,16 +304,26 @@ static rsRetVal wallmsg(uchar *pMsg, instanceData *pData) {
             sdRet = sd_session_get_uid(sessions_list[j], &uid);
             if (sdRet >= 0) {
                 pws = getpwuid(uid);
+
+                if (pws == NULL) {
+                    LogError(0, NO_ERRCODE, "failed to get passwd for userid '%d'\n", uid);
+                    free(sessions_list[j]);
+                    continue;
+                }
+
                 user = pws->pw_name; /* DO NOT FREE, OS/LIB internal memory! */
 
                 if (user == NULL) {
-                    dbgprintf("failed to get username for userid '%d'\n", uid);
+                    LogError(0, NO_ERRCODE, "failed to get username for userid '%d'\n", uid);
+                    free(sessions_list[j]);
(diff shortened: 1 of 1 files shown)


rgerhards:
Thx for the PR, much appreciated,. Pls also have a look at the AI review, it sounds valid. I'll review now as well, as this looks like we are close to merge-ready.

tools/omusrmsg.c
```diff
@@ -304,16 +304,25 @@ static rsRetVal wallmsg(uchar *pMsg, instanceData *pData) {
             sdRet = sd_session_get_uid(sessions_list[j], &uid);
             if (sdRet >= 0) {
                 pws = getpwuid(uid);
+
+                if (pws == NULL) {
+                    dbgprintf("failed to get passwd for userid '%d'\n", uid);
```
rgerhards: While you correctly mimic the other "error reporting", this is actually kind of a pre-existing bug. The `dbgprintf` is not visible to the user. I suggest you change this (and the other occcurences) to `LogError`. I am fine if you do not do this, then I'll step in after merge. So your decision :-)

rgerhards:
The PR validation failed check, as it fails, would be acceptable for me and does not block merge.

tools/omusrmsg.c
century6: OK, I will make modifications according to the review comments.

rgerhards: many thanks for your quick turn-around :-)

century6: I have updated the PR. Please check it again when you have a chance :-)

rgerhards:
The CI fail was a flake, I have re-initiated. I'll check and I have - just to be sure - initiated a final AI review. Looks much like we get a same-day merge. Thx again @century6 for your great work. This not only fixes the original issue, but also helped improve the user experience by better error reporting. Great!

rgerhards:
@century6 Thx again for the PR, I just merged it. Great cooperation, many thanks!

Issue #5060, opened 2023-01-10
omelasticsearch cannot ship logs to elasticsearch 8.3. in bulkmode


celinaT:
### Actual behavior
Omelasticsearch does not ship logs to Elastic Search 8.3. in bulk mode, but it works when bulk mode is not enabled.
There is the same issue reported https://github.com/rsyslog/rsyslog/issues/4908 which should be fixed, but it does not seem that the fix works in  8.2212.0

### Environment
- rsyslog version: 8.2212.0
- platform: Ubuntu 18.04.3 LTS

Config:

ruleset(name="send") {
  action(
    type="omelasticsearch"
    server="servername"
    serverport="443"
    usehttps="on"
    tls.cacert="tls.cert"
    allowunsignedcerts="on"
    template="json"
    searchIndex="indexname"
    dynSearchIndex="on"
    searchType="_doc"
    bulkmode="on"
    queue.dequeuebatchsize="5000"
    queue.size="100000"
    queue.workerthreads="5"
    action.resumeretrycount="-1"
    errorfile="/var/log/omelasticsearch.log"
    uid="userid"
    pwd="pwd"
  )
}

Error:

contains an unknown parameter [_type]" } ], "type": "illegal_argument_exception", "reason": "Action\/metadata line [1] contains an unknown parameter [_type]" }, "status": 400 }

StoryDevOffical:
same here, from my research with elastic search >= v8, type have to be deleted.
I still don't know how to

EnZee1:
I see the same behavior on my system. 
With "bulkmode=on" I get the error message mentioned above, with "bulkmode=off" the messages can be sent to my ElasticSearch.

I have also tried setting the parameter "esVersion.major=8" as mentioned [here](https://github.com/rsyslog/rsyslog/issues/5059#issue-1522198095), however, it seems that rsyslog is not using it, I continue to receive the same error message.

- rsyslog version: 8.2212.0
- Ubuntu version: 22.10

julienkosinski:
Hello,

On my side, I have tested bulkmode with Elasticsearch 8.7 and 8.8 and it works well.
My Rsyslog is in version 8.2304.0.

Can you still reproduce your problem with the latest Rsyslog?

plantroon:
@EnZee1  I might be late to comment here but have you tried `esVersion.major="8"` (the quotes around 8) ? That's what worked for me.

sshmanko:
[...]

PR #6177, opened 2025-09-25
omelasticsearch: drop search_type and type segment by default

rgerhards:
Newer Elasticsearch rejects the search_type parameter. This change aligns defaults with current ES behavior and lays groundwork for more automatic backend/version handling. We also add OpenSearch specifics.

Before/After: previously URLs included /<index>/<type> and often sent search_type; now URLs use /<index>/_doc and search_type is omitted by default (and always suppressed for Elasticsearch).

Impact: user-visible; configs that set searchType now only work against OpenSearch and only for two supported values. Tests and docs updated.

Technical details:
- Stop appending the type path segment; always post to /<index>/_doc.
- Only add the search_type query parameter when explicitly configured and its expansion is one of: query_then_fetch, dfs_query_then_fetch.
- Detect backend (Elasticsearch vs OpenSearch) via an initial HTTP GET, cache result under a mutex, log detected type/version, and force- suppress search_type for Elasticsearch regardless of config.
- Validate static/dynamic searchType; on invalid values log a config error and ignore. Remove _type from bulk metadata lines.
- Adjust tests to no longer rely on searchType; refresh docs to reflect the new semantics and defaults.

Fixes: https://github.com/rsyslog/rsyslog/issues/5060


rgerhards:
Take note: this PR may contain too many not really related changes, I am also not clear if I am happy with the code quality. So I may or may not merge it. Currently working on a more focused and structured approach.

rgerhards:
as i already hinted, this is now actually being replaced by a set of less intrusive patches. Not happy with code quality over here.

PR #6185, opened 2025-09-26
docs: split properties into reference pages

rgerhards:
- extract message, system, and time-related property docs into individual reference pages with summaries, metadata, and usage anchors
- replace the overview lists with list-tables that include per-property summaries and add a hidden toctree for navigation
- preserve all explanatory notes, alias details, and case-sensitivity guidance while pointing See also links to the overview

With the help of AI-Agent: ChatGPT


rgerhards:
I screwed up working on this one. A new one will be upcoming, easier than to fix this one.

PR #6202, opened 2025-09-30
imkafka: try to reproduce potential mem leak

rgerhards:


rgerhards:
worked, made clear no fix is needed.

PR #6207, opened 2025-10-01
Fix: Subtree Templates Generate Empty Output

alorbach:
### Summary

This PR fixes a bug where `type="subtree"` templates produce empty output when used with modules like `omfwd` that rely on the `lenStr` field to determine message length.

### Problem Description

The `tplToString()` function in `template.c` correctly extracts and copies JSON subtree data into the output buffer, but fails to set `iparam->lenStr` to record the actual data length. This causes downstream modules to believe the buffer is empty (lenStr=0), resulting in empty frames being sent over the network or written to files.

### Impact

**Before this fix:**
- Configurations using `type="subtree"` templates with `omfwd` send **empty TCP/UDP frames**
- Data is correctly extracted but silently dropped due to lenStr=0
- Affects any module that checks lenStr before processing (omfwd, potentially others)
- No error messages are generated, making this a silent data loss bug

**After this fix:**
- Subtree templates work correctly with all output modules
- lenStr is properly set to match the actual JSON data length
- Network forwarding delivers complete JSON content as expected

### Example Scenario

**Configuration:**
```conf
template(name="eventSubtree" type="subtree" subtree="$!event")

if $msg contains "error" then {
    set $!event!level = "error";
    set $!event!code = 500;
    action(type="omfwd" template="eventSubtree" 
           target="192.168.1.100" port="10514" protocol="tcp")
}
```

**Before:** Receiver gets empty frame (lenStr=0)  
**After:** Receiver gets `{"level":"error","code":500}`

### Technical Details

**File:** `template.c`  
**Function:** `tplToString()`  
**Change:** Added `iparam->lenStr = iLenVal;` after the subtree memcpy() operation

The fix is a single line that ensures the length metadata matches the copied data. Other template code paths (regular templates, jsonftree, string generators) already set lenStr correctly; the subtree branch was simply overlooked.

### DIFF
--- a/template.c
+++ b/template.c
@@ -759,8 +759,8 @@ rsRetVal tplToString(struct template *__restrict__ const pTpl,
      */
     pTpe = pTpl->pEntryRoot;
     iBuf = 0;
-    const int extra_space = (pTpl->optFormatEscape == JSONF) ? 1 : 3;
-    if (pTpl->optFormatEscape == JSONF && !pTpl->bJsonTreeEnabled) {
+    const int isJsonFlat = (pTpl->optFormatEscape == JSONF && !pTpl->bJsonTreeEnabled);
+    if (isJsonFlat) {
         if (iparam->lenBuf < 2) /* we reserve one char for the final \0! */
             CHKiRet(ExtendBuf(iparam, 2));
         iBuf = 1;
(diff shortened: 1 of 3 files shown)


alorbach:
@codex review

template.c
```diff
@@ -793,14 +792,15 @@ rsRetVal tplToString(struct template *__restrict__ const pTpl,
...
-
             if (need_comma) {
+                if (iBuf + 2 >= iparam->lenBuf) CHKiRet(ExtendBuf(iparam, iBuf + 2 + 1));
                 memcpy(iparam->param + iBuf, ", ", 2);
                 iBuf += 2;
             }
+            /* first, make sure buffer fits */
+            if (iBuf + iLenVal >= iparam->lenBuf) /* we reserve one char for the final \0! */
+                CHKiRet(ExtendBuf(iparam, iBuf + iLenVal + 1));
+
```
rgerhards, line 810: I would also suggest to combine the computation and use extendBuf only once.

rgerhards (changes requested):
See requested change.

I also recommend to check if there is any interaction witht the patch in https://github.com/rsyslog/rsyslog/pull/5978

This seeems to have overlooked the `jsonf` formatting (my guess). If so, there may also be a similiar problem with other options like `json. I recommend to review these cases and also add testcases like done in #5978 so that we will detect regressions in the future.

template.c
alorbach: Changed according to your suggestion

PR #6209, opened 2025-10-01
core/ratelimit: add shared ratelimit() object and name binding

rgerhards:
Provide a central place to define rate-limiting, improving UX and keeping
configs consistent across modules. This also sets the stage for future
rate-limit enhancements and makes containerized setups easier to tune.

Impact: New ratelimit() object and ratelimit.name param; tcpsrv IF bump
to v29; mixing name with inline ratelimit.* is now rejected.

Before: Each module had ad-hoc ratelimit.interval/burst; no reusable spec.
After: A ratelimit() object can be declared once and referenced via
ratelimit.name; inline params still work and are promoted to shared cfg.

Technical details:
- Add ratelimit config store to rsconf (init/destruct). New immutable
  ratelimit_config_t and helpers: ratelimitConfig{Create*,Lookup}, and
  ratelimitResolveFromValues() to resolve name vs inline with guard rails.
- New ratelimitNewFromConfig() binds runtime limiters to shared configs;
  legacy ratelimitSetLinuxLike()/SetSeverity() remain for inline cases.
- RainerScript: add CNFOBJ_RATELIMIT; implement ratelimitProcessCnf();
  lexer/parser accepts ratelimit(...).
- Global: support internalmsg.ratelimit.name (falls back to inline if
  unset); rsyslogd wires shared config for internal messages.
- Modules: imtcp, imudp, imptcp, imhttp, imjournal, imklog, imuxsock,
  omhttp, omelasticsearch, omfwd accept ratelimit.name; resolve configs
  at load and construct ratelimiters from them. imuxsock maps severity
  from the object when used.
- tcpsrv: SetLinuxLikeRatelimiters() now takes optional cfg pointer;
  interface version bumped to 29.
- Tests: add imtcp-ratelimit-object/legacy/impstats plus diag helpers to
  wait for and assert burst behavior. Docs updated across module pages,
  reference params, and new rainerscript/ratelimit page.


rgerhards:
not fully happy with code - solution seems unnecessarily complex

PR #6211, opened 2025-10-02
Optimize testbench runtime by improving CI parallelization

alorbach:
# Optimize Testbench Runtime by Improving CI Parallelization

## Summary
This PR optimizes the CI testbench runtime by increasing parallel test execution across different environments, with expected 20-40% performance improvements.

### 1. **Increased Default Parallelization**
- Changed default `CI_MAKE_CHECK_OPT` from `-j4` to `-j8` in `run_checks.yml`

### 2. **Environment-Specific Optimizations**
- **ubuntu_20**: Added `-j8` parallelization (was using default `-j4`)
- **ubuntu_24**: Added `-j8` parallelization (was using default `-j4`)
- **ubuntu_22_distcheck**: Added `-j6` parallelization (new addition)
- **ubuntu_24_tsan**: Conservative increase from `-j2` to `-j3` (CPU-intensive environment)

### 3. **Fixed Configuration Issues**
- **Elasticsearch**: Removed conflicting `-j1` setting that was being overwritten by `-j8`
- **Codecov workflow**: Increased from `-j5` to `-j8` for faster test execution

### 4. **Code Quality**
- Fixed shellcheck SC2155 warnings by separating variable assignment from export statements

## Performance Impact

Based on current runtime data from CI screenshots:

| Environment | Current Runtime | Expected Improvement |
|-------------|----------------|---------------------|
| ubuntu_24 | 21m 50s | ~30-40% faster (13-15min) |
| ubuntu_20 | 21m 41s | ~30-40% faster (13-15min) |
| ubuntu_22_distcheck | 20m 42s | ~25% faster (15-16min) |
| ubuntu_24_tsan | 23m 1s | ~15% faster (19-20min) |
| codecov workflow | N/A | ~35% faster test execution |

**Overall CI pipeline**: 20-25% faster completion

### DIFF
--- a/.github/workflows/run_checks.yml
+++ b/.github/workflows/run_checks.yml
@@ -63,15 +63,17 @@ jobs:
       - name: run container CI pipeline
         run: |
           chmod -R go+rw .
-          export CODECOV_repo_slug="$(git config --get remote.origin.url | sed -E 's#.*[:/]([^/]+/[^/.]+)(\\.git)?$#\1#')"
-          export CODECOV_commit_sha="$(git rev-parse HEAD)"
+          CODECOV_repo_slug="$(git config --get remote.origin.url | sed -E 's#.*[:/]([^/]+/[^/.]+)(\\.git)?$#\1#')"
+          export CODECOV_repo_slug
+          CODECOV_commit_sha="$(git rev-parse HEAD)"
+          export CODECOV_commit_sha
           export RSYSLOG_CONTAINER_UID="" # use default
           export RSYSLOG_STATSURL='http://build.rsyslog.com/testbench-failedtest.php'
           export CFLAGS='-g'
           export CC='gcc'
           export USE_AUTO_DEBUG='off'
           export CI_MAKE_OPT='-j20'
-          export CI_MAKE_CHECK_OPT='-j4'
+          export CI_MAKE_CHECK_OPT='-j8'
           export CI_CHECK_CMD='check'
           export VERBOSE=1
           case "${{ matrix.config }}" in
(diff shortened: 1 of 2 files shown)


rgerhards:
I do not like this idea. We came from higher concurrency levels, but that causes even more resources stress on the CI builders and thus more timing-related flakes. I see two other CI PRs, and they seem to try to "fix" such timing issues. I do not think this is the right path.

The bare truth is that some tests are inherently racy, and so we need sufficient headroom with resources.

rgerhards:
I need to add: I am not totally opposed to that idea, but than we should first address all those flaky tests that we know about. Remember that for some we work for several years now to remove timing dependency, but it was so far not possible without going to unrealistic tests pathes. Again, if we can remove the flakiness, we can increase concurrency.

One idea may be to split test runs between known-flaky (w/ limited concurrency) and others. But that means we would need to carefully select and double most CI runs (with an overall saving in runtime but increased manual work, I guess).

rgerhards:
> Elasticsearch: Removed conflicting -j1 setting that was being overwritten by -j8

very bad idea

alorbach:
Lets close this for now, and look into more specific performance improvements once the particular rare failing tests are stabelized (if possible).

rgerhards:
> Lets close this for now, and look into more specific performance improvements once the particular rare failing tests are stabelized (if possible).

We could actually try to temporarily add more concurrency to one builder, just to see what fails very frequently. But that still means we need to monitor and improve the CI runs. It's something that would further stress the time budget (albeit it seems tempting).

PR #6214, opened 2025-10-03
mmjsontransform: add experimental JSON dotted-key (un)flatten

rgerhards:
Real-world need: unflatten dotted JSON into nested objects, and optionally flatten back for downstream tools. This introduces a general transformer with a dedicated output tree. Interface is intentionally unstable.

Impact: New module behind --enable-mmjsontransform; no default behavior changes. New tests and docs added. Parameters and behavior may change.

Add mmjsontransform, a message modification module that rewrites dotted JSON keys. By default it "unflattens" an input object to nested containers and stores the result in a configured output property. A mode parameter also supports "flatten" to collapse nested trees into dotted keys. The action refuses to overwrite an existing destination, validates that input is a JSON object, and reports conflicts with precise key paths. Per-action config is immutable; workers hold pointers only, so no extra locking. Docs (Sphinx + parameter refs) and doxygen coverage included, plus a regression test exercising nested arrays/objects. Build system and CI scripts gain --enable-mmjsontransform and a basic test hook. An experimental companion mmjsonrewrite module is wired similarly for dotted-key expansion.

Before/After: Previously no built-in JSON un/flatten; now an action can unflatten (default) or flatten JSON into a separate message property.

With the help of AI Agents: ChatGPT codex, gemini


rgerhards:
AI comments noted; they are valid, but no concerns for current experimental module. I will most probably be refactored and functionality be extended. so good enough to merge.

PR #6218, opened 2025-10-04
CI: enhance doc workflow to check tarball completeness

rgerhards:
Real-world tarballs must be self-contained for doc builds. This change verifies release completeness by building the docs from the tarball, not the checkout.
    
Impact: Release tarballs now include additional doc assets; CI fails early if a doc-required file is missing.
    
Before: CI built Sphinx docs directly from the repo tree. Missing files could be masked by in-tree paths. After: CI creates a 'make dist' tarball, unpacks it, and builds docs from that tree to catch omissions.
    
Technically, the doc workflow installs build deps, runs autoreconf and a minimal './configure', issues 'make dist', unpacks into
'doc-builder/', and runs 'make html' under 'doc-builder/doc'. The artifact path is updated accordingly.
    
For distribution, 'configure.ac' now composes DOC_FILES from 'doc/source' (*.rst, *.conf, *.jpg, *.png) and explicitly adds 'tutorials/cert-script.tar.gz', 'doc/Makefile', and 'doc/ai/module_map.yaml'. The duplicate listing of  'doc/ai/module_map.yaml' is removed from 'Makefile.am' EXTRA_DIST to avoid drift; DOC_FILES remains the single source of truth for doc payloads.

### DIFF
--- a/.github/workflows/doc_build.yml
+++ b/.github/workflows/doc_build.yml
@@ -40,6 +40,60 @@ jobs:
           files: |
             doc/**/*.rst
 
+      # we build from tarball to ensure no missing files!
+      - name: setup make dist build env
+        if: ${{ (github.event_name != 'pull_request' && github.event_name != 'pull_request_target') || steps.doc_changes.outputs.any_changed == 'true' }}
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y \
+            libestr-dev \
+            libglib2.0-dev \
+            libgnutls28-dev \
+            liblognorm-dev \
+            liblz4-dev \
+            libnet1-dev \
+            librelp-dev \
+            libssl-dev \
+            libsystemd-dev \
+            libtool \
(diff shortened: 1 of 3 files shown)


rgerhards:
@mbiebl FYI, I have enhanced CI to build doc from a dist tarball, this hopefully ensures that we detect missing files and other things where the tarball is unsuitable for doc building.

I Initially thought to integrate this in the overrall make/make dist workflow, but TBH this looks like it creates a big mess (or I did it wrong), so I resorted to the place which keeps everything else working that was setup in the past ~2 month.

If you have further suggestions, they are welcome.  If not, I would send you a pre-8.2510 tarball to verification. If your workflow differs, pls let me know so that I can apply further changes to CI.

mbiebl:
It did test-build a `make dist` generated tarball from current git and the results look good so far. The only issue I ran into was #6224 which has already been fixed (thanks!)

PR #6220, opened 2025-10-05
tests: retry omfwd load-balancer flake before failing

rgerhards:
- wrap omfwd-lb-1target-retry-full_buf.sh around the shared skeleton so the scenario is retried once before reporting failure, easing TCP buffer timing races
- document follow-on stabilisation ideas for minitcpsrvr coordination and omfwd diagnostics

AI-Agent: Codex


rgerhards:
@alorbach I think this is probably a solution for inherently flaky tests - i f we have real issues, it will fail reliably.

I just do it for this one, let's see how it evovles.

PR #6236, opened 2025-10-07
mysql: stabilize async ommysql workers

alorbach:
mysql: stabilize async ommysql workers; skip TSAN for mysql-asyn

ThreadSanitizer flagged a race inside libmysql (memcmp in parser paths)
when multiple workers issue queries concurrently. The module now keeps
per-worker transaction serialization and per-worker SQL buffers, while
we avoid penalizing production by skipping the multi-worker test only
under TSAN.

What changed
- ommysql:
  - Serialize each transaction per worker via txMutex (START/INSERTs/COMMIT)
  - Use a per-worker reusable SQL buffer; overflow-safe growth
  - Log pthread_* return codes (not errno)
  - Removed prior global libmysql locks/workarounds
- tests/mysql-asyn.sh:
  - Skip under TSAN (exit 77) with a documented rationale, preserving CI stability
    without adding global serialization to production builds

Rationale
- The TSAN report originates in libmysql internals; with one MYSQL handle per
  worker and proper per-worker locking, production is safe. Skipping the test
  under TSAN avoids CI noise while keeping runtime performance unchanged.
  
  closes https://github.com/rsyslog/rsyslog/issues/5884

### DIFF
--- a/plugins/ommysql/ommysql.c
+++ b/plugins/ommysql/ommysql.c
@@ -26,6 +26,7 @@
 #include "config.h"
 #include "rsyslog.h"
 #include <stdio.h>
+#include <stdbool.h>
 #include <stdarg.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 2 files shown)


alorbach:
@codex review

PR #6279, opened 2025-10-31
PoC: rsyslog ETL pipeline container for sending to vespa.ai

rgerhards:
Interfacing to [vespa.ai](https://vespa.ai) and the vespa engine.

This is a first and definitely incomplete PoC, it just lays out the core ideas and plumbing.

I think it should ideally be ready-to-roll for the public vespa cloud interface.

### DIFF
--- a/packaging/docker/rsyslog/Makefile
+++ b/packaging/docker/rsyslog/Makefile
@@ -4,7 +4,7 @@
 # New Naming Scheme:
 # Standard Image: rsyslog/rsyslog:<VERSION>
 # Functional Variants: rsyslog/rsyslog-<function>:<VERSION>
-# Functions: minimal, standard, collector, dockerlogs (extensible)
+# Functions: minimal, standard, collector, dockerlogs, etl (extensible)
 # Version: rsyslog version (e.g., 2025-04, matches rsyslog release)
 # ==============================================================================
 
@@ -40,23 +40,25 @@ RSYSLOG_FUNCTION_PREFIX = $(strip $(ORG_NAME)/rsyslog-)
 MINIMAL_IMAGE_NAME = $(strip $(RSYSLOG_FUNCTION_PREFIX)minimal)
 COLLECTOR_IMAGE_NAME = $(strip $(RSYSLOG_FUNCTION_PREFIX)collector)
 DOCKERLOGS_IMAGE_NAME = $(strip $(RSYSLOG_FUNCTION_PREFIX)dockerlogs)
+ETL_IMAGE_NAME = $(strip $(RSYSLOG_FUNCTION_PREFIX)etl)
 
 MINIMAL_IMAGE_TAG = $(strip $(MINIMAL_IMAGE_NAME):$(VERSION))
 COLLECTOR_IMAGE_TAG = $(strip $(COLLECTOR_IMAGE_NAME):$(VERSION))
 DOCKERLOGS_IMAGE_TAG = $(strip $(DOCKERLOGS_IMAGE_NAME):$(VERSION))
+ETL_IMAGE_TAG = $(strip $(ETL_IMAGE_NAME):$(VERSION))
 
 
 # --- Phony Targets (commands, not actual files) ---
 .PHONY: all build clean push all_push tag_latest push_latest help \
-        minimal standard collector dockerlogs \
-        build_minimal_image build_standard_image build_collector_image build_dockerlogs_image \
-        push_minimal push_standard push_collector push_dockerlogs \
+        minimal standard collector dockerlogs etl \
+        build_minimal_image build_standard_image build_collector_image build_dockerlogs_image build_etl_image \
+        push_minimal push_standard push_collector push_dockerlogs push_etl \
         rebuild_all
 
 # Default target: Builds all functional images.
-# Assumed layering: minimal -> standard -> (collector, dockerlogs)
-all: build_collector_image build_dockerlogs_image
-# build_standard_image is a dependency for collector and dockerlogs
+# Assumed layering: minimal -> standard -> (collector, dockerlogs, etl)
+all: build_collector_image build_dockerlogs_image build_etl_image
+# build_standard_image is a dependency for collector, dockerlogs, and etl
 # build_minimal_image is a dependency for standard
 
 # --- User-Friendly Build Targets ---
@@ -71,6 +73,9 @@ collector: build_collector_image
 
 dockerlogs: build_dockerlogs_image
 	@echo "Convenience target: Dockerlogs image build triggered."
+
+etl: build_etl_image
+	@echo "Convenience target: ETL image build triggered."
 
 
 # --- Core Build Logic ---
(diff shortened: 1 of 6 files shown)


rgerhards:
@radu-gheorghe I am not sure it this is too rough... Anyhow, if you have a moment, can you please have a look and maybe even try to build and run? Some info on how to make it talk to vespa official would also be great. I know, I am lazy and should look more into doc and setup, but I want something really going as a base. Best yesterday ;-)

radu-gheorghe:
It looks good to me, thanks for putting it together. I don't know how to run it, I assume `make etl`? Though I don't think it would work for me, as I have Podman instead of Docker.

But let me share my own PoC, which should be similar. This is the directory structure I have:
Screenshot 2025-11-02 at 16 09 04

`rsyslog.conf` is very similar to yours. I use it by starting rsyslog manually and running simple `logger` commands interactively from the container:
```
# reads data
module(load="imuxsock")

# write to a file using the template
action(type="omfile"
       file="/tmp/messages")


module(load="omhttp")

# template that outputs the message field in a JSON document
# go wild and make your own documents
template(name="json-message" type="list" option.json="on") {
    constant(value="{ \"fields\": { \"message\": \"")
    property(name="msg")
    constant(value="\" }}")
}

# template that outputs Vespa path + random doc ID
template(name="vespaDocUrlUuidHex" type="string"
  string="/document/v1/mynamespace/message/docid/%uuid%"
)

action(type="omhttp"
        # host and port of Vespa
       server="vespa"
       serverport="8080"
       # for secure endpoints (e.g., Vespa Cloud), you can provide certs and all
       useHttps="off"
       # no batching, but maybe HTTP/2 in the future
       batch="off"
       # point to the template that outputs random doc ID
       dynrestpath="on"
       restpath="vespaDocUrlUuidHex"
       # template making JSON documents to send to Vespa
       template="json-message"
)
```

Dockerfile is simply:
```dockerfile
# Use Ubuntu 22.04 LTS as base image
FROM ubuntu:22.04

# Update package lists and install prerequisites
RUN apt-get update && \
    apt-get install -y \
    software-properties-common

# Add Adiscon (official) PPA repository
RUN add-apt-repository ppa:adiscon/v8-stable

# Update package lists with new repository
RUN apt-get update

# Install rsyslog from Official PPA
RUN apt-get install -y rsyslog rsyslog-omhttp
```

Then, the docker-compose makes sure Vespa is ready and has the application package deployed:
```yaml
version: '3'
services:
  # Vespa container
  vespa:
    image: vespaengine/vespa
    container_name: rsyslog-demo-vespa
    hostname: rsyslog-demo-vespa
    ports:
      - "8080:8080"
      - "19071:19071"

  # rsyslog container
  rsyslog:
    build: .
    depends_on:
      vespa
    container_name: rsyslog-demo
    hostname: rsyslog-demo
    volumes:
      # rsyslog config file
      - ./rsyslog.conf:/etc/rsyslog.conf
      # application package. Not needed if you deploy from the host machine with `vespa deploy` (see below)
      - ./test_app:/tmp/test_app
    command: |
      bash -c "
        ###### deploy application package ######
        ### only needed once
        ### can also be done from the host machine with \`vespa deploy\`
        ### see https://docs.vespa.ai/en/vespa-cli.html
        ###########################################
        # install curl and zip for deployment
        apt-get update && apt-get install -y curl zip
        #
        while ! curl -f http://vespa:19071/ ; do echo 'Waiting for config server to be healthy...'; sleep 5; done
        #
        # deploy the application package
        (cd /tmp/test_app && zip -r - .) | \
          curl --header Content-Type:application/zip --data-binary @- \
            vespa:19071/application/v2/tenant/default/prepareandactivate
        #
        while ! curl -f http://vespa:8080/ ; do echo 'Waiting for Vespa endpoint after package deploy...'; sleep 5; done
        #
        ###########################################
        # now wait for interactive use
        sleep infinity
      "

```

Finally, the application package has a schema that defines that `message` field (the only field we send):
```
schema message {
    document message {
        # TODO more real data, add some parsing...
        field message type string {
            indexing: summary | index
        }
    }
}
```

And a services.xml that defines one container and one content node (all running in the same Docker/Podman container) and pointing to that `message` schema. I'll leave the comments, in case they are helpful:
```xml
<?xml version="1.0" encoding="utf-8" ?>
<!-- Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root. -->
<services version="1.0" xmlns:deploy="vespa" xmlns:preprocess="properties">

    <!--
        A container cluster handles incoming requests to the application and processes those requests,
        and their results. The processing to do and the API's to expose can be provides by Vespa
        or by the application through Java components supplied as part of the application.

        See:
          - Reference: https://docs.vespa.ai/en/reference/services-container.html
    -->
    <container id="default" version="1.0">

          <!-- <clients>
            <!- - 
              Required for mTLS to still work, the below
              configuration is equivalent to using a single
              security/clients.pem
            - ->
            <client id="mtls" permissions="read,write">
              <certificate file="security/clients.pem"/>
            </client>
        
            <!- -
              The below sets up a client with read permissions.
              Note that the "myToken" token must be created in
              the console before deploying.
            - ->
            <client id="tokenClient" permissions="read,write">
              <token id="test-token"/>
            </client>
          </clients> -->
        <!--
            <document-api> tells the container that it should accept documents for indexing. Through the
            Document REST API you can PUT new documents, UPDATE existing documents, and DELETE documents
            already in the cluster.

            Documents sent to the Document REST API will be passed through document processors on the way
            to the content cluster.

            See:
             - Reference: https://docs.vespa.ai/en/reference/services-container.html#document-api
             - Operations: https://docs.vespa.ai/en/document-v1-api-guide.html
        -->
        <document-api/>

        <!--
            <search> tells the container to answers queries and serve results for those queries.
            Inside the <search /> cluster you can configure chains of "searchers" -
            Java components processing the query and/or result.

            See:
             - Reference: https://docs.vespa.ai/en/query-api.html
             - Searchers: https://docs.vespa.ai/en/searcher-development.html
        -->
        <search/>

        <!--
            <nodes> specifies the nodes that should run this cluster.
        -->
        <nodes>
            <node hostalias="node1" />
        </nodes>
    </container>

    <!--
        <content/> content clusters store application data, maintain indexes and executes the
        distributed parts of a query.

        See:
          - Reference: https://docs.vespa.ai/en/reference/services-content.html
    -->
    <content id="blog" version="1.0">
        <min-redundancy>2</min-redundancy>
        <documents>
            <document type="message" mode="index" />
        </documents>
        <nodes>
            <node hostalias="node1" distribution-key="0" />
        </nodes>
    </content>

</services>
```

When experimenting with the application package, I would use Vespa CLI (https://docs.vespa.ai/en/vespa-cli.html) and do `vespa deploy` from the application package directory, instead of the curl gymnastics from docker-compose :)

Let me know if I can help with anything else. I'd gladly help push this forward. We could also use Vespa Cloud with the application package already there, which would get rid of a lot of Vespa-related stuff above :)

rgerhards:
Radu, thx a lot!

I got side-stepped again, hey, its quite a bit going on :-) [and, actually I began this note 3 days ago ;-)]

Also a good reminder in podman, we should definitely add a dedicated section for it in rsyslog repo.

I think we will follow your approach. My goal is to create something for rsyslog users that is really helpful for them to get started with vespa. And it should be very easy to do the first step (we are learning from past comlexities ;-)).

radu-gheorghe:
You're welcome, Rainer. Sounds good, especially the learning from the past part :)

Oh, and thanks again for your help in getting people started with Vespa!

PR #6312, opened 2025-11-21
mmsnareparse: add ignoreTrailingPattern parameter

alorbach:
Add `ignoreTrailingPattern` parameter to mmsnareparse for removing trailing extra-data sections

**Description:**

Adds a configurable parameter to the `mmsnareparse` plugin to detect and remove trailing extra-data sections from messages before parsing. This addresses cases where third-party enrichers append non-standard data (e.g., `enrichment_section: fromhost-ip=...`) that can interfere with parsing.

**Changes:**
- Added `ignoreTrailingPattern` parameter at both module and action levels (action-level overrides module default)
- Implemented trailing pattern detection that only matches patterns appearing after the last tab-separated token
- When a pattern is found, the message is truncated at the start of the last token (after the last tab), removing the entire trailing section including any preceding content in that token
- The truncated extra-data section is optionally exposed as a `!extradata_section` message property for downstream use
- Added test case `mmsnareparse-trailing-extradata.sh` validating the feature with anonymized sample data
- Updated documentation in `doc/source/configuration/modules/mmsnareparse.rst` with parameter description

**Technical Details:**
- Pattern matching is literal string-based (not regex)
- Truncation only occurs when pattern appears in trailing positions (after last tab or in trailing portion for non-tab messages)
- Existing parsing behavior is unchanged when parameter is not set
- Memory management: proper allocation and cleanup of pattern strings and extra-data sections

**Testing:**
- New test case validates correct parsing of Event ID 13 (Registry value set) with trailing enrichment section
- Verifies enrichment section is removed from parsed fields but stored in `!extradata_section` property
- Test passes successfully

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -274,14 +274,84 @@ Minimum setup requires:
   - Autotools toolchain: `autoconf`, `automake`, `libtool`, `make`, `gcc`
   - Side libraries: `libestr`, `librelp`, `libfastjson`, `liblognorm` (must be installed or built manually)
 
-Example commands (swap the final step for the most relevant smoke test):
+### Complete Dependency Installation (Ubuntu/Debian WSL)
+
+For a full development environment with all common dependencies:
 
 ```bash
-./autogen.sh
-./configure --enable-debug --enable-testbench
-make -j$(nproc)
+sudo apt-get update
+sudo apt-get install -y \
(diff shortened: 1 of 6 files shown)


rgerhards:
OK to merge as is, as it improves the code. Not perfect, and some follow-up is expected next week.

PR #6313, opened 2025-11-21
build: Add code coverage targets

frikilax:

- Add new configure flag to enable code coverage using gcc and lcov/genhtml
- Add Makefile targets to run coverage after or during checks, as well as cleaning coverage artefacts

### References

None

### Notes (optional)

---

#### Quick check (optional)

---

  ```
  git commit --amend
  ```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,3 +100,7 @@ tools/logctl
 tools/rscryutil
 tools/rscryutil.1
 /bin/
+# Coverage artifacts
+*.gcda
+*.gcno
+coverage*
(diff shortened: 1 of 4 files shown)


frikilax:
example of generated HTML (after runing all "rscript_*" tests)
image

rgerhards:
Thx for this PR. I think it is useful for local testing (and as such for digital sovereignty) but keep on your mind that to get a full coverage report, one needs to run all tests and on all platforms (like Solaris). As such, the local view will only be partial. This is the reason we use CodeCov here on github.

rgerhards:
@frikilax thx again for the PR!

PR #6340, opened 2025-12-04
mmanon: better document externalized function part

rgerhards:
was externalized because of cognitive load, but calling contract and specialization was not specified, made AI mistake helper semantics.


rgerhards:
@jgerhards FYI

Issue #6201, opened 2025-09-29
[Feature]: Per-source rate limiting with template sender key (YAML, HUP)


rgerhards:
# History & Issue content

**NOTE: while we keep the issue, the implementation has notably changed - we now use the shared ratelimiter system*

The shared ratelimiter is used as a global object that implements the here-mentioned feature set.

#  Details
**Summary**
Introduce a per-source rate limiter governed by an external YAML policy (reload on HUP; optional file-watch with debounce). Sender identification is **not hardcoded**: it is computed via a RainerScript **template** (similar to `omsendertrack`). Provide a global default and exact overrides (initially by hostname values). Add an optional adaptive baseline (learning approach to be decided **after** a simple static implementation). Aim for a **module-independent core** implementation that `imtcp` uses first; only scope it to `imtcp` if a general solution adds prohibitive complexity.

**Motivation / Problem**

* Global input limits let a noisy sender starve others.
* Operators need a global default plus exceptions.
* Policy should live outside main config and be reloadable without restart.
* Avoid duplicating sender-tracking logic; evaluate sharing with/replacing `omsendertrack`.

**Scope (initial)**

* First consumer: `imtcp` (TCP inputs).
* Design goal: reusable core helper for other inputs (e.g., `imudp`, `imptcp`, `imrelp`). If generalization adds considerable complexity, start with `imtcp` only.
* Overrides: exact match on the **evaluated sender key** (default template mirrors hostname).
* Policy + optional learner snapshot: YAML.

---

### Proposed behavior

1. **Template-keyed sender identification**

   * New `perSourceKeyTpl` option on the input selects a template that produces the sender key (default equivalent to `$hostname`).
   * Performance impact of templating in hot path must be measured; provide a fast path for common templates.

2. **Global default + per-sender overrides**

   * YAML defines a `default` `{max, window}` and a list of `overrides` keyed by exact value of the evaluated sender key (commonly a hostname).

3. **Limiter action**

   * On exceed: drop (counted). Optionally a delay mode if input queues have headroom (Phase 2).

4. **Adaptive baseline (optional)**

   * When enabled, adjust effective caps based on observed behavior. **Algorithm intentionally unspecified** until after the static MVP lands.

5. **Policy reload & snapshot**

   * Reload YAML policy on HUP; optional file-watch with debounce.
   * If adaptive mode is on, optionally persist learner state to a YAML snapshot for warm restarts (feasibility to be validated).

6. **Observability**

[...]

PR #6399, opened 2025-12-29
ratelimit: add named rate limit support for input modules

rgerhards:
Introduce named rate limit objects to separate policy definition from input configuration. This allows administrators to define reusable traffic shaping rules (interval, burst) and apply them consistently across multiple 'imtcp' and 'imptcp' listeners, simplifying management of high- volume log streams.

Impact: New 'ratelimit()' configuration object available.

This change extends the Rainerscript grammar with a top-level 'ratelimit' object, stored in a central registry within 'runtime/ratelimit.c'. The 'imtcp' and 'imptcp' modules are updated to accept a 'ratelimit.name' parameter. When set, the listener instantiation logic in 'tcpsrv' retrieves the configured policy from the registry instead of using legacy ad-hoc values. 'libyaml' support is added to the build system to enable future loading of external policy files.

Refs: https://github.com/rsyslog/rsyslog/issues/6201
With the help of AI-Agents: Antigravity


rgerhards:
@cubic-dev-ai review

Issue #4016, opened 2019-12-05
Possible NULL dereference in omelasticsearch


jvymazal:
https://github.com/rsyslog/rsyslog/blob/92d01f4446bd388984b60f9c9e7d7768c660f382/plugins/omelasticsearch/omelasticsearch.c#L280

I am not exactly sure how the exact logic of freeing instance is intended to work, however if `root == pData`, it will try to access member of NULL, thus crashing.

https://github.com/rsyslog/rsyslog/blob/92d01f4446bd388984b60f9c9e7d7768c660f382/plugins/omelasticsearch/omelasticsearch.c#L286

PR #6400, opened 2025-12-30
omelasticsearch fix: potential NULL dereference

rgerhards:
Under some exotic cases, omelasticsearch could dereference a NULL pointer during shutdown cleanup. This would lead to undefined behaviour and most likely an abort. This is fixed in this patch.

closes https://github.com/rsyslog/rsyslog/issues/4016

With the Help of AI Agent: ChatGPT Codex


rgerhards:
CI failures are flakes

PR #6408, opened 2026-01-01
core, runtime, omfwd: plumb action name into modules and logs

rgerhards:
Non-technical: actions need to obtain their name for better reporting. This wires a safe way for OMs to learn their owning action and use its human-readable name in diagnostics.

Impact: omfwd log messages now include the action name in some errors. Also, a generic internal API has been added to enable actions to obtain their assigned name. The omfwd change is primarily a template of how this can be done (but also useful in itself).

Before: output modules had no supported way to access the action name. omfwd rate-limit errors did not identify the action instance. After: core calls an optional OM callback with the action pointer and omfwd includes the action name in its error logs.

Technical details:
- Introduce optional OM entry point `setActionInfo(void *pData, action_t*)` and query it during module init; absent modules get NULL (no change).
- Call `setActionInfo()` from `addAction()` after finalize, passing the module's instance data and the new `action_t`.
- Provide `actionGetName(const action_t*)` that never returns NULL and falls back to "[action]".
- Extend `module-template.h` with BEGIN/CODESTART/END macros for the new callback and a query hook macro.
- omfwd implements `setActionInfo` to cache the action pointer and uses `actionGetName()` inside commitTransaction() logging.
- No OMODTX, batching, or HUP behavior changes; pointer lifetime is tied to the action instance managed by config.

With the help of AI Agents: ChatGPT Codex

### DIFF
--- a/action.c
+++ b/action.c
@@ -18,7 +18,7 @@
  *
  * File begun on 2007-08-06 by Rainer Gerhards (extracted from syslogd.c).
  *
- * Copyright 2007-2022 Rainer Gerhards and Adiscon GmbH.
+ * Copyright 2007-2026 Rainer Gerhards and Adiscon GmbH.
  *
  * This file is part of rsyslog.
  *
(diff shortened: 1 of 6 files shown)


frikilax:
@rgerhards what about input modules? Can this be used as well for them?
I feel input modules' stats and logs could also benefit from this feature

PR #6419, opened 2026-01-05
imhttp: add address input parameter for explicit IPv4/IPv6 binding

ankitkumardev:
Improve usability and container readiness by allowing imhttp to bind to a specific IP address instead of always listening on 0.0.0.0.

Impact: Users can now explicitly bind imhttp to IPv4 or IPv6 addresses, including IPv6-only setups commonly used in Kubernetes.

Before/After:
- Before: imhttp always listened on 0.0.0.0, ignoring address intent
- After: imhttp respects the address parameter and binds accordingly

Technical details:
- Register and parse a new `address` input parameter in imhttp.
- Combine `address` and `ports` into the CivetWeb `listening_ports` configuration during activation.
- Automatically format IPv6 addresses using bracket notation as required by CivetWeb.
- Eliminates the need to use raw `liboptions` for dual-stack or IPv6-only configurations.

### DIFF
--- a/contrib/imhttp/imhttp.c
+++ b/contrib/imhttp/imhttp.c
@@ -95,6 +95,7 @@ struct modConfData_s {
     instanceConf_t *root, *tail;
     struct option ports;
     struct option docroot;
+    struct option address;
     struct option *options;
     int nOptions;
     char *pszHealthCheckPath;


rgerhards:
@ankitkumardev thx for the PR, much appreciated and the description sounds very useful.

However, it seems to contain just a single line change. Is this a draft where you tried the submission?

rgerhards:
@ankitkumardev any interest in replying or actually implementing this PR?

rgerhards:
@ankitkumardev ok, i will treat this now as a feature request

rgerhards:
@ankitkumardev this functionality is already present via the `port` parameter. I will update the documentation accordingly.

rgerhards:
@ankitkumardev replaced by #6447

ankitkumardev:
> @ankitkumardev this functionality is already present via the `port` parameter. I will update the documentation accordingly.
@rgerhards 
Sorry for the delayed response to the comment due to other commitments.
As per my understanding, the port only specifies which TCP port number to listen on; it does not control which IP address or network interface the server binds to.

ankitkumardev:
> @ankitkumardev replaced by #6447

@rgerhards Oh, I see. Thanks. We already have an existing feature that binds to an IP address and port.

PR #6420, opened 2026-01-05
Refactor duplicate string copy logic in msg.c

rgerhards:
Refactored the duplicate code for generating string properties in `MsgGetProp` into a new static helper function `msgPropStrGen`. This reduces code duplication in both the substring extraction and regex matching sections.

The new helper function handles memory allocation (using `CHKmalloc`), copying, null-termination, and proper management of the `pbMustBeFreed` flag.

With the help of AI-Agents: Jules


rgerhards:
CI fail is a know current flake

Issue #4906, opened 2022-06-22
Documentation for sparseArray lookup type is misleading


JPvRiel:
### Expected behavior

<https://www.rsyslog.com/doc/master/configuration/lookup_tables.html#sparsearray> states:

> A typical use case would be the matching of IPv4 address information.

An example should follow.

### Actual behavior

No matching examples are given. In fact, a "string" based lookup example is given for IP lookups. So the examples contridict the suggestion of the `sparseArray` use-case for IPv4 address information.

### Steps to reproduce the behavior

Create a lookup table:

```
{ "version" : 1,
  "nomatch" : "unk",
  "type" : "sparseArray",
  "table" : [
    {"index" : "127.0.1.1", "value" : "A" },
    {"index" : "127.0.2.1", "value" : "B" },
    {"index" : "127.0.3.1", "value" : "C" }]}
```

Without converting to integers, using the sparseArray for IPv4 lookups is awkward:

- Matching ""127.0.1.1" will return "unk".
- Matching "127.0.4.1" will return "C", not "unk" which may not be intuitive.

Most people would rather understand matching IPs via sub-netting.

Related issues:

- https://github.com/rsyslog/rsyslog/issues/1391
- https://github.com/rsyslog/rsyslog/issues/4734
- https://github.com/rsyslog/rsyslog/issues/1279
- https://github.com/rsyslog/rsyslog/issues/3394

Using `ip4tonum()` might be a way to coherce IPv4 strings into integers for use with sparseArrays, but even then, the greater than matching the last item in the array would fail to be intuitive.

The lookup table documentation needs a more robust example showcasing how to match IPv4 subnets, because, as per the related issues, it's a common question/feature request for rsyslog.

### Environment
- rsyslog version: 8.2102.0-7.el8
- platform: Alma Linux 8.6 (RHEL-like clone)

rgerhards:
While this was overlooked, it still is useful. Have finally picked it up (sry for the "delay")...

doc on how this can be acomplished is in #6421 - I am also working on a function.

PR #6421, opened 2026-01-05
doc: add sparseArray lookup IPv4 subnet matching example

rgerhards:
Documenting how to use sparseArray with ipv42num() for efficient IPv4 subnet matching.

Added a regression test to verify this functionality.

see also: https://github.com/rsyslog/rsyslog/issues/4906


rgerhards:
CI fail are known flakes

PR #6424, opened 2026-01-05
tcpsrv: fix race condition in session table slot allocation

alorbach:
Fix race condition in TCPSessTblFindFreeSpot() where multiple worker threads could select the same free slot in the TCP session table, causing sessions to be lost or overwritten.

The race occurred because:
- TCPSessTblFindFreeSpot() searched for free slots without synchronization
- Multiple threads could find the same slot between the search and assignment operations
- The gap between finding a slot (line 674) and assigning it (line 765) allowed concurrent threads to select the same index

Solution:
- Add pthread_mutex_t mutSessions to tcpsrv structure to protect session table access
- Initialize mutex in TCPSessTblInit() and destroy in deinit_tcp_listener()
- Protect all session table operations:
  * Finding and reserving a slot (with temporary marker)
  * Assigning the actual session pointer
  * Removing sessions on close
  * Error path cleanup

The fix uses a temporary marker value to reserve slots during construction, ensuring no two threads can select the same slot even if construction takes time. Mutex is held only during table modifications, not during expensive operations.

Only active in non-EPOLL builds (EPOLL mode doesn't use the session table). Fixed compile errors for EPOLL builds by conditionally compiling unused code.

Resolves TSAN race condition warnings like detected on macOS ARM64 CI: https://github.com/rsyslog/rsyslog/actions/runs/20708776690/job/59444439825?pr=6415

Addresses race condition listed in tests/tsan.supp line 15.

### References

### Notes (optional)

---

#### Quick check (optional)

---

  ```
  git commit --amend
  ```

### DIFF
--- a/runtime/tcpsrv.c
+++ b/runtime/tcpsrv.c
@@ -84,6 +84,11 @@ PRAGMA_IGNORE_Wswitch_enum MODULE_TYPE_LIB MODULE_TYPE_NOKEEP;
 /* defines */
 #define TCPSESS_MAX_DEFAULT 200 /* default for nbr of tcp sessions if no number is given */
 #define TCPLSTN_MAX_DEFAULT 20 /* default for nbr of listeners */
+/* Temporary marker value used to reserve a session slot during allocation.
+ * This is set before the actual session object is created to prevent
+ * other threads from selecting the same slot. Must be a non-NULL pointer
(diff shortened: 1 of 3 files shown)


rgerhards:
closed via #4626, which has a better approach via atomic operations

PR #6429, opened 2026-01-06
lmsig_ksi: remove unused REPORT_PARAM_MISSING macro

rgerhards:
The macro REPORT_PARAM_MISSING was defined but never used in the codebase. It also contained a TODO noting that the ABORT_FINALIZE within it was useless because the return value was unchecked.

This commit removes the unused macro definition.


rgerhards:
known flakes, good to merge

PR #6430, opened 2026-01-06
ommysql: very experimental locking in case of non-threadsafe lib

rgerhards:
If libmysql is not thread-safe, we need to avoid doing any parallel calls. This is quite complicated and should not be necessary any longer. Actually, it looks like it never was necessary.

So just giving this a try, but do not merge or consider unless we know we need to look into it.


rgerhards:
root cause was libmysql issues, so closing this

PR #6432, opened 2026-01-07
runtime: fix inconsistent synchronization of iCurNumWrkThrd

rgerhards:
Fixed inconsistent access to iCurNumWrkThrd where the shutdown loop used a plain read while all other accesses used atomic operations. This could result in the loop reading a stale value and not properly detecting when all workers have terminated.

The variable iCurNumWrkThrd is updated atomically by worker threads via ATOMIC_DEC when they terminate. The shutdown code must use atomic reads to ensure it sees these updates immediately, even though it holds mutWtp.

Before fix (inconsistent):
  Line 278: while (pThis->iCurNumWrkThrd > 0) // Plain read
  Line 338: ATOMIC_DEC(&pThis->iCurNumWrkThrd, ...) // Atomic update

After fix (consistent):
  Line 282: while (ATOMIC_FETCH_32BIT(&pThis->iCurNumWrkThrd, ...) > 0)
  Line 342: ATOMIC_DEC(&pThis->iCurNumWrkThrd, ...)

Impact:
- Ensures shutdown loop sees worker terminations immediately
- Prevents potential shutdown hangs due to stale counter value
- Maintains consistent synchronization pattern throughout codebase
- Properly pairs atomic writes with atomic reads

Testing:
- Builds successfully
- Passes smoke test (imtcp-basic.sh)
- All accesses to iCurNumWrkThrd now use atomic operations

With the help of AI Agents: ChatGPT Copilot

### DIFF
--- a/runtime/wtp.c
+++ b/runtime/wtp.c
@@ -275,7 +275,11 @@ PRAGMA_IGNORE_Wempty_body
     d_pthread_mutex_lock(&pThis->mutWtp);
     pthread_cleanup_push(mutexCancelCleanup, &pThis->mutWtp);
     bTimedOut = 0;
-    while (pThis->iCurNumWrkThrd > 0 && !bTimedOut) {
+    /* Use atomic read for consistency - iCurNumWrkThrd is updated atomically by
+     * worker threads. Even though we hold mutWtp, workers update this counter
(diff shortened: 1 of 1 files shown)


rgerhards:
CI is known flake.

PR #6433, opened 2026-01-07
runtime: fix race condition in worker thread cancellation

rgerhards:
Fixed a race condition in wtiCancelThrd() where thrdID could be accessed after the thread had already terminated and been joined by another caller, potentially resulting in pthread_cancel/pthread_kill operating on an invalid or reused thread ID.

The race occurred when:
1. wtiCancelThrd() checks thread state (sees RUNNING)
2. Thread terminates and transitions to WAIT_JOIN
3. Another thread joins the worker (thrdID becomes invalid)
4. wtiCancelThrd() accesses thrdID (undefined behavior)

Additionally, the function had a duplicate pthread_cancel() call that has been removed.

The fix:
- Cache thread state in local variable to avoid multiple atomic reads
- Only access thrdID when state is RUNNING or INITIALIZING
- Skip pthread_cancel/pthread_kill if thread is in WAIT_JOIN state
- Remove duplicate pthread_cancel() call
- Add comments explaining the race prevention

This ensures thrdID is only accessed when the thread is guaranteed to exist, preventing signals from being sent to wrong or invalid threads.

Impact:
- Eliminates potential for sending signals to wrong thread
- Prevents undefined behavior from accessing invalid thrdID
- More robust cancellation logic
- Cleaner code (no duplicate operations)

Testing:
- Builds successfully
- Passes smoke test (imtcp-basic.sh)
- Thread state checked before every thrdID access

With the help of AI Agents: GitHub Copilot

### DIFF
--- a/runtime/wti.c
+++ b/runtime/wti.c
@@ -162,40 +162,69 @@ rsRetVal wtiWakeupThrd(wti_t *pThis) {
  */
 rsRetVal ATTR_NONNULL() wtiCancelThrd(wti_t *pThis, const uchar *const cancelobj) {
     DEFiRet;
+    int state;
+    wtp_t *pWtp;
+    pthread_t thrdID;
 
     ISOBJ_TYPE_assert(pThis, wti);
+    pWtp = pThis->pWtp;
+    ISOBJ_TYPE_assert(pWtp, wtp);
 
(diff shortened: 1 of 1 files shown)


rgerhards:
CI Fail is known flake

PR #6439, opened 2026-01-09
Fix race in worker instance creation (action.c)

rgerhards:
Motivation: CI surfaced TSAN-related flakes suggesting a threading edge case. This change keeps the atomic fast path efficient and makes the non-atomic slow path portable and safe. No real-world bug reports.
    
Impact: Non-atomic builds avoid a nested-lock scenario in worker init. Atomic builds retain the lock-free fast path. No API/ABI change.
    
Before/After: non-atomic builds could check instance existence via a mutex-backed fallback while holding the same mutex / the check happens under a single mutex (or via atomics), preventing self-deadlock.
    
Technical: when HAVE_ATOMIC_BUILTINS is set, perform a fast pointer read and re-check under mutWrkrDataTable (double-checked locking). On non-atomic toolchains, take mutWrkrDataTable once and read actWrkrData directly, avoiding fallback atomics that would re-lock the same mutex. Provide pointer atomics for both __sync-based and mutex-backed paths, and return type fix for ATOMIC_FETCH_32BIT_unsigned. Also reverts a recent fix-attempt from earlier this week that introduced the nested-lock risk on non-atomic builds. Add ommysql README with TSAN context and extend tsan-rt.supp for libmysqlclient. Minor whitespace tweaks in diag.sh.
    
 With the help of AI-Agents: Claude (Anthropic), Codex

### DIFF
--- a/action.c
+++ b/action.c
@@ -945,6 +945,29 @@ static rsRetVal actionCheckAndCreateWrkrInstance(action_t *const pThis, const wt
     if (actionIsDisabled(pThis)) {
         FINALIZE;
     }
+
+    /* Fast path: check without lock if instance already exists.
+     * This optimization avoids mutex contention in the common case where
+     * the worker instance has already been created.
+     */
+    if (pWti->actWrkrInfo[pThis->iActionNbr].actWrkrData != NULL) {
+        FINALIZE;
+    }
+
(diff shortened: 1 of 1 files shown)


rgerhards:
will most probably be replaced by #6441, which has a better fix.

rgerhards:
replaced by #6441

PR #6449, opened 2026-01-12
core: make action state changes output better debug info

rgerhards:
The helper actionSetState() now outputs the function name which requests the state change into the debug log. This makes it easier to follow control-of-flow while analyzing the debug log.

Many thanks to Flos Lonicerae for suggesting this feature and providing a prototype PR.

With the help of AI Agents: ChatGPT Codex

see also https://github.com/rsyslog/rsyslog/pull/5120


rgerhards:
replaces #5120

PR #6461, opened 2026-01-16
Harden disk queue recovery after invalid .qi seeka

alorbach:
Hardens disk queue recovery after dirty shutdowns by resetting read/delete pointers and size accounting when an invalid `.qi` file is detected, and by repositioning the read stream so cleanup continues reliably for the daqueue dirty-shutdown scenario.

### References
Refs: https://github.com/alorbach/rsyslog/actions/runs/21024971278

### Notes (optional)
- Tests could not run because `libestr` was missing during configure (`make -j$(nproc) check TESTS=""` and `./tests/daqueue-dirty-shutdown.sh`).
- `devtools/format-code.sh` was not run to avoid reformatting the entire tree.

---

#### Quick check (optional)


rgerhards:
@cubic-dev-ai review

rgerhards:
Thx, this looks like a very welcome hardening!

PR #6470, opened 2026-01-19
testbench: robustly stop elasticsearch using kill -9 on timeout

rgerhards:
The previous implementation of stop_elasticsearch in diag.sh used wait_pid_termination which aborts the test script execution if the process does not terminate within the timeout, leaving the Elasticsearch process running. This caused subsequent tests to fail or hang when they tried to reuse the stuck instance.

This change replaces wait_pid_termination with a loop that waits for the process to exit and, if it does not exit within TB_TIMEOUT_STARTSTOP, forcefully terminates it with kill -9. This ensures that a stuck Elasticsearch instance is properly cleaned up, preventing it from interfering with future test runs.


rgerhards:
@cubic-dev-ai review

PR #6491, opened 2026-01-28
testbench: Scope macOS sanitizer reporting to rsyslog

alorbach:
macOS CI intermittently fails because TSAN/ASAN report issues from system libraries rather than rsyslog. This adjusts the macOS workflow to keep sanitizer reports focused on rsyslog code paths.

### References
Refs: https://github.com/rsyslog/rsyslog/actions/runs/21432360489

### Notes (optional)
- **Workflow: run_macos.yml**
  - Add `ignore_noninstrumented_modules=1` to ASAN/TSAN options so only instrumented rsyslog code is reported.
- **Workflow: run_macos_weekly.yml**
  - Mirror the same sanitizer scoping for weekly macOS runs.

### DIFF
--- a/tests/imtcp-msg-truncation-on-number.sh
+++ b/tests/imtcp-msg-truncation-on-number.sh
@@ -18,7 +18,7 @@ tcpflood -m1 -M "\"<120> 2011-03-01T11:22:12Z host tag: this is a way too long m
 shutdown_when_empty
 wait_shutdown
 
-grep "Framing Error in received"  $RSYSLOG_OUT_LOG > /dev/null
+grep -E "Framing Error in received|received oversize message from peer" $RSYSLOG_OUT_LOG > /dev/null
 if [ $? -ne 0 ]; then
         echo
         echo "FAIL: expected error message from imtcp not found.  $RSYSLOG_OUT_LOG is:"


.github/workflows/run_macos.yml
```diff
@@ -164,11 +164,13 @@ jobs:
             echo "Configuring AddressSanitizer options"
             export ASAN_OPTIONS="verbosity=1:halt_on_error=1:\
```
rgerhards: Isn't `verbosity=1` the root issue of the overwhelming and useless output volume?

alorbach: Yes, you're right. verbosity=1 is contributing to the overwhelming output. The SanitizerCommonFlags docs show verbosity=0 (default) is silent, while verbosity=1 adds diagnostic output. The Linux CI workflows don't use verbosity=1 at all, so it's not needed here either.  I will update the PR

alorbach:
@cubic-dev-ai review

rgerhards:
Instead of requesting squash, proper committer ID and commit message, I have re-created the minimal change in https://github.com/rsyslog/rsyslog/pull/6492 and will merge that one. I think that's faster and simpler for all parties involved :-)

Issue #3404, opened 2019-01-09
imtcp-tls-ossl-x509fingerprint.sh fails with "Bad File Descriptor"


rgerhards:
This also happens occasionally with some other tests. A sample log is available here:
https://build.rsyslog.com/#/builders/177/builds/314/steps/4/logs/test-suite_log

I traced this down to a multithreaded race *on open and close calls*. I am very surprised to see this type of issue, as I think the kernel guarantees that it does not happen. Or did I do some programming error (more probable, but...)?

Here is what I see in strace -f:
openssl accepts a socket:
```
[pid 66386] accept(4, {sa_family=AF_INET, sin_port=htons(59054), sin_addr=inet_addr("127.0.0.1")}, [128->16]) = 10
```
then, it works a bit with that socket, detects a failure and shuts it down. Sometimes, at the very same instant omfile on another thread tries to open on output file. Then the following happens:
```
[pid 66386] close(10)                   = 0
[pid 66389] openat(AT_FDCWD, "./rstb_356100_31fa9d20.out.log", O_WRONLY|O_CREAT|O_NOCTTY|O_APPEND|O_CLOEXEC, 0644 <unfinished ...>
[pid 66386] close(10 <unfinished ...>
[pid 66389] <... openat resumed> )      = 10
[pid 66386] <... close resumed> )       = 0
[pid 66386] poll([{fd=4, events=POLLIN}, {fd=5, events=POLLIN}], 2, -1 <unfinished ...>
[pid 66389] write(2, "file './rstb_356100_31fa9d20.out"..., 66file './rstb_356100_31fa9d20.out.log' opened as #10 with mode 420
) = 66
[pid 66389] ioctl(10, TCGETS, 0x7f59aeb89540) = -1 EBADF (Bad file descriptor)
```
This is **literally** from the log, without deleting or reordering lines. I read it so that there is a race between `open` and `close` where fd 10 is reused, but seemingly closed - resulting in the `EBADF`

Does that really point to an issue inside the kernel? Any other explanation?

Note that I can reproduce the problem locally by running `./imtcp-tls-ossl-x509fingerprint.sh` in a loop until it fails. It does NOT fail if debug logging is active. That also points to a race like the one seen by strace,

rgerhards:
@mbiebl @Whissi @jvymazal I would appreciate if you could have a look at this issue. I am really more than a bit clueless ;-) Any idea is appreciated.

jvymazal:
[...]

PR #6501, opened 2026-01-30
stream: lock open/init and defer error log

rgerhards:
Why: Avoid a rare async close/open race and prevent lock-order inversion when internal logging runs during open failures.

Impact: Error-path logging timing changes; tests:
./tests/imtcp-tls-ossl-x509fingerprint.sh

Before/After: Before, async open released its lock before init and could log while locked; after, open/init stays locked and LogError is deferred until unlock.

Technical Overview:
- Hold the async writer mutex for open(2), isatty, and crypto init.
- Defer LogError emission until after mutex release.
- Keep errno and error code for accurate post-unlock logging.
- Add a clarifying doc comment for doPhysOpen.

Closes https://github.com/rsyslog/rsyslog/issues/3404

### DIFF
--- a/runtime/stream.c
+++ b/runtime/stream.c
@@ -239,11 +239,20 @@ finalize_it:
  * strm instance object.
  */
 
-/* do the physical open() call on a file.
+/**
+ * @brief Open the underlying file descriptor and initialize stream state.
+ *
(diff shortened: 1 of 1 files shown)


rgerhards:
@cubic-dev-ai review

PR #6531, opened 2026-02-06
omhttp: reuse reply buffer safely

rgerhards:
Why: reduce response handling overhead and avoid leaks

Impact: adds replymaxbytes cap; adjusts curl sizing logic

Before/After: reply buffers reallocated per chunk -> reused with limits

Technical Overview:

- Track reply length as size_t and reuse the buffer

- Enforce replymaxbytes and overflow guards during curl callbacks

- Keep replies NUL-terminated for logging/metadata

- Use CURLOPT_POSTFIELDSIZE_LARGE for large payloads

- Add defaults and config parsing for replymaxbytes

### DIFF
--- a/contrib/omhttp/omhttp.c
+++ b/contrib/omhttp/omhttp.c
@@ -48,6 +48,7 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <fcntl.h>
+#include <limits.h>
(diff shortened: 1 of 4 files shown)


rgerhards:
@cubic-dev-ai review

PR #6542, opened 2026-02-09
configure: default-enable impstats push build support

rgerhards:
Why:
Impstats remote-write support should be built by default so users get it without extra configure flags.

Before/After:
Before: --enable-impstats-push defaulted to no.
After: --enable-impstats-push defaults to yes.

Technical Overview:
Switch impstats-push default to yes in AC_ARG_ENABLE help and parsing. Keep strict dependency checks so missing protoc-c/libcurl/protobuf-c/ snappy abort configure when push is enabled.
Auto-enable impstats when push is enabled and impstats was not explicitly set by the user.
Preserve explicit conflict behavior: if user disables impstats while push is enabled, configure aborts with a clear error.

Impact: Configure now fails by default when push deps are missing unless users install deps or pass --disable-impstats-push.

### References

### Notes (optional)

---

#### Quick check (optional)

---

  ```
  git commit --amend
  ```

### DIFF
--- a/.github/workflows/run_checks.yml
+++ b/.github/workflows/run_checks.yml
@@ -1583,6 +1583,7 @@ jobs:
                 ./configure --enable-silent-rules --enable-testbench \
                   --enable-imdiag --disable-imdocker --disable-imfile \
                   --disable-default-tests --disable-impstats \
+                  --disable-impstats-push \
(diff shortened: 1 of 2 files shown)


rgerhards:
codefactor is not really relevant here. all pre-existing and not related

PR #6549, opened 2026-02-12
ci: extend GitHub Actions for ARM (armhf/arm64) QEMU builds and tests

alorbach:
### Summary
Adds ARM CI to the GitHub Actions workflow so rsyslog is built and tested on
armhf and arm64. arm64 uses the native GitHub runner (ubuntu-24.04-arm);
armhf uses QEMU emulation (no armhf runner available). Includes a dedicated
ARM Docker image and graceful skipping of network-namespace tests when they
cannot run (e.g. under QEMU user-mode).

### Changes
- **run_checks.yml**: New `arm_CI` job for armhf and arm64.
  - **armhf**: ubuntu-24.04 + QEMU, reduced test set (--disable-default-tests,
    many modules disabled) for faster QEMU runs.
  - **arm64**: ubuntu-24.04-arm (native), expanded test set (default tests,
    imfile, impstats, imptcp, relp, gnutls, etc.). Skips when no relevant
    files changed.
- **devtools/ci/Dockerfile.arm**: Ubuntu 24.04 with build tools and deps
  (gnutls, libestr, libfastjson, zlib, libgcrypt, librelp, uuid, libyaml,
  iproute2).
- **diag.sh**: Add `require_netns_capable()` and `skip_ASAN()`; netns tests and
  ASan-incompatible tests (empty-hostname, omfile-read-only*) skip when needed.
- **ratelimit.c**: Fix double-free in ratelimitFreeShared; `shared->name` is the
  hashtable key (freed by hashtable_destroy), remove redundant free.
- **run_checks.yml**: Quote `$GITHUB_OUTPUT` and `$GITHUB_STEP_SUMMARY` in
  clang static analyzer steps.

### Testing
- ARM CI runs on PRs when C sources, tests, Makefiles, Dockerfile.arm, or
  configure.ac change.
- Netns tests skip cleanly (exit 77) when namespace creation is unavailable.
- ASan-incompatible tests skip (empty-hostname, omfile-read-only*).

PR #6550, opened 2026-02-12
ci: add ubuntu 24 arm64 matrix run in checks

rgerhards:
Why
We need parity coverage on GitHub's arm64 runner for the Ubuntu 24 CI profile, and we cannot rely on arm dev containers here.

Impact
The CI matrix now runs the Ubuntu 24 profile natively on arm64 with manual dependency installation.

Before/After
Before: Ubuntu 24 CI checks ran only in the container path. After: A dedicated ubuntu_24_arm64 matrix entry runs the same profile on ubuntu-24.04-arm.

Technical Overview
- Added ubuntu_24_arm64 to the CI job matrix.
- Added a matrix include override to select runs-on: ubuntu-24.04-arm only for that config.
- Added an arm64-native dependency install step derived from packaging/docker/dev_env Ubuntu 24 base dependencies.
- Reused the same configure/test flags as ubuntu_24 for arm64.
- Switched execution path for arm64 to devtools/run-ci.sh while keeping container-based execution for all other matrix entries.
- Moved skip behavior earlier in the matrix flow by checking for relevant changes immediately and short-circuiting before heavy steps.

With the help of AI-Agents: Codex (GPT-5)

### DIFF
--- a/.github/workflows/run_checks.yml
+++ b/.github/workflows/run_checks.yml
@@ -60,6 +60,8 @@ jobs:
             tests/*.sh
             diag.sh
             **/Makefile.am
+          files_ignore: |
+            doc/Makefile.am
 
       - name: run compile check (container)
         if: steps.code_changes.outputs.any_changed == 'true'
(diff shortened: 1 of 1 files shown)


rgerhards:
replaced by https://github.com/rsyslog/rsyslog/pull/6549

Issue #3395, opened 2019-01-04
[Improvement] omkafka - enabling librdkafka statistics can potentially cause large memory growth


n2yen:
[Improvement] omkafka - enabling **statistics.interval.ms** can potentially cause large memory growth if a configured input source is not receiving input.

### Background
librdkafka appears to handle statistics callbacks as _stats_ events. These _stats_ events, which include the json string that should be processed are enqueued into an internal queue. The expectation from librdkafka is that the client will call rd_kafka_poll() periodically to process these events.
Thus for each configured omkafka instance with statistics interval/callbacks enabled, there could be a back-log of these stats-callback events if the configured input source is dormant for some reason.

Is there a built-in path in rsyslog to consume these statistic callback events periodically? Perhaps periodically wake up dormant omkafka instances?

### Expected behavior
N/A
### Actual behavior
N/A
### Steps to reproduce the behavior
1. create a input configuration with omkafka as the output destination, for example imfile -> omkafka.
2. add statistics callback to the omkafka configuration with a short interval, for example 30000ms.
3. generate some logs to the configured imfile input file, and confirm the statistics callback is executed.
4. pause log generation for a few minutes or some multiple of configured statistics callback interval.
5. verify statistics callback is not executed during this pause.
6. resume generating logs, and observe multiple statistics callback executed in succession.

### Environment
- rsyslog version:  all versions?
- platform:  all platforms?
- for configuration questions/issues, include rsyslog.conf and included config files

PR #6551, opened 2026-02-12
omkafka: poll callbacks while action is idle

rgerhards:
Why
When statistics.interval.ms is enabled, librdkafka queues stats events until rd_kafka_poll() is called. Idle omkafka actions can go long periods without polling, so callbacks pile up and consume memory.

Impact
Stats callbacks are drained continuously even when no new messages arrive, preventing callback queue buildup.

Before/After
Before: callbacks were mostly drained only during message processing. After: a lightweight poll thread drains callbacks during idle periods.

Technical Overview
Add a dedicated callback poller thread per omkafka instance. The thread uses mut_doAction and rkLock so callback execution stays serialized with doAction and handle recreation.
The thread wakes every 100ms and calls rd_kafka_poll(..., 0) only when the kafka handle is open.
Start the thread once openKafka succeeds and stop it during freeInstance. Thread lifecycle state is tracked in instanceData.

closes https://github.com/rsyslog/rsyslog/issues/3395

With the help of AI-Agents: GPT-5.2-Codex

### DIFF
--- a/plugins/omkafka/omkafka.c
+++ b/plugins/omkafka/omkafka.c
@@ -217,6 +217,8 @@ static uint64 getClockTopicAccess(void) {
 static int closeTimeout = 1000;
 static pthread_mutex_t closeTimeoutMut = PTHREAD_MUTEX_INITIALIZER;
 
+#define OMKAFKA_POLL_THREAD_SLEEP_USEC 100000
+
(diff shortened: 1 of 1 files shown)


rgerhards:
@cubic-dev-ai review

PR #6570, opened 2026-02-19
ci: add DOC_BASE_URL variable for docs Sitemap base

alorbach:
Adds DOC_BASE_URL repository variable so the docs Sitemap in robots.txt can use https://docs.rsyslog.com instead of rsyslog.github.io. Forks without the variable keep the github.io URL.


rgerhards (approved):
fine now

alorbach:
> now that the rsyslog doc lives on docs.rsyslog.com, shouldn't we update RSYSLOG_BASE_URL to that URL? This would be consistent with current state. The override is still very good to have, so I could also merge as-is (I need to apply at least one other change to the doc to cleanly finalize the move).

Yes - updated the URL - we still keep the override (Which is useful for forked repo testing).

Issue #5771, opened 2025-07-12
queue: Harden Disk Queue Recovery Against State Corruption


rgerhards:
### Problem Statement

The rsyslog disk queue (`queue.type="disk"` or a DA queue that has spooled) is not resilient to state file corruption. If the queue's metadata file (`.qi`) is lost, corrupt, or becomes inconsistent with the on-disk segment files, rsyslog may fail to start, silently lose the spooled data, or enter an unpredictable state.

This is a critical reliability issue, especially in microservice environments where pods can be terminated ungracefully, increasing the likelihood of state file corruption.

### Current Behavior (The "Pain")

A fresh analysis of the startup logic in `queue.c` (`qqueueConstruct` -> `qqueueLoad`) reveals the following failure modes:

1.  **`.qi` File Missing, Segment Files Exist:** If the `.qi` file is deleted but segment files (e.g., `qname.00000001`) remain, `qqueueLoad` fails to find the `.qi` file. The system then assumes it's starting fresh. It creates a *new* `.qi` file with pointers starting at segment `00000001`. If an old `qname.00000001` file exists, it will be overwritten, and its data is silently lost.
2.  **`.qi` File Corrupt (Partial Write):** If the `.qi` file is smaller than `sizeof(qqueue_t)`, the `read()` call in `qqueueLoad` will return a short read. The current code treats this as a file-not-found error and proceeds as in scenario #1, leading to silent data loss.
3.  **Inconsistent State:** If rsyslog crashes after creating a new segment file but *before* the `.qi` file is checkpointed, that segment becomes an "orphan." The current recovery logic does not detect or handle these orphaned files.

The core issue is that the system **trusts the `.qi` file implicitly** and lacks a robust mechanism to **verify its state against the ground truth of the files on disk.**

### Desired Behavior (The "Solution")

The queue subsystem must be hardened to always start in a predictable, safe state.

1.  **Corruption Detection:** On startup, the queue must perform a sanity check to detect any inconsistencies between the `.qi` file and the segment files in its spool directory.
2.  **Safe Mode Recovery (Default):** If any corruption or inconsistency is detected, the default behavior should be to prioritize service availability and prevent data contamination. The system should:
    a. Create a new directory, e.g., `{workDir}/{qname}.bad/{timestamp}`.
    b. Move all existing queue files (`.qi` and `qname.*`) for the corrupted queue into that directory for later forensic analysis.
[...]

PR #6572, opened 2026-02-19
queue: harden disk queue corruption recovery

rgerhards:
Why:
Queue recovery must fail safe and remain compatible with existing startup flows, including disk queue tests that rely on deferred spool setup.

Impact:
Disk queue startup validates corruption more safely, and imjournal CI is more resilient to transient readiness timing under journal churn.

Before/After:
Before, some startup paths could incorrectly switch to emergency mode when spool scanning could not run, and imjournal readiness could fail flakily. After, startup keeps legacy behavior when scan is unavailable and imjournal retries readiness before failing.

Technical Overview:
Expose queue.oncorruption as a user parameter and document it. Set explicit default mode to safe for queue defaults/construct paths. Harden disk queue scan parsing with bounds, overflow, and alloc checks. If spool directory/prefix is unset, return original load status unchanged. If spool scan cannot run (ENOENT/ENOTDIR), skip corruption scan. On recovery-path failures, switch to emergency in-memory mode with alerts. Retry imjournal readiness probes with bounded attempts and timeout. If readiness is still not observed, skip test as environment-limited.

Closes https://github.com/rsyslog/rsyslog/issues/5771

With the help of AI-Agents: codex

### DIFF
--- a/doc/source/rainerscript/queue_parameters.rst
+++ b/doc/source/rainerscript/queue_parameters.rst
@@ -318,6 +318,38 @@ be turned on without a good reason. Note that the penalty also depends on
 *queue.checkpointInterval* frequency.
 
 
+queue.onCorruption
+------------------
+
+.. csv-table::
+   :header: "type", "default", "mandatory", "|FmtObsoleteName| directive"
+   :widths: auto
+   :class: parameter-table
+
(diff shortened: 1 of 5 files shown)


rgerhards:
wrong repo. this is just a test

PR #6573, opened 2026-02-19
ci: restrict doc deployment to official repository

rgerhards:
This prevents documentation deployment workflows from running on forks, saving GitHub Actions minutes and reducing noise for contributors. The workflow will now only execute when the repository is 'rsyslog/rsyslog'.


rgerhards:
@alorbach this is best practice and we should strongly consider it.

PR #6579, opened 2026-02-23
queue/runtime/tests: harden diskqueue recovery and shutdown race handling

rgerhards:
Why:
This branch combines two related hardening steps for disk queue reliability:
- robust corruption detection/recovery handling in disk queue state/file validation
- worker startup cancellation-race closure that could lead to shutdown wait loops

Impact:
- disk queue scan now rejects out-of-range segment sequence numbers early and reports corruption deterministically.
- worker startup no longer exposes a cancellation window before cleanup registration.
- test/CI diagnostics preserve timeout backtraces (gdb) in ARM jobs and print them to stdout for post-mortem debugging.
- test script cleanup removes redundant operations and uses a macOS-friendlier segment enumeration path.

Technical Overview:
- runtime/queue.c:
  - add out-of-range sequence-number rejection during spool scan
  - keep orphan-loop range check as defensive fallback
- runtime/wtp.c:
  - disable cancellation and register cleanup before publishing RUNNING
  - document startup/cancellation invariant inline
- runtime/wti.c:
  - add concise cancellation-contract comment
- devtools/ci/Dockerfile.arm:
  - install gdb for CI timeout diagnostics
- tests/diskqueue-oncorruption-missing-segment.sh:
  - emit timeout gdb backtraces to stdout
  - drop redundant STARTED_LOG truncate
  - avoid GNU find -printf/mapfile dependency in segment listing

### DIFF
--- a/devtools/ci/Dockerfile.arm
+++ b/devtools/ci/Dockerfile.arm
@@ -7,5 +7,5 @@ RUN apt-get update \
        libgnutls28-dev libestr-dev libfastjson-dev zlib1g-dev \
        libgcrypt20-dev librelp-dev uuid-dev libyaml-dev \
        libcurl4-gnutls-dev libprotobuf-c-dev protobuf-c-compiler libsnappy-dev \
-       iproute2 \
+       iproute2 gdb \
     && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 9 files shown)


rgerhards:
gemini review items are all acceptable

PR #6581, opened 2026-02-23
Omotel protobuf

rgerhards:


rgerhards:
@cubic-dev-ai review

rgerhards:
> would just check the findings from "cubic-dev-ai", then it can be commited.
> Testcase was also added.

Indeed, I admit I overlooked that part. Sry.

PR #6593, opened 2026-02-25
tests: Makes all imfile-logrotate* tests pass with logrotate 3.8+

alorbach:
## Problem

- logrotate 3.8+ refuses to rotate files in world/group-writable parent
  directories unless a `su` directive is present
- `su root root` triggers "Operation not permitted" in non-root CI
  (e.g. uid 1000), since switching to root is not allowed

## Solution

Add dynamic `su $(id -un) $(id -gn)` to all five imfile-logrotate test
configs. This satisfies logrotate’s security check without requiring root.

## Tests Updated

- imfile-logrotate.sh
- imfile-logrotate-async.sh
- imfile-logrotate-copytruncate.sh
- imfile-logrotate-multiple.sh
- imfile-logrotate-nocopytruncate.sh

## Testing

- Tests pass locally under root and non-root
- CI succeeds with logrotate 3.8+ in non-root environments

### DIFF
--- a/tests/imfile-logrotate-async.sh
+++ b/tests/imfile-logrotate-async.sh
@@ -13,6 +13,7 @@ export RETRIES=50
 # Write logrotate config file
 echo '"./'$RSYSLOG_DYNNAME'.input*.log"
 {
+    su '"$(id -un)"' '"$(id -gn)"'
     #daily
     rotate 60
     missingok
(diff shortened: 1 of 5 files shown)


alorbach:
The current implementation is correct. The '"$(id -un)"' '"$(id -gn)"' quoting is appropriate for embedding command substitution inside an otherwise single-quoted string. Gemini’s proposed change would break the tests, especially imfile-logrotate-async.sh. No code change is required to address these review comments.

PR #6595, opened 2026-02-26
ratelimit: wire ratelimit.name into remaining modules

rgerhards:
Why: Commit 3ee31a8d0 added shared named ratelimit support (ratelimit.name + ratelimitNewFromConfig) but only wired it into imtcp, imptcp, and imudp. Seven modules with existing ratelimit.interval/ratelimit.burst support were left without the ability to reference shared ratelimit() objects.

Impact: Adds ratelimit.name parameter to omfwd,
omelasticsearch, omhttp, imhttp, imjournal, imklog, and imuxsock. Existing configurations remain unchanged.

Before: These seven modules could only use inline
ratelimit.interval/ratelimit.burst parameters.
After: All modules can now reference shared ratelimit() configuration objects via ratelimit.name.

Technical Overview:
For each module, the same pattern from the imudp reference implementation is applied:
- Add uchar *pszRatelimitName to config struct
- Add ratelimit.name to cnfparamdescr
- Change interval/burst defaults to -1 (sentinel)
- Parse ratelimit.name in config handler
- Add mutual-exclusivity check (name vs interval/burst)
- Branch ratelimiter creation on pszRatelimitName: if set, call ratelimitNewFromConfig(); else legacy path
- Free pszRatelimitName in destructor

imuxsock is the most complex: it supports both per-socket ratelimit.name and syssock.ratelimit.name, with the per-source hashtable ratelimiter path also updated.

imklog uses legacy parameter names (ratelimitinterval, ratelimitburst) but the new parameter uses the dotted form (ratelimit.name) for consistency.

Each module receives a parameter doc page, module doc update, and a config-validation test.

With the help of AI-Agents: GitHub Copilot

### DIFF
--- a/ai/rsyslog_memory_auditor/audit_leaks.txt
+++ b/ai/rsyslog_memory_auditor/audit_leaks.txt
@@ -1,0 +1,17 @@
+You are a senior C security and performance auditor. Analyze the following rsyslog module code specifically for memory leaks on error paths.
+
+**Audit Focus: Error Path Leaks & Ownership**
+
+1. **Error Path Leaks**: 
+   - Trace every `RS_RET` return path. 
+   - Ensure that any memory allocated via `malloc`, `calloc`, or `strdup` (common in `setInstParam`) is freed before an error return.
+   - Check if `pData` or `WID` sub-elements are leaked during partial initialization failure.
+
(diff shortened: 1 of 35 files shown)


rgerhards:
as usual codefactor is way off --> ignore

PR #6610, opened 2026-03-02
docs: remove -c5 from troubleshooting example

anargam:
The -c flag has been removed. The deprecation warning in the manpages was removed in 2014, with v8.6.0.

Commit: https://github.com/rsyslog/rsyslog/commit/79b7c05331a0fd97be2790ad40519de5473cabac

Diff between tagged versions: https://github.com/rsyslog/rsyslog/compare/v8.5.0...v8.6.0


anargam:
The replacement for `-c5` is `-iNONE`, which Debian users might recognise from https://salsa.debian.org/debian/rsyslog/-/blob/debian/master/debian/rsyslog.service

rgerhards (approved):
well spotted - many thx for the PR!

PR #6627, opened 2026-03-16
ci: extend focused repo policy checks

rgerhards:
Why: the first deterministic policy workflow covers the highest-value rules, but it still leaves new-module build wiring and parameter reference doc coverage to human review.

Impact: the focused policy review now also checks top-level build wiring for new modules and warns when new module parameters lack matching reference docs.

Before/After: before the workflow only checked tests, doc distribution, and basic module onboarding; after it also covers module build manifests and parameter doc sync.

Technical Overview:
Extend the review package builder with deterministic facts for new module build wiring against the top-level `Makefile.am` and `configure.ac`.

Add a `parameter-doc-sync` rule that compares newly introduced `cnfparamdescr` names against the expected files under `doc/source/reference/parameters/`.

Teach the deterministic evaluator about the two new checks and update the workflow trigger and documentation to match the expanded rule set.

### DIFF
--- a/.github/workflows/repo-policy-focus-review.yml
+++ b/.github/workflows/repo-policy-focus-review.yml
@@ -10,6 +10,8 @@ on:
   pull_request:
     paths:
       - ".github/workflows/repo-policy-focus-review.yml"
+      - "Makefile.am"
+      - "configure.ac"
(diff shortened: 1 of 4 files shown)


rgerhards:
@cubic-dev-ai review

Issue #6576, opened 2026-02-21
[General]: test-suite failure: FAIL: imfile-symlink-ext-tmp-dir-tree


mbiebl:
Version: 8.2602.1
OS: Debian sid

A full build log is attached but can also be found at
https://salsa.debian.org/debian/rsyslog/-/jobs/9088796

[build.txt](https://github.com/user-attachments/files/25458971/build.txt)

The problem is reproducible here. So I can run further diagnostics if you want.

mbiebl:
I ran `git bisect` and it shows 262b22d82a40811ee14ed2cc3ff930d8eb45c9d4 as the first faulty commit.
This commit does indeed look related.

mbiebl:
```
commit 262b22d82a40811ee14ed2cc3ff930d8eb45c9d4 (HEAD)
Author: Rainer Gerhards <rgerhards@adiscon.com>
Date:   Thu Jan 22 17:21:40 2026 +0000

    imfile: add inode generation check for file rotation detection
    
    This commit adds a check for inode generation number in the imfile module.
    This addresses an issue where file rotation is not detected when a new file
    reuses the same inode number as the old one (which can happen on some
    filesystems or scenarios).
    
    The check is implemented by using `FS_IOC_GETVERSION` ioctl on Linux.
    It is guarded by `HAVE_FS_IOC_GETVERSION` and `HAVE_LINUX_FS_H`.
    
    To minimize performance impact, the check is only performed if `st_ctime`
    has changed while the inode number remains the same.
    
    This also updates the state file format to include `inode_gen` to detect
    inode reuse across rsyslog restarts.
    
    With the help of AI Agents: Google Jules, github copilot
    
    closes https://github.com/rsyslog/rsyslog/issues/6292
```

mbiebl:
@rgerhards did the test-suite catch a regression or does the test-suite simply need to be updated?

rgerhards:
@mbiebl 
> did the test-suite catch a regression or does the test-suite simply need to be updated?

I am actually quite concerned. The testbench did not detect anything in our CI. And you know it is extensive. We often have this - testbench works, but on deb build it does not work. I specifically added a Debian 13 checker to avoid this, but that also seems not to work. I am currently trying again to create one workflow that should be identical to what you do, but I begin to wonder if this is all racy and so we cannot really reproduce it - maybe a test harness issue.

[...]

PR #6630, opened 2026-03-16
imfile: fix FD leak when symlink and target both enter pending_destroy

alorbach:
When act_obj_destroy() was called for a deleted symlink, it searched act->edge->active for the target to call detect_updates() on the parent directory edge. However, after commit 262b22d, detect_updates() may move both the symlink AND its target to pending_destroy in the same pass (when both have expired deletion delays). In that case, the target is no longer in act->edge->active, so the search fails, detect_updates(links_edge) is never called, and the target's parent directory FD is never released.

Fix: instead of searching act->edge->active for the target, directly call detect_updates(act->edge->parent->root->edges). This is always correct because the target is always added to the same edge as the symlink by process_symlink(), making act->edge->parent->root->edges equivalent to target_act->edge->parent->root->edges.

Fixes: imfile-symlink-ext-tmp-dir-tree test failure (issue #6576)

### References

### Notes (optional)

---

#### Quick check (optional)

---

  ```
  git commit --amend
  ```

### DIFF
--- a/plugins/imfile/imfile.c
+++ b/plugins/imfile/imfile.c
@@ -1117,14 +1117,19 @@ static void act_obj_destroy(act_obj_t *const act, const int is_deleted) {
     DBGPRINTF("act_obj_destroy: act %p '%s' (source '%s'), wd %d, pStrm %p, is_deleted %d, in_move %d\n", act,
               act->name, act->source_name ? act->source_name : "---", act->wd, act->pStrm, is_deleted, act->in_move);
     if (act->is_symlink && is_deleted) {
-        act_obj_t *target_act;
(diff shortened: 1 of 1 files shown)


rgerhards:
@cubic-dev-ai this is a regression that occurred in commit 262b22d82a40811ee14ed2cc3ff930d8eb45c9d4. That commit fixed a potential use-after-free condition. IMHO this condition is partly re-introduced here. Read and understand commit 262b22d82a40811ee14ed2cc3ff930d8eb45c9d4 and thoroughly check this PR/patch if part of the undesired behavior reoccurs.

rgerhards:
after careful evaluation, I close this in favor of #6629. The latter has less overhead and handles some edge cases better.

PR #6647, opened 2026-03-20
ci: stop PR docs from overwriting Pages

rgerhards:
Why: the PR docs workflow deploys a Pages artifact that contains only
`pr-<num>/`, which replaces the production site and turns `/` and
`/doc/` into 404s.

Impact: PR doc workflows still build and upload HTML artifacts and post
review comments, but only the main docs deploy workflow updates GitHub
Pages.

Before/After: before a later PR preview run could overwrite the live
docs site; after PR workflows no longer publish to Pages, so production
docs remain owned by `doc_deploy_main.yml`.

Technical Overview:
- Remove top-level Pages/id-token permissions from `doc_build.yml`.
- Convert the PR preview job into a comment-only job.
- Drop the `configure-pages`, `upload-pages-artifact`, and
  `deploy-pages` steps from PR runs.
- Keep the PR comment, but link to the workflow run and downloadable
  HTML artifacts instead of a Pages preview URL.

Validation:
- Parsed `.github/workflows/doc_build.yml` with `yaml.safe_load`.
- Reproduced the Pages overwrite behavior by correlating the later PR
  workflow run with the return to GitHub Pages 404 responses.


rgerhards:
@alorbach I merged this to solve the big doc trouble. I did not yet check if the workflow continues to work otherwise. In any case, publishing a preview should not be done anymore.

PR #6652, opened 2026-03-20
packaging/docker: define safer image version contract

rgerhards:
## Summary
- change the default container image tag to a clearly non-release local value
- require an explicit stable `VERSION` for publish and `latest` tagging targets
- document the version/tag contract for both users and AI agents
- clarify that container CI uses `ci-<sha>` validation tags by design

## Why
This makes local container builds safe by default while preserving explicit local release rehearsals and preparing the image family for later release automation.

## Validation
- `make -C packaging/docker/rsyslog check_publish_version` fails for the default local version
- `make -C packaging/docker/rsyslog check_publish_version VERSION=2026-03` passes
- `make -C packaging/docker/rsyslog all`

### DIFF
--- a/.github/workflows/container_build.yml
+++ b/.github/workflows/container_build.yml
@@ -43,6 +43,8 @@ jobs:
       - name: Build rsyslog container family
         # Keep the workflow green for unrelated PRs while still exercising the
         # full image family whenever container packaging inputs change.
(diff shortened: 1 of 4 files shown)


rgerhards:
gemin review is not valid, so ready to merge

PR #6676, opened 2026-03-30
mmjsontransform: reload transformation mode from policy

rgerhards:
## Why

The YAML policy file should cover the full transformation behavior that
`mmjsontransform` exposes, so operators can update policy-driven JSON
normalization in one place and reload it on HUP.

## What Changed

- added support for a top-level `mode` key in the YAML policy file
- kept the action parameter `mode` as the fallback default when the
  policy omits it
- reload policy-owned mode together with `map.rename` and `map.drop` on
  HUP
- retain the previous in-memory policy when a HUP reload fails
- updated the policy docs and module docs to describe the expanded
  schema and precedence
- added a ChangeLog entry for the user-visible policy enhancement

## Behavior

Before this change, policy files could only rename or drop keys and the
transform mode stayed fixed until restart.

After this change, policy files can also select `flatten` or
`unflatten`, and a successful HUP reload updates the effective mode and
preprocessing rules together.

`input` and `output` remain action parameters because they control which
message properties are read and written rather than the transformation
behavior itself.

## Validation

- `./devtools/format-code.sh --git-changed`
- `make -j$(nproc)`
- `./tests/mmjsontransform-policy-basic.sh`

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,9 @@
 --------------------------------------------------------------------------------------
 Scheduled Release 8.2604.0 (aka 2026.04) 2026-04-??
+- 2026-03-30: mmjsontransform: add YAML policy preprocessing
+  mmjsontransform can now load a YAML policy file to rename or drop JSON
+  keys before flatten/unflatten processing. Policy files are documented
+  and reloaded on HUP when they change.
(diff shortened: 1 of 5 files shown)


rgerhards:
solaris CI is not needed, mmjsontransform is not built there

Issue #6523, opened 2026-02-04
[Modernization] Replace Internal Backticks with Shell Expansion


rgerhards:
## Goal
Modernize the rsyslog test suite by replacing legacy internal backtick expansion (e.g., ``file=`echo $RSYSLOG_OUT_LOG` ``) with standard shell expansion (e.g., `file="'$RSYSLOG_OUT_LOG'"`). 

**Why?** 
The legacy pattern forces rsyslog to fork a shell at runtime for every expansion. Modern shell expansion is more efficient, faster, and follows current project best practices.

## For Contributors (AI-Assisted Workflow)
This issue is designed to be completed using an AI agent (like Gemini CLI, Aider, Cursor, or ChatGPT with codebase access). 

**Instructions:**
1. Use the **Mission Prompt** below to guide your AI agent.
2. **Limit your work to ONE module or batch (max 10-15 files)** per Pull Request to ensure a smooth review process.
3. Check off the module you are working on in the "Progress Tracking" section below.

---

### 🤖 The Mission Prompt
Copy and paste this prompt into your AI agent:

> "Mission: Modernize rsyslog test scripts by replacing internal backticks with shell expansion.
>
> 1. **Discovery & Selection**: 
>    - Search the `tests/` directory for `.sh` files containing the pattern '`echo $'. 
>    - Pick a specific category or module (e.g., 'imfile', 'imuxsock', 'rscript', etc.).
>    - **Limit this session to a maximum of 10-15 files** to keep the PR manageable.
>    - List the files you found and ask me: 'I suggest starting with these X files. Should I proceed?'
>
> 2. **Filtering (Safety)**: 
>    - **EXCLUDE** any file where the filename contains 'backtick' (e.g., `rscript_backticks.sh`).
>    - **EXCLUDE** any file where comments indicate it specifically tests the 'backtick feature'.
>    - *Reason: We must keep actual feature tests in their original format.*
>
> 3. **Transformation**: 
>    - For the selected files, update the `add_conf` blocks.
>    - Replace: `file=`echo $VAR``  ==>  `file="'$VAR'"`
>    - Replace: `binary=`echo $VAR`` ==> `binary="'$VAR'"`
>    - *Note: The quoting format "'$VAR'" is critical so the bash shell expands the variable before passing it to rsyslog.*
>
> 4. **Verification**: 
>    - For every modified file, run its specific test: `make -C tests [FILENAME].log`
[...]

PR #6692, opened 2026-04-07
[Modernization] Replace Internal Backticks with Shell Expansion

aditiprakash:
### Summary

Modernize the  manon / mmnormalize test suite to remove legacy backtick-based
expansion and align with current shell best practices. This
reduces overhead and improves maintainability, especially in
CI and containerized environments.

Impact: test behavior may change in edge quoting cases.

BEFORE: tests used `echo $VAR` via backticks, spawning shells.
AFTER: tests use direct '$VAR' expansion without subshells.

The update replaces command substitution patterns used to
populate file paths and parameters with direct variable
expansion. This avoids per-expansion process forks and keeps
evaluation within the shell, improving performance and
readability.

Care is taken to preserve quoting semantics where variables
are passed to rsyslog configs. No rsyslog runtime behavior is
changed; only test execution mechanics are affected.

AI-Agent: Copilot 2026-04

Refs: https://github.com/rsyslog/rsyslog/issues/6523

### DIFF
--- a/tests/mmanon_recognize_ipembedded.sh
+++ b/tests/mmanon_recognize_ipembedded.sh
@@ -12,7 +12,7 @@ input(type="imtcp" port="0" listenPortFileName="'$RSYSLOG_DYNNAME'.tcpflood_port
 
 ruleset(name="testing") {
 	action(type="mmanon" ipv4.enable="off" ipv6.enable="off" embeddedipv4.bits="128" embeddedipv4.anonmode="zero")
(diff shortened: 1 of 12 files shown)


rgerhards:
@aditiprakash sry for the delay, I think this was just a CI flake. will re-check soon!

rgerhards (approved):
good cleanup, thx!
this also helps with AI understanding and makes it craft better tests :-)

PR #6695, opened 2026-04-08
mmutf8fix: also fix invalid UTF-8 in RFC 5424 structured data

camathieu:
## Summary
- Extend mmutf8fix to sanitize pszStrucData in addition to msg and tag
- Fixes invalid UTF-8 bytes in structured data values (e.g. [user@51719 h_ua="Brain NBSP Twist"])
- Must be placed before mmpstrucdata in the ruleset so parsed properties are clean
- Includes new test mmutf8fix_sd.sh guarded by both ENABLE_MMUTF8FIX and ENABLE_MMPSTRUCDATA

## Problem

RFC 5424 structured data values can contain invalid UTF-8 sequences — for example, haproxy logs user-agent strings that may include Latin-1 encoded characters like 0xa0 (non-breaking space) or 0xed (accented characters in Spanish app names). Since mmutf8fix only sanitized msg and tag, these bytes passed through mmpstrucdata into the rfc5424-sd property tree and then into downstream outputs like omelasticsearch.

Elasticsearch's JSON parser (Jackson) strictly validates UTF-8 and rejects the **entire** bulk HTTP body when it encounters a single invalid byte — causing all documents in the batch to fail, not just the one with the bad data.

The root cause is that libfastjson's char_needsEscape[256] table has entries 128-255 set to 0, so bytes >= 0x80 pass through the JSON serializer without escaping.

## Fix

pszStrucData is a mutable buffer populated by the RFC 5424 input parser before any action runs. By applying doUTF8()/doCC() to this buffer in doAction, invalid bytes are replaced in place before mmpstrucdata parses the SD into individual rfc5424-sd properties.

```
action(type="mmutf8fix")       # sanitizes msg, tag, AND structured data
action(type="mmpstrucdata")    # parses now-clean SD into $! properties
```

MsgHasStructuredData() macro is used as guard (consistent with mmpstrucdata.c). The lenStrucData field is uint16_t and is explicitly cast to int to match doCC/doUTF8 signatures.

## Test plan

New test mmutf8fix_sd.sh covers:
- [x] Invalid 0xa0 (Latin-1 NBSP) in SD values is replaced with ?
- [x] Invalid 0xed (Latin-1 accented char) in SD values is replaced with ?
- [x] Valid multi-byte UTF-8 in SD values (e.g. España) is preserved
- [x] Valid ASCII in SD values is preserved
- [x] Messages without structured data (SD = "-") are unaffected


rgerhards:
@camathieu thx for the PR. Pls check the I reviews and pontentially upcoming CI failures. I think at there may be some points right in AI review.

camathieu:
Hello @rgerhards, many thanks for having a look to the PR 🙏

Coding style has been fixed, the new tests run only when ENABLE_MMUTF8FIX and ENABLE_MMPSTRUCDATA are true and tests are passing.

The failure of `ratelimit_exclusivity.sh` for arm64 seems unrelated to the change.

Please let me know if there is anything I overlooked, or would you have any suggestion for the change.

rgerhards (approved):

rgerhards:
@camathieu sry for the delay in merge, I was unfortunately swamped.

And thx for the great PR.

PR #6722, opened 2026-04-15
imhttp: add route-scoped API key authentication

rgerhards:
Why: Elastic Agent style HTTP ingestion needs a lighter auth mode than Basic auth, and imhttp also needs route-level flexibility for future endpoint-specific behavior.

Impact: imhttp inputs, health checks, and metrics can now be run as no auth, Basic only, API key only, or Basic plus API key, with new shell and YAML regression coverage.

Before/After: Basic auth was the only supported auth mode; imhttp now supports per-route API key auth and mixed auth selection.

Technical Overview:
Introduce a small route auth context so each registered CivetWeb handler can carry its own auth policy instead of relying on one module-wide Basic auth setting.

Add apiKeyFile support for input endpoints plus matching healthCheckApiKeyFile and metricsApiKeyFile module settings.

Accept API keys from Authorization: ApiKey and X-API-Key headers, with files parsed as one token per line while ignoring comments and blank lines.

Trim leading whitespace for X-API-Key values and add an auth-header limits regression so oversized auth headers cannot silently bypass validation.

Keep the implementation extensible for a larger route table so full Elasticsearch mock mode can build on the same registration path.

Fix nearby imhttp defects worth addressing before further expansion, including safer auth parsing, better cleanup on partial failures, proper submit/flush error handling, and safer metadata string use.

Document the new parameters and examples, and add focused shell and YAML tests for API-key and mixed-auth behavior.

### DIFF
--- a/contrib/imhttp/imhttp.c
+++ b/contrib/imhttp/imhttp.c
@@ -59,7 +59,14 @@ DEFobjCurrIf(glbl) DEFobjCurrIf(prop) DEFobjCurrIf(ruleset) DEFobjCurrIf(statsob
 #define MAX_READ_BUFFER_SIZE 16384
 #define INIT_SCRATCH_BUF_SIZE 4096
 /* General purpose buffer size. */
+/**
+ * Basic-auth credentials are capped at 2 KiB of Base64 text. That is well
+ * above practical user-id/password sizes, yet still small enough to keep the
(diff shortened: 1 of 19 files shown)


rgerhards:
@cubic-dev-ai review

rgerhards:
pkgbuild CI is an ubuntu issue, so ready to merge

PR #6727, opened 2026-04-16
gssapi: harden framing and fix imgssapi settings

rgerhards:
The GSSAPI helper and imgssapi paths benefit from more defensive framing, clearer shared limits, and better coverage for longstanding listener configuration behavior.

Impact
GSS token processing is more robust, and imgssapi now honors its existing max-sessions and dynamic port file settings as documented.

Before/After
Before, token framing checks were less consistent across the GSS helper and modules, and imgssapi silently ignored some listener settings. After, framing logic is centralized and bounded, and the documented listener settings take effect.

Technical Overview
Add shared helpers for decoding framed GSS token lengths and deriving bounded message limits for wrapped payloads.

Update gss-misc to apply token-size limits, use a whole-token timeout, use the correct select() nfds value, and return an error instead of aborting on oversized outbound tokens.

Update imgssapi and omgssapi to use the bounded recv_token() interface. Keep the imgssapi listener and receive-path cleanup, including the size_t token-length check and the tcpsrv wiring for existing settings.

Add focused tests for listener port-file and session-limit handling, plus YAML include coverage for the legacy imgssapi configuration path and a unit test for the shared token helpers.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -537,17 +537,43 @@ solaris) GSSLIB=gss ;;
 *)       GSSLIB=gssapi_krb5 ;;
 esac
 
+gssapi_found=no
+GSS_CFLAGS=""
+GSS_LIBS=""
 if test $enable_gssapi_krb5 = yes; then
-	AC_CHECK_LIB($GSSLIB, gss_acquire_cred, [
-		AC_CHECK_HEADER(gssapi/gssapi.h, [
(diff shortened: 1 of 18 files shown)


rgerhards:
ubuntu pkg build fail is a known env issue - no merge blocker

PR #6739, opened 2026-04-20
queue: delete orphaned disk queue write files

### LINKED ISSUE #5085: SEGV on startup with truncated files in spool
This is forwarded from https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1031360 and reproduced with 8.2212.0-1

The simplified config files and spool files can be found in the attached zip
[rsyslog_reproducer.zip](https://github.com/rsyslog/rsyslog/files/10758052/rsyslog_reproducer.zip)

Full output of running `rsyslogd -d -n` : 
[out.txt](https://github.com/rsyslog/rsyslog/files/10758069/out.txt)

Backtrace attached as well 
[gdb.txt](https://github.com/rsyslog/rsyslog/files/10758072/gdb.txt)

rgerhards:
## Summary

This fixes disk queue teardown after corruption recovery when the queue has
already become logically empty.

## Root Cause

After restart recovery from a corrupted persisted disk queue, rsyslog can emit
late internal messages during shutdown. Those messages land in the active write
stream even though the queue is already logically empty and the `.qi` file has
been removed. The old destructor only deleted the write file when its offset was
zero, which left an orphaned `mainq.*` file behind.

## What Changed

- delete the active disk queue write file on close when the queue is logically
  empty
- add a reproducer that corrupts bytes inside a persisted disk queue segment and
  verifies restart recovery does not leave live queue files behind

## Impact

Disk queue recovery after corrupted persisted state no longer leaves orphaned
queue files behind at shutdown.

## Validation

- `./tests/diskqueue-truncated-segment-startup.sh`
- `./tests/diskqueue-oncorruption-missing-segment.sh`
- `./tests/daqueue-dirty-shutdown-recovery.sh`
- `./tests/daqueue-invld-qi.sh`

Closes https://github.com/rsyslog/rsyslog/issues/5085

### DIFF
--- a/runtime/msg.c
+++ b/runtime/msg.c
@@ -1122,7 +1122,7 @@ finalize_it:
 static void reinitVar(var_t *pVar) {
     rsCStrDestruct(&pVar->pcsName); /* no longer needed */
     if (pVar->varType == VARTYPE_STR) {
-        if (pVar->val.pStr != NULL) rsCStrDestruct(&pVar->val.pStr);
+        rsCStrDestruct(&pVar->val.pStr);
     }
     memset(&pVar->val, 0, sizeof(pVar->val));
     pVar->varType = VARTYPE_NONE;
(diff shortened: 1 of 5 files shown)


rgerhards:
@cubic-dev-ai review

PR #6744, opened 2026-04-21
fmhash: guard FNV/DJB with USE_HASH_XXHASH

midnya:
Adjust build-time selection of hash implementations to improve consistency across configurations and reduce maintenance overhead.

Impact: behavior may vary by build flag; no test updates included.

Before: FNV 32-bit hash implementation was always compiled in.
After: FNV code is excluded when USE_HASH_XXHASH is defined.

Wrap the FNV 32-bit implementation in a conditional so it is only compiled when xxhash is not selected. This aligns the source with the intended single-hash backend model and avoids compiling unused code paths. The public hash32() entry point remains, with behavior determined by the selected backend at compile time.

Same reasoning applies to DJB 64-bit hash function.

No changes to interfaces; selection is controlled via USE_HASH_XXHASH. This reduces code footprint and clarifies which implementation is active in a given build.

Refs: no issue

### References

### Notes (optional)

---

#### Quick check (optional)

---

  ```
  git commit --amend
  ```

### DIFF
--- a/contrib/fmhash/fmhash.c
+++ b/contrib/fmhash/fmhash.c
@@ -77,9 +77,9 @@ struct hash_context_s {
 
 #ifndef USE_HASH_XXHASH
     /*
-    * Fowler–Noll–Vo hash 32 bit
-    * http://www.isthe.com/chongo/src/fnv/hash_32.c
-    */
+     * Fowler–Noll–Vo hash 32 bit
+     * http://www.isthe.com/chongo/src/fnv/hash_32.c
+     */
     #if defined(__clang__)
         #pragma GCC diagnostic ignored "-Wunknown-attributes"
     #endif
@@ -108,9 +108,9 @@ static hash_t
 
 
     /*
-    * Modified Bernstein
-    * http://www.eternallyconfuzzled.com/tuts/algorithms/jsw_tut_hashing.aspx
-    */
+     * Modified Bernstein
+     * http://www.eternallyconfuzzled.com/tuts/algorithms/jsw_tut_hashing.aspx
+     */
     #if defined(__clang__)
         #pragma GCC diagnostic ignored "-Wunknown-attributes"
     #endif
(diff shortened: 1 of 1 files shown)


midnya:
Followed the AI boilerplate but basically: compiling with --enable-fmhash and --enable-fmhash-xxhash with `-Werror=unused-function` fails; I see no reason why those two functions should be defined in this case.

rgerhards:
> Followed the AI boilerplate but basically: compiling with --enable-fmhash and --enable-fmhash-xxhash with `-Werror=unused-function` fails; I see no reason why those two functions should be defined in this case.

boilerplate is usually good, but for smaller changes often overdone ;)

There is an issue with code formatting. If you like, I can also fix it, otherwise follow instruction in failed CI. Let me know your preferrence.

And- thx a lot for the PR!

midnya:
> boilerplate is usually good, but for smaller changes often overdone ;)

Yeah as per instructions, I just pasted what your ChatGPT wrapper gave me.

> There is an issue with code formatting

Should be fixed; pushed it as a --fixup to keep the two other CI failures in convo, as they look both interesting and entirely unrelated to my commit.

> And- thx a lot for the PR!

No problem.

Btw, since opening the PR, I identified that without it, with those flags, compilation reliably fails if `autoconf-archive` is installed on a Debian system, and reliably succeeds if it is not. Don't know if you want to add this to your CI build suite, but you have the info.

rgerhards (approved):
lgtm

rgerhards:
@midnya thx again for the PR! you will notice that I changed the boilerplate to be more up to the point in the merge ;-)

rgerhards:
@midnya Completely **unrelated** side note:

I recently noticed that there was already a third-party rsyslog-related entry on G2 which IMHO "impersonated" like the real rsyslog, so I set up the official rsyslog listing there as well. G2 is a software review platform many people use to look up and compare tools, including open-source ones.

If would feel comfortable leaving an honest review there, that would be very helpful. No pressure at all - I just wanted to mention it.

https://www.g2.com/products/rsyslog/reviews

PR #6755, opened 2026-04-23
[codex] tests: harden diskqueue corruption recovery test

rgerhards:
## What changed

This hardens `tests/diskqueue-truncated-segment-startup.sh` and the
common spool-directory setup path used by restart-based tests.

## Why it changed

The truncated-segment recovery test was flaky in CI because it spent a
large fixed amount of time building the persisted queue and then used a
recovery flow that depended on the remaining global harness runtime.
It also emitted a misleading `mkdir ... File exists` message during the
same-test restart even though reusing the spool directory is correct for
this scenario.

## Impact

The test now prepares its persisted queue faster, retries preparation if
it fails to create enough disk queue segments, waits for explicit
recovery outcomes, and reuses the same spool directory without false
errors.

## Root cause

The original test combined `15000` messages with a fixed `20ms`
`omtesting:sleep` delay and then called `shutdown_when_empty` on the
recovery leg. That pushed the test close to the harness-wide timeout on
slower CI jobs. Separately, `generate_conf` unconditionally called
`mkdir $RSYSLOG_DYNNAME.spool`, which produced a false error-looking
message on the intentional phase-2 restart.

## Checks

- `bash -n tests/diskqueue-truncated-segment-startup.sh`
- `./tests/diskqueue-truncated-segment-startup.sh`

### DIFF
--- a/tests/diag.sh
+++ b/tests/diag.sh
@@ -382,7 +382,7 @@ generate_conf() {
 		export RSYSLOG_OUT_LOG="${RSYSLOG_DYNNAME}.out.log"
 		export RSYSLOG2_OUT_LOG="${RSYSLOG_DYNNAME}_2.out.log"
 		export RSYSLOG_PIDBASE="${RSYSLOG_DYNNAME}:" # also used by instance 2!
-		mkdir $RSYSLOG_DYNNAME.spool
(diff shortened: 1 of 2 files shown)


rgerhards:
closing in favor of https://github.com/rsyslog/rsyslog/pull/6759 which also introduces a fix

PR #6759, opened 2026-04-24
queue: quarantine runtime disk corruption safely

rgerhards:
## What changed

This teaches the disk queue runtime path to handle in-file corruption in
`queue.onCorruption="safe"` without weakening the existing startup
checks.

When dequeue-time corruption is hit, rsyslog now:
- preserves already dequeued valid records in the current batch
- quarantines the unread tail into `mainq.bad.*` when bounded resync does
  not recover safely
- reconstructs a fresh active disk queue state
- emits explicit user-visible runtime corruption and quarantine messages

The targeted regression test now requires the strict contract that was
requested during analysis:
- valid output before the corrupted point must still be processed
- quarantine must happen
- zero live `mainq*` files may remain afterwards

## Root cause

Startup validation only checks queue topology and pointer consistency. It
cannot cheaply scan every record payload in every queue segment.

The failing s390x case corrupted bytes inside a queue segment, so startup
still accepted the queue. The old runtime path then surfaced that damage
through generic skipped-message handling, which could report it like an
`invalid .qi file` style loss and did not provide a dedicated runtime
corruption quarantine flow with a clear zero-live-segments end state.

This change closes that gap at runtime instead of weakening startup
behavior.

## User impact

Users now get explicit rsyslog messages when runtime disk queue
corruption is detected, when unread tail data is quarantined, and how
many messages were skipped during runtime corruption handling.

## Validation

Local s390x QEMU/container reruns:
- `./tests/diskqueue-truncated-segment-startup.sh` passed 3 consecutive
  times
- `./tests/diskqueue-oncorruption-missing-segment.sh` passed once to
  confirm the startup corruption path was not weakened

Artifact root:
- `/home/rger/proj/rsyslog-s390x-artifacts/other-bin-runtime-20260424T112114Z`

### DIFF
--- a/doc/ai/module_map.yaml
+++ b/doc/ai/module_map.yaml
@@ -2,6 +2,12 @@
 # Per-module metadata (support status, maturity, contacts) lives in each module
 # directory as `MODULE_METADATA.yaml`; update both the metadata file and this map
 # when concurrency guidance changes.
+runtime_queue:
+  paths: ["runtime/queue.c", "runtime/queue.h"]
+  requires_serialization: true
+  notes:
+    - Disk-assisted mode is an in-memory parent plus a first-class disk child; disk queue semantics apply to the child.
+
 omfile:
   paths: ["plugins/omfile/"]
   requires_serialization: true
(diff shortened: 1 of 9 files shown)


rgerhards:
@cubic-dev-ai review

PR #6761, opened 2026-04-24
tests: stabilize uxsock receiver reset

rgerhards:
## Summary

This stabilizes the uxsock reset tests by synchronizing the scripts with the helper receiver after startup and after `SIGHUP` resets, and by ensuring `SIGHUP` is sent to the actual receiver process.

PR 6756 CI showed isolated failures in `uxsock_simple.sh` and `uxsock_multiple.sh` on slower CI jobs. PR 6761 then exposed the same race in `uxsock_multiple-vg.sh` and `uxsock_multiple.sh` while validating the first fix, which confirmed the reset path was still not deterministic enough under slow CI.

## Root Cause

`uxsock_simple.sh` and `uxsock_multiple.sh` sent `SIGHUP` to `uxsockrcvr` and immediately injected the next message batch. The helper handles reset asynchronously in its poll loop, so on slower or loaded CI systems the next batch could be sent while the receiver was still closing and rebinding its socket.

The first version of this PR also left one hidden problem in place: the scripts launched the receiver as `timeout 30s ./uxsockrcvr ... &`. In that form `$!` is the `timeout` wrapper PID, not reliably the `uxsockrcvr` PID. The test then sent `SIGHUP` to the wrapper, so the receiver did not always reset and write ready generation 2. This was the direct cause of the PR 6761 CI failures that reported missing ready generation 2.

For connected `STREAM` and `SEQPACKET` sockets in `uxsock_multiple.sh`, listener readiness alone was not enough. `omuxsock` may still need a reconnect-triggering send after the receiver reset. If the full post-reset sequence was injected immediately, extra sequence values could be lost during the reconnect window.

## What Changed

- Added a `-r readyfile` option to `uxsockrcvr`.
- `uxsockrcvr` writes an incrementing ready generation after each bind/listen cycle.
- `uxsock_simple.sh` waits for receiver readiness before startup and after reset.
- `uxsock_multiple.sh` waits for all receiver generations and sends the expected connected-socket reconnect trigger separately before the remaining sequence.
- The uxsock scripts start `uxsockrcvr` directly and rely on its existing `-t 60` timeout, so `$!` is the real helper PID that receives `SIGHUP`.
- `uxsock_multiple-vg.sh` resolves its base script through `srcdir` or its own directory so direct and VPATH-style invocations both work.

## Validation

- `make -j$(nproc) check TESTS=""` passed.
- `./tests/uxsock_simple.sh` passed.
- `./tests/uxsock_multiple.sh` passed.
- `./tests/uxsock_multiple-vg.sh` passed with valgrind `ERROR SUMMARY: 0`.
- Earlier worker validation: 20 focused iterations of `uxsock_simple.sh` plus `uxsock_multiple.sh` passed, followed by 5 additional post-format/post-build iterations.
- AI review portability comments addressed: helper wait loops now avoid brace expansion/local variables, and uxsock_multiple.sh uses plain shell lists instead of Bash arrays.
- `git diff --check` passed.

### DIFF
--- a/tests/uxsock_multiple-vg.sh
+++ b/tests/uxsock_multiple-vg.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 export USE_VALGRIND="YES"
-. ${srcdir:-.}/uxsock_multiple.sh
+. "${srcdir:-$(dirname "$0")}/uxsock_multiple.sh"
--- a/tests/uxsock_multiple.sh
+++ b/tests/uxsock_multiple.sh
@@ -6,27 +6,41 @@
 # Based on uxsock_simple.sh added 2010-08-06 by Rgerhards
 # Updated 2025-04-16 for abstract sockets
 . ${srcdir:=.}/diag.sh init
-check_command_available timeout
-
 uname
 if [ $(uname) != "Linux" ] ; then
     echo "This test requires Linux (AF_UNIX abstract addresses)"
     exit 77
 fi
 
-SOCKET_NAMES=(
-    "$RSYSLOG_DYNNAME-testbench-dgram-uxsock.0"
-    "$RSYSLOG_DYNNAME-testbench-dgram-uxsock.DGRAM"
-    "$RSYSLOG_DYNNAME-testbench-dgram-uxsock.STREAM"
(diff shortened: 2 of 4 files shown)


rgerhards:
@cubic-dev-ai review

PR #6762, opened 2026-04-25
testbench: wait for abstract uxsock receiver reset

rgerhards:
## What changed

`uxsock_simple_abstract.sh` now starts `uxsockrcvr` directly and waits for the
receiver readiness file both at initial startup and after the SIGHUP receiver
reset.

## Root cause

The abstract-socket variant still used a `timeout` wrapper around `uxsockrcvr`.
The test reset path killed the wrapper process and then immediately continued
with the next message batch. Under high scheduler pressure, the actual receiver
child could still be exiting or the replacement receiver had not rebound the
abstract socket yet. That made the sender race the receiver reset and could
drop part of the second batch, producing sequence gaps such as 19723/20000.

The plain uxsock test already used a receiver-ready generation file after
restart. The abstract-socket test lacked that synchronization.

## Validation

- Direct worker validation for `uxsock_simple_abstract.sh` passed.
- Parent `make clean && make check -j80` validation did not fail
  `uxsock_simple_abstract.sh`; unrelated full-suite flakes were recorded in the
  campaign artifacts and are not acceptance criteria for this branch.
- Later parent validations for other branches passed both
  `uxsock_simple_abstract.sh` and `uxsock_simple_abstract-vg.sh`.

## Campaign note

The unrelated failures observed during full-suite stress runs remain tracked by
the flake campaign and are intentionally not bundled into this PR.

### DIFF
--- a/tests/uxsock_simple_abstract.sh
+++ b/tests/uxsock_simple_abstract.sh
@@ -5,13 +5,27 @@
 # added 2010-08-06 by Rgerhards
 # Updated 2025-04-16 for abstract sockets
 . ${srcdir:=.}/diag.sh init
-check_command_available timeout
-
 uname
 if [ $(uname) != "Linux" ] ; then
     echo "This test requires Linux (AF_UNIX abstract addresses)"
     exit 77
 fi
+
+wait_receiver_ready() {
+    expected_generation=$1
+    ready_file=$2
+    count=1
+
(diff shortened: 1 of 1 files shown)


rgerhards:
Addressed the AI review timeout-portability comment: `wait_receiver_ready` now uses the testbench `TB_TIMEOUT_STARTSTOP` limit instead of a hardcoded 10-second loop. Focused local validation after the change: `./tests/uxsock_simple_abstract.sh` passed.

PR #6763, opened 2026-04-25
imfile test: stabilize async logrotate drain

rgerhards:
## What changed

`imfile-logrotate-async.sh` now synchronizes the asynchronous logrotate test on
observable imfile output progress instead of assuming every rotated-file line is
available before the final file tail has been closed and processed.

The test still requires the final exact 10000-line result.

## Root cause

The test rotated an input file while `inputfilegen` was still writing and then
used intermediate assertions that could count lines in a rotated input file
before imfile had emitted all of those lines. Under high load, one record can
remain pending until a later close/modify notification. The late HUP cycle could
also rotate after generated input had entered the final tail range, creating a
window where the sentinel append did not reliably force imfile to revisit the
right file before the fixed timeout.

The evidence points to test synchronization around asynchronous generation and
rotation, not to a proven production imfile data-loss bug. Local failures
stalled near the final line count, for example 9811/10000 and 9983/10000, while
the generated input files still accounted for all 10000 messages.

## Validation

- Worker direct validation passed repeatedly for `imfile-logrotate-async.sh`.
- Worker stress loops passed under CPU load.
- Parent `make clean && make check -j80` validation on
  `parent-j80-imfile-20260425-120443` passed `imfile-logrotate-async.sh`.
  Remaining failures in that run were unrelated program, imfile-endregex, and
  omfwd candidates recorded in the campaign artifacts.

## Campaign note

The unrelated full-suite flakes remain tracked separately and are not acceptance
criteria for this branch.

### DIFF
--- a/tests/imfile-logrotate-async.sh
+++ b/tests/imfile-logrotate-async.sh
@@ -5,6 +5,7 @@
 check_command_available logrotate
 export NUMMESSAGES=10000
 export RETRIES=50
+export ROTATE_STOP_VISIBLE_LINES=$((NUMMESSAGES - 1000))
 
 # Uncomment fdor debuglogs
 #export RSYSLOG_DEBUG="debug nostdout noprintmutexaction"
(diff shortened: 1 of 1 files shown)


rgerhards:
@cubic-dev-ai review

rgerhards:
Addressed the AI review portability comment: inode reads now use `ls -i | awk` instead of GNU-only `stat -c %i`. Focused local validation after the change: `./tests/imfile-logrotate-async.sh` passed.

PR #6764, opened 2026-04-25
tests: synchronize prog lifecycle flakes

rgerhards:
## What changed

This stabilizes the program lifecycle tests that rely on external helper
processes:

- improg tests now wait for concrete helper output before calling
  `shutdown_when_empty`
- the multiline improg test uses the same output-before-shutdown synchronization
- `omprog-feedback-timeout.sh` now waits for the timeout retry to begin before
  shutdown
- the omprog timeout helper exits shortly after the intentional timeout instead
  of depending on killing a long-sleeping child
- the omprog timeout action uses explicit resume scheduling for this test path

## Root cause

Under high parallel load, queue-empty was not a sufficient synchronization point
for these tests. The main queue could be empty while an external helper had not
yet produced the output asserted later by the test. That allowed shutdown to
cancel or kill the helper before the expected transcript was complete.

`omprog-feedback-timeout.sh` had an additional timing problem. It intended to
prove that a message is retried after `confirmTimeout`, but the test still
depended on default action resume timing. Under full `-j80` load the action
could remain suspended long enough that the restarted helper never received the
retried `msgnum:00000004` within the test wait window.

## Validation

- Worker validation:
  - `./omprog-feedback-timeout.sh` direct repeated runs passed
  - 80-way parallel `omprog-feedback-timeout.sh` stress passed
  - owned direct program test set passed
- Coordinator validation:
  - `make clean && make check -j80` on
    `parent-j80-prog-20260425-125543` passed all owned program tests,
    including `omprog-feedback-timeout.sh`
  - the only remaining failure in that run was unrelated
    `imfile-logrotate-async.sh`, which is tracked separately in PR 6763

## Negative-control proof

A separate proof worktree temporarily changed the omprog timeout helper's retry
ACK transcript from `<= OK` to `<= BAD-ACK`. The test failed at the final
`cmp_exact` check with:

```text
< <= OK
---
> <= BAD-ACK
FAIL: Test ./tests/omprog-feedback-timeout.sh
```

The temporary mutation was fully undone before finishing, and the proof worktree
was left clean. This confirms the test still catches invalid retry behavior and
was not weakened into an always-pass test.

## Campaign note

Unrelated full-suite flakes found during validation are recorded in the flake
campaign memory but are not bundled into this PR.

### DIFF
--- a/tests/improg_prog_confirm.sh
+++ b/tests/improg_prog_confirm.sh
@@ -13,6 +13,7 @@ ruleset(name="ruleset") {
 }
 '
 startup
+wait_content "program data" "$RSYSLOG_OUT_LOG"
 shutdown_when_empty
 wait_shutdown
 content_check "program data"
--- a/tests/improg_prog_confirm_killonclose.sh
+++ b/tests/improg_prog_confirm_killonclose.sh
@@ -13,6 +13,7 @@ ruleset(name="ruleset") {
 }
 '
 startup
+wait_content "program data" "$RSYSLOG_OUT_LOG"
 shutdown_when_empty
 wait_shutdown
 content_check "program data"
--- a/tests/improg_prog_killonclose.sh
+++ b/tests/improg_prog_killonclose.sh
@@ -13,6 +13,7 @@ ruleset(name="ruleset") {
 }
 '
 startup
+wait_content "program data" "$RSYSLOG_OUT_LOG"
 shutdown_when_empty
 wait_shutdown
 content_check "program data"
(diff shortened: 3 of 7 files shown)


rgerhards:
@cubic-dev-ai review

rgerhards:
Addressed the AI review simplification comment: removed the redundant `wait_content` before `content_check_with_count`, since `content_check_with_count` already waits for the two retry transcript entries. Focused local validation after the change: `./tests/omprog-feedback-timeout.sh` passed.

PR #6765, opened 2026-04-25
imtcp test: avoid early diagnostic capture race

rgerhards:
## Failure Mode

`imtcp_framing_regex_maxmsg_overflow.sh` intermittently failed even though rsyslogd emitted the expected protection diagnostic.

The recurring signature was:

- stderr showed `framing.delimiter.regex requires maxMessageSize <= 1073741823`
- `${RSYSLOG_DYNNAME}.started` was missing or had not captured that early message
- `content_check` failed against `.started`

This was seen repeatedly in unrelated CI contexts, including Buildbot 234 build 4809. That build tested PR 6793, which only changed docs/Sphinx config and an impstats analyzer helper, but still hit the same imtcp test failure.

## Root Cause

The test asserted an early rsyslogd startup diagnostic through the normal testbench `.started` action output. That file is produced asynchronously by the configured rsyslog action path. The diagnostic under test is emitted by rsyslogd itself during startup, before that action capture path is a reliable assertion target.

So the product validation fired correctly, but the test sometimes looked in the wrong capture channel.

## Fix

Set `RS_REDIR` before startup and assert the diagnostic in a dedicated per-test rsyslogd stdout/stderr log instead of `${RSYSLOG_DYNNAME}.started`.

Skip Valgrind mode explicitly because `startup_vg` does not honor this `RS_REDIR` capture path. That avoids a misleading Valgrind-only failure while preserving normal-mode regression coverage.

## Proof

Negative-control proof: temporarily changed the expected string to include `BROKEN`. The test failed at `content_check`, and the captured process log still contained the real diagnostic. The temporary break was undone before final validation.

## Validation

```sh
./tests/imtcp_framing_regex_maxmsg_overflow.sh
for i in $(seq 1 10); do ./tests/imtcp_framing_regex_maxmsg_overflow.sh || exit $?; done
USE_VALGRIND=YES ./tests/imtcp_framing_regex_maxmsg_overflow.sh; rc=$?; test $rc -eq 77
make -j24 check TESTS=imtcp_framing_regex_maxmsg_overflow.sh
```

All passed. The focused Automake run reported `TOTAL: 1`, `PASS: 1`.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,16 @@
 --------------------------------------------------------------------------------------
 Scheduled Release 8.2606.0 (aka 2026.06) 2026-06-??
+- 2026-05-06: omrelp: fix OpenSSL TLS auth failure CPU spin
+  omrelp with OpenSSL TLS and an action queue could consume a full CPU
+  after TLS authentication failed, for example when the action omitted
+  tls.caCert. The action was disabled, but queued messages from the
+  already-dequeued batch were retried immediately against the same
+  permanently disabled action.
+  omrelp now classifies librelp TLS authentication return codes as
+  permanent auth failures, and the action queue commits affected batch
+  entries from the queue's perspective when an action is disabled.
+  Transient suspend/retry behavior is unchanged.
+  Fixes https://github.com/rsyslog/rsyslog/issues/6612
 - 2026-05-06: ossl drvier: match PermittedPeer wildcards against cert identities
   OpenSSL x509/name authorization only applied rsyslog wildcard matching to
   the full OpenSSL subject string and then fell back to X509_check_host() for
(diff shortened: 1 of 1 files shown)


rgerhards:
Addressed the Valgrind-mode concern in e285ac89f84b423ff046c65753e8d0a7c0d40f98. The test now explicitly skips `USE_VALGRIND=YES` and `USE_VALGRIND=YES-NOLEAK`, because `startup_vg` does not honor `RS_REDIR`. Normal mode still captures rsyslogd stderr/stdout directly and checks the diagnostic. Validation included direct pass, focused Automake pass, and a `USE_VALGRIND=YES` skip check returning 77.

Issue #6768, opened 2026-04-28
[General]: 7–9 digit sub-second precision corrupts RFC 3339 output with garbage bytes


blemoallic:
### 📝 Description:
It sems that when rsyslog receives an RFC 5424 syslog message whose TIMESTAMP field contains 7, 8, or 9 sub-second digits (nanosecond precision), the RFC 3339 formatter (probably `formatTimestamp3339` / `formatTimestampSecFrac` in `runtime/datetime.c`) produces a corrupted fractional-seconds field, the first output character is a garbage byte caused by a modular-arithmetic wrap and integer-to-char overflow.

I think it's a bug, but not sure if I'm using this properly.

### 💻 Environment:

* **Rsyslog Version:** `8.2504.0 (package rsyslog 8.2504.0-1 on Debian stable-slim)` 
* **Platform:** `Debian`
* **Deployment Method:** `Docker`

### ➕ Additional Context:

Input message (TCP syslog, RFC 5424):
```
<171>1 2026-04-27T15:46:41.000006930Z myhostname myapp - - - my message body
```
The timestamp field `2026-04-27T15:46:41.000006930Z` has 9 sub-second digits.

rsyslog output (`timereported` property, `dateFormat="rfc3339"`):
```
2026-04-27T15:46:41.u30Z
```

#### More examples

| Input secfrac | Digits | Expected output | Actual output |
|---|---|---|---|
| `.0300000` | 7 | `.030000` | `.\x10Z` (0x10 = DLE control char) |
| `.03000000` | 8 | `.030000` | `.\x100Z` |
| `.000006930` | 9 | `.000006` | `.u30Z` (0x75 = `u`) |
| `.000022054` | 9 | `.000022` | `.\x0c54Z` (0x0C = FF control char) |

---
**Thank you for your contribution!**

PR #6770, opened 2026-04-28
Fix RFC3339 formatting for high-precision fractional seconds

alorbach:
Summary:
- Fixes issue #6768 by normalizing parsed fractional seconds to the six-digit precision supported by rsyslog’s RFC3339 formatter.
- Removes the old modulo-based divisor selection that could emit corrupted fractional output for 7-9 digit secfrac values.
- Refactors the shared secfrac normalization/divisor logic into a helper used by both `formatTimestampSecFrac()` and `formatTimestamp3339()`.
- Adds regression coverage for the reported 7, 8, and 9 digit timestamp cases.

### DIFF
--- a/runtime/datetime.c
+++ b/runtime/datetime.c
@@ -831,6 +831,24 @@ static int formatTimestampToPgSQL(struct syslogTime *ts, char *pBuf) {
 }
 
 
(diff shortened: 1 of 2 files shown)


alorbach:
@codex review

PR #6771, opened 2026-04-29
packaging/rpm: Add omotel OpenTelemetry Sub-package

AppleTechy:
Provides RPM package of omotel module. This reduces the friction for adoption of the plugin while in preview.

### References
N/A

### Notes (optional)
Packaging-only change

### DIFF
--- a/packaging/rpm/rpmbuild/SPECS/rsyslog-v8-stable.spec
+++ b/packaging/rpm/rpmbuild/SPECS/rsyslog-v8-stable.spec
@@ -300,6 +300,12 @@ Group: System Environment/Daemons
 Requires: %name = %version-%release
 BuildRequires: libcurl-devel
 
+%package omotel
+Summary: omotel output module for rsyslog
+Group: System Environment/Daemons
+Requires: %name = %version-%release
+BuildRequires: libcurl-devel
+
 %package mmdblookup
 Summary: mmdblookup MaxMind/GeoIP DB lookup for rsyslog
 Group: System Environment/Daemons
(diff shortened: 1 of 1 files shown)


alorbach (approved):
Thanks for the contrib

AppleTechy:
Hi @alorbach @rgerhards,  I was expecting with the above change that the rsyslog-omotel rpm would be built as part of the v8-stable-daily build. However, I am not seeing it. Would it be possible to get it built as part of the daily? Thanks!

alorbach:
@AppleTechy We are currently transitioning the full package generation from our dedicated packaging repositories:

- https://github.com/rsyslog/rsyslog-pkg-rhel-centos
- https://github.com/rsyslog/rsyslog-pkg-ubuntu

into the main rsyslog repository.

However, this is a complex migration, so for now we need to maintain both setups in parallel: the main rsyslog repository and the existing dedicated packaging repositories.

I will backport the omotel package changes top the dedicated packaging repositories

rgerhards:
@alorbach yeah, I think we should do it as quickly as possible. I was unfortunately not aware it's currently not packaged. scheduled stable should also get a rebuild.

alorbach:
Merged omotel changes into the other projects, next daily package build should build them

AppleTechy:
Thank you!  I was unaware of the dedicated package repos still being in use and that I needed to open a PR there as well to get it built.  🙃

alorbach:
> Thank you! I was unaware of the dedicated package repos still being in use and that I needed to open a PR there as well to get it built. 🙃

Don't worry, those old repos will become obsolete very soon (and archived once we are done). Your path was the right one ;).

Issue #6612, opened 2026-03-05
[General]: omrelp+tls still 100% CPU usage after tls auth failed


kbcz1989:
### 📝 Description / What's on your mind:
Rsyslog is still using all the CPU when relp tls fails due to the wrong certificate

### 💻 Environment (if applicable):
* **Rsyslog Version:8.2602.0``
* **Platform:Linux RHEL ``
* **Deployment Method:RPM ``

### ➕ Additional Context:
https://github.com/rsyslog/rsyslog/issues/5423

This issue can be reliably reproduced using minimal config for client and server.

Generate cert to use for server:
```bash
openssl genrsa -out ca.key 2048
openssl req -x509 -new -nodes -key ca.key -sha256 -days 3650 -out ca.crt -subj "/CN=Test CA"
openssl genrsa -out server.key 2048
openssl req -new -key server.key -out server.csr -subj "/CN=localhost"
openssl x509 -req -in server.csr -CA ca.crt -CAkey ca.key -CAcreateserial -out server.crt -days 3650 -sha256
```

Create minimal server config file rsyslog-server.conf (replace or set $TEST_DIR):
```bash
module(load="imrelp" tls.tlslib="openssl")

input(type="imrelp"
      name="5514_imrelp"
      port="5514"
      tls="on"
      tls.cacert="$TEST_DIR/cert/ca.crt"
      tls.mycert="$TEST_DIR/cert/server.crt"
      tls.myprivkey="$TEST_DIR/cert/server.key"
)

*.* $TEST_DIR/test_server.log
```

Create minimal client config file rsyslog-client.conf (replace or set $TEST_DIR):
```bash
module(load="imfile")
module(load="omrelp" tls.tlslib="openssl")

global (
    workDirectory="$TEST_DIR/state"
)

input(type="imfile"
  File="$TEST_DIR/log/test.log"
  Tag="TEST"
)

action(type="omrelp"
    name="test_action"
    Target="localhost"
    Port="5514"
    queue.type="LinkedList"
    tls="on"
)
```

Run server and run client in debug mode (replace or set $TEST_DIR):
```bash
rsyslogd -f "$TEST_DIR/rsyslog-server.conf" -i "$TEST_DIR/rsyslog-server.pid"
RSYSLOG_DEBUGLOG="$TEST_DIR/log/rsyslog.debug" rsyslogd -d -f "$TEST_DIR/rsyslog-client.conf" -i "$TEST_DIR/rsyslog-client.pid"
[...]

PR #6772, opened 2026-04-30
Fix omrelp TLS auth failure CPU spin

alorbach:
## Summary

Fixes issue #6612: `omrelp` with OpenSSL TLS could spin at 100% CPU after TLS authentication failed, especially with an action queue such as `queue.type="LinkedList"`.

The failure mode is not a crash. The action reaches an authentication failure, becomes disabled, and then the queued action worker can keep retrying the same already-dequeued messages against an action that can never recover.

## Changes

- Add `tests/omrelp-tls-ossl-authfail-no-cacert.sh`.
- Register the new test under RELP + OpenSSL coverage in `tests/Makefile.am`.
- Update `omrelp` to classify librelp TLS authentication return codes as permanent auth failures.
- Update action queue handling so `RS_RET_DISABLE_ACTION` commits the affected queued batch entries from the queue’s perspective instead of reprocessing them forever.
- Make the action commit retry loop honor immediate shutdown before starting another retry attempt.

## Why `action.c` is part of the fix

The `omrelp` side can correctly return `RS_RET_DISABLE_ACTION`, but the CPU spin happened in the action queue layer.

For queued actions, `processBatchMain()` receives a batch from the action queue. Before this change, if processing returned `RS_RET_DISABLE_ACTION`, the action was disabled but the dequeued batch entries were not marked committed. That allowed the queue to retry the same messages immediately, which just hit the same disabled action again. With no network wait or retry sleep left, this became a tight CPU loop.

The new queue behavior is limited to permanent action disable. Transient failures still use the normal `RS_RET_SUSPENDED` retry path.

## Test Coverage

The new regression test:

- starts an OpenSSL TLS `imrelp` receiver using existing `tests/tls-certs`
- starts an OpenSSL TLS `omrelp` sender with `queue.type="LinkedList"`
- intentionally omits client `tls.caCert`
- injects messages into the sender
- checks that the auth failure disables the action
- samples `/proc/<pid>/stat` CPU ticks after auth failure and after shutdown starts
- fails if the sender consumes CPU like a tight loop

## Validation

Passed:

- `./tests/omrelp-tls-ossl-authfail-no-cacert.sh`
- `./tests/sndrcv_relp_tls_chainedcert.sh`
- `./tests/omrelp_errmsg_no_connect.sh`

### DIFF
--- a/action.c
+++ b/action.c
@@ -1678,6 +1678,9 @@ static rsRetVal ATTR_NONNULL() actionCommit(action_t *__restrict__ const pThis, 
     DBGPRINTF("actionCommit[%s]: unhappy, we still have %d uncommitted messages.\n", pThis->pszName, nMsgs);
     int bDone = 0;
     do {
+        if (*pWti->pbShutdownImmediate) {
+            ABORT_FINALIZE(RS_RET_FORCE_TERM);
+        }
         iRet = actionTryCommit(pThis, pWti, iparams, nMsgs);
         DBGPRINTF("actionCommit[%s]: in retry loop, iRet %d\n", pThis->pszName, iRet);
         if (iRet == RS_RET_FORCE_TERM) {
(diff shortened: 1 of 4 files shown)


alorbach:
@codex review

alorbach:
@codex review
/gemini review
/cursor review

rgerhards (approved):
I checked that the patch is valid and, most importantly, disabling the action is the right path. For a running instance, there is little chance that it will auto-recover, as that would require cert changes at the remote peer. So good to go.

PR #6773, opened 2026-04-30
ossl: match PermittedPeer wildcards against cert identities

alorbach:
## Summary

The OpenSSL `x509/name` authorization path did not handle wildcard `PermittedPeer` entries consistently with the other TLS drivers. Exact peer names worked, but entries such as `PermittedPeer=["*.example.com"]` could fail because the OpenSSL path only applied rsyslog wildcard matching to the full OpenSSL subject string, then fell back to `X509_check_host()` for configured peer names.

That fallback is useful for exact host checks, but it does not implement rsyslog's configured wildcard allowlist semantics for matching certificate identities. By contrast, `gtls` and `mbedtls` extract certificate SAN/CN identities and run them through the shared `PermittedPeerWildcardMatch()` helper.

## Changes

- Extract and authorize certificate identities in the OpenSSL `x509/name` path:
  - DNS SAN names.
  - IP SAN names as printable IPv4/IPv6 strings.
  - CN values when CN fallback is allowed.
- Match extracted identities through the shared `PermittedPeerWildcardMatch()` helper.
- Preserve legacy full subject-DN matching for existing configurations using values like `/CN=.../OU=...`.
- Keep `X509_check_host()` as a compatibility fallback for non-wildcard peer entries.
- Preserve `PrioritizeSAN` behavior:
  - If SAN entries exist, CN fallback is suppressed.
  - If no SAN entries exist, CN fallback remains allowed.
- Share SAN parsing across DNS/IP SAN handling to avoid decoding the SAN extension twice.
- Keep the SAN parser available for wolfSSL builds as well, fixing the wolfSSL `PrioritizeSAN` bypass/compile issue.
- Add OpenSSL regression coverage for wildcard allow and deny cases.

## Tests

Validated with:

```text
git diff --check -- runtime/net_ossl.c
./tests/imtcp-tls-ossl-x509name.sh
./tests/imtcp-tls-ossl-x509name-wildcard.sh
./tests/imtcp-tls-ossl-x509name-wildcard-invld.sh
./tests/imtcp-tls-gtls-x509name.sh

### DIFF
--- a/runtime/net_ossl.c
+++ b/runtime/net_ossl.c
@@ -34,6 +34,7 @@
 #include <fcntl.h>
 #include <pthread.h>
 #include <netdb.h>
+#include <arpa/inet.h>
 #include <sys/select.h>
 #include <sys/time.h>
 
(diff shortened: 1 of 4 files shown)


alorbach:
@codex review
@cursor review

PR #6795, opened 2026-05-04
improg: stop on child stdout EOF

### LINKED ISSUE #6791: 100% CPU usage on improg after its external program killed
### 📝 Description:
improg spinning in an infinite loop if its binary/external program is killed or exits early.

strace showing 99.98% time in `read`

```
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 99.98    9.314065           9    988664           read
  0.02    0.001742        1742         1           execve
  0.00    0.000413          45         9           write
------ ----------- ----------- --------- --------- ----------------
100.00    9.316220           9    988674           total
```
Identifying the spinning fd

```
[pid  5429] read(9, "", 1)              = 0
[pid  5429] read(9, "", 1)              = 0
[pid  5429] read(9, "", 1)              = 0
...
```
Thread 5429 (`in:improg`) spinning on fd 9, receiving EOF (`= 0`) on every call.

$ sudo lsof 2>/dev/null | grep 78425
rsyslogd  5425  5429 in:improg  root  9r  FIFO  0,11  0t0  78425 pipe

### Steps to reproduce
1. Start rsyslog with improg and specify a binary program prog1
2. SIGKILL prog1
3. Observe the rsyslog is now comsuming 100% CPU by top commands

It looks like the improg code lacks handling for child death or for EOF.

rgerhards:
## Summary
- Treat EOF on the improg child stdout pipe as child termination.
- Close/reap the child on EOF or read errors so the fd is removed from the select loop.
- Add a regression test that kills the improg child and checks rsyslogd CPU ticks to catch the old spin.

## Issue validity
I reproduced the issue before applying the fix with the new regression test. The test failed on the old code with `rsyslogd CPU ticks after improg child kill: 120`, which matches the reported busy loop after the child exits. With the fix, the same test reports `0` CPU ticks in the sampling window.

Closes https://github.com/rsyslog/rsyslog/issues/6791

## Validation
- `./autogen.sh --enable-debug --enable-testbench --enable-improg`
- `./tests/improg_child_killed.sh` (failed before fix: 120 CPU ticks)
- `./tests/improg_child_killed.sh`
- `./tests/improg_prog_simple.sh`
- `./tests/improg_prog_confirm.sh`
- `./tests/improg_prog_killonclose.sh`
- `./tests/improg_prog_confirm_killonclose.sh`
- `./tests/improg_simple_multi.sh`
- `./tests/improg_errmsg_no_params.sh`
- `make distcheck TEST_RUN_TYPE=MOCK-OK -j$(nproc)`
- `git diff --check`

### DIFF
--- a/contrib/improg/improg.c
+++ b/contrib/improg/improg.c
@@ -401,7 +401,7 @@ static rsRetVal readChild(instanceConf_t *const pInst) {
             cstrAppendChar(pInst->ppCStr, c);
         }
     }
-    return (retval == 0) ? RS_RET_OK : RS_RET_IO_ERROR;
+    return (retval == 0) ? RS_RET_EOF : RS_RET_IO_ERROR;
 }
 
 /* create input instance, set default parameters, and
@@ -553,6 +553,7 @@ BEGINrunInput
     struct timeval tv;
     int retval;
     instanceConf_t *pInst;
+    rsRetVal readRet;
     CODESTARTrunInput;
     FD_ZERO(&rfds);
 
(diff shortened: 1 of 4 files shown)


rgerhards:
@cubic-dev-ai review

PR #6802, opened 2026-05-05
imtcp: wait for mbedtls diagnostics in tests

rgerhards:
## Summary

This replaces the temporary PR 6798 process-output capture with the proper test synchronization. The affected mbedtls negative tests now wait until rsyslog internal diagnostics reach the configured omfile output before shutting rsyslogd down.

## Root cause

The diagnostics in these tests are emitted through `LogMsg`/`LogError`, enter the internal rsyslog message path, and are intended to reach working omfile actions in testbench mode. On slow or loaded systems the tests could terminate rsyslogd before that internal diagnostic line was written to `$RSYSLOG_OUT_LOG`. Capturing stderr with `RS_REDIR` avoided the symptom, but it no longer checked the user-visible omfile/internal-message path.

## Changes

- Remove the temporary `RS_REDIR` capture from seven mbedtls negative tests.
- Keep each test's existing tcpflood `NUMMESSAGES` value unchanged.
- Add a local `QUEUE_EMPTY_CHECK_FUNC` that waits for at least one line in `$RSYSLOG_OUT_LOG` via `wait_file_lines`.
- Use `shutdown_when_empty` before `wait_shutdown`.
- Restore `content_check` assertions to the default `$RSYSLOG_OUT_LOG`.

## Validation

- `make -j24 check TESTS=""`
- Direct runs of all seven affected mbedtls negative tests
- `make -j24 check TESTS="imtcp-tls-mbedtls-error-cert.sh imtcp-tls-mbedtls-error-ca.sh imtcp-tls-mbedtls-error-key.sh imtcp-tls-mbedtls-error-key2.sh imtcp-tls-mbedtls-x509invalid.sh imtcp-tls-mbedtls-x509fingerprint-invld.sh imtcp-tls-mbedtls-x509name_invalid.sh"` with `TOTAL: 7`, `PASS: 7`

Negative-control proof: temporarily changed one expected diagnostic string in `imtcp-tls-mbedtls-error-ca.sh`; the test failed while `$RSYSLOG_OUT_LOG` contained the real `nsd mbedtls: error parsing crypto config` diagnostic. The temporary break was reverted before commit.

### DIFF
--- a/tests/imtcp-tls-mbedtls-error-ca.sh
+++ b/tests/imtcp-tls-mbedtls-error-ca.sh
@@ -1,12 +1,12 @@
 #!/bin/bash
 # This file is part of the rsyslog project, released under ASL 2.0
 . ${srcdir:=.}/diag.sh init
-export RS_REDIR=">${RSYSLOG_DYNNAME}.rsyslog.log 2>&1"
 
 # uncomment for debugging support:
 #export RSYSLOG_DEBUG="debug nostdout noprintmutexaction"
 #export RSYSLOG_DEBUGLOG="log"
-export NUMMESSAGES=10
+export NUMMESSAGES=3
+export QUEUE_EMPTY_CHECK_FUNC=wait_file_lines
 generate_conf
 add_conf '
 global(	defaultNetstreamDriverCAFile="'$srcdir/tls-certs/ca-fail.pem'"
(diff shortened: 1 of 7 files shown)


rgerhards:
Thanks for the review. I am leaving the small helper local in these tests for this PR. The shared primitive remains `wait_file_lines`; the wrapper only fixes this mbedtls-negative-test family by waiting for one `$RSYSLOG_OUT_LOG` line before shutdown. Moving that one-line wrapper into `diag.sh` would broaden the shared testbench API for a very narrow use case. If the same lower-bound diagnostic wait pattern appears in more families, extracting a named common helper would make sense as a follow-up.

rgerhards:
Updated the PR after policy review: the branch now uses the existing testbench pattern `QUEUE_EMPTY_CHECK_FUNC=wait_file_lines` directly, with `NUMMESSAGES=3` as the wait target. I measured the affected output files locally at five or six lines; three avoids the startup-only race without pinning the tests to the exact current diagnostic count. Revalidated all seven direct tests, the negative-control proof, and the focused Automake run (`TOTAL: 7`, `PASS: 7`).

rgerhards:
@cubic-dev-ai review

PR #6820, opened 2026-05-06
tests: validate diagtalker port

rgerhards:
## Summary

Adds controlled handling for invalid DiagTalker port arguments.

`DiagTalker.java` now catches invalid integer input, reports the bad port, and
exits before opening a socket. The historical replacement note is also made a
Java comment so the helper can compile for validation.

## Impact

Default port behavior is unchanged. Invalid custom ports now fail with a clear
error instead of an uncaught `NumberFormatException` stack trace.

## Validation

- `git diff --check origin/main..HEAD`
- `javac -d /tmp/rsyslog-javac-diagtalker-maincheck tests/historical/DiagTalker.java`


rgerhards:
Addressed Gemini review feedback: added explicit port range validation for parsed DiagTalker port values before socket creation.

rgerhards:
closed via #6824

PR #6821, opened 2026-05-06
doc: avoid unused python imports

rgerhards:
## Summary

Removes unused Python imports from documentation configuration and lexer code.

Optional Sphinx extension checks now use a small `importlib.util.find_spec()`
helper instead of importing modules only to test availability.

## Impact

No documentation behavior change is intended. Sitemap and Google Analytics
extensions are still enabled only when configured and available; the existing
GA metatag fallback remains in place when the extension is missing.

## Validation

- `python3 -m py_compile doc/source/conf.py doc/source/_ext/rsyslog_lexer.py`
- `git diff --check origin/main..HEAD`

### DIFF
--- a/doc/source/_ext/rsyslog_lexer.py
+++ b/doc/source/_ext/rsyslog_lexer.py
@@ -1,11 +1,8 @@
 # source/_ext/rsyslog_lexer.py
 
-import re
-
 from pygments.lexer import RegexLexer, bygroups
 from pygments.token import (
     Text, Comment, Operator, Keyword, Name, String, Number, Punctuation,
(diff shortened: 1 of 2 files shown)


rgerhards:
Addressed Gemini review feedback: `has_optional_extension()` now catches `ImportError`, matching the previous optional import fallback semantics.

rgerhards:
Addressed cubic review feedback: `has_optional_extension()` now imports the optional extension inside the probe and catches `ImportError`, so broken optional installs fall back cleanly while avoiding unused import bindings.

PR #6822, opened 2026-05-06
tests: stabilize imptcp maxsessions guard

rgerhards:
Root cause:
The original test intended to prove that `imptcp` enforces `maxsessions=10` by opening 20 client sessions and expecting 10 refusal diagnostics. Kernel-level TCP `connect()` success is not the same as imptcp session acceptance: over-limit sockets may connect at the OS level, then be rejected when imptcp accepts them and checks the session table. The old test also used the exact diagnostic count as `QUEUE_EMPTY_CHECK_FUNC`, so a timing miss could hold shutdown until the global testbench abort timeout.

Fix:
Keep the original threshold semantics, but make the accepted session state deterministic with project testbench plumbing. `tcpflood` now has short option `-K seconds`; after normal `openConnections()` and message sending it keeps still-open plain TCP sockets open before the normal close path. `-K0` holds until the test terminates the holder process.

The test now starts a background tcpflood holder for 10 accepted sessions, waits until those 10 messages are written, then runs a second tcpflood probe for 10 extra connections and requires exactly 10 `too many tcp sessions` diagnostics. The exact diagnostic count is no longer used as the queue-empty shutdown predicate.

Additional checks retained/added:
- `MAXSESSIONS=10`, `CONNECTIONS=20`, and `EXPECTED_DROPS=10` remain explicit.
- The output file is checked for exactly the 10 accepted holder messages.
- `seq_check 0 9` verifies the accepted messages are the intended holder sessions.
- Error cleanup kills and waits for the tcpflood holder.
- `tcpflood -K` is restricted to plain TCP, one run, and no random connection dropping; invalid combinations fail before sockets are opened.

Proof that the test still fails for invalid behavior:
- Temporarily setting test config to `maxsessions="20"` failed with `expected exactly 10 dropped sessions, got 0`.
- Temporarily injecting an off-by-one guard in `plugins/imptcp/imptcp.c` by changing `>=` to `>` failed with `expected exactly 10 dropped sessions, got 9`.
- Both temporary negative-control edits were reverted before commit.

Local validation before PR update:
- Worker: `make -C tests -j$(nproc) tcpflood`
- Worker: `./tests/imptcp_maxsessions.sh`, 3 iterations before negative controls
- Worker: `./tests/imptcp_maxsessions.sh`, 2 iterations after negative controls
- Worker: `make check TESTS='imptcp_maxsessions.sh'`
- Worker: `git diff --check`
- Worker: `-K` invalid-combo smoke checks rejected UDP, `-R2`, and `-D`
- Coordinator: signed commit and `git diff --check HEAD~1..HEAD`

### DIFF
--- a/tests/imptcp_maxsessions.sh
+++ b/tests/imptcp_maxsessions.sh
@@ -12,21 +12,77 @@ CONNECTIONS=20
 EXPECTED_DROPS=$((CONNECTIONS - MAXSESSIONS))
 
 EXPECTED_STR='too many tcp sessions - dropping incoming request'
-wait_too_many_sessions()
+STARTED_LOG="${RSYSLOG_DYNNAME}.started"
+HOLDER_PID=
+
+stop_holder()
 {
-  test "$(grep "$EXPECTED_STR" "$RSYSLOG_OUT_LOG" | wc -l)" = "$EXPECTED_DROPS"
+	if [ -n "$HOLDER_PID" ]; then
+		if kill -0 "$HOLDER_PID" 2>/dev/null; then
+			kill "$HOLDER_PID" 2>/dev/null || true
+		fi
+		wait "$HOLDER_PID" 2>/dev/null || true
+		HOLDER_PID=
+	fi
 }
 
-export QUEUE_EMPTY_CHECK_FUNC=wait_too_many_sessions
+test_error_exit_handler()
+{
+	stop_holder
+}
+
+count_dropped_sessions()
+{
+	if [ -f "$STARTED_LOG" ]; then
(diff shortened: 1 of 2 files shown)


rgerhards:
Addressed the AI review hardening point by normalizing the drop-count helper output to a numeric value and making the failure dump tolerate a missing started log. Revalidated locally with 10 direct `./tests/imptcp_maxsessions.sh` runs and `make check TESTS="imptcp_maxsessions.sh"` before force-pushing the amended signed commit.

rgerhards:
@cubic-dev-ai review

PR #6831, opened 2026-05-07
python: make file cleanup lexical

rgerhards:
Why:
CodeQL could not prove that several long-lived Python file handles were closed on every path when cleanup was hidden behind lifecycle hooks.

Impact:
The scripts keep the same behavior while making file ownership explicit.

Before/After:
Before, some files were opened in setup code and closed later indirectly. After, each file is owned by a nearby context manager or direct cleanup block.

Technical Overview:
- Move sample plugin log files into context managers around their loops.
- Open Solr and SPM error logs in context managers that own the plugin loop.
- Keep HTTP connection cleanup in onExit while the file context owns the log.
- Wrap the SNMP receiver output and log files in nested context managers.
- Close partially initialized impstats CSV files if header setup fails.
- Keep the existing final impstats close pass for all retained CSV handles.

Validation:
- python3 -m py_compile plugins/external/skeletons/python/plugin.py plugins/external/skeletons/python/plugin-with-feedback.py tests/snmptrapreceiver.py
- python3 -m tokenize plugins/external/solr/rsyslog_solr.py plugins/external/spm/rsyslog_spm_custom_metrics.py plugins/impstats/statslog-analyzer.py
- git diff --check

Stale reports checked separately:
- SimplePlugin.java no longer calls File.createNewFile().
- tests/historical/DiagTalker.java no longer exists on main.

### DIFF
--- a/plugins/external/skeletons/python/plugin-with-feedback.py
+++ b/plugins/external/skeletons/python/plugin-with-feedback.py
@@ -72,11 +72,6 @@ def onInit():
     # emitted you must set 'level' to logging.DEBUG above.)
     logging.debug("onInit called")
 
-    # For illustrative purposes, this plugin skeleton appends the received logs
-    # to a file. When implementing your plugin, remove the following code.
-    global outfile
-    outfile = open("/tmp/logfile", "w")
-
 
 def onMessage(msg):
     """Process one log message received from rsyslog (e.g. send it to a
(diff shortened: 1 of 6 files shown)


rgerhards:
Addressed Gemini's review by changing the new statslog-analyzer error cleanup handler from bare `except:` to `except Exception:`, so it no longer catches system-exiting exceptions while still closing the just-opened file on header setup failures.

PR #6850, opened 2026-05-08
rosi-collector: prevent Loki 400 queue stalls

alorbach:
ROSI Collector could get stuck when Loki rejected a pushed sample with
HTTP 400, for example because old or clock-skewed sender timestamps were
outside Loki's accepted time window. In that state one bad backlog entry
could keep the omhttp action retrying instead of advancing to newer logs.

Impact:
Loki rejects no longer block the collector queue indefinitely.

Before/After:
Before, ROSI used sender-reported time for Loki timestamps and retried
400 responses. After, it uses collector receive time and drops 400 rejects.

Technical Overview:
Use rsyslog timegenerated for the Loki push API timestamp so queued,
backlogged, or clock-skewed messages are ordered by collector processing
time. The original syslog timestamp remains preserved in the message body.

Configure omhttp with httpIgnorableCodes=["400"] so Loki reject-sample
responses are treated as processed instead of retried forever. Keep other
failure handling bounded by changing action.resumeRetryCount from -1 to
100.

Remove the stray /var/log/debug selector from the Loki ruleset.

Teach init.sh to hash the installed rsyslog.conf directory before and
after copying/generating collector config. If the installed collector
rsyslog config changed and an rsyslog Compose container already exists,
init.sh recreates only that service with --force-recreate so Docker
refreshes the bind-mounted config. First installs still defer to the
normal docker compose up -d flow.

Document the Loki 400 troubleshooting path, the collector timestamp
choice, the automatic init.sh recreate behavior, and the manual checksum
check for stale config suspicions.

### DIFF
--- a/deploy/docker-compose/rosi-collector/README.md
+++ b/deploy/docker-compose/rosi-collector/README.md
@@ -577,6 +577,31 @@ docker inspect rosi-collector-grafana-1 --format '{{.State.Status}} {{.State.Err
    rosi-monitor health  # Includes "Log flow (rsyslog -> omhttp -> Loki)" check
    ```
 
+If rsyslog logs show `omhttp: checkResult error http status code: 400`
+or Loki reports entries as "too far behind", Loki rejected the sample and
+the collector should advance to newer messages. The shipped
+`rsyslog.conf/30-send-loki-http.conf` marks HTTP 400 as non-retriable with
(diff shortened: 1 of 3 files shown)


alorbach:
@cubic-dev-ai review
@gemini review
@codex review

alorbach:
@cubic-dev-ai review
@gemini review
@codex review

PR #6853, opened 2026-05-09
hardning: config parameter ownership and OOM cleanup

rgerhards:
Why
Some config parameter blocks still retained es_str2cstr() results without checking for NULL first. That only shows up on OOM, but an OOM path during startup still needs to fail cleanly.

Impact
Mostly cosmetic in the normal case. The change only affects startup behavior when allocation fails while parsing config parameters.

Before/After
Before: retained config strings could dereference NULL on OOM.
After: the affected config parameter blocks fail cleanly on OOM.

Technical Overview
Wrap retained es_str2cstr() results in CHKmalloc() inside config parameter blocks only.
Leave temporary validation strings alone when they are freed in-place.
Keep the behavior unchanged for successful allocations.
Audit the touched modules for remaining unchecked retained paths.
Validate the tree with make -j$(nproc) check TESTS="".

### DIFF
--- a/contrib/imbatchreport/imbatchreport.c
+++ b/contrib/imbatchreport/imbatchreport.c
@@ -652,20 +652,20 @@ BEGINnewInpInst
     for (i = 0; i < inppblk.nParams; ++i) {
         if (!pvals[i].bUsed) continue;
         if (!strcmp(inppblk.descr[i].name, "reports")) {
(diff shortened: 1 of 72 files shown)


rgerhards:
@cubic-dev-ai review

rgerhards:
AI comments have been adressed

PR #6859, opened 2026-05-10
[HARDENING] imuxsock: harden trusted property annotation buffer

rgerhards:
## Summary

This hardens imuxsock trusted-property annotation assembly by replacing fixed slack-space assumptions with explicit size accounting before each append.

The annotation path can add process metadata such as PID, UID, GID, command name, executable path, and escaped command line to the received message. The old code relied on a fixed extra allocation and direct `memcpy()` calls. This change computes the maximum trusted-property expansion, tracks the output offset as `size_t`, and verifies each append before writing.

## Impact

Normal imuxsock behavior is unchanged. Oversized annotation data is handled through the existing error path instead of relying on implicit buffer slack.

## Validation

- `./devtools/format-code.sh`
- `./tests/imuxsock_annotate_large.sh`

### DIFF
--- a/plugins/imuxsock/imuxsock.c
+++ b/plugins/imuxsock/imuxsock.c
@@ -871,6 +871,34 @@ static int copyescaped(uchar *dstbuf, uchar *inbuf, int inlen) {
 }
 
 
+static size_t escapedLen(uchar *inbuf, int inlen) {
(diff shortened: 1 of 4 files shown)


rgerhards:
CI is environment fail.

PR #6866, opened 2026-05-11
imbeats: join session threads during teardown

rgerhards:
## Summary

- make imbeats session threads joinable instead of detached
- transfer accepted netstream ownership explicitly into sessions
- stop and reap open sessions during instance teardown before freeing config
- add a shutdown regression test for idle and partial open Beats sessions

## Validation

- `git diff --check`
- `bash -n tests/imbeats-shutdown-open-session.sh`
- previously validated before this PR split with `autogen.sh`, `make -j$(nproc) check TESTS=""`, focused imbeats tests, and mock `distcheck`

### DIFF
--- a/plugins/imbeats/imbeats.c
+++ b/plugins/imbeats/imbeats.c
@@ -17,6 +17,7 @@
 #include <poll.h>
 #include <arpa/inet.h>
 #include <netinet/in.h>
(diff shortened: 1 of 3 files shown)


rgerhards:
CI fail is unrelated

PR #6874, opened 2026-05-11
imbeats: retry interrupted poll calls

rgerhards:
## Summary

- retry `poll()` in the imbeats session wait helper when interrupted by signals
- handle listener-loop `EINTR` explicitly before applying timeout/error handling

## Why

A signal can interrupt `poll()` while imbeats is waiting for socket readiness.
The session path previously treated that as an I/O error, which could close an
otherwise healthy Beats connection.

## Validation

- `./autogen.sh --enable-debug --enable-testbench --enable-imdiag --enable-imbeats --enable-omstdout`
- `./tests/imbeats-basic.sh`
- `./tests/imbeats-compressed.sh`
- `./tests/imbeats-shutdown-open-session.sh`
- `./tests/imbeats-seq-cumulative-two-windows.sh`

### DIFF
--- a/plugins/imbeats/imbeats.c
+++ b/plugins/imbeats/imbeats.c
@@ -355,10 +355,16 @@ finalize_it:
 
 static rsRetVal sessionPollWait(const int fd, const short events, const int timeout_ms) {
     struct pollfd pfd;
+    int pollRet;
+
     pfd.fd = fd;
     pfd.events = events;
     pfd.revents = 0;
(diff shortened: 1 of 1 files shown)


rgerhards:
Updated in c895a5cef: sessionPollWait now treats EINTR as a non-fatal interruption and returns to the caller, so the surrounding read/write loops can re-check termination state before retrying. I also reran the focused imbeats basic and shutdown-open-session tests after the amend.

PR #6903, opened 2026-05-14
java: use nio writer in simple plugin skeleton

rgerhards:
## Summary

This updates the Java external plugin skeleton to open `out.txt` with `Files.newBufferedWriter` using explicit `CREATE` and `APPEND` options.

## Rationale

GitHub CodeQL reports an ignored `File.createNewFile` status for `plugins/external/skeletons/java/SimplePlugin.java` on `main`, even though the current source no longer contains a direct `createNewFile()` call. Switching away from `FileOutputStream` construction makes the file creation and append behavior explicit for the analyzer.

## Impact

No rsyslog runtime behavior changes. This only affects the sample Java skeleton used for external plugin examples.

## Validation

No local runtime testing was run, as requested. I ran `git diff --check` for the touched Java file.

### DIFF
--- a/plugins/external/skeletons/java/SimplePlugin.java
+++ b/plugins/external/skeletons/java/SimplePlugin.java
@@ -19,13 +19,21 @@ See the License for the specific language governing permissions and
 limitations under the License.
 
 */
-import java.io.*;
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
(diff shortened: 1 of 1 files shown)


rgerhards:
no ci tests need as there is nothing to test except for the usual linters (and CI is running at capacity)

Issue #6308, opened 2025-11-19
[bug]: omfwd-lb-1target-retry-* tests still flaky / racy


mbiebl:
Version: 8.2510.0
OS: Debian sid

Looks like despite the attempts made in #5881,  the omfwd-lb-1target-retry-* related tests are still flaky and suceptible to races.

A prominent example is arm64 on reproducible-builds.org , see e.g.
https://tests.reproducible-builds.org/debian/rb-pkg/unstable/arm64/rsyslog.html

```
FAIL: omfwd-lb-1target-retry-1_byte_buf-TargetFail
==================================================

------------------------------------------------------------
11:43:12[0]  Test: ./omfwd-lb-1target-retry-1_byte_buf-TargetFail.sh
------------------------------------------------------------
waiting for file rstb_731893_ca200ac76tKx.minitcpsrvr_port1
writing to file rstb_731893_ca200ac76tKx.out.log
listen on target port 0
### background minitcpsrv process id is 1557293 port 35093 ###
Note: intentionally not started any local TCP receiver!
config rstb_731893_ca200ac76tKx_.conf is:
     1	module(load="../plugins/imdiag/.libs/imdiag")
     2	global(inputs.timeout.shutdown="60000"
     3	       default.action.queue.timeoutshutdown="20000"
     4	       default.action.queue.timeoutEnqueue="20000"
     5	       debug.abortOnProgramError="on")
     6	# use legacy-style for the following settings so that we can override if needed
     7	$MainmsgQueueTimeoutEnqueue 20000
     8	$MainmsgQueueTimeoutShutdown 10000
     9	$IMDiagListenPortFileName rstb_731893_ca200ac76tKx.imdiag.port
    10	$IMDiagServerRun 0
    11	$IMDiagAbortTimeout 580
    12	
    13	:syslogtag, contains, "rsyslogd"  ./rstb_731893_ca200ac76tKx.started
    14	###### end of testbench instrumentation part, test conf follows:
    15	
    16	$MainMsgQueueTimeoutShutdown 10000
    17	$MainMsgQueueDequeueBatchSize 100
    18	
    19	global(allMessagesToStderr="on")
    20	
    21	template(name="outfmt" type="string" string="%msg:F,58:2%\n")
    22	module(load="builtin:omfwd" template="outfmt" iobuffer.maxSize="1000")
    23	
    24	if $msg contains "msgnum:" then {
[...]

PR #6925, opened 2026-05-17
tests: harden imbeats shutdown session setup

rgerhards:
## Summary

The de-flake campaign repeatedly found `imbeats-shutdown-open-session.sh` failing under broad `-j250` stress. The test was marking clients ready after TCP connect, which did not prove that `imbeats` had accepted and registered the sockets it was supposed to close during shutdown.

This changes the setup so each client first sends a valid Lumberjack batch and waits for the expected ack. That ack is the oracle that the session is owned by `imbeats`; only then does the test leave the sockets open in idle/partial-frame states and start shutdown.

Related: https://github.com/rsyslog/rsyslog/issues/6308

## Validation

- `./tests/imbeats-shutdown-open-session.sh` in dev container: pass, 00:02
- Ubuntu 26.04 full `-j250` stress, ack-only candidate: reproduced imbeats failure, 05:24
- Ubuntu 26.04 full `-j250` stress, final patch: imbeats passed; unrelated `tcp_forwarding_retries.sh` and `imuxsock_unlink_off_stale_socket.sh` failed, 05:12
- Ubuntu 26.04 full `-j250` stress, final patch: imbeats passed; unrelated `imuxsock_unlink_off_stale_socket.sh` failed, 04:46

The unrelated failures were recorded in the flake-campaign repo.

### DIFF
--- a/tests/imbeats-shutdown-open-session.sh
+++ b/tests/imbeats-shutdown-open-session.sh
@@ -2,6 +2,12 @@
 . ${srcdir:=.}/diag.sh init
 require_plugin imbeats
 
+# This checks that imbeats shutdown closes client sessions that are already
+# accepted and left open in idle or partial-frame protocol states. The client
+# first completes one valid Lumberjack batch on each socket and waits for the
+# expected ack; that ack is the oracle that the session is owned by imbeats
+# before the test starts rsyslog shutdown.
+
 generate_conf
 add_conf '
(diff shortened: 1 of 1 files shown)


rgerhards:
Addressed the AI review note by replacing the bash arithmetic for-loop with a portable while loop and an explicit testbench timeout default. Re-ran ./tests/imbeats-shutdown-open-session.sh in the dev container successfully.

PR #6948, opened 2026-05-18
docker: add container regex noise filter

### LINKED ISSUE #5821: docker: improve base container with filtering capabilities
We use the new regexp support in lookup tables to provide a basic, externally-configurable, noise drop rule for the base container. This permits to remove noise events right at the source before they go into further processing. This capability also extends to all containers build from the base container.

rgerhards:
Add a native regex lookup-table filter to the base container image so operators can drop known noisy events before packaged outputs process them. Document the inherited behavior and cover the default and named-ruleset paths with a testbench scenario.

closes https://github.com/rsyslog/rsyslog/issues/5821


rgerhards:
Addressed the shell-portability part by replacing the bash here-string in \	ests/container-noise-drop.sh\ with a \printf\ pipeline. For dockerlogs and ETL: both images inherit \/etc/rsyslog.d/02-noise-drop.conf\ from the minimal/standard base and their packaged inputs/actions stay in the default ruleset, so the inherited filter runs before their output snippets. The only image that binds inputs directly to a named ruleset is collector, and this PR adds the filter at the start of that \
s-main\ ruleset.

rgerhards:
CI fail is unrelated

PR #7052, opened 2026-05-26
tests: fix pipe and imfile statistics sync oracles

rgerhards:
## Summary

This fixes two test synchronization bugs exposed by the local deflake stress campaign:

- `pipeaction.sh` now waits for the external FIFO reader output file to contain every injected message before shutdown.
- `imfile-statistics.sh` now waits for the aggregate output from all three input files before shutdown and sequence validation.

Both tests also now document their intent and oracle at the top of the script.

## Root Cause

`pipeaction.sh` treated rsyslog main queue empty as sufficient before shutdown. That does not prove the external FIFO reader has drained and copied all pipe output, because that reader is outside rsyslog queue accounting.

`imfile-statistics.sh` injects three files with 1,000 lines each, but bare `wait_file_lines` only waited for `NUMMESSAGES`, i.e. 1,000 lines. The later `seq_check 0 2999` already expected all 3,000 records, so the wait oracle was weaker than the actual test invariant.

## Validation

Full Ubuntu 26.04 dev-container check run at `-j400`:

```text
CI_MAKE_CHECK_OPT=-j400
1280 total, 1187 pass, 93 skip, 0 fail
runtime: 5m32.61s
```

The stressed full run included passing `pipeaction.sh`, `imfile-statistics.sh`, and `imfile-endmsg.regex-with-example.sh`.

### DIFF
--- a/tests/imfile-statistics.sh
+++ b/tests/imfile-statistics.sh
@@ -1,7 +1,16 @@
 #!/bin/bash
 # This is part of the rsyslog testbench, licensed under ASL 2.0
+#
+# Test imfile per-file statistics for two wildcard-matched files and one
+# explicitly configured file. The oracle is twofold: all three input files must
+# produce one contiguous output sequence, and impstats must report the expected
+# byte and line counters for each input file. The output wait uses the aggregate
+# message count so shutdown does not race ahead after only one file's data was
+# observed.
 . ${srcdir:=.}/diag.sh init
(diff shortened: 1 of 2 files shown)


rgerhards:
AI review follow-up: addressed Gemini’s maintainability suggestion by naming the three input files explicitly as `NUM_INPUT_FILES=3` and deriving `TOTALMESSAGES` from that value.

PR #7060, opened 2026-05-26
ai: harden local review workflow against prompt-injection tool abuse

rgerhards:
### Motivation
- Prevent prompt-injection and sensitive-data leakage from local developer tooling by eliminating paths that send attacker-controlled diffs/untracked content to AI CLIs with unrestricted tool execution flags.
- Reduce blast radius while preserving the scripted allocation-policy check for tracked/staged diffs.

### Description
- Removed the Codex `--dangerously-bypass-approvals-and-sandbox` flag from the allocation-policy invocation so the CLI is no longer asked to run in an explicitly unrestricted sandbox-bypass mode.
- Removed the Copilot `--allow-all-tools` flag from the allocation-policy invocation to avoid granting blanket tool access to the remote CLI invocation.
- Eliminated the logic that appended untracked file contents into the temporary diff used to build the AI prompt and added a comment explaining that untracked files are intentionally excluded.
- Preserved existing behavior for generating diffs from tracked/staged changes and the JSON-format issue output expected by downstream stages.

### Testing
- Ran a shell syntax check with `bash -n ai/local-review-workflow.sh`, which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a15f9d0680c833293a8c35414d7f489)

### DIFF
--- a/ai/local-review-workflow.sh
+++ b/ai/local-review-workflow.sh
@@ -177,17 +177,15 @@ check_allocation_policy() {
       die "Forced --subcli codex but codex not found"
     fi
     debug "Using codex for policy check"
-    debug "Command: codex exec --model $POLICY_CHECK_MODEL --dangerously-bypass-approvals-and-sandbox <PROMPT>"
+    debug "Command: codex exec --model $POLICY_CHECK_MODEL <PROMPT>"
     if $DEBUG; then
       codex exec \
         --model "$POLICY_CHECK_MODEL" \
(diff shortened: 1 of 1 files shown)


rgerhards:
We need to review first if this workflow is still needed. I doubt,

rgerhards:
this script was an old experiment and has been removed by #7154 - so this PR is no longer needed.

PR #7104, opened 2026-05-28
imptcp: harden auto compression probe

rgerhards:
### Motivation

- Fix misclassification in `compression.mode="stream:auto"` where two-byte zlib-looking prefixes could permanently route a plaintext session to the compressed path and cause the session to be closed on zlib errors.  
- Ensure an imptcp listener that must accept both compressed and plain peers can tolerate plaintext sessions whose first bytes happen to match a zlib header.

### Description

- Replace the single fixed sniff window with a bounded probe window by introducing `COMPRESS_AUTO_MIN_SNIFF_BYTES` (2) and `COMPRESS_AUTO_MAX_SNIFF_BYTES` (4096) and enlarge the per-session sniff buffer accordingly.  
- Rework `compressionAutoDetect()` so the two-byte zlib header is treated only as a candidate: when a candidate header is seen, a throw-away zlib `inflate()` probe is run on the buffered bytes (using a temporary `z_stream`) and the result is used to classify the session; the function returns `0` while the verdict is still ambiguous.  
- Update `DataRcvd()` to buffer ambiguous session-start bytes up to the bounded maximum, call the revised `compressionAutoDetect()`, lock `compressionMode` only after the probe returns a verdict, and re-inject the buffered bytes into either the compressed or uncompressed path as before.  
- Add a regression test `tests/imptcp-stream-compression-auto-zlib-prefix-plain.sh` (and register it in `tests/Makefile.am`) that verifies a plaintext octet-stuffed session beginning with ASCII `x^` is accepted and that subsequent writes on the same connection are processed.

### Testing

- Ran `./configure --enable-imptcp --enable-testbench --disable-impstats-push` and performed an incremental host build with `make -j${RSYSLOG_LOCAL_BUILD_JOBS:-10} check TESTS=""` which completed successfully.  
- Executed the relevant test scripts with `make check TESTS="imptcp-stream-compression-auto.sh imptcp-stream-compression-auto-zlib-prefix-plain.sh"` and both tests passed.  
- Ran `devtools/check-test-antipatterns.sh` on the new test and `shellcheck -S warning` against the new test script with no failures.  
- Ran the repository local validation helper `devtools/local-validation-plan.sh --base HEAD --run`; available local checks completed, but full container-style validation (Cubic / Docker / Podman lanes and the Ubuntu 26.04 analyzer container runs) was not available in this environment and should be exercised in CI or by a container-capable reviewer.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a1865d03d788332ba34c13c9cac1e57)

### DIFF
--- a/plugins/imptcp/imptcp.c
+++ b/plugins/imptcp/imptcp.c
@@ -1452,21 +1452,19 @@ static int compressionAutoDetect(const uchar *const buf, const size_t len, sbool
 
     probe.next_in = (Bytef *)buf;
     probe.avail_in = len;
-    do {
-        probe.next_out = out;
-        probe.avail_out = sizeof(out);
-        zRet = inflate(&probe, Z_SYNC_FLUSH);
-        if (sizeof(out) - probe.avail_out != 0 || zRet == Z_STREAM_END) {
-            *pIsCompressed = 1;
-            inflateEnd(&probe);
-            return 1;
-        }
-        if (zRet != Z_OK && zRet != Z_BUF_ERROR) {
-            *pIsCompressed = 0;
-            inflateEnd(&probe);
-            return 1;
-        }
-    } while (probe.avail_out == 0);
+    probe.next_out = out;
+    probe.avail_out = sizeof(out);
+    zRet = inflate(&probe, Z_SYNC_FLUSH);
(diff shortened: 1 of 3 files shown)


rgerhards:
@jjourdin The zlib check has become a bit more elaborate to gain confidence. Just so that you know.

PR #7147, opened 2026-05-31
tests: add incremental doc sample coverage

rgerhards:
## Summary
- add a first tranche of `doc-sample-*` tests for user-facing configuration examples
- fix the database tutorial sample that the new coverage exposed
- keep this branch open for incremental follow-up commits that extend coverage across more doc samples

## Validation
- rendered docs build passed locally
- local Cubic review passed locally
- focused first-tranche doc-sample runs are clean with the expected host-local result of 6 pass / 1 skip
- the skip is `doc-sample-database.sh` on hosts where `ommysql` exists in-tree but cannot load because a shared-library dependency such as `libmysqlclient.so.24` is missing
- `devtools/local-validation-plan.sh --run` is not yet fully green because it also hit unrelated existing failures in `imdtls-basic-vg` and `omprog-diskq-restart-save`

## Notes
This is intended as an incremental PR. I will keep extending coverage, including simpler documentation examples, in additional commits on the same branch.

### DIFF
--- a/doc/source/tutorials/database.rst
+++ b/doc/source/tutorials/database.rst
@@ -126,9 +126,10 @@ like this:
    template(
        name="sqlInsertMessage"
        type="list"
+       option.sql="on"
    ) {
        constant(value="insert into syslog(message) values ('")
-       property(name="msg" sql="on")
+       property(name="msg")
        constant(value="')")
    }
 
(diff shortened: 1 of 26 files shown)


rgerhards:
this does not look like the right approach. to validate the doc samples, we better use a dynamic method. but it may still be worth having some functionality integration tests so that we can ensure coverage for the samples.

Keeping it open, but not for immediate merge.

rgerhards:
Closing this PR because it uses the older hand-copied `tests/doc-sample-*.sh` approach.

The replacement is #7351, which validates marked documentation samples dynamically by extracting `rsyslog` code blocks from the RST source, running them through `rsyslogd -C -N1`, and feeding the resulting coverage into Codecov via the `docs-samples` flag.

PR #7149, opened 2026-06-01
ci: upload Codecov from daily focused tests

rgerhards:
## Summary
- re-enable Codecov upload from the trusted daily focused tests workflow
- add lcov capture and Codecov flags for the existing base, journal, and kafka daily lanes
- fail the daily lane if coverage generation/upload fails so the existing daily focused issue reporter records it

## Validation
- python3 YAML parse for .github/workflows/daily_focused_tests.yml
- actionlint .github/workflows/daily_focused_tests.yml
- git diff --check

### DIFF
--- a/.github/workflows/daily_focused_tests.yml
+++ b/.github/workflows/daily_focused_tests.yml
@@ -69,6 +69,7 @@ jobs:
               default-jre-headless \
               faketime \
               gdb \
+              lcov \
               libaprutil1-dev \
               libbson-dev \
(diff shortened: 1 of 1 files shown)


rgerhards:
Addressed the issue identified by cubic: coverage filtering now fails the daily lane if \lcov --remove\ fails instead of masking the error with \|| true\. Re-ran local YAML parse, actionlint, and git diff --check successfully.

PR #7193, opened 2026-06-08
diag.sh: fix content-pattern-check and assert-first-column-sum-greate…

bogoyavlensky:

This change improves test suite robustness by adding quotes around `$4` in `cat` commands. Previously, filenames with spaces would cause command failures and mask real test errors. Now such filenames are handled correctly.

### References

Refs: none (test hardening)

### Notes

Defensive quoting fix; no behavior change for normal filenames.


bogoyavlensky:
Hi @rgerhards . 
Am I correct in understanding that my pr will be merged into milestone v8.2606? 
Is there any other action required from me? 
Thanks.

PR #7228, opened 2026-06-18
config: warn on constant boolean operands

### LINKED ISSUE #1046: Incorrect filter not detected
When having incorrect filter operations, the configuration check doesn't produce a warning/error. Also, the debug output does not show this. 

Config example:
`if $msg contains 'bla1' or 'bla2' then {
        action(type="omfile" file="/var/log/bla12.log")
} else {
        if $msg contains 'bla3' or 'bla4' then {
                action(type="omfile" file="/var/log/bla34.log")
        } else {
                if $msg contains 'bla5' then {
                        action(type="omfile" file="/var/log/bla5.log")
                } else {
                        if $msg contains 'bla6' then {
                                action(type="omfile" file="/var/log/bla6.log")
                        } else {
                                action(type="omfile" file="/var/log/blarest.log")
                                }
                        }
                }
        }
`

When reading this, it seems logical. The problem is, that after the "or" operator a complete operation is expected with property, comparison and value. Instead, just a value is provided, which results in a boolean comparison that always evaluates to true. Thus, all operations further down the configuration will not be carried out, even if the values from the message did not trigger the first rule. 

To have this working correctly, the syntax for the filters with multiple values must be different. It must be either of these:
`if $msg contains ["bla1", "bla2"] then {}

if $msg contains 'bla1' or $msg contains 'bla2' then {}`

rgerhards:
## Summary
- warn during config processing when RainerScript `and`/`or` receives a bare string or number constant operand
- keep the existing accepted syntax and runtime truthiness behavior unchanged
- add a config-check regression test for the reported `$msg contains "a" or "b"` pattern

Closes https://github.com/rsyslog/rsyslog/issues/1046

## Validation
- `bash -n tests/rscript_bool_constant_warning.sh`
- `devtools/check-test-antipatterns.sh tests/rscript_bool_constant_warning.sh`
- `git diff --check`
- `devtools/format-code.sh --git-changed --check --check-if-available`
- `make -j80 check TESTS=""`
- `./tests/rscript_bool_constant_warning.sh`
- `make -j80 check TESTS="rscript_bool_constant_warning.sh"`
- `cubic review --print-logs --base upstream/main`
- `RSYSLOG_LOCAL_CHECK_JOBS=80 RSYSLOG_LOCAL_BUILD_JOBS=80 timeout 120m devtools/local-validation-plan.sh --run`

<a href="https://cubic.dev/pr/rsyslog/rsyslog/pull/7228?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/grammar/rainerscript.c
+++ b/grammar/rainerscript.c
@@ -5448,6 +5448,15 @@ static struct cnfexpr *cnfexprOptimize_NOT(struct cnfexpr *expr) {
 static struct cnfexpr *cnfexprOptimize_AND_OR(struct cnfexpr *expr) {
     struct cnffunc *funcl, *funcr;
 
+    if (expr->l->nodetype == 'N' || expr->l->nodetype == 'S') {
+        parser_warnmsg("boolean operator '%s' has constant left operand; did you mean to repeat the comparison?",
+                       tokenToString(expr->nodetype));
+    }
+    if (expr->r->nodetype == 'N' || expr->r->nodetype == 'S') {
+        parser_warnmsg("boolean operator '%s' has constant right operand; did you mean to repeat the comparison?",
+                       tokenToString(expr->nodetype));
+    }
+
     if (expr->l->nodetype == 'F') {
         if (expr->r->nodetype == 'F') {
             funcl = (struct cnffunc *)expr->l;
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -474,6 +474,7 @@ TESTS_DEFAULT = \
 	lookup_table_rscript_reload.sh \
 	lookup_table_rscript_reload_without_stub.sh \
 	config_multiple_include.sh \
+	rscript_bool_constant_warning.sh \
 	include-obj-text-from-file.sh \
 	include-obj-text-from-file-noexist.sh \
 	multiple_lookup_tables.sh \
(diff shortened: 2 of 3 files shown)


julthomas:
Hello @rgerhards,

The warnings introduced in this PR are annoying when using legitimate use of constants. For instance I have such constructs in my config set.

```
[...] if (`echo $ZLC_USE_DNS` == "on" and $app-name == "nxlog-msvistalog") then { [...]
[...] if (`echo $ZLC_ARCHIVES_FQDN_SRC` != "on" and $.fromhost != $.fromhost-ip) then [...]
[...] if (`echo $ZLC_ARCHIVES_NXLOG_HOSTNAME_SRC` == "on" and $.app-name == "nxlog-msvistalog") then { [...]
```

Would you consider an option to disable such warning ? or remove the patch entirely ? I believe admins should know what they write and it is not rsyslog job to guess what they should have done or want to do ;-) What do you think ?

Have a nice day.

Regards,
Julien

julthomas:
@rgerhards 

Note: the warning triggers only if there is an "and' or an "or" in the condition. For instance the following expression does NOT trigger the warning, which is good for me.

```
[...]   if (`echo $ZLC_NETWORK_STOP_AFTER_OUTPUTS` == "on") then {   [...]
```

For that, I tend to think the warn should never happen when using backticks. It may have been resolved to constant string at that point where the check is enforced, this is another story ;-)

rgerhards:
> For that, I tend to think the warn should never happen when using backticks. It may have been resolved to constant string at that point where the check is enforced, this is another story ;-)

thats's a good point...

PR #7273, opened 2026-06-24
imuxsock: clarify missing Socket diagnostics

### LINKED ISSUE #4295: The default system socket is not actually overwritten for systemd.
According to [(https://www.rsyslog.com/doc/v8-stable/configuration/modules/imuxsock.html#syssock-use)](url), 

> if rsyslog is running under systemd AND /run/systemd/journal/syslog exists, (AND the user has not explicitly set SysSock.Use="off") then the default listener socket name is overwritten with /run/systemd/journal/syslog.

It turns out this is not the case.

Instead (according to messages in the operating state file, see below):

- imuxsock logs an error to the operating state file complaining of a missing 'socket' parameter--obviously, according to the documentation, this should not be the case
- imuxsock then logs that it has acquired the systemd system socket--OK, but why the error in the first place?
- imuxsock nevertheless then fails to write any log files notwithstanding having acquired the system socket--why would this be if the socket has been acquired?

Operating state file (note also the timestamp error, as per my issue posted a day or two ago):

```
20200422-150817: STATE CLEAN CLOSE
20200422-150902: STATE INITIALIZING 8.2004.0
20200422-150902: MSG   error during parsing file /etc/rsyslog.conf, on or before line 169: parameter 'socket' required but not specified - fix config
20200422-150902: MSG   imuxsock: required parameter are missing

20200422-150902: MSG   debug log file is '/var/log/rsyslog.debug', fd 6
20200422-150902: MSG   imuxsock: Acquired UNIX socket '/run/systemd/journal/syslog' (fd 3) from systemd.
```

The solution is to believe the error message notwithstanding the documentation, and include the system socket as a mandatory input parameter.  (Note also that setting syssock.name="/run/systemd/journal/syslog" has no effect.)

The trouble is, rsyslog doesn't log any messages until this is corrected, so the problem is invisible.  Only by a chance encounter with a passing reference to the operating state file, which is unset (not created) by default, and then configuring it in the hope that it would work as configured even though nothing else is, would anyone have a clue.  This took me months.

Probably:
- Either (a) the source code needs to be revised so as to acquire the system socket as documented and not throw an error; or (b) the documentation needs to be revised to reflect that the imuxsock.socket input parameter is mandatory, as the error message states, even for the systemd system log.
- The source code needs to be revised not to fail writing imuxsock messages to logs as configured after it acquires the system socket, even if the imuxsock.socket input parameter is omitted.

### Expected behavior

rsyslog executes if configured consistent with the documentation.

### Actual behavior

rsyslog fails if configured consistent with the documentation.

Configure rsyslog according to the documentation.

### Environment
- rsyslog version: 8.2004
- platform: Ubuntu 16.04 & 18.04

Configuration file attached:

[rsyslog.conf.pdf](https://github.com/rsyslog/rsyslog/files/4670698/rsyslog.conf.pdf)

rgerhards:
## Summary
- add an imuxsock-specific config error when input(type="imuxsock") omits Socket
- clarify that input objects are for additional Unix sockets and SysSock.* configures the system socket
- add a focused regression test using diag.sh content_check

Closes https://github.com/rsyslog/rsyslog/issues/4295

## Validation
- devtools/format-code.sh --git-changed --check --check-if-available
- git diff --check
- bash -n tests/imuxsock_missing_socket.sh
- devtools/check-test-antipatterns.sh tests/imuxsock_missing_socket.sh
- make -j60 check TESTS='imuxsock_missing_socket.sh'
- RSYSLOG_LOCAL_BUILD_JOBS=60 RSYSLOG_LOCAL_CHECK_JOBS=60 CI_MAKE_OPT=-j60 CI_MAKE_CHECK_OPT=-j60 devtools/local-validation-plan.sh --run
- Local Cubic review: no issues found

Container validation used Docker image rsyslog/rsyslog_dev_base_ubuntu:26.04 with image id sha256:17e57e817596410451a48b11dd4388a6e69ae9affd2a45cfb38f6cf87c48fc2d. No manual lane relaxations beyond the helper relevance selection.

<a href="https://cubic.dev/pr/rsyslog/rsyslog/pull/7273?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/doc/source/configuration/modules/imuxsock.rst
+++ b/doc/source/configuration/modules/imuxsock.rst
@@ -445,6 +445,11 @@ messages from applications running on the local system.
 
 This only needs to be done once.
 
+Do not add a bare ``input(type="imuxsock")`` for the system log socket.
+Additional ``imuxsock`` input objects are only for extra Unix sockets and must
+set the :ref:`param-imuxsock-socket` parameter. Configure the system log socket
+with module-level ``SysSock.*`` parameters instead.
+
 
 Enable flow control
 -------------------
@@ -572,4 +577,3 @@ system log socket.
    ../../reference/parameters/imuxsock-unlink
    ../../reference/parameters/imuxsock-usespecialparser
    ../../reference/parameters/imuxsock-parsehostname
-
--- a/doc/source/reference/parameters/imuxsock-socket.rst
+++ b/doc/source/reference/parameters/imuxsock-socket.rst
@@ -20,12 +20,15 @@ This parameter applies to :doc:`../../configuration/modules/imuxsock`.
 :Scope: input
 :Type: string
 :Default: input=none
-:Required?: no
+:Required?: yes
 :Introduced: at least 7.0.0, possibly earlier
 
 Description
 -----------
-Adds additional unix socket. Formerly specified with the ``-a`` option.
+Adds an additional unix socket. Formerly specified with the ``-a`` option.
+This parameter is required for every ``input(type="imuxsock" ...)`` object,
+because input objects configure additional sockets only. The system log socket
+is configured on the module object with ``SysSock.*`` parameters.
 
 Input usage
 -----------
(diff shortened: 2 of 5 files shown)


rgerhards:
@cubic-dev-ai does this cause backward compatibility issues?

PR #7394, opened 2026-07-17
rainerscript: warn on constant AND/OR operand at parse time

julthomas:
# rainerscript: warn on constant AND/OR operand at parse time

## Summary

Moves the AND/OR constant-operand warning (issue #1046) from the optimizer to
expression construction time, so it no longer fires on legitimate constant
comparisons and reports the correct file/line.

Follows up the discussion with @rgerhards in #7228 (2026-07-17).

## Background

The warning "boolean operator '%s' has constant left/right operand; did you mean
to repeat the comparison?" was added to catch the classic mistake
`$msg contains "a" or "b"` (#1046). It ran in `cnfexprOptimize_AND_OR()`, i.e.
during optimization, which has two consequences:

1. **False positives on folded comparisons.** `constFoldCmp()` reduces a
   constant comparison to a single constant node *before* the check runs. A
   common, legitimate pattern — feature flags driven by backtick-expanded
   environment variables — is flagged even though nothing is wrong:

   ```
   if (`echo $MY_FEATURE` == "on" and $app-name == "myapp") then { ... }
   ```

   expands to `("on" == "on" and ...)`; the left comparison folds to a constant,
   which the optimizer-stage check then reports as a "constant left operand".

2. **Wrong location.** The optimizer runs from `rulesetOptimizeAll()` after
   `yyparse()` has completed, so `cnfcurrfn`/`yylineno` point at the last line of
   the master config file for every occurrence, not at the offending expression.

## Change

The check is moved to `cnfexprNew()`, which is the single construction point for
every expression node and runs only at parse time. There:

- the operand is still the comparison node *as written* (constant folding has not
  happened yet), so a folded comparison is no longer mistaken for a bare literal;
- `cnfcurrfn`/`yylineno` still point at the expression, so the reported file and
  line are accurate.

The warning message text is unchanged, so tooling that matches it keeps working.

## Note on backtick operands

A backtick expansion used *directly* as a bare boolean operand still warns, and
this is intended:

```
if ($msg contains "alpha" or `echo $MY_FLAG`) then { ... }
```

The lexer expands the backtick before the expression tree is built, so this is
identical to a hand-written `... or "value"` — exactly the mistake the warning
targets, and treated the same as `$a contains "b" or "c"`. It is *not* the same
as the fixed case, where the backtick is an operand of a *comparison*
(`` `echo $FLAG` == "on" ``): there the boolean operand is the comparison, which
is legitimate and no longer warns.

## Testing

`tests/rscript_bool_constant_warning.sh` is extended: the positive case
(`$msg contains "a" or "b"` still warns) is unchanged, and a negative case using
a backtick-expanded feature flag verifies the folded comparison stays silent.
Manual `rsyslogd -N1` on a config using backtick-based feature flags confirms the
warnings are gone and that the bare-literal case is still reported, now at the
correct location.

<a href="https://cubic.dev/pr/rsyslog/rsyslog/pull/7394?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/grammar/rainerscript.c
+++ b/grammar/rainerscript.c
@@ -1413,6 +1413,26 @@ void cnfobjPrint(struct cnfobj *o) {
 struct cnfexpr *cnfexprNew(unsigned nodetype, struct cnfexpr *l, struct cnfexpr *r) {
     struct cnfexpr *expr;
 
+    /* Warn on bare constant AND/OR operands as written in the config, e.g.
+     * `$msg contains "a" or "b"` (issue #1046). This must happen here, at
+     * construction time, rather than in the optimizer: constFoldCmp() later
+     * reduces legitimate constant comparisons (typically backtick-expanded
+     * environment variables, e.g. `echo $FLAG` == "on") to constant operands,
+     * which must not warn. At this point such an operand is still a comparison
+     * node, and cnfcurrfn/yylineno still point at the offending expression, so
+     * the reported file and line are accurate.
+     */
+    if (nodetype == AND || nodetype == OR) {
+        if (l != NULL && (l->nodetype == 'N' || l->nodetype == 'S')) {
+            parser_warnmsg("boolean operator '%s' has constant left operand; did you mean to repeat the comparison?",
+                           tokenToString(nodetype));
+        }
+        if (r != NULL && (r->nodetype == 'N' || r->nodetype == 'S')) {
+            parser_warnmsg("boolean operator '%s' has constant right operand; did you mean to repeat the comparison?",
+                           tokenToString(nodetype));
+        }
+    }
+
     /* optimize some constructs during parsing */
     if (nodetype == 'M' && r->nodetype == 'N') {
         ((struct cnfnumval *)r)->val *= -1;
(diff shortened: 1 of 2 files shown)


julthomas:
Re/ gemini-code-assist: added `l != NULL` / `r != NULL` guards before the nodetype checks.

For the record on the mechanism: there is no `expr: error` production in the
grammar, so parser error recovery never hands a NULL operand to
`cnfexprNew(AND/OR, …)` — the rule only reduces once both sub-expressions have
parsed. The only way an operand can be NULL is an OOM in a sub-expression
constructor, and the surrounding code already relies on the non-NULL invariant
(the optimizer-stage check this patch replaced dereferenced `expr->l`/`expr->r`
the same way, and so does `cnfexprEval`). Since the check now runs at
construction time for every AND/OR node, the guard is cheap and keeps it
strictly additive, so let's add it.

rgerhards:
@codex review

PR #7403, opened 2026-07-21
mmkubernetes: reload SA token proactively and on HTTP 401

### LINKED ISSUE #7402: [Feature]:  mmkubernetes: reload ServiceAccount token so short-lived/rotated tokens don't cause 401s
**👋 Before you suggest, check the rsyslog Assistant!**

While the rsyslog Assistant at [https://rsyslog.ai](https://rsyslog.ai) is primarily for support questions, it can sometimes offer insights or existing solutions that might relate to your feature idea. It's an experimental AI-powered tool that draws from our documentation.

**👉 It might be worth a quick check at [https://rsyslog.ai](https://rsyslog.ai) to see if similar functionality exists or has been discussed!**

---

### 💡 Feature Description:

Make `mmkubernetes` re-read the Kubernetes API ServiceAccount bearer token from `tokenfile` after worker startup, instead of reading it exactly once and caching it forever. Two mechanisms: a proactive periodic reload controlled by a new `tokenreloadinterval` parameter (integer seconds, default `3600`, `0` disables), and a reactive reload-and-retry-once when a lookup returns HTTP 401.

### ❓ Problem it Solves:

`mmkubernetes` reads the token from `tokenfile` once, in `createWrkrInstance`, and caches it in the worker's persistent curl `Authorization` header. It is never re-read for the life of the worker (no mtime check, no inotify, no per-request rebuild), and on a 401/403 it only logs and aborts.

This was harmless while projected ServiceAccount tokens were valid for months — longer than a pod typically lives, so the cached token never went stale before the pod was recreated. With shorter-lived projected tokens this breaks. Kubernetes 1.34 caps projected-token validity at roughly one day (External JWT Signer), and the kubelet rotates the on-disk token at ~80% of its lifetime. The module keeps using the original cached copy, so about a day after a pod starts every cache-miss metadata lookup returns 401 and never recovers until the pod is restarted. Log data is unaffected; only the Kubernetes metadata enrichment (pod/namespace labels and annotations, and anything derived from them) is lost or misattributed.

### 🚀 Proposed Solution / Detailed Design:

Factor the `Authorization`-header construction out of `createWrkrInstance` into a small helper (`mmk_build_token_header()`) that reads the token fresh from `token`/`tokenfile` and rebinds the header to the worker's curl handle. Then reload the token two ways:

1. **Proactive** — in `queryKB`, re-read the token every `tokenreloadinterval` seconds so the worker stays inside the validity window and avoids the 401 entirely. Because validity greatly exceeds the interval, a single failed/missed reload is harmless. Default `3600`; `0` disables.
2. **Reactive** — on a **401**, rebuild the header from the current token on disk and retry the request exactly once before falling through to the existing error path. This covers the startup / identity-warm-up window the timer cannot (e.g. a pod that comes up before its node identity is authorized). A **403 is not retried**: it means the request authenticated but the identity lacks RBAC access, so re-reading the same token cannot help.

If a reload cannot obtain a token (e.g. the projected-token volume is briefly unavailable during rotation), the existing header is kept and the reload is retried on the next lookup rather than tearing down a working header. The healthy steady-state path is unchanged — cache hits never reach `queryKB`.

I have a working implementation with a testbench case (rotates the token file after the worker has cached the old value, then confirms a subsequent lookup 401s, reloads the fresh token, retries, and succeeds) and a doc update for the new parameter, and can open a PR.

### 🔄 Alternatives Considered:

- **Reactive-only (reload just on 401, no timer):** simplest, but leaves a predictable once-per-rotation 401 blip on every worker; the proactive reload avoids the failure entirely.
- **Hardcoded reload interval (no parameter):** rejected in favor of a configurable knob so operators can tune it to their token lifetime, or disable it (`0`) and rely solely on the reactive path.
- **inotify/mtime watch on the token file:** more moving parts and platform-specific; a periodic re-read plus the 401 backstop is simpler and sufficient given the token read is cheap and off the hot path.

### 💻 Environment (if applicable):

* **Rsyslog Version:** current `master` (`contrib/mmkubernetes`); behavior is unchanged from 8.23.x through today
* **Platform:** Kubernetes 1.34 (External JWT Signer enforced; projected-token validity ~24h). Not observed on 1.32, where token validity was months.
* **Deployment Method:** in-cluster, projected ServiceAccount token via `tokenfile` (the standard `/var/run/secrets/kubernetes.io/serviceaccount/token` projection)

### ➕ Additional Context:

Reproduce: run rsyslog with `mmkubernetes` in a pod on K8s 1.34 using a projected SA token via `tokenfile`; let it run past the token rotation point (~1 day); trigger a metadata cache miss (a new pod/namespace). The lookup fails with HTTP 401 and `mmkubernetes: Unauthorized ...` even though the on-disk token file holds a valid, freshly rotated token. Restarting the pod fixes it until the next rotation.

---
**Thank you for contributing to rsyslog! Your ideas help us grow.** ✨

prad23:
Fixes #7402

## What

`mmkubernetes` reads the ServiceAccount token from `tokenfile` once, at worker creation, and caches it in the worker's persistent curl `Authorization` header. It is never re-read (no mtime check, no inotify, no per-request rebuild), and a 401/403 only logs and aborts.

This was harmless while projected tokens were valid for months — longer than a pod typically lives. With shorter-lived projected tokens (e.g. Kubernetes 1.34 caps projected-token validity at ~1 day via the External JWT Signer), kubelet rotates the on-disk token at ~80% of its lifetime but the module keeps using the original cached copy. About a day after a pod starts, every cache-miss metadata lookup returns 401 and never recovers until the pod is restarted. Log data is unaffected; only the Kubernetes metadata enrichment is lost.

## How

The `Authorization`-header construction is factored out of `createWrkrInstance` into `mmk_build_token_header()`, which reads the token fresh and rebinds the header to the worker's curl handle. The token is then reloaded via two mechanisms:

1. **Proactive** — in `queryKB`, re-read the token every `tokenreloadinterval` seconds (new integer parameter, default `3600`, `0` disables) so the worker stays inside the validity window and avoids the 401 entirely. Because validity greatly exceeds the interval, a single failed/missed reload is harmless.
2. **Reactive** — on a **401**, rebuild the header from the current token on disk and retry the request once before falling through to the existing error path. This covers the startup / identity-warm-up window the timer cannot. A **403 is not retried**: it means the request authenticated but the identity lacks access, so re-reading the same token cannot help.

If a reload cannot obtain a token (e.g. the projected-token volume is briefly unavailable during rotation), the existing header is kept and `lastTokenReload` is not advanced, so the next lookup retries the reload rather than tearing down a working header. `createWrkrInstance` releases the curl handle on its own error path, since rsyslog does not call `freeWrkrInstance` when `createWrkrInstance` fails. The healthy steady-state path is unchanged — cache hits never reach `queryKB`. The reload logic lives inside `queryKB`, so it composes with the existing `kubernetesurl` failover loop.

## Tests

Adds `mmkubernetes-token-reload.sh` (plus a valgrind variant) that rotates the token file after the worker has cached the old value, then confirms a subsequent lookup 401s, reloads the fresh token, retries, and succeeds. The `mmkubernetes_test_server.py` mock gains an opt-in bearer-token check (enabled via `MMK8S_EXPECTED_TOKEN_FILE`) used by that test. Registered in `tests/Makefile.am` via `TESTS_MMKUBERNETES` / `TESTS_MMKUBERNETES_VALGRIND`.

## Docs

Adds the `tokenreloadinterval` parameter reference page and links it from the mmkubernetes module doc.

## Notes

- Formatted with `clang-format-18` per `.clang-format` / `devtools/format-code.sh`.
- I could not run the full C build/testbench locally (no rsyslog build deps on my machine); relying on CI for the compile and test run.

AI-Agent: Claude 2026-07

<a href="https://cubic.dev/pr/rsyslog/rsyslog/pull/7403?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/mmkubernetes-token-reload-vg.sh
+++ b/tests/mmkubernetes-token-reload-vg.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+# valgrind variant of mmkubernetes-token-reload.sh - see that script for details.
+export USE_VALGRIND="YES"
+. ${srcdir:=.}/mmkubernetes-token-reload.sh
--- a/contrib/mmkubernetes/mmkubernetes.c
+++ b/contrib/mmkubernetes/mmkubernetes.c
@@ -41,6 +41,7 @@
 #include <assert.h>
 #include <errno.h>
 #include <unistd.h>
+#include <time.h>
 #include <sys/stat.h>
 #include <libestr.h>
 #include <liblognorm.h>
@@ -106,6 +107,10 @@ DEFobjCurrIf(regexp) DEFobjCurrIf(statsobj) DEFobjCurrIf(datetime)
     -1 /* delete all expired entries from the cache every N seconds \
 -1 disables cache expiration/ttl checking                           \
 0 means - run cache expiration for every record */
+#define DFLT_TOKEN_RELOAD_INTERVAL                                         \
+    3600 /* proactively re-read the bearer token from tokenfile every N    \
+seconds so a rotated projected ServiceAccount token is picked up without a \
+restart; 0 disables the proactive reload */
 
 /* only support setting the partial chain flag on openssl platforms that have the define */
 #if defined(ENABLE_OPENSSL) && defined(X509_V_FLAG_PARTIAL_CHAIN)
@@ -159,6 +164,8 @@ struct modConfData_s {
     sbool sslPartialChain; /* if true, allow using intermediate certs without root certs */
     int cacheEntryTTL; /* delete entries from the cache if they are older than this many seconds */
     int cacheExpireInterval; /* delete all expired entries from the cache every this many seconds */
+    int tokenReloadInterval; /* proactively re-read the token from tokenFile every this many seconds
+                                (0 disables); guards against short-lived rotated SA tokens going stale */
 };
 
 /* action (instance) configuration data */
@@ -192,12 +199,15 @@ typedef struct _instanceData {
     sbool sslPartialChain; /* if true, allow using intermediate certs without root certs */
     int cacheEntryTTL; /* delete entries from the cache if they are older than this many seconds */
     int cacheExpireInterval; /* delete all expired entries from the cache every this many seconds */
+    int tokenReloadInterval; /* proactively re-read the token from tokenFile every this many seconds
+                                (0 disables); guards against short-lived rotated SA tokens going stale */
 } instanceData;
 
 typedef struct wrkrInstanceData {
     instanceData *pData;
     CURL *curlCtx;
     struct curl_slist *curlHdr;
+    time_t lastTokenReload; /* when curlHdr's bearer token was last (re)read from disk */
     char *curlRply;
     size_t curlRplyLen;
     statsobj_t *stats; /* stats for this instance */
@@ -238,7 +248,8 @@ static struct cnfparamdescr modpdescr[] = {{"kubernetesurl", eCmdHdlrArray, 0},
                                            {"busyretryinterval", eCmdHdlrInt, 0},
                                            {"sslpartialchain", eCmdHdlrBinary, 0},
                                            {"cacheentryttl", eCmdHdlrInt, 0},
-                                           {"cacheexpireinterval", eCmdHdlrInt, 0}
+                                           {"cacheexpireinterval", eCmdHdlrInt, 0},
+                                           {"tokenreloadinterval", eCmdHdlrInt, 0}
 #if HAVE_LOADSAMPLESFROMSTRING == 1
                                            ,
                                            {"filenamerules", eCmdHdlrArray, 0},
@@ -266,7 +277,8 @@ static struct cnfparamdescr actpdescr[] = {{"kubernetesurl", eCmdHdlrArray, 0},
                                            {"busyretryinterval", eCmdHdlrInt, 0},
                                            {"sslpartialchain", eCmdHdlrBinary, 0},
                                            {"cacheentryttl", eCmdHdlrInt, 0},
-                                           {"cacheexpireinterval", eCmdHdlrInt, 0}
+                                           {"cacheexpireinterval", eCmdHdlrInt, 0},
+                                           {"tokenreloadinterval", eCmdHdlrInt, 0}
 #if HAVE_LOADSAMPLESFROMSTRING == 1
                                            ,
                                            {"filenamerules", eCmdHdlrArray, 0},
@@ -645,6 +657,7 @@ BEGINsetModCnf
     loadModConf->sslPartialChain = DFLT_SSL_PARTIAL_CHAIN;
     loadModConf->cacheEntryTTL = DFLT_CACHE_ENTRY_TTL;
     loadModConf->cacheExpireInterval = DFLT_CACHE_EXPIRE_INTERVAL;
+    loadModConf->tokenReloadInterval = DFLT_TOKEN_RELOAD_INTERVAL;
     for (i = 0; i < modpblk.nParams; ++i) {
         if (!pvals[i].bUsed) {
             continue;
@@ -776,6 +789,8 @@ BEGINsetModCnf
             loadModConf->cacheEntryTTL = pvals[i].val.d.n;
         } else if (!strcmp(modpblk.descr[i].name, "cacheexpireinterval")) {
             loadModConf->cacheExpireInterval = pvals[i].val.d.n;
+        } else if (!strcmp(modpblk.descr[i].name, "tokenreloadinterval")) {
+            loadModConf->tokenReloadInterval = pvals[i].val.d.n;
         } else {
             dbgprintf(
                 "mmkubernetes: program error, non-handled "
@@ -804,6 +819,14 @@ BEGINsetModCnf
                      loadModConf->cacheEntryTTL);
             ABORT_FINALIZE(RS_RET_CONFIG_ERROR);
         }
+    }
+
+    if (loadModConf->tokenReloadInterval < 0) {
+        LogError(0, RS_RET_CONFIG_ERROR,
+                 "mmkubernetes: tokenreloadinterval value [%d] is invalid - "
+                 "value must be 0 (disabled) or greater",
+                 loadModConf->tokenReloadInterval);
+        ABORT_FINALIZE(RS_RET_CONFIG_ERROR);
     }
 
     /* set defaults */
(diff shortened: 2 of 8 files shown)


rgerhards (changes requested):
Thx for the PR, but the AI comments need to be addressed (either by fixes or giving reason why they are invalid).

prad23:
Thanks for the reviews — addressed in the latest push (branch amended to keep the PR a single commit):

**Fixed**
- **OOM on `curl_slist_append`** (Content-Type and Authorization appends): both now capture the return value and `ABORT_FINALIZE(RS_RET_OUT_OF_MEMORY)` on NULL, so the previously installed list is never leaked or silently dropped.
- **Partial token-file read**: the read now requires the full file (`fread == statbuf.st_size`); a truncated read during a rotation race leaves `tokenHdr` NULL, so the existing (still-valid) header is preserved rather than replaced with a bad token.
- **Inline `token` gating**: both the proactive interval reload and the reactive 401 reload+retry now run only when `tokenFile` is set and `token` is not. An inline token can't change, so a reload can't help — this matches the documented behavior and avoids needless header rebuilds / an extra request.
- **Doc `:Introduced:`**: corrected to `8.2608.0` to match `configure.ac`.
- **Grammar**: "which ensure" → "which ensures".

**Not changed (respectfully disagree)**
- The P1 that the test greps the wrong file: the module's internal `LogMsg` diagnostics *do* land in `$RSYSLOG_OUT_LOG` in this harness — the existing `mmkubernetes-basic.sh` greps that same file for the same class of message (`mmkubernetes: Not Found: ...`, `mmkubernetes: Too Many Requests: ...`) and passes in CI. The new test follows that established pattern, so the reload-and-retry assertion is effective, not vacuous.

I could not run the full C build/testbench locally (no rsyslog build deps on my machine), so I'm relying on CI for the compile and test run. `clang-format-18` (per `.clang-format` / `devtools/format-code.sh`) reports no changes.

prad23:
@rgerhards — done. Every AI comment is now either fixed or has a stated reason; summary mapping each one:

| # | Source | Comment | Resolution |
|---|--------|---------|------------|
| 1 | cubic P2 / Copilot | `curl_slist_append` OOM (Content-Type + Authorization) leaks/drops the list | **Fixed** — capture return, `ABORT_FINALIZE(RS_RET_OUT_OF_MEMORY)` on NULL |
| 2 | cubic P2 | Partial token-file read installs a truncated token | **Fixed** — require a complete `fread` (`== statbuf.st_size`); otherwise keep existing header |
| 3 | Copilot ×2 | Reload runs for inline `token` too | **Fixed** — both proactive and reactive reload now gated on `tokenFile != NULL && token == NULL` |
| 4 | Copilot | Doc `:Introduced:` off by a year | **Fixed** — now `8.2608.0` (matches `configure.ac`) |
| 5 | Copilot | Grammar "which ensure" | **Fixed** — "ensures" |
| 6 | cubic P1 | Test greps the wrong file for internal messages | **Invalid** — internal `LogMsg` output lands in `$RSYSLOG_OUT_LOG` in this harness; the existing `mmkubernetes-basic.sh` greps that same file for the same class of message (`mmkubernetes: Not Found: …`, `Too Many Requests: …`) and passes in CI. The new test follows that established pattern. |

All fixes are in the current commit (`clang-format-18` clean; both buildbot lanes green). One disclosure: I could not run the C build/testbench locally (no rsyslog build deps on my machine), so I relied on CI for the compile and test run.

rgerhards (changes requested):
thx for the changes.

But cubic has a valid point. The phrasing is just fuzzy. In fact, no manual grep should be used but the established harness functions like

```
content_check --regex 'error .* message'
```

Also ensure you fully utilize the imdiag.sh functions instead of re-inventing code.

prad23:
@rgerhards thanks — done. The test no longer uses manual `grep`/`rc` bookkeeping; it now uses the harness helpers from `diag.sh`:

- `content_check --regex 'mmkubernetes: got \[401\].*reloaded SA token and retrying once'` — asserts the reload-and-retry fired
- `assert_content_missing 'mmkubernetes: Unauthorized'` — asserts the retry recovered (no surfaced auth error)
- `content_check 'pod-name1-id'` / `content_check 'pod-name2-id'` — asserts enrichment actually happened. I switched from checking `pod_name`/`namespace_name` (which come from the log filename via `filenamerules`, so they'd be present even without a successful lookup) to `pod_id` (`pod-nameN-id`), which is derived from the Kubernetes API response `uid` — so its presence proves the API lookup succeeded, which is the behavior under test.

Both the main test and the valgrind variant were updated. The ad-hoc Python oracle and all raw `grep` calls are gone.

tests/mmkubernetes-token-reload-vg.sh
```diff
@@ -0,0 +1,108 @@
+#!/bin/bash
+# added 2026 for https://github.com/rsyslog/rsyslog/issues/7402, released under ASL 2.0
+#
+# valgrind variant of mmkubernetes-token-reload.sh - see that script for details.
```
rgerhards, line 2: do not duplicate the script. the modern and most common way is to source the orignal file. check other *-vg.sh tests, Some older ones may duplicate code or may be different in other tasks.

rgerhards (changes requested):
I usually do the manual review only after everthing else is green, but I noticed this while looking at the other issue. pls also fix.

prad23:
@rgerhards good point — fixed. `mmkubernetes-token-reload-vg.sh` is now the modern 3-line variant that sources the base test rather than duplicating it:

```bash
#!/bin/bash
# valgrind variant of mmkubernetes-token-reload.sh - see that script for details.
export USE_VALGRIND="YES"
. ${srcdir:=.}/mmkubernetes-token-reload.sh
```

The base `mmkubernetes-token-reload.sh` no longer hardcodes `USE_VALGRIND`; it uses plain `startup`/`wait_shutdown` (which dispatch to the valgrind path automatically) and only sets `EXTRA_VALGRIND_SUPPRESSIONS` when `USE_VALGRIND=YES`. Same pattern as `es-basic-vg.sh` / `imfile-basic-vg.sh`.

rgerhards:
@prad23 Thx - much appreciated. I how now started full CI, please check for the results. Note: CI can take up to 50 minutes to complete fully, depending on current load.

prad23:
Thanks for kicking off full CI! Two fast-failing lanes flagged issues in my additions, both now fixed and pushed:

- **`pycodestyle`** — `tests/mmkubernetes_test_server.py`: E128 (continuation-line indent) and E231 (missing whitespace after comma). Fixed; `pycodestyle` now passes clean locally.
- **`repo policy focus review`** (`doc-dist-sync`) — the new parameter page was missing its `EXTRA_DIST` entry. Added `source/reference/parameters/mmkubernetes-tokenreloadinterval.rst` to `doc/Makefile.am`.

The remaining full-CI lanes were still running when these two failed fast; the push re-triggers them. Please re-run / let me know if anything else turns up.

prad23:
Full CI is essentially green: **49 pass, 2 skip, 1 fail**. The single failure is on `CI (fedora_44)`:

```
FAIL: Test ./segmented-da-idle-cleanup-failure.sh (took 8 seconds)
```

That test is unrelated to this PR — it exercises disk-assisted queue idle-cleanup, is not part of this change, and the same `fedora_44` job shows both of the tests added here passing:

```
PASS: mmkubernetes-token-reload.sh
PASS: mmkubernetes-token-reload-vg.sh
```

pycodestyle and repo-policy (the two lanes I fixed after the first full run) are now green as well. Looks like a flaky/unrelated failure — could you re-run the `fedora_44` job when you get a chance? Happy to look further if it reproduces.

prad23:
Gentle nudge on this one @rgerhards — the only red is `CI (fedora_44)`, which failed on `segmented-da-idle-cleanup-failure.sh` (unrelated to this PR; both tests added here pass, and it looks like a known flaky test). Everything else is green (49 pass / 2 skip).

When you have a moment, could you re-run the `fedora_44` job and take another look? I believe all your review points are addressed. Thanks!

prad23:
Rebased onto the latest `main` to resolve the merge conflict that had developed (the branch had fallen ~188 commits behind). The only real overlap was in `contrib/mmkubernetes/mmkubernetes.c` and `tests/Makefile.am`, where the new `includenamespacemetadata` parameter landed adjacent to this PR's `tokenreloadinterval` additions — both now coexist. `clang-format-18` is clean and the change set is unchanged in intent. Ready for another CI run whenever you have a moment, @rgerhards.

rgerhards:
@codex review

rgerhards:
@prad23 thx again - enabled CI run

rgerhards:
@prad23 thx for your patience and the great contribution - and sorry for me being late again in review.

prad23:
Nope, you're all good. Thank you for reviewing.

PR #7410, opened 2026-07-22
imptcp: guard regex framing match at line start

rgerhards:
## Summary

- reject an invalid zero-offset regex-framing transition in `imptcp`
- make the existing oversize regex-framing test assert its diagnostics with the supported testbench helper
- document the regression-test oracle

## Root cause

After regex-framing oversize recovery, a zero line offset could be treated as a valid delimiter match. Subtracting one from that offset produced an invalid negative message length before submission.

The fix mirrors the guard already used by the shared `imtcp` parser.

## Impact

This prevents a configuration-dependent remote denial of service for `imptcp` listeners using the non-default `framing.delimiter.regex` option.

## Validation

- focused `imptcp_framing_regex-oversize.sh`
- Ubuntu 26.04 static analyzer
- change-gated Ubuntu 26.04 container check
- focused ASan/UBSan imptcp regression
- formatting and diff checks

The broad sanitizer suite had one unrelated `imfile-logrotate-async` timeout; the affected `imptcp` test passed under sanitizers.

<a href="https://cubic.dev/pr/rsyslog/rsyslog/pull/7410?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/plugins/imptcp/imptcp.c
+++ b/plugins/imptcp/imptcp.c
@@ -1053,7 +1053,7 @@ static rsRetVal ATTR_NONNULL() processDataRcvd_regexFraming(ptcpsess_t *const __
         pThis->iCurrLine = pThis->iMsg;
     } else {
         const int isMatch = !regexec(&inst->start_preg, (char *)pThis->pMsg + pThis->iCurrLine, 0, NULL, 0);
-        if (isMatch) {
+        if (pThis->iCurrLine > 0 && isMatch) {
             DBGPRINTF("regex match (%d), framing line: %s\n", pThis->iCurrLine, pThis->pMsg);
             memmove(pThis->pMsg_save, pThis->pMsg + pThis->iCurrLine, ustrlen(pThis->pMsg + pThis->iCurrLine) + 1);
             pThis->iMsg = pThis->iCurrLine - 1;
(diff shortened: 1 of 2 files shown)


mbiebl:
Hi @rgerhards ,

when security issues like this are published, I, as downstream, always wonder which versions are affected. Looking at https://tracker.debian.org/pkg/rsyslog, these are the currently supported versions in Debian.

 It would be awesome if future security issues in the upstream bug tracker included a hint when the issue was introduced (in which released version/commit).

mbiebl:
> It would be awesome if future security issues in the upstream bug tracker included a hint when the issue was introduced (in which released version/commit).

Is this something you could automate (via AI)?

rgerhards:
the advisory is here: https://github.com/rsyslog/rsyslog/security/advisories/GHSA-cj5r-wh2m-7w29

mbiebl:
Oh, that looks perfect. Sorry for the noise.

rgerhards:
some background: I requested a CVE via github, aksed after a week when they assign, they said 2+ weeks due to AI overload, I can publish anyhow. Did so a week later. Red Hat seems to have gone impatient after another 2 weeks and assigned CVE, but seemed not to have linked to the GHSA... A week later Github said it's a duplicate and they cannot assign CVE ;-)

So much to the state of responsible disclosure in the AI age (IMHO the whole process is considerably broken now).

Timing is out of my head, but roughly correct.

PR #7413, opened 2026-07-22
config: centralize explicit parameter tracking

rgerhards:
## Summary

- add a generic parse-time query for whether a named parameter was explicitly supplied
- remove one-off explicitness fields from omfwd, omrelp, and imrelp
- retain value-sensitive TLS conflict classification inside each module, keeping neutral generated values quiet
- preserve legacy omfwd selector behavior

This is a prototype follow-up to https://github.com/rsyslog/rsyslog/pull/7393. It focuses only on simplifying the current implementation, not on defining a broader future policy for conflicting configuration parameters.

Because the prototype is based directly on the current head of PR 7393, this draft includes the commits from that PR and is intended for comparing the implementation direction.

## Validation

- clang-format-18 check passed
- incremental host build passed
- RainerScript and YAML compatibility tests passed
- Ubuntu 26.04 change-gated container suite passed: 1343 passed, 33 skipped, 0 failed
- clang static analyzer found no bugs
- local Cubic review found no issues
- manual parse-state and lifetime audit found no issues

### DIFF
--- a/grammar/rainerscript.c
+++ b/grammar/rainerscript.c
@@ -1349,6 +1349,16 @@ int cnfparamvalsIsSet(struct cnfparamblk *params, struct cnfparamvals *vals) {
     return 0;
 }
 
+/* Check whether a named parameter was explicitly supplied. */
+int cnfparamvalsIsSetByName(const struct cnfparamblk *params, const struct cnfparamvals *vals, const char *name) {
+    int i;
+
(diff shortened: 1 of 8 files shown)


rgerhards:
@julthoma this is just explorative and most probably not a final solution. I just need full CI visibility. Posibly another candidate will follow. Have not yet verified anything myself.

PR #7474, opened 2026-08-07
msg: ignore null values in msgSetPropViaJSON() instead of crashing

julthomas:
Why: setting a core message property to a JSON null crashed rsyslogd. jsonToString() returns NULL for a JSON null and the string setters (msg, rawmsg, syslogtag, hostname/source, fromhost, fromhost-ip) then called strlen(NULL). Reachable from mmexternal (a program returning {"msg": null}), from mmlua and from pmnormalize.

Impact: no crash; a null core property is ignored -- it can be neither null nor unset -- leaving the current value unchanged.

Technical Overview: the value is null exactly when the json argument is NULL, so a single guard at the top of msgSetPropViaJSON() covers every property (string and numeric). It skips the property with a DBGPRINTF, mirroring the existing "unknown property ignored" policy. Two mmexternal regression tests cover the msg and syslogtag setters.

With the help of Claude AI

<a href="https://cubic.dev/pr/rsyslog/rsyslog/pull/7474?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/runtime/msg.c
+++ b/runtime/msg.c
@@ -4749,6 +4749,13 @@ static rsRetVal msgSetPropViaJSON(smsg_t *__restrict__ const pMsg,
      *       string. So we MUST NOT free it!
      */
     dbgprintf("DDDD: msgSetPropViaJSON key: '%s'\n", name);
+    if (json == NULL) {
+        /* Ignore a JSON null value: a core property can be neither null nor
+         * unset, and the setters below would otherwise crash on
+         * strlen(jsonToString(NULL)). */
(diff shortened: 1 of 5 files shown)


rgerhards (approved):

rgerhards:

PR #7478, opened 2026-08-10
stream: keep async writer alive across size-limit rotation

### LINKED ISSUE #7477: omfile: rotation.sizeLimitCommand with asyncWriting="on" stops output and hangs shutdown
### expected behavior

an omfile action with `rotation.sizeLimit` and `rotation.sizeLimitCommand` keeps writing
across rotations and rsyslogd shuts down cleanly, whatever `asyncWriting` is set to.

### actual behavior

with `asyncWriting="on"` the action rotates once, then stops writing permanently. the
process stays alive and reports `active`; only a restart recovers it. `systemctl stop`
exceeds `TimeoutStopSec` and gets `SIGKILL`ed.

with no dedicated action queue the stalled worker is the main queue worker, so all local
omfile output stops, not just output to the file that rotated.

`asyncWriting="off"` on the same action is unaffected.

measured with 1500 messages plus 50 markers at a 100KiB limit on `main` `a3a2a2882`:

| asyncWriting | delivered | rotations | SIGTERM |
|---|---|---|---|
| `off` | 1550/1550 | 3 | clean, 2s |
| `on` | 397/1550 | 1 | hung, SIGKILL after 20s |

the delivered count varies with where the size limit lands. what the size-limit command
does makes no difference: a three-line `mv` script and a full archiving script behave the
same.

### steps to reproduce the behavior

```
global(workDirectory="/tmp/repro/wrk")
module(load="imtcp")
input(type="imtcp" port="13514")

template(name="outfmt" type="string" string="%msg%\n")

if ($msg contains "TESTMSG") then {
    action(type="omfile"
           file="/tmp/repro/out.log"
           template="outfmt"
           asyncWriting="on"
           flushOnTXEnd="on"
           ioBufferSize="64k"
           rotation.sizeLimit="102400"
           rotation.sizeLimitCommand="/tmp/repro/rotate.sh")
    stop
}
```

`rotate.sh` only has to move the oversized file aside:

```sh
#!/bin/bash
mv -f /tmp/repro/out.log "/tmp/repro/out.log.$(date +%s%N)"
```

send ~1500 messages of ~250 bytes over tcp, enough for several rotations at the 100KiB
limit, then a few more as markers. the first archive is created, then nothing further is
written and the markers never arrive.

### environment

* rsyslog version: 8.2510.0-2.el9 where it was first hit, and `main` `a3a2a2882`
  (v8.2608.0.daily) where it was reproduced from source
* platform: rhel 9 x86_64 (distro package), and ubuntu 24.04 x86_64 (source build)
* modules: imjournal, imfile, omfile and omfwd on the original host; imtcp plus omfile in
  the reduced repro above

### root cause

line numbers from `main` `a3a2a2882`.

`strmWrite()` takes `pThis->mut` when `bAsyncWrite` is set (`runtime/stream.c:1983`) and
calls `doSizeLimitProcessing()` inside that locked region (`:2005`).
`doSizeLimitProcessing()` calls `strmCloseFile()` (`:226`), which calls `stopWriter()`
whenever `bAsyncWrite` is set (`:487`). `stopWriter()` sets `bStopWriter = 1` and
`pthread_join()`s the writer thread (`:454`).

the writer thread is created in one place, `strmConstructFinalize()` (`:1258`), and
`bStopWriter` is reset only there, so after the rotation the stream has no writer.
`STREAM_ASYNC_NUMBUFS` is 2 (`runtime/stream.h:97`), so `doAsyncWriteInternal()` accepts
one more buffer and then waits on `notFull` (`:1499`), which only the joined writer could
signal.

from a `-d` run at the rotation:

```
main Q:Reg/w0 : stream.c: file 9(out.log) doAsyncWriteInternal at exit: iCnt 1, iEnq 6
<writer>      : stream.c: file 9(out.log) asyncWriterThread terminated
main Q:Reg/w0 : stream.c: file 9(out.log) closing
main Q:Reg/w0 : srutils.c: exec program '/tmp/repro/rotate.sh' with 1 args
main thread   : glbl.c: waitpid for child 32417 returned status: 00
main Q:Reg/w0 : stream.c: file -1(out.log) doAsyncWriteInternal at exit: iCnt 1, iEnq 7
main Q:Reg/w0 : stream.c: file -1(out.log) doAsyncWriteInternal at begin: iCnt 1, iEnq 7
```

that last line is the final stream entry in the log. the rotation command itself runs to
completion and the archive is created. no `rs:<filename>` writer thread remains in the
process afterwards; a healthy async run has one.

`rotation.sizeLimitCommand` is needed to see the hang. without it,
`resolveFileSizeLimit()` returns `RS_RET_NON_SIZELIMITCMD` and sets `pThis->bDisabled`, so
`strmWrite()` returns early and never reaches the blocking call. output still stops, but
shutdown stays clean.

### fix

pr to follow: keep the async writer thread alive across the size-limit rotation, so the
stream still has a consumer when it is written to again, and drain the pending buffers
with `strmWaitAsyncWriterDone()` before the fd is closed. it comes with a regression test
using `asyncWriting="on"`.

owen-rote:
fixes https://github.com/rsyslog/rsyslog/issues/7477

## problem

an omfile action with `rotation.sizeLimit`, `rotation.sizeLimitCommand` and
`asyncWriting="on"` rotates once, then stops writing permanently, and shutdown hangs.

`doSizeLimitProcessing()` reaches `stopWriter()` through `strmCloseFile()`
(`runtime/stream.c:226` -> `:487`), which joins the stream's async writer thread. the
writer is created only in `strmConstructFinalize()` (`:1258`), so the stream is left
without a consumer. with `STREAM_ASYNC_NUMBUFS` at 2, the next buffer is accepted and the
one after that waits on `notFull` (`:1499`) forever.

## fix

`strmCloseFile()` becomes a thin wrapper around
`strmCloseFileInternal(pThis, bStopAsyncWriter)` and passes 1, so every existing caller
keeps its current behavior. `doSizeLimitProcessing()` passes 0, which:

* keeps the writer thread alive, so the stream still has a consumer after the rotation
* drains the buffer `strmFlushInternal()` just handed over with a second
  `strmWaitAsyncWriterDone()`, so the data is on disk before the fd is closed and the
  rotation command runs
* keeps the caller's mutex held for the rest of `strmWrite()`, since `stopWriter()` is
  what would otherwise release it

the stream reopens its file descriptor on the next write through the existing
`strmPhysWrite()` -> `strmOpenFile()` path (`:1689`).

scope is limited to omfile: it is the only caller that enables async streams
(`tools/omfile.c:737`) and it uses `STREAMTYPE_FILE_SINGLE`, so `strmNextFile()` and
`rereadTruncated()` cannot reach the async branch.

## test

`tests/omfile-sizelimitcmd-async.sh` mirrors `tests/omfile-sizelimitcmd-many.sh` with
`asyncWriting="on"` added. it fails deterministically on unpatched `main` (timeoutGuard
abort, ~44k messages stranded in the queue) and passes in 2s with the fix, across 8
rotations.

## before/after

standalone reproducer from the issue, 1500 messages plus 50 markers at a 100KiB limit, on
`main` `a3a2a2882`:

| build | delivered | rotations | SIGTERM |
|---|---|---|---|
| `asyncWriting="off"` | 1550/1550 | 3 | clean, 2s |
| `asyncWriting="on"` | 397/1550 | 1 | hung, SIGKILL after 20s |
| `asyncWriting="on"` + this pr | 1550/1550 | 3 | clean, 2s |

## validation

ubuntu 24.04 host with docker.

* clang static analyzer via `devtools/run-static-analyzer.sh` on
  `rsyslog/rsyslog_dev_base_ubuntu:26.04`: no bugs found
* change-gated `devtools/devcontainer.sh --rm devtools/run-ci.sh` on the same image,
  `CC=gcc`, `CFLAGS=-g`, `CI_MAKE_OPT=-j10`, `CI_MAKE_CHECK_OPT=-j10`, default pr service
  suppressions from `devtools/apply-service-relevance.sh`: total 1466, pass 1433, skip 33,
  fail 0, error 0 in 888s
* `make distcheck TEST_RUN_TYPE=MOCK-OK -j16`: pass
* `devtools/format-code.sh --git-changed --check --check-if-available` and
  `shellcheck -S warning` on the new test: clean
* host-side runs of `omfile-outchannel.sh`, `omfile-outchannel-many.sh`, `omfile_hup.sh`,
  `omfile-module-params.sh`, `diskqueue.sh`, `queue-persist.sh`, `gzipwr_*`,
  `imfile-basic.sh`, `imfile-truncate.sh`: pass

local cubic was not run, it is not installed on this machine.

`tests/zstd.sh` fails on this host, but it fails identically on unpatched `main`, verified
by stashing the patch and rebuilding. unrelated to this change.

happy to squash the two commits or adjust anything here.

### DIFF
--- a/doc/ai/module_map.yaml
+++ b/doc/ai/module_map.yaml
@@ -57,6 +57,18 @@ runtime_msg:
   notes:
     - MsgGetRcvFromProp returns fromhost, fromhost-ip, or fromhost-port with the stored prop length; it resolves DNS only when NEEDS_DNSRESOL is still set.
 
+runtime_stream:
+  paths: ["runtime/stream.c", "runtime/stream.h"]
+  requires_serialization: true
+  locks:
+    - type: recursive-mutex
+      field: strm_t.mut
+  notes:
+    - strm_t.mut, notFull, notEmpty and isEmpty exist only when bAsyncWrite is set; omfile is the only caller that enables async streams and it serializes writes behind its own pData.mutWrite.
+    - The async writer thread is created once in strmConstructFinalize(); no other code path recreates it, so any caller that stops it must be done with the stream.
+    - strmCloseFile() stops the async writer and, as a documented side effect, leaves pThis->mut unlocked on return; callers must either hold no expectation of the lock or use the non-stopping variant.
+    - Size-limit rotation uses the non-stopping close: the writer thread survives, pending buffers are drained with strmWaitAsyncWriterDone() before the fd is closed, and the caller's mutex stays held across the rotation command.
+
 mmnormalize:
   paths: ["plugins/mmnormalize/"]
   requires_serialization: true
--- a/runtime/AGENTS.md
+++ b/runtime/AGENTS.md
@@ -37,6 +37,10 @@ collection, and process orchestration).
   - Directly invoke the most relevant shell test under `tests/` (e.g. `./tests/queue-persist.sh`).
   - Use `make check TESTS='<script>.sh'` when you need automake’s harness or Valgrind variants (`*-vg.sh`).
   - For configuration validation edits, run `./tests/validation-run.sh`.
+- **Existing smoke coverage:** `tests/omfile-sizelimitcmd-async.sh` covers
+  `stream.c` size-limit rotation with `asyncWriting="on"`, i.e. that the async
+  writer thread survives a rotation. `tests/omfile-sizelimitcmd-many.sh` covers
+  the same rotation path with the default synchronous writer.
 - When changing exported symbols, update `runtime/Makefile.am` and ensure the library version script (if touched) remains consistent with existing SONAME policies.
 
 ## Coding expectations
--- a/runtime/stream.c
+++ b/runtime/stream.c
@@ -224,11 +224,11 @@ static rsRetVal doSizeLimitProcessing(strm_t *pThis) {
          * need to preserve it.
          */
         CHKmalloc(pszCurrFName = ustrdup(pThis->pszCurrFName));
-        /* The stream keeps being written to after the rotation, so the async
-         * writer thread must survive it. Stopping the writer here would leave
-         * the stream without a consumer for its buffer queue, and the producer
-         * would block on notFull forever. Keeping it alive also preserves the
-         * caller's mutex, which stopWriter() would otherwise release.
+        /* the stream is written to again after the rotation, so the async
+         * writer thread must survive it. stopping it here leaves the buffer
+         * queue without a consumer and the producer blocks on notFull forever.
+         * keeping it also preserves the caller's mutex, which stopWriter()
+         * would release.
          */
         CHKiRet(strmCloseFileInternal(pThis, 0));
         CHKiRet(resolveFileSizeLimit(pThis, pszCurrFName));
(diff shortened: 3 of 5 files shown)


rgerhards:
@codex review

owen-rote:
@codex review

rgerhards:
@codex review

owen-rote:
codex: both fixed in f461b844c.

cubic 1: LogError() ends in iminternalAddMsg(), which appends under a 1s
timedlock and drops on timeout. never takes a stream lock, and nothing holds
mutList while waiting on one. so no cycle.

cubic 2: pre-existing. fdDir is only set at stream.c:1188 and :1248 and never
reopened, so sync mode already loses directory fsync after a rotation today.
separate PR?

can squash whenever.

rgerhards:
@codex review

PR #7489, opened 2026-08-12
tests: retire obsolete TSan function suppressions

rgerhards:
## Summary

- initialize action suspension-reporting defaults before queue workers start
- make impcap shutdown signaling async-signal-safe
- remove obsolete function-level TSan suppressions and add deterministic concurrency coverage

## Why

The old function-wide suppressions masked more than their original findings. The action reporting race remained real: workers could concurrently resolve the same configuration defaults after startup. The new testbench barrier reproduces that race deterministically, and the fix resolves the values before workers begin.

The impcap signal handler also performed logging and configuration traversal from signal context. The controller now calls `pcap_breakloop()` directly; the empty handler only interrupts capture on platforms where a cross-thread break does not wake the blocking syscall.

`processWorksetItem` and `doLogMsg` are covered by their already-merged synchronization fixes, while rsyslog no longer calls `setlocale`.

## Validation

- focused TSan: `action-reporting-concurrency-tsan.sh`
- focused TSan: `manytcp.sh`
- focused TSan: `impcap_basic_http.sh`
- focused TSan: `impcap_basic_dns.sh`
- full unsuppressed TSan run: 668 tests, 639 passed, 19 skipped, 10 failures from independent existing races in `debug.c`, `queue.c`, and `imdiag.c`; none of the removed function suppressions was a reported race site
- `./devtools/format-code.sh --git-changed --check`
- `git diff --check`
- ShellCheck on the new test using the testbench's established exclusions

The independent full-suite TSan findings are intentionally not hidden or included in this PR.

<a href="https://cubic.dev/pr/rsyslog/rsyslog/pull/7489?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/contrib/impcap/impcap.c
+++ b/contrib/impcap/impcap.c
@@ -585,18 +585,10 @@ void packet_parse(uchar *arg, const struct pcap_pkthdr *pkthdr, const uchar *pac
 /* This is used to terminate the plugin.
  */
 static void doSIGTTIN(int __attribute__((unused)) sig) {
-    pthread_t tid = pthread_self();
-    const int bTerminate = PREFER_LOAD_INT(&bTerminateInputsSigSafe);
-    DBGPRINTF("impcap: awoken via SIGTTIN; bTerminateInputsSigSafe: %d\n", bTerminate);
-    if (bTerminate) {
-        for (instanceConf_t *inst = runModConf->root; inst != NULL; inst = inst->next) {
-            if (pthread_equal(tid, inst->tid)) {
-                pcap_breakloop(inst->device);
-                DBGPRINTF("impcap: thread %lx, termination requested via SIGTTIN - telling libpcap\n",
-                          (long unsigned int)tid);
-            }
-        }
-    }
+    /* pcap_breakloop() is called by the controller before this signal is sent.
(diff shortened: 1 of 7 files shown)


rgerhards:
Cubic review follow-up for 17a8b5f38:

- Restored fun:doSIGTTIN; it remains required for the impcap SIGTTIN handler policy.
- Added mutexCancelCleanup around the barrier condition wait, so cancellation cannot leave its mutex locked.
- The undersubscribed-barrier concern does not apply: queue.workerThreads=2 with queue.workerThreadMinimumMessages=1 targets both workers immediately; the two test messages supply the two arrivals.
- Expanded the test header with the two-worker/two-message barrier setup and TSan/shutdown oracle.
- Renamed the new barrier fields, condition variable, and helper to snake_case.

Threaded replies were not permitted for this fork account, so this records the responses without resolving the threads.

Issue #7406, opened 2026-07-21
tests: track load-sensitive local validation flakes


rgerhards:
## Summary

Two different tests failed once under the parallel local PR-ready Ubuntu 26.04 suite and passed immediately when rerun alone under the unchanged configuration. Neither test is related to the imtcp zlib change that triggered the validation.

## Evidence

Image: rsyslog/rsyslog_dev_base_ubuntu:26.04
Image ID: sha256:32ade478a405e4f27f077b5268ec5ecc59dd572843ad67ca2b6723594960ae09
Local concurrency: build -j10, check -j10

First broad run:
- TOTAL 1392, PASS 1357, SKIP 34, FAIL 1
- Failure: segmented-da-idle-cleanup-failure.sh
- Immediate isolated rerun passed.

Second broad run:
- TOTAL 1392, PASS 1357, SKIP 34, FAIL 1
- segmented-da-idle-cleanup-failure.sh passed in this run.
- Failure: ratelimit-policy-persource-key-reload.sh
- Immediate isolated rerun passed.

All imtcp zlib and zstd tests passed in both broad runs. The behavior is consistent with load-sensitive test flakes rather than a deterministic product regression.

## Reproduction commands

Broad lane: devtools/local-validation-plan.sh --run

Isolated reruns:
- make -C tests check TESTS=segmented-da-idle-cleanup-failure.sh
- make -C tests check TESTS=ratelimit-policy-persource-key-reload.sh

rgerhards:
PR #7407 hosted arm64 ASan also produced two unrelated segmented disk-queue failures under the broad parallel lane:

- segmented-da-idle-crash-state.sh
- segmented-da-persist.sh

The new imtcp-stream-always-zlib-completion-release.sh test passed in that same arm64 ASan job. These failures are recorded as additional load/architecture-sensitive flake evidence; no imtcp source path was implicated.

rgerhards:
Additional hosted flake evidence from PR #7407, workflow run 29841467419, arm64 CI (native, asan) job 88678417187: `lookup_table-hup-backgrounded.sh` timed out after 60 seconds waiting for the imdiag port. The PR-specific `imtcp-stream-always-zlib-completion-release.sh` test passed in this job. This appears unrelated to the imtcp/zlib change.

PR #7530, opened 2026-08-25
core: fix spelling errors in comments and messages

prownd:
## What this is

A `codespell` sweep over `runtime/`, `tools/` and `grammar/`, plus a manual
pass over the findings codespell offers several corrections for and therefore
does not auto-apply.

396 corrections across 102 files. Comment text, debug output and user-facing
message wording only - no functional or behavioral change.

Examples: `ressources` -> `resources`, `successfull` -> `successful`,
`postion.from` -> `position.from`, `ERROR/UNKNWON` -> `ERROR/UNKNOWN`,
`enqueing` -> `enqueuing`, `litteral` -> `literal`.

## Care taken

- No identifier and no configuration parameter name is touched.
- Every modified string literal was checked against `tests/` and `doc/` first;
  none of them is matched on by the testbench or the documentation.
- Identifier-shaped false positives (`pRes`, `aNULL`, `pres`, `PRIs`, ...) were
  excluded rather than "corrected".
- Findings with several plausible corrections were reviewed by hand in context
  instead of being auto-applied.

## Formatting

Verified formatting-neutral: for every touched `.c`/`.h` file, `clang-format`
reports exactly the same result before and after the change. No modified line
exceeds the 120 column limit, and no modified line is part of a macro
continuation block, so `AlignEscapedNewlines` is unaffected.

Note: the spellcheck CI job only covers `doc/source`, so this change does not
alter CI behavior; it just removes the backlog from the core sources.

## Validation

Full build clean. Testbench sample around templates, parsers, queues and
`imtcp` passes unchanged.

### DIFF
--- a/grammar/grammar.y
+++ b/grammar/grammar.y
@@ -2,7 +2,7 @@
  * Please note: this file introduces the new config format, but maintains
  * backward compatibility. In order to do so, the grammar is not 100% clean,
  * but IMHO still sufficiently easy both to understand for programmers
- * maitaining the code as well as users writing the config file. Users are,
+ * maintaining the code as well as users writing the config file. Users are,
  * of course, encouraged to use new constructs only. But it needs to be noted
  * that some of the legacy constructs (specifically the in-front-of-action
  * PRI filter) are very hard to beat in ease of use, at least for simpler
@@ -125,7 +125,7 @@ extern int yyerror(const char*);
 %nonassoc UMINUS NOT
 
 %expect 1 /* dangling else */
-/* If more erors show up, Use "bison -v grammar.y" if more conflicts arise and
+/* If more errors show up, Use "bison -v grammar.y" if more conflicts arise and
  * check grammar.output for were exactly these conflicts exits.
  */
 %%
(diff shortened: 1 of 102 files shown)


prownd:
CI note: `arm64 CI (native, asan)` failed on a single unrelated test,
`imptcp-stream-always-corrupt.sh`. This PR changes comment text only -
the three files it touches under that code path (`runtime/stream.c`,
`runtime/stream.h`, `runtime/tcpsrv.c`) have no non-comment changes at all:

```
-    accross files. For other cases we accept the very remote risk.
+    across files. For other cases we accept the very remote risk.
-    objSerializeSCALAR(pStrm, iCurrFNum, INT); /* implicit cast is OK for persistance */
+    objSerializeSCALAR(pStrm, iCurrFNum, INT); /* implicit cast is OK for persistence */
-        char *pIOBuf_truncation; /* iobuffer used during trucation detection block re-reads */
+        char *pIOBuf_truncation; /* iobuffer used during truncation detection block re-reads */
```

The same arm64 asan job passes on my other three open PRs (#7527, #7528,
#7529), two of which do change `runtime/msg.c` behavior. Looks like one of
the load-sensitive flakes tracked in
https://github.com/rsyslog/rsyslog/issues/7406 - a rerun should be enough,
but I can't trigger one from a fork.

The `package build (Ubuntu, ...)` failures on the sibling PRs are also
infrastructure: Launchpad returned `HTTP Error 500 / GPGKeyTemporarilyNotFoundError`
for the adiscon PPA key on all three retries.

PR #7568, opened 2026-09-06
nsd_ossl: keep shared-group queries in debug output

### LINKED ISSUE #7325: False-positive "no shared curve" log on successful TLS handshake breaks security certification compliances
### Component Affected
`rsyslog-openssl` (`lmnsd_ossl` network stream driver module / `runtime/nsd_ossl.c`)

### Environment
* **OS Distribution:** Rocky Linux 9 / RHEL 9 (or any distro running modern OpenSSL 3.x)
* **Rsyslog Version:** v8.2510.0-2.el9 (and present in current upstream code)
* **OpenSSL Version:** OpenSSL 3.5.x

### Issue Description: False-Positive Cryptographic Logs Violating Security Standards

When rsyslog is configured with secure cryptographic constraints over TLS, it floods the application log with a false-positive warning:

```text
nsd_ossl: Information, no shared curve between syslog client '*****' and server
```

#### Why This is a Critical Operational Problem:
Even though the connection is fully functional and successfully established, **this log message is causing severe compliance violations during security certification testing.** To end users and security auditors, a message stating `"no shared curve"` looks explicitly like a cryptographic negotiation failure or an unencrypted/insecure fallback. 

If a shared curve exists and the handshake is completely successful, **this log should not appear.** It disrupts audit trails and fails strict enterprise certification testing behaviors.

### Technical Proof of Success (The Mismatch)
Running a direct cryptographic interrogation bypassing the rsyslog wrapper proves that the client and server negotiate perfectly over native OpenSSL libraries:

```text
\$ openssl s_client -connect <SERVER_IP>:6514 -tls1_2 -brief
CONNECTION ESTABLISHED
Protocol version: TLSv1.2
Ciphersuite: ECDHE-ECDSA-AES256-GCM-SHA384
Signature type: ecdsa_secp256r1_sha256
Peer Temp Key: X25519, 253 bits
```

**Analysis:**
The handshake establishes perfectly using `X25519` for the key exchange and `secp256r1` for the certificate signature validation. However, because OpenSSL 3.x loops through internal multi-stage evaluation checks, the rsyslog `nsd_ossl` wrapper mistakenly captures an intermediate transitional check macro and prints a false `"no shared curve"` alert right before the connection successfully binds.

### Expected Behavior
If OpenSSL successfully finishes the handshake and resolves an active cipher session (`SSL_get_current_cipher`), rsyslog should suppress any intermediate loop warnings. The logging output must remain entirely clean so as not to trigger false alarms for security compliance systems.

---

### Request to Maintainers

Can the core engineering team confirm if this intermediate event loop log bug has been addressed in the latest OpenSSL 3.x codebase adjustments? If not, we urgently need a path to silence or correct this logic so that successful connections do not throw deceptive error-like messages.

@rgerhards @alorbach @davidelang 
*(Tagging core architecture maintainers and active network module reviewers to help analyze this logic block).*

be-student:

Successful TLS connections should not look like negotiation failures in normal operator logs. Keep the shared-group query in neutral debug output instead of emitting the misleading informational `no shared curve` message. Actual handshake, cipher, and certificate-validation error paths remain intact.

The query is not a handshake-success check: it can return zero for a successful client-side ECDHE session and for group-free TLS 1.2 RSA key exchange. This change does not alter TLS negotiation or authentication settings.

### References

Fixes: https://github.com/rsyslog/rsyslog/issues/7325

### Notes

- Adds normal Automake regressions that exchange 100 messages over real TLS for ECDHE and RSA, then inspect normal logs on both endpoints. Both fail against the original diagnostic and pass with the change. The existing incompatible-cipher regression also passes.
- Documents the diagnostic contract in the operator guide and runtime agent guidance.
- Passed `make distcheck TEST_RUN_TYPE=MOCK-OK -j4` through archive rebuild, install/uninstall, and cleanup. Mock mode validates packaging and scheduling; it does not count as runtime regression evidence.
- Passed direct clang-format 18 validation, documentation build, static analyzer, and the repository's independent security-review process with no findings.
- Broad Linux container run: 1,495 test entries, 1,402 passed, 74 skipped, 19 failed, zero errors. Both added regressions passed. This is **not** an all-green broad run. Failures include connection tests reporting `Too many open files`, an inherited `/mnt/lima-rosetta/rosetta` descriptor in omprog descriptor assertions, missing `libgcc_s.so.1` diagnostics, and queue/helper timeout cases. The stalled `mmexternal-InvldProg-vg` helper was interrupted after more than 51 minutes and is included among the failures. These failures have not all been reproduced on an untouched baseline.
- Isolated rerun of `imtcp-tls-ossl-manycon.sh` also hit `Too many open files` in tcpflood at connection 519, followed by connection resets. This remains unresolved in this environment; no test thresholds or TLS failure handling were changed.
- Validation environment: upstream Ubuntu 26.04 amd64 development image under Rosetta on macOS, using native Linux-volume storage. Hosted native CI remains necessary to resolve broad-suite uncertainty.

AI assistance: implemented and validated with Codex. No production deployment or live security-certification claim.

<a href="https://cubic.dev/pr/rsyslog/rsyslog/pull/7568?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


be-student:
I checked the format-specifier finding against the actual C expression. `SSL_get_shared_curve()` is a macro: in OpenSSL 1.0.2 it expands directly to `SSL_ctrl(...)`; in 1.1.1 and 3.x it aliases `SSL_get_shared_group()`, which also expands to `SSL_ctrl(...)`. `SSL_ctrl` returns `long`, so `%ld` matches the vararg type here.

Sources: [OpenSSL 1.0.2 macro](https://github.com/openssl/openssl/blob/OpenSSL_1_0_2u/ssl/ssl.h#L2084), [OpenSSL 3.0 macro](https://github.com/openssl/openssl/blob/openssl-3.0.20/include/openssl/ssl.h.in#L1428), and [its long return type](https://github.com/openssl/openssl/blob/openssl-3.0.20/include/openssl/ssl.h.in#L1910).

I also compiled a C11 `_Generic(SSL_get_shared_curve((SSL *)0, -1), long: 1, default: 0)` static assertion and the `%ld` call against the installed OpenSSL 3.6.3 headers with `clang -Wformat=2 -Werror -fsyntax-only`; both passed. Keeping `%ld` preserves the correct type. No source change is needed for this finding.

